PR #1, opened 2016-12-22
ECIES crpypto implementation

adamsmo:


src/main/scala/io/iohk/ethereum/crypto/ECIESPublicKeyEncoder.scala
```diff
@@ -0,0 +1,13 @@
+package io.iohk.ethereum.crypto
+
+import org.spongycastle.crypto.KeyEncoder
+import org.spongycastle.crypto.params.AsymmetricKeyParameter
+import org.spongycastle.crypto.params.ECPublicKeyParameters
+
+/**
+  * Created by Anton Nashatyrev on 01.10.2015.
```
rtkaczyk: Who is that? 😉

src/main/scala/io/iohk/ethereum/crypto/ECIESCoder.scala
```diff
@@ -0,0 +1,165 @@
...
+import org.spongycastle.crypto.modes.SICBlockCipher
+import org.spongycastle.crypto.params._
+import org.spongycastle.crypto.parsers.ECIESPublicKeyParser
+import org.spongycastle.crypto.{BufferedBlockCipher, InvalidCipherTextException}
+import org.spongycastle.math.ec.ECPoint
+
+object ECIESCoder {
+  val KEY_SIZE = 128
+  val params: X9ECParameters = SECNamedCurves.getByName("secp256k1")
+  val CURVE = new ECDomainParameters(params.getCurve, params.getG, params.getN, params.getH)
```
rtkaczyk: Do you think we could use the naming conventions as proposed here: http://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods? That is upper camel case.

src/main/scala/io/iohk/ethereum/crypto/ECIESCoder.scala
```diff
@@ -0,0 +1,165 @@
...
+
+  @throws[IOException]
+  @throws[InvalidCipherTextException]
+  def decrypt(privKey: BigInteger, cipher: Array[Byte], macData: Option[Array[Byte]] = None): Array[Byte] = {
+    var plaintext: Array[Byte] = null
```
rtkaczyk: No need for this, is there?

src/test/scala/io/iohk/ethereum/crypto/ECIESCoderSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+    expectedPayload shouldBe Hex.toHexString(payload)
+  }
+
+  "ECIESCoder" should "past tests from ethereumJ - test2" in {
+
+    val privKey = new BigInteger("5e173f6ac3c669587538e7727cf19b782a4f2fda07c1eaa662c593e5e85e3051", 16)
+    val payload = Hex.decode("1122334455")
+    val pubKeyPoint = ECIESCoder.CURVE.getG.multiply(privKey)
+    val cipher = ECIESCoder.encrypt(pubKeyPoint, payload)
+    val decrypted_payload = ECIESCoder.decrypt(privKey, cipher)
```
rtkaczyk: Let's use camelCase

rtkaczyk:
What's your opinion on working with the `ByteString`? Should we use them everywhere or only where necessary? Should they also be used for RLP in #2?

Certainly there are many pros like immutability, easier comparison, hashcodes and monadic combinators without implicits. On the other hand, I understand that underlying libs rely on byte arrays, so there's always a conversion needed - will that impact the efficiency?

Anyone's thoughts appreciated.

src/main/scala/io/iohk/ethereum/crypto/ECIESPublicKeyEncoder.scala
adamsmo: Probably author of this class in EthereumJ, since I ported it 1:1 I decided that it should stay, as it turns out this class is simple and can be "inlined"

adamsmo:
I think that we should use `ByteString` only when it is necessary. Problem is that encryption algorithms from `spongycastle` does not returns result as separate array but writes them into provided one, so we have to use mutable structure. For me using `ByteString` makes sense where we can replace concatenation/slicing because it is easier to read than done with `System.arraycopy`.

ntallar:
@adamsmo about:
> For me using ByteString makes sense where we can replace concatenation/slicing because it is easier to read than done with `System.arraycopy`.

Do you think it would be a good idea to use `ByteString` for RLP? It seems to make sense as we mainly interact with it by slicing it (and also to transform it to and from primitive types). We did some testing and it even seems to increase the decoding performance a little.

adamsmo:
@ntallar if this is main thing that we do than yes i think we should use `ByteStrig`, I added comment to PR

adamsmo:
@rtkaczyk other than things you mentioned is there something else to fix?

LukasGasior1:
LGTM :+1:

rtkaczyk:
👍

PR #2, opened 2016-12-22
RLP Encoding

AlanVerbner:
## Description

This PR includes a [RLP binary encoding](https://github.com/ethereum/wiki/wiki/RLP) scala implementation.

As [EthereumJ](https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/test/java/org/ethereum/util/RLPTest.java), [RubyRLP](https://github.com/cryptape/ruby-rlp), and [PyRLP](https://github.com/ethereum/pyrlp) tests there might be some "duplicated" tests that we can remove sometime in the future but we wanted to keep them to make it easier for us to compare it against other implementations. 

Also, you will find some "real life" encoding decoding examples (like an encoded by EthereumJ Ping Message decoding or a sample Block with TX composition).


LukasGasior1:
Are the protocol messages supposed to extend RLPValue? How about going more decoupled way, maybe something like json encoders/decoders in Play! framework (and "clean" message classes)?

rtkaczyk:
Ditto on that. 
We would then create new typeclass instances in `RLPImplicits` for whatever we need to encode.

And while on implicits, why do we need the implicit conversion in the other directions, i.e. `RLPEncodable` -> primitive types?

src/main/scala/io/iohk/ethereum/utils/RLP.scala
```diff
@@ -0,0 +1,276 @@
...
+  /**
+    * Reason for threshold according to Vitalik Buterin:
+    * - 56 bytes maximizes the benefit of both options
+    * - if we went with 60 then we would have only had 4 slots for long strings
+    * so RLP would not have been able to store objects above 4gb
+    * - if we went with 48 then RLP would be fine for 2^128 space, but that's way too much
+    * - so 56 and 2^64 space seems like the right place to put the cutoff
+    * - also, that's where Bitcoin's varint does the cutof
+    */
+  val SIZE_THRESHOLD: Int = 56
```
rtkaczyk: Do you think we could use the naming conventions as proposed here: http://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods. That it upper camel case?

build.sbt
```diff
@@ -4,4 +4,9 @@ version := "0.1"
 
 scalaVersion := "2.11.8"
 
-
+libraryDependencies ++= Seq(
+  "org.scorexfoundation" %% "scorex-core" % "2.0.0-M3",
+  "com.madgag.spongycastle" % "core" % "1.54.0.0",
+  "org.scalatest" %% "scalatest" % "2.+" % "test",
```
rtkaczyk: Let's use the newest testing libs

src/main/scala/io/iohk/ethereum/utils/RLP.scala
```diff
@@ -0,0 +1,276 @@
...
+  * For example, an empty string is an item, as is the string containing the word
+  * "cat", a list containing any number of strings, as well as more complex data
+  * structures like ["cat",["puppy","cow"],"horse",[[]],"pig",[""],"sheep"]. Note
+  * that in the context of the rest of this article, "string" will be used as a
+  * synonym for "a certain number of bytes of binary data"; no special encodings
+  * are used and no knowledge about the content of the strings is implied.
+  * <p>
+  * See: https://github.com/ethereum/wiki/wiki/%5BEnglish%5D-RLP
+  *
+  * @author Roman Mandeleil
```
AlanVerbner: lol....we wanted to keep main comments to help other to compare this impl against EtehereumJ but this one is a misstake...fixed!

src/main/scala/io/iohk/ethereum/utils/RLP.scala
AlanVerbner: @rtkaczyk definetly...fixed

AlanVerbner:
@LukasGasior1 about:

> Are the protocol messages supposed to extend RLPValue? How about going more decoupled way, maybe something like json encoders/decoders (typeclasses) in Play! framework (and "clean" message classes)?

Ok, we will take a second look at it. We first tried type classes but the code got dirty very quickly. Feel free to point us some example / code regarding this if you want to.

@rtkaczyk not sure if I understand: 

> And while on implicits, why do we need the implicit conversion in the other directions, i.e. RLPEncodable -> primitive types?

You need that in order to decode messages to known types (for example [this](https://github.com/input-output-hk/etc-client/blob/feature/rlp/src/test/scala/io/iohk/ethereum/utils/RLPSuite.scala#L472) PingMessage example)

LukasGasior1:
@AlanVerbner 
I mean something in this way:

```
trait RLPEncoder[T] {
	def encode(obj: T): RLPValue
}

trait RLPDecoder[T] {
	def decode(rlp: RLPValue): T
}


case class Ping(...)

object Ping {
	implicit val encDec = new RLPEncoder[Ping] with RLPDecoder[Ping] {
		override def encode = ...
		override def decode = ...	
	}
}

```

then you can just have encode (and decode) method:
```
def encode[T: RLPEncoder](obj: T): RLPValue
```

capable of encoding any T as long as there's implicit encoder in scope

@rtkaczyk ^ does that make sense?

rtkaczyk:
@LukasGasior1 yes, that's also what I was thinking. `RLPEncoder/RLPDecoder` companions should provide some primitives for encoding simple values, that the instances would rely on.

src/main/scala/io/iohk/ethereum/utils/RLP.scala
```diff
@@ -0,0 +1,274 @@
...
+        case p if p == OffsetShortItem => (new RLPValue {
+          override def toBytes: Array[Byte] = Array.emptyByteArray
+        }, pos + 1)
+        case p if p < OffsetShortItem => (new RLPValue {
+          override def toBytes: Array[Byte] = Array(data(pos))
+        }, pos + 1)
+        case p if p <= OffsetLongItem =>
+          val length = p - OffsetShortItem
+          val res: Array[Byte] = new Array[Byte](length)
+          Array.copy(data, pos + 1, res, 0, length)
```
adamsmo: maybe we can use `ByteString` here to to slice consecutive parts from `data`?
maybe than we could commit position and only pass reminder to recurrent call, sth like:
`val (res, reminder) = data.splitAt(pos+1)` ?

build.sbt
AlanVerbner: Fixed!

AlanVerbner:
@LukasGasior1 @rtkaczyk Just pushed the improvements you suggested. Please let us know what do you think :smile:

LukasGasior1:
@AlanVerbner I think that's a step in good direction :)

It seems RLPImplicits are not really implicits ATM:
```
 override def encode(obj: PingMessage): RLPEncodeable = new RLPList {
        override def items: Seq[RLPEncodeable] = Seq(
          intEncDec.encode(obj.version),
          Endpoint.encDec.encode(obj.from),
          Endpoint.encDec.encode(obj.to),
          longEncDec.encode(obj.expiration)
        )
      }
```

it would be way better if we could just write:
```
override def encode(obj: PingMessage) = {
  import obj._  
  RLPList(version, from, to, expiration)
}
```
this should be easily doable

AlanVerbner:
Ok, will take another look @LukasGasior1 :smile:

AlanVerbner:
@LukasGasior1 I guess it's done now :smile:

src/main/scala/io/iohk/ethereum/utils/RLP.scala
AlanVerbner: @adamsmo will try this now :smile:

AlanVerbner: @adamsmo Just pushed a version with `ByteString`. Tried different alternatives but finally decided not to "_commit position and only pass reminder to recurrent call_". That suggestion was in terms of speed or code clarity?

adamsmo: it was about code clarity and readability
auto correct changed `omit` to `commit`

AlanVerbner: @adamsmo it seems it works a little bit slower going for the _omit position_ alternative (about 10-20%, I can run better tests to be more precise if you want to). I would vote to keep it as it is assuming the code is not too unreadable and we won't need to dig/touch this code too much. What do you think?

adamsmo: @AlanVerbner sure, if this approach is more efficient

LukasGasior1:
For the protocol handshake (frame codec actually) I'm also going to need following RLP-related methods (from ethereumj):
```
public static RLPElement decode2OneItem(byte[] msgData, int startPos) {
        RLPList rlpList = new RLPList();
        fullTraverse(msgData, 0, startPos, startPos + 1, 1, rlpList);
        return rlpList.get(0);
    }
```
and something like:
```
RLP.getNextElementIndex(buffer, pos)
```
see:
https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/net/rlpx/FrameCodec.java#L195 and
https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/net/rlpx/FrameCodec.java#L166

would it be difficult to add these in your implementation?

AlanVerbner:
@LukasGasior1 checking it atm

AlanVerbner:
@LukasGasior1 about `decode2OneItem` our RLP implementation only decodes one single item at a time  so it's already supported if you send the array starting at the correct position. For example, the [following test](https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/test/java/org/ethereum/util/RLPTest.java#L1122) taken from EthereumJ 

```java
 @Test
public void partialDataParseTest() {
    String hex = "000080c180000000000000000000000042699b1104e93abf0008be55f912c2ff";
    RLPList el = (RLPList) decode2OneItem(Hex.decode(hex), 3);
    assertEquals(1, el.size());
    assertEquals(0, Util.rlpDecodeInt(el.get(0)));
}
```

will be translated as:


```scala
test("Partial Data Parse Test") {
    val hex: String = "000080c180000000000000000000000042699b1104e93abf0008be55f912c2ff"
    val data = Hex.decode(hex)
    val decoded: Try[Seq[Int]] = RLP.decode[Seq[Int]](data.splitAt(3)._2)
    assert(decoded.isSuccess)
    assert(1 == decoded.get.length)
    assert(0 == decoded.get.head)
  }
```

We can change the method signature in order to avoid doing the split before calling the rlp decoder you can call something like 

```scala
val decoded: Try[Seq[Int]] = RLP.decode[Seq[Int]](data, 3)
```

That being said, it's important to note that our implementation does not support decoding arrays made like `encode(msg1) ++ encode(anotherMessage)` because those are not items of a single `RLPList`. Does it make sense?

LukasGasior1:
@AlanVerbner Sure it does. Didn't know it already works that way :) No need to change the signature, `data.drop(index)` is sufficient. And how about `RLP.getNextElementIndex(buffer, pos)`?
https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/net/rlpx/FrameCodec.java#L195

AlanVerbner:
@LukasGasior1 `RLP.getNextElementIndex(buffer, pos)` it's not supported atm but easily doable. We are trying to find a way to avoid repeating that code (in EthereumJ is at least done 3 times if i'm not mistaken)...will get back to you that coded :smile:

LukasGasior1:
Thanks :)

AlanVerbner:
@LukasGasior1 Sorry for the delay but tests were messing around with me. I've just pushed `RLP.getNextElementIndex(buffer, pos)`. There are several tests starting from [here](https://github.com/input-output-hk/etc-client/pull/2/commits/6ff6006b2091ac770d1c9977d5a1da4d36d5631c#diff-7e1721917071c960a9a335644a0ef45fR498)

Please check if it fits networking needs. Also let me know if you want me to create tests for encoding / decoding handshake massages.

LukasGasior1:
Thanks, LGTM :+1:

mcsherrylabs:
@AlanVerbner Can you if you didn't already add tests for the 2 fixes you mention? Ta!

AlanVerbner:
@mcsherrylabs I've already did but will add a couple more

src/main/scala/io/iohk/ethereum/utils/RLP.scala
```diff
@@ -0,0 +1,276 @@
...
+  def apply(enc: RLPEncodeable*): RLPList = new RLPList {
+    override def items: Seq[RLPEncodeable] = enc.toSeq
+  }
+
+  def apply[E](seq: Seq[E])(implicit convert: E => RLPEncodeable): RLPList = new RLPList {
+    override def items: Seq[RLPEncodeable] = seq.map(convert)
+  }
+}
+
+trait RLPValue extends RLPEncodeable {
```
AlanVerbner: @LukasGasior1 let me see. I think it was left after a refactor and might be changed to case classes as you say.

PR #3, opened 2017-01-02
Auth and proto handshake

LukasGasior1:
Brief overview of design choices (to discuss/change, of course):
- AuthHandshaker is immutable and extends the interface that makes it usable inside Scorex (see my approach here: https://github.com/input-output-hk/scorex-etc/pull/2)
- it creates a **mutable** Secrets instance (that also extends new Scorex interface to help make it usable there)
   the class is mutable, because KeccakDigests are (egressMac, ingressMac) I'm not sure if there's anything we can do about it.
- same goes for FrameCodec, ideally this could be immutable case class, perhaps similar in design to AuthHandshaker, 
  but because of mutable Secrets it also is mutable.
- since I'm still working on getting everything running with Scorex, I've created a simple socket-based "Handshaker" app: TestSocketHandshaker


src/main/scala/io/iohk/ethereum/crypto/package.scala
```diff
@@ -0,0 +1,45 @@
...
+import fr.cryptohash.Keccak256
+import org.spongycastle.asn1.sec.SECNamedCurves
+import org.spongycastle.crypto.AsymmetricCipherKeyPair
+import org.spongycastle.crypto.generators.ECKeyPairGenerator
+import org.spongycastle.crypto.params.{ECKeyGenerationParameters, ECDomainParameters}
+
+package object crypto {
+
+  val curveParams = SECNamedCurves.getByName("secp256k1")
+  val curve = new ECDomainParameters(curveParams.getCurve, curveParams.getG, curveParams.getN, curveParams.getH)
```
LukasGasior1: as these are constants I should probably change them back to PascalCase

mcsherrylabs: You can leave them.

src/main/scala/io/iohk/ethereum/network/AuthInitiateMessage.scala
```diff
@@ -0,0 +1,33 @@
...
+}
+
+case class AuthInitiateMessage(
+    signature: ECDSASignature,
+    ephemeralPublicHash: ByteString,
+    publicKey: ECPoint,
+    nonce: ByteString,
+    knownPeer: Boolean) {
+
+  def encode(): ByteString = {
```
mcsherrylabs: There are no side effects here? Could be a lazy val?

src/main/scala/io/iohk/ethereum/network/AuthResponseMessage.scala
```diff
@@ -0,0 +1,27 @@
...
+    AuthResponseMessage(
+      ephemeralPublicKey = curve.getCurve.decodePoint(Array(4.toByte) ++ input.take(64)),
+      nonce = ByteString(input.drop(64).take(32)),
+      knownPeer = input(96) == 1)
+  }
+}
+
+case class AuthResponseMessage(ephemeralPublicKey: ECPoint, nonce: ByteString, knownPeer: Boolean) {
+
+  def encode(): ByteString = {
```
mcsherrylabs: Also lazy val...?

src/main/scala/io/iohk/ethereum/network/package.scala
```diff
@@ -0,0 +1,20 @@
+package io.iohk.ethereum
+
+import io.iohk.ethereum.crypto._
+import org.spongycastle.crypto.params.{AsymmetricKeyParameter, ECPublicKeyParameters}
+import org.spongycastle.util.encoders.Hex
+
+package object network {
+
+  def nodeIdFromPublicKey(pubKey: AsymmetricKeyParameter) = {
```
mcsherrylabs: What about an implicit class extends AnyVal { def toNodeId: String } ?
Same for nodeId to public key?

src/test/scala/io/iohk/ethereum/network/p2p/SecureChannelSetup.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  val (initPacket, handshakerInitiated) = handshaker.initiate(remoteUri)
+  val (responsePacket, AuthHandshakeSuccess(remoteSecrets: Secrets)) = remoteHandshaker.handleInitialMessage(initPacket)
+  val AuthHandshakeSuccess(secrets: Secrets) = handshakerInitiated.handleResponseMessage(responsePacket)
+
+  def randomNonce() = {
+    val arr = new Array[Byte](AuthHandshaker.NonceSize)
+    new SecureRandom().nextBytes(arr)
+    ByteString(arr)
+  }
```
mcsherrylabs: Generating random bytes happens a lot, can you add a util class that provides arrays of bytes using secureRandom  and regular Random? Such that we can do 

val mySecureBytes: Array[Byte] = SecureSeedBytes(sizeOfArray) and 
val myBytes: Array[Byte] = SeedBytes(sizeOfArray) 

(I think Scorex does this)

src/main/scala/io/iohk/ethereum/network/AuthInitiateMessage.scala
LukasGasior1: could be, but I don't see much improvement in such change, do you?

mcsherrylabs: By convention methods with no side effects and no params should be declared without '()'. When you see a method with () and no params you are not free to replace it with a val as it may have side effects. In this case, there are none, so it should be replaced with a lazy val so that the effort of calculating it is never repeated unnecessarily.

src/main/scala/io/iohk/ethereum/network/AuthResponseMessage.scala
adamsmo: as far as I know lazy val have disadvantage of synchronized initialization

mcsherrylabs: Are you recommending using a 'val' here instead Adam? Or to never use lazy val?

adamsmo: I mentioned disadvantage of `lazy val` to take it into consideration
for me `val` fits better here because calculations are not heavy and we will probably need encoded version anyway

LukasGasior1: True, lazy has some disadvantages. I think we can go with whatever here, it really doesn't matter. I'll make it `lazy val encoded`.

mcsherrylabs: Yes it matters Lukasz, keep an eye out for opportunities to take advantage of immutability.

LukasGasior1: Out of these approaches lazy val is the least "immutable", at least technically cause practically they're all on the same level ;)

src/main/scala/io/iohk/ethereum/network/p2p/FrameCodec.scala
```diff
@@ -0,0 +1,186 @@
...
+
+          val frameSize = totalSizeToRead - macSize
+          secrets.ingressMac.update(buffer, 0, frameSize)
+          dec.processBytes(buffer, 0, frameSize, buffer, 0)
+
+          val `type` = RLP.decode[Int](buffer)
+
+          val pos = RLP.nextElementIndex(buffer, 0)
+          val payload = buffer.drop(pos).take(header.bodySize - pos)
+          val size = header.bodySize - pos
```
AlanVerbner: is this variable used?

src/main/scala/io/iohk/ethereum/network/p2p/FrameCodec.scala
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  private var unprocessedData: ByteString = ByteString.empty
+
+  private var headerOpt: Option[Header] = None
+
+  def readFrames(data: ByteString): Seq[Frame] = {
+    unprocessedData ++= data
+
+    def readRecursive(): Seq[Frame] = {
```
ntallar: @LukasGasior1 What about transforming into this function into tail recursive and using @tailrec annotation? It can be rapidly done using an accumulator as a parameter.

src/main/scala/io/iohk/ethereum/network/AuthHandshaker.scala
```diff
@@ -0,0 +1,204 @@
...
+      initiateMessageOpt = Some(message),
+      initiatePacketOpt = Some(data),
+      responseMessageOpt = Some(response),
+      responsePacketOpt = Some(encryptedPacket)).finalizeHandshake()
+
+    (encryptedPacket, handshakeResult)
+  }
+
+  private def createAuthInitiateMessage(remotePubKey: ECPoint) = {
+    // TODO: handle the case when the peer is known
```
LukasGasior1: Yes. The logic is slightly different when the peer is known (session tokens must be stored somewhere).

src/main/scala/io/iohk/ethereum/network/AuthHandshaker.scala
```diff
@@ -0,0 +1,204 @@
...
+        mac = sha3(agreedSecret, aesSecret),
+        token = sha3(sharedSecret),
+        egressMac = egressMacSecret,
+        ingressMac = ingressMacSecret))
+    }
+
+    successOpt getOrElse AuthHandshakeError
+  }
+
+  private def bigIntegerToBytes(b: BigInteger, numBytes: Int): Array[Byte] = {
```
LukasGasior1: I'll add `ByteUtils` as they have in ethereumj

src/main/scala/io/iohk/ethereum/crypto/package.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  def xor(a: Array[Byte], b: Array[Byte]): Array[Byte] = {
```
LukasGasior1: moved to `ByteUtils`

mcsherrylabs:
LGTM. (Didn't see the util for generating secure/random bytes? If it's not there, not to worry, we'll get it in next time)

LukasGasior1:
@mcsherrylabs forgot to push :) Added these methods to ByteUtils

PR #4, opened 2017-01-04
RLP Refactor

AlanVerbner:
## Description

The following changes were made to RLP code:

- RLPValue and RLPList traits replaced by case classes (breaking)
- Match cases replaced by if statements if no unboxing
- all exceptions replaced by RuntimeExceptions
- ByteString repalced by Array[Byte]
- Prevent Long values to be decoded if they have more than 8 Bytes (breaking - low impact)
- Tests and Scoverage plugin added
- RLPEncoding package object created and code moved into it (breaking)
- RPL decoding of an empty array now results in an exception (breaking - low impact)

### DIFF
--- a/src/main/scala/io/iohk/ethereum/rlp/package.scala
+++ b/src/main/scala/io/iohk/ethereum/rlp/package.scala
@@ -1,0 +1,21 @@
+package io.iohk.ethereum.rlp
+
+package object RLPEncoding {
+  val Implicits = RLPImplicits
+
+  def encode[T](input: T)(implicit enc: RLPEncoder[T]): Array[Byte] = RLP.encode(enc.encode(input))
+
+  def decode[T](data: Array[Byte])(implicit dec: RLPDecoder[T]): T = dec.decode(RLP.rawDecode(data))
+
+  /**
+    * This function calculates the next element item based on a previos element starting position. It's meant to be
+    * used while decoding a stream of RLPEncoded Items.
+    *
+    * @param data Data with encoded items
+    * @param pos  Where to start. This value should be a valid start element position in order to be able to calculate
+    *             next one
+    * @return Next item position
+    * @throws RuntimeException if there is any error
+    */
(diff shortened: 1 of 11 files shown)


mcsherrylabs:
Hi Alan, I made a few changes, just trying to get the API into the package object. Please have a look, and if I haven't broken anything, LGTM...

AlanVerbner:
@mcsherrylabs reviewing now

AlanVerbner:
@mcsherrylabs LGTM thanks! :smile:

src/main/scala/io/iohk/ethereum/rlp/package.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  trait RLPEncoder[T] {
+    def encode(obj: T): RLPEncodeable
+  }
+
+  trait RLPDecoder[T] {
+    def decode(rlp: RLPEncodeable): T
+  }
+
+  def encode[T](input: T)(implicit enc: RLPEncoder[T]): Array[Byte] = RLP.encode(enc.encode(input))
```
LukasGasior1: why do we need these methods? Can't we just have `def encode[T](input: T)(implicit enc: RLPEncoder[T])` in `RLP` obj?

AlanVerbner: @LukasGasior1 the goal is to have all public RLP api in `rlp` package object, does that make sense?

LukasGasior1: I don't know why that's the goal, but it's not a big deal anyway

AlanVerbner: @mcsherrylabs :arrow_up:

mcsherrylabs: @lukasGasior1, in this case the API for RLP is small enough to be declared in the package class. That helps clients of the module understand what's available at a glance. When clients import the package they have all the functionality available bar the supporting implicits.

PR #5, opened 2017-01-09
Blockchain download

adamsmo:
This PR is for asking for blocks to download and than print out messages that contains blocks


src/main/scala/io/iohk/ethereum/network/p2p/Message.scala
```diff
@@ -60,44 +64,427 @@ object Hello {
...
+        GetBlockHeaders(Right(ByteString(block: Array[Byte])), maxHeaders, skip, reverse)
+
+      case _ => throw new RuntimeException("Cannot decode GetBlockHeaders")
+    }
+  }
+
+  val code: Int = 0x10 + 0x03
+}
+
+case class GetBlockHeaders(block: Either[BigInt, ByteString], maxHeaders: BigInt, skip: BigInt, reverse: Int) extends Message {
```
LukasGasior1: is it `Either[BigInt, ByteString]` because you want to give more options to the caller? Maybe we can just stick to one of these?

src/main/scala/io/iohk/ethereum/network/p2p/Message.scala
```diff
@@ -60,44 +64,427 @@ object Hello {
...
+          ByteString(mixHash: Array[Byte]),
+          ByteString(nonce: Array[Byte]))
+
+      case _ => throw new RuntimeException("Cannot decode BlockHeaders")
+    }
+  }
+}
+
+case class BlockHeader(
+  parentHash: ByteString,
```
LukasGasior1: as far as I remember the guide suggests to use 2 indentation levels here, i.e 4 spaces

src/main/scala/io/iohk/ethereum/network/p2p/Message.scala
```diff
@@ -60,44 +64,427 @@ object Hello {
...
+//ETH yellow paper section 4.3
+case class Transaction(
+  nonce: BigInt,
+  gasPrice: BigInt,
+  gasLimit: BigInt,
+  receivingAddress: ByteString,
+  value: BigInt,
+  payload: Either[ContractInit, TransactionData],
+  //yellow paper appendix F
+  pointSign: Byte, //v - 27 or 28 according to yellow paper, but it is 37 and 38 in ETH
```
LukasGasior1: does it make sense to use ECDSASignature for these fields?

adamsmo: it does, I think I will implement it in next iteration, because we have to address situation when sign is 37/38 instead of 27/28 which happens in some transactions

src/main/scala/io/iohk/ethereum/network/p2p/Message.scala
adamsmo: no we can not, there is explicit need to get blocks either by block hash or by block number

PR #6, opened 2017-01-10
Feature/patricia trie amcs

mcsherrylabs:
Just some example changes, no need to merge. (Also while the tests pass I did not check that the logic is *exactly* the same by stepping through.)


AlanVerbner:
thanks @mcsherrylabs ... logic might be improved but as far as the final hashes return the same values as other implementation (EthereumJ, parity, etc) we are good.

PR #7, opened 2017-01-11
Minor sundry code changes (class to case class, intro implicit value …

mcsherrylabs:
Tiny TINY changes that I noticed....if you could have a look.
@LukasGasior1 an implicit value class should avoid the instantiation cost
@adamsmo class to case class....

### DIFF
--- a/src/main/scala/io/iohk/ethereum/network/AuthHandshaker.scala
+++ b/src/main/scala/io/iohk/ethereum/network/AuthHandshaker.scala
@@ -15,12 +15,12 @@ import org.spongycastle.math.ec.ECPoint
 import scorex.core.network
 import scorex.core.network._
 
-class Secrets(
-    val aes: Array[Byte],
-    val mac: Array[Byte],
-    val token: Array[Byte],
-    val egressMac: KeccakDigest,
-    val ingressMac: KeccakDigest)
+case class Secrets(
+    aes: Array[Byte],
+    mac: Array[Byte],
+    token: Array[Byte],
+    egressMac: KeccakDigest,
+    ingressMac: KeccakDigest)
   extends scorex.core.network.Secrets
 
 object AuthHandshaker {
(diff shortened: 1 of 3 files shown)


src/main/scala/io/iohk/ethereum/network/AuthHandshaker.scala
```diff
@@ -15,12 +15,12 @@ import org.spongycastle.math.ec.ECPoint
 import scorex.core.network
 import scorex.core.network._
 
-class Secrets(
-    val aes: Array[Byte],
-    val mac: Array[Byte],
-    val token: Array[Byte],
-    val egressMac: KeccakDigest,
-    val ingressMac: KeccakDigest)
+case class Secrets(
```
rtkaczyk: I see a small problem with this. A `case class` implements `equals` and `hashCode` et. al. with the intent of storing and comparison of immutable data. However because this one consists of `Arrays` it will not work:

```
scala> case class A(arr: Array[Int])
defined class A

scala> A(Array(1)) == A(Array(1))
res0: Boolean = false
```

Maybe it would be better to leave it as a regular class so as not give a false impression?
Alternatively maybe those secrets could be stored within immutable structures like `ByteString`?

LukasGasior1: Yes, I made it regular class on purpose. Beside the Arrays, KeccakDigest is mutable (as I wrote in my previous PR description).

src/main/scala/io/iohk/ethereum/network/package.scala
```diff
@@ -6,7 +6,7 @@ import org.spongycastle.util.encoders.Hex
 
 package object network {
 
-  implicit class ECPublicKeyParametersNodeId(pubKey: ECPublicKeyParameters) {
+  implicit class ECPublicKeyParametersNodeId(val pubKey: ECPublicKeyParameters) extends AnyVal {
```
LukasGasior1: Good point :+1:

src/main/scala/io/iohk/ethereum/network/AuthHandshaker.scala
mcsherrylabs: Yes.As Adam says somewhere, he did it on purpose. Forget this one...

PR #8, opened 2017-01-11
enabling coveralls reports

rtkaczyk:


AlanVerbner:
@rtkaczyk this seems to be ok, isn't it? is there any missing work to be done?

rtkaczyk:
As far as coverage goes this is done. However I've also been using this branch to experiment with travis, it's not building again 😞

PR #9, opened 2017-01-11
Message versioning

adamsmo:
This PR contains proposition on how to split messages and how to support different versions of protocol


src/main/scala/io/iohk/ethereum/network/p2p/messages/CommonMessages.scala
```diff
@@ -0,0 +1,117 @@
...
+      }
+
+      override def decode(rlp: RLPEncodeable): Status = rlp match {
+        case RLPList(protocolVersion, networkId, totalDifficulty, bestHash, genesisHash) =>
+          Status(protocolVersion, networkId, totalDifficulty, ByteString(bestHash: Array[Byte]), ByteString(genesisHash: Array[Byte]))
+        case _ => throw new RuntimeException("Cannot decode Status")
+      }
+    }
+
+    val code: Int = 0x10 + 0x00
```
rtkaczyk: Why the addition?

LukasGasior1: messages 0x00-0x10 are for ÐΞVp2p messages. "regular" messages start from 0x10 (see: https://github.com/ethereum/wiki/wiki/%C3%90%CE%9EVp2p-Wire-Protocol)

adamsmo: to indicate that code is addition of 0x10 and sub-protocol message number which in this case is 0x00

rtkaczyk: ok, thanks

mcsherrylabs: Any reason we don't create a val - something like 'RegularMessageOffset'?

adamsmo: @mcsherrylabs because we have same super class for wire protocol and for sub protocol and sub wire protocol messages do not have this value
But I can add this to sub protocol messages

AlanVerbner:
@adamsmo code looks good so far (still looking at it). Besides that, is there any easy way to test older versions compatibility? Maybe we can encode with other clients old versions messages and make sure we can decode them. What do you think?

adamsmo:
@AlanVerbner encoding and decoding are same for all messages despite version, what we can do is try to decode for example NewBlockHashes with PV61 and PV62 since this message changed, I can add test for that, did not done it because it was only proposal

AlanVerbner:
@adamsmo yes...`NewBlockHashes` is the one that has changed so probably it's a good idea to have tests for different versions. Also `BlockHashesFromNumber` is PV61 only and it seems not to be implemented...do you think we should add it now or maybe when we focus on nodes sync?

adamsmo:
we can add this one now

src/main/scala/io/iohk/ethereum/network/p2p/messages/CommonMessages.scala
```diff
@@ -0,0 +1,117 @@
...
+          Transaction(nonce, gasPrice, gasLimit, ByteString(), value, Left(ContractInit(ByteString(payload: Array[Byte]))),
+            pointSign, ByteString(signatureRandom: Array[Byte]), ByteString(signature: Array[Byte]))
+
+        case _ => throw new RuntimeException("Cannot decode Transaction")
+      }
+    }
+  }
+
+  //ETH yellow paper section 4.3
+  case class Transaction(
```
AlanVerbner: @adamsmo I know it's a gray area but, does a Transaction belong to `CommonMessages`?

adamsmo: they are used by all version of protocol and were not redefined in any version so I assumed that they are common to all versions of protocol

AlanVerbner: I'm saying that those are maybe entities not only related to messages. For example `Block` is not strictly a message and it's composed by BlockHeader and BlockBody. For example, in `EthereumJ` is located under `org.ethereum.core;`

adamsmo: ok, I get it now
I think that this case class `Transaction` belongs to messages because it is used to decode messages about transactions received from other nodes

adamsmo:
@AlanVerbner / @mcsherrylabs 
I pushed unit tests and added constant for `0x10`

PR #10, opened 2017-01-11
Merkle Patricia Trie

AlanVerbner:
## Description

This branch has the MPT implementation. It also has a DataSource implementation based on [IODB](https://github.com/input-output-hk/iodb).

We tried to have as most coverage as we could for this code because of it's critical nature so please keep special attention at it while reviewing.

## Design choices

- There is no any defined hash function to be used, it can be configured. The only restriction there is that it should match the iodb's `LSMStore` `keySize` [config](https://github.com/input-output-hk/iodb/blob/master/src/main/scala/io/iohk/iodb/LSMStore.scala#L32). 
- Ideally `MerklePatriciaTrie` class could be immutable but as long as it needs to access a mutable `DataSource` it's also mutable.
- In order to be used with key/value pairs of different types we have created a `ByteArraySerializable` that will be responsible for their byte array encoding. There is no specific guideline regarding how that should be done as long as the merkle root returns the same hash


src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
```diff
@@ -0,0 +1,487 @@
...
+      case extension: ExtensionNode =>
+        RLPList(HexPrefix.encode(nibbles = extension.sharedKey, t = false), extension.next match {
+          case Right(node) => this.encode(node)
+          case Left(bytes) => bytes
+        })
+      case branch: BranchNode =>
+        val toEncode: Seq[RLPEncodeable] = branch.children.map {
+          case Some(Right(node)) => this.encode(node)
+          case Some(Left(bytes)) => RLPValue(bytes)
+          case _ => RLPValue(Array.emptyByteArray)
```
adamsmo: minor: replace `_` with `None`?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/Node.scala
```diff
@@ -0,0 +1,102 @@
...
+import io.iohk.ethereum.merklepatriciatrie.MerklePatriciaTrie.HashFn
+
+/**
+  * Trie elements
+  */
+private[merklepatriciatrie] sealed trait Node {
+
+  import MerklePatriciaTrie._
+
+  private var hashedCache: Option[Array[Byte]] = None
```
adamsmo: what is the purpose of this hash?

AlanVerbner: @adamsmo when you are working with a node you might need to access several times to it's hash. Either to store the reference inside another node or to use it as a key to store it in the DataSource.

adamsmo: so the hash changes?

AlanVerbner: Hash does not change but it might be used several times within a single put / remove operation, that's why the its being _cached_.

Hash function is defined within the Trie so you cannot change the hashFn. By cannot I mean that if you do so, your tree will be unusable because you won't be able to access nodes stored in DataSource (as you won't be able to fetch nodes).
MPT uses Keccak256 but it might be useful to use it with another hashFn. Maybe the generalization makes everything more complex and unclear and we should remove hashFn as a trie parameter.

About thread safety, yes, I understand your point (we cannot be sure this optimization will be used as it is) . Maybe a `lazy val` makes more sense but as long as `hashFn` is a parameter we cannot do so.

adamsmo: if `hashFn` should not change maybe we can add it as `val` in `Node` trait?
this way it should be easier to introduce lazy val

src/main/scala/io/iohk/ethereum/merklepatriciatrie/HexPrefix.scala
```diff
@@ -0,0 +1,64 @@
...
+  /**
+    * Transforms an array in hexadecimal format to a byte array
+    *
+    * @param nibbles byte[]
+    * @return array with bytes combining pairs of nibbles
+    *
+    */
+  def nibblesToBytes(nibbles: Array[Byte]): Array[Byte] = {
+    require(nibbles.length % 2 == 0)
+    (0 until nibbles.length/2).map(i => (16*nibbles(2*i) + nibbles(2*i + 1)).toByte).toArray
```
ntallar: I agree that I would be better not having to play with indexes. While looking at your suggestion I found that doing ´a.grouped(2).map{case Array(a,b) => (16*a+b).toByte}.toArray´ would also let me get the pairs of consecutive elements without the pairs like ´(2,3)´ that wouldn't be used, what do you think of this option?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/Node.scala
AlanVerbner: @adamsmo let me see if I can work on that but i'm not sure if it might work as nodes are also created while decoding (and decoder is used as an implicit created within companion object).

Also, let me rephrase. `Node` objects are only created while doing a put or remove operations. They are not part of the Trie state (at least in the form of `Node`) nor shared between operations. This might (and i'm not sure if it will) be a problem if we decide by any reason keep cached in-memory versions of MPT (for example to speedup blocks processing). 
In terms of thread safety, we may need not to allow multiple `put` and `remove` operations from different threads, it's not about nodes. Does this make sense?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/HexPrefix.scala
```diff
@@ -0,0 +1,64 @@
...
+
+object HexPrefix {
+  /**
+    * Pack nibbles to binary
+    *
+    * @param nibbles sequence
+    * @return hex-encoded byte array
+    *
+    */
+  def encode(nibbles: Array[Byte], t: Boolean): Array[Byte] = {
```
ntallar: Fixed!

src/main/scala/io/iohk/ethereum/merklepatriciatrie/HexPrefix.scala
```diff
@@ -0,0 +1,64 @@
...
+    nibblesWithFlag(0) = firstByteFlag
+    if(!hasOddLength) nibblesWithFlag(1) = 0
+    nibblesToBytes(nibblesWithFlag)
+  }
+
+  /**
+    * Unpack a binary string to its nibbles equivalent
+    *
+    * @param src of binary data
+    * @return array of nibbles in byte-format
```
ntallar: Fixed, added description on function comment.

src/main/scala/io/iohk/ethereum/merklepatriciatrie/HexPrefix.scala
```diff
@@ -0,0 +1,64 @@
...
+    val hasOddLength = (srcNibbles(0) & 1) != 0
+    val flagLength = if(hasOddLength) 1 else 2
+
+    val res = new Array[Byte](srcNibbles.length - flagLength)
+    Array.copy(srcNibbles, flagLength, res, 0, srcNibbles.length-flagLength)
+    (res, t)
+  }
+
+  /**
+    * Transforms a binary array to hexadecimal format
```
ntallar: You are right, fixed the description to include that.

src/main/scala/io/iohk/ethereum/merklepatriciatrie/HexPrefix.scala
```diff
@@ -0,0 +1,64 @@
...
+    * @return array with each individual nibble
+    *
+    */
+  def bytesToNibbles(bytes: Array[Byte]): Array[Byte] = {
+    bytes.foldRight[List[Byte]](List()){(elem, rec) =>
+      ((elem >> 4) & 0xF).toByte :: (elem & 0xF).toByte :: rec}.toArray
+  }
+
+  /**
+    * Transforms an array in hexadecimal format to a byte array
```
ntallar: You are also right, fixed the description to include that!

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
```diff
@@ -0,0 +1,487 @@
...
+  def apply[K, V](rootHash: Array[Byte], source: DataSource, hashFn: HashFn)
+                 (implicit kSerializer: ByteArraySerializable[K], vSerializer: ByteArraySerializable[V])
+  : MerklePatriciaTrie[K, V] = new MerklePatriciaTrie[K, V](Some(rootHash), source, hashFn)(kSerializer, vSerializer)
+
+  private def getNode(nodeId: Array[Byte], source: DataSource)(implicit nodeDec: RLPDecoder[Node]): Node = {
+    val nodeEncoded = tryGetNode(nodeId, source)
+    decodeRLP[Node](nodeEncoded)
+  }
+
+  private def tryGetNode(key: Array[Byte], source: DataSource): Array[Byte] =
```
AlanVerbner: @adamsmo i think this is a good point and should be changed. It originally returned a try and we left it as it is.

src/main/scala/io/iohk/ethereum/merklepatriciatrie/HexPrefix.scala
adamsmo: great, even better solution 👍

src/main/scala/io/iohk/ethereum/merklepatriciatrie/Node.scala
adamsmo: maybe solution that I proposed will not work her, my main concern here is that we have method and var for something that may be replaced by `lazy val`

about case of decoding, maybe we can have 
```
trait T {val a: Option[hashFn[}
case class Foo(a: Option[hashFn]) extends T
```
? (Just another idea that may work)

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
ntallar: Fixed! Removed tryGetNode function.

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
```diff
@@ -0,0 +1,487 @@
...
+    if (key.length < 32) key
+    else source.get(key).getOrElse(throw MPTException("Node not found, trie is inconsistent"))
+
+  private def matchingLength(bytes1: Array[Byte], bytes2: Array[Byte]): Int = bytes1.zip(bytes2).takeWhile(t => t._1 == t._2).length
```
ntallar: I agree, I changed parameters names according to your suggestion.

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
ntallar: Fixed!

src/main/scala/io/iohk/ethereum/merklepatriciatrie/Node.scala
AlanVerbner: @adamsmo let me try to work around this idea to see if I can come with a clean solution

AlanVerbner: @adamsmo i've just pushed to a test branch how a new approach about this might work. Here https://github.com/input-output-hk/etc-client/compare/feature/patriciaTrie...feature/patriciaTrieAdamChangeSuggestion you can find the branch comparison....please let me know your thoughts about this.

adamsmo: @AlanVerbner thanks!
why do we need `enc: RLPEncoder[Node]` in Nodes ?

adamsmo: @AlanVerbner thank for update, this looks better for me because we do not have var and we have one place where we access node hash

AlanVerbner: @adamsmo great, will merge it here :smile: thanks!!

adamsmo:
@AlanVerbner we have `MerklePatriciaTrie.scala` which has ~500 lines of code and methods like `private def put` which have ~120 , will this cause issue with scalastyle?

AlanVerbner:
@adamsmo I think it's better once scalastryle is merged (and style defined) we will run it on this branch

build.sbt
```diff
@@ -8,5 +8,6 @@ libraryDependencies ++= Seq(
   "org.scorexfoundation" %% "scorex-core" % "2.0.0-M3-feature-handshake",
   "com.madgag.spongycastle" % "core" % "1.53.0.0",
   "org.scalatest" %% "scalatest" % "3.0.1" % "test",
-  "org.scalacheck" %% "scalacheck" % "1.13.4" % "test"
-)
+  "org.scalacheck" %% "scalacheck" % "1.13.4" % "test",
+  "org.scorexfoundation" %% "iodb" % "0.+"
```
adamsmo: I think exact version is better

AlanVerbner: sure, will change it now @adamsmo

AlanVerbner: @adamsmo btw, this is how Scorex Example projects included it...maybe it's because they won't break the API but will release fixes....i can fix the version though

AlanVerbner: @adamsmo done!

adamsmo: thx

rtkaczyk:
So I still don't understand why the tree has to depend on a `DataSource`. Why can't we have the `DataSource` and `MerklePatriciaTree` be implemented independently, and some class like `Storage` that will utilise both the DataSource and the MPT to provide persistence - **separation of concerns**.

src/main/scala/io/iohk/ethereum/merklepatriciatrie/Node.scala
```diff
@@ -0,0 +1,98 @@
+package io.iohk.ethereum.merklepatriciatrie
```
rtkaczyk: Maybe the package could renamed to `mpt`? Especially that this name is referred to in so many places.

AlanVerbner:
@rtkaczyk how will you handle value retrieve in that case?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
```diff
@@ -0,0 +1,486 @@
...
+
+  implicit val defaultByteArraySerializable = new ByteArraySerializable[Array[Byte]] {
+    override def toBytes(input: Array[Byte]): Array[Byte] = input
+
+    override def fromBytes(bytes: Array[Byte]): Array[Byte] = bytes
+  }
+}
+
+class MerklePatriciaTrie[K, V](private val rootHash: Option[Array[Byte]],
+                               val dataSource: DataSource,
```
rtkaczyk: So I still don't understand why the tree has to depend on a DataSource. Why can't we have the DataSource and MerklePatriciaTree be implemented independently, and some class like Storage that will utilise both the DataSource and the MPT to provide persistence - **separation of concerns**.

AlanVerbner: @rtkaczyk If you do so you will need to do the operations against the storage, are you thinking about that? You cannot traverse the tree without any `DataSource`

src/main/scala/io/iohk/ethereum/merklepatriciatrie/Node.scala
AlanVerbner: @rtkaczyk yes we can! I guess it's not too much effort to do so

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
```diff
@@ -0,0 +1,486 @@
...
+      else None
+    case branch@BranchNode(children, terminator, _) =>
+      if (searchKey.isEmpty) terminator
+      else getChild(branch, searchKey(0), dataSource) match {
+        case Some(child) => get(child, searchKey.slice(1, searchKey.length))
+        case None => None
+      }
+  }
+
+  private def put(node: Node, searchKey: Array[Byte], value: Array[Byte]): NodeInsertResult = {
```
rtkaczyk: Looks quite complicated. I've been comparing with https://github.com/ethereum/pyethereum/blob/develop/ethereum/trie.py#L972-L992 implementation and that one looks more readable because it's split into several methods, rather than 1 uber-method.

Also, some vertical spacing might help improve readability - and this pertains to all of this file.

AlanVerbner: @rtkaczyk yes, we can split this into several shorter methods (our version looks similar to parity one https://github.com/ethcore/parity/blob/master/util/src/trie/triedbmut.rs#L429)..

Do you want us to do the changes now or do you prefer to finish reviewing?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
rtkaczyk: Yeah, I was thinking that the `Storage` would fetch the nodes by the hashes, building along an immutable tree structure. But the question is do we even need such a structure or is just about getting/putting values?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
rtkaczyk: Whatever suits you best. I do want give it a bit more review, but you can go on with the changes if you like.

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
AlanVerbner: @rtkaczyk No, you don't need that structure atm. But you will need a way to access to the db that stores the nodes within the Trie. Not only while getting a value but when inserting / removing (besides navigating in order to fix nodes in case a remove leaves an inconsistent state). So, what's your proposal about how to let the trie know about other nodes (expressed in terms of `Storage`, `MPT` and `DataSource`)?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
AlanVerbner: @rtkaczyk will work on this now, it should be easier to read after that so I guess it won't be a problem

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
```diff
@@ -0,0 +1,486 @@
...
+            val temporalExtensionNode = ExtensionNode(sharedKeySuffix, next, hashFn)
+            val NodeInsertResult(newBranchNode: BranchNode, toDeleteFromStorage, toUpdateInStorage) = put(temporalExtensionNode, searchKey.drop(ml), value)
+            val newExtNode = ExtensionNode(sharedKeyPrefix, newBranchNode, hashFn)
+            NodeInsertResult(
+              newNode = newExtNode,
+              toDeleteFromStorage = node +: toDeleteFromStorage,
+              toUpdateInStorage = newExtNode +: toUpdateInStorage
+            )
+        }
+      case branchNode@BranchNode(children, optStoredValue, _) =>
```
adamsmo: it looks like `optStoredValue` is never used

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
```diff
@@ -0,0 +1,486 @@
...
+    }
+  }
+
+  private def remove(node: Node, searchKey: Array[Byte]): NodeRemoveResult = (node, searchKey.isEmpty) match {
+    // They key matches a branch node but it's value doesn't match the key
+    case (BranchNode(children, None, _), true) => NodeRemoveResult(hasChanged = false, maybeNewChild = None)
+    // We want to delete Branch node value
+    case (BranchNode(children, _, _), true) =>
+      // We need to remove old node and fix it because we removed the value
+      val fixedNode = fix(BranchNode(children, None, hashFn), dataSource, Seq(), hashFn)
```
adamsmo: why we are passing `hashFn` twice here?

AlanVerbner: @adamsmo we will check it, maybe it's an error. Thx for pointing it out!

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
```diff
@@ -0,0 +1,486 @@
...
+          }
+        }
+    }
+  }
+
+  private def remove(node: Node, searchKey: Array[Byte]): NodeRemoveResult = (node, searchKey.isEmpty) match {
+    // They key matches a branch node but it's value doesn't match the key
+    case (BranchNode(children, None, _), true) => NodeRemoveResult(hasChanged = false, maybeNewChild = None)
+    // We want to delete Branch node value
+    case (BranchNode(children, _, _), true) =>
```
adamsmo: we want to remove BranchNode that have children, and to do so we call `fix` method that fixes `Branch node where there is only a single entry;` 
what `fix()` is doing here?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
```diff
@@ -0,0 +1,486 @@
...
+              toUpdateInStorage = Seq(newLeafNode, newBranchNode)
+            )
+          }
+        }
+    }
+  }
+
+  private def remove(node: Node, searchKey: Array[Byte]): NodeRemoveResult = (node, searchKey.isEmpty) match {
+    // They key matches a branch node but it's value doesn't match the key
+    case (BranchNode(children, None, _), true) => NodeRemoveResult(hasChanged = false, maybeNewChild = None)
```
adamsmo: `children` is not used

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
```diff
@@ -0,0 +1,486 @@
...
+        // Child not found in this branch node, so key is not present
+        NodeRemoveResult(hasChanged = false, maybeNewChild = None)
+      }
+    case (LeafNode(existingKey, storedValue, _), _) =>
+      if (existingKey sameElements searchKey) {
+        // We found the node to delete
+        NodeRemoveResult(hasChanged = true, maybeNewChild = None, toDeleteFromStorage = Seq(node))
+      }
+      else NodeRemoveResult(hasChanged = false, maybeNewChild = None)
+    case (extensionNode@ExtensionNode(sharedKey, next, _), _) =>
```
adamsmo: `next` is unused

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
```diff
@@ -0,0 +1,486 @@
...
+              hasChanged = false,
+              maybeNewChild = None,
+              toDeleteFromStorage = nodesToRemoveFromStorage,
+              toUpdateInStorage = nodesToUpdateInStorage)
+        }
+      } getOrElse {
+        // Child not found in this branch node, so key is not present
+        NodeRemoveResult(hasChanged = false, maybeNewChild = None)
+      }
+    case (LeafNode(existingKey, storedValue, _), _) =>
```
adamsmo: `storedValue` is unused

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
rtkaczyk: I see. If we don't really need the structure, then I guess this discussion is irrelevant. But I think my concerns stemmed from the name `MerklePatriciaTree` which suggests a logical data structure. Looking at parity impl I can see they named it `TrieDBMut`. Not sure what the `Mut` is for, but I like the `DB` part. What do you think about renaming?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
AlanVerbner: @adamsmo as in that case we are removing branch node value  `BranchNode(children, None, hashFn)` (check the _None_ as second parameter) you need to check if you have a single node within your 16 positions, if so, that's an invalid state (Branch node where there is only a single entry). Check out this case https://github.com/input-output-hk/etc-client/blob/feature/patriciaTrie/src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala#L450

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
```diff
@@ -0,0 +1,486 @@
...
+      }
+      (usedIndexes, optStoredValue) match {
+        case (Nil, None) => throw MPTException("Branch with no subvalues")
+        case (index :: Nil, None) =>
+          val temporalExtNode = ExtensionNode(Array[Byte](index.toByte), children(index).get, hashFn)
+          fix(temporalExtNode, dataSource, notStoredYet, hashFn)
+        case (Nil, Some(value)) => LeafNode(Array.emptyByteArray, value, hashFn)
+        case _ => node
+      }
+    case extensionNode@ExtensionNode(sharedKey, next, _) =>
```
adamsmo: `next` is unused

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
```diff
@@ -0,0 +1,486 @@
...
+      newNode
+    case _ => node
+  }
+}
+
+private case class NodeInsertResult(newNode: Node,
+                                    toDeleteFromStorage: Seq[Node] = Seq(),
+                                    toUpdateInStorage: Seq[Node] = Seq())
+
+private case class NodeRemoveResult(hasChanged: Boolean, maybeNewChild: Option[Node],
```
rtkaczyk: `NodeInsertResult` and `NodeRemoveResult` are use internally in `MerklePatriciaTree`, right? So maybe it would better to place them in the companion object?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/package.scala
```diff
@@ -0,0 +1,34 @@
+package io.iohk.ethereum
+
+package object merklepatriciatrie {
+  trait ByteArraySerializable[T] {
+    def toBytes(input: T): Array[Byte]
+
+    def fromBytes(bytes: Array[Byte]): T
+  }
+
+  trait DataSource {
```
rtkaczyk: Why are those traits in a package object?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
AlanVerbner: I think of MPT as any other Trie that instead of using an in memory structure to store it's info might (or not) use a persistent storage so ok, we agreed not to create Storage, MPT and DataSource.

About naming. In EthereumJ, go-ethereum and python versions it's not named under DB. So i'm not sure. I don't mind renaming it though.

ps: lol, I don' t know either what' Mut stands for.

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
adamsmo: thanks for explanation,
so in this case `fix()` method does not fix (change) anything but is checking if node is in valid state?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
ntallar: Yes, they are only used for the output of the remove and put functions. I agree, I'm currently fixing it.

src/main/scala/io/iohk/ethereum/merklepatriciatrie/package.scala
AlanVerbner: @rtkaczyk Because they are part of the MPT public API

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
adamsmo: `Mut` => mutable?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
AlanVerbner: it fixes in terms that return a new valid node. The naming ispiration was taken from https://github.com/ethcore/parity/blob/master/util/src/trie/triedbmut.rs#L695

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
AlanVerbner: @adamsmo interesting theory :thinking:

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
ntallar: Fixed!

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
ntallar: Fixed!

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
ntallar: Fixed!

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
ntallar: Fixed!

src/main/scala/io/iohk/ethereum/merklepatriciatrie/package.scala
ntallar: Yes, we decided to put them there as they should be commonly implemented in order to use the trie (if any our implementations of them are not used). Do you think it would be better to put them elsewhere?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
AlanVerbner: @adamsmo because one is used within BranchNode's apply and the second one is used by fix fn when creating a `LeafNode`. An alternative is to make Node's hashFn public and take it from there, what do you think?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
ntallar: Fixed!

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
adamsmo: @AlanVerbner since it is `val` I think we can make it public

src/main/scala/io/iohk/ethereum/merklepatriciatrie/package.scala
rtkaczyk: Sure, but why `package object`? Package objects are usually used for things that cannot be placed at a top-level of a package, like vals and defs. But traits can :)

src/test/scala/io/iohk/ethereum/ObjectGenerators.scala
```diff
@@ -1,7 +1,7 @@
 package io.iohk.ethereum
 
 import java.math.BigInteger
-import org.scalacheck.{Arbitrary, Gen}
+import org.scalacheck.{Arbitrary, Gen, _}
```
rtkaczyk: This is equivalent to `import org.scalacheck._`. Isn't it?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/package.scala
AlanVerbner: @mcsherrylabs ^

src/test/scala/io/iohk/ethereum/ObjectGenerators.scala
ntallar: You are right. As we are only using `org.scalacheck.{Arbitrary, Gen}` I'll change it to that.

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
rtkaczyk: > In EthereumJ, go-ethereum and python versions it's not named under DB

Since it's well established that this Trie uses an external data source, then I guess I can live with the current name 😀

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
```diff
@@ -0,0 +1,485 @@
...
+    * @param key
+    * @return New trie with the (key-value) pair associated with the key passed deleted from the trie.
+    * @throws MPTException if there is any inconsistency in how the trie is build.
+    */
+  def remove(key: K): MerklePatriciaTrie[K, V] = {
+    rootHash map { rootId =>
+      val keyNibbles = HexPrefix.bytesToNibbles(bytes = kSerializer.toBytes(key))
+      val root = getNode(rootId, dataSource)
+      remove(root, keyNibbles) match {
+        case NodeRemoveResult(true, Some(newRoot), nodesToRemoveFromStorage, nodesToUpdateInStorage) =>
```
adamsmo: about second parameter it is called here `newRoot` but in case class this field is named `maybeNewChild`
https://github.com/input-output-hk/etc-client/pull/10/files#diff-2c1aab8e8ca156fbe430bbb586ecfc01R17
should it be that way?

ntallar: In `NodeInsertResult` and `NodeRemoveResult` we named it that way as it is more general, the `maybeNewChild` could be the an inner node of the trie, as is the case in the uses of both case classes in the put and remove function. Maybe we should rename `maybeNewChild` as `maybeNewNode` as the use of a the child word can be mixed with it meaning in the BranchNode.

adamsmo: maybe just `newNode`, as it is in `NodeInsertResult`?
I think `maybe` is implied by `Option` type
What do you think?

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
AlanVerbner: ok, so we are good then :metal:

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
ntallar: I agree, I'll change the variable's name to `newNode`.

src/main/scala/io/iohk/ethereum/merklepatriciatrie/Node.scala
AlanVerbner: fixed!

src/main/scala/io/iohk/ethereum/merklepatriciatrie/MerklePatriciaTrie.scala
ntallar: Do you think it will be clear enough to have the code split into 3 function, one per node type? Python's code takes advantage of working with lists and not node types to remove some duplicated code lines but that's not possible in our case.

rtkaczyk:
LGTM!

src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
```diff
@@ -0,0 +1,619 @@
...
+    val trieWithOneElement = emptyTrie.put(1, 5)
+    val obtained = trieWithOneElement.get(1)
+    assert(obtained.isDefined)
+    assert(obtained.get == 5)
+    val trieAfterDelete = trieWithOneElement.remove(1)
+    val obtainedAfterDelete = trieAfterDelete.get(1)
+    assert(obtainedAfterDelete.isEmpty)
+  }
+
+  ignore("IODB test - Insert of the first 5000 numbers hashed and then remove half of them"){
```
rtkaczyk: Why are these tests ignored?

ntallar: We set some of the IODB and EthereumJ Compatibility tests to be ignored so that they are not run every time, as they take a while to run.

rtkaczyk: I have several remarks about those tests:

1. If we're ignoring tests in such manner it deserves at least a comment. Otherwise an unknowing developer might come at those tests and wonder what he has to do to fix them.
2. For such long running tests I think a better solution would be to create them in a separate test configuration. Then in Circle CI we could easily run them conditionally with `sbt long-running:test`, e.g. only when merging to master.
  Another reason for a separate config is that, unless I'm missing something, they don't look like regular unit tests. Are they stress tests? Does running them with 40000 numbers give us a big boost of confidence as compared to running with 4 numbers?
  Lastly, if we don't run those tests automatically how are we going maintain them? Are we going to nominate a person who's going to run them occasionally to see if they're not broken? I certainly do not volunteer 😁 
3. I ran one of those tests: `"IODB Test - PatriciaTrie insert and get"` and it takes ~600 ms. I wouldn't say that's terribly slow.
4. Could we test `IodbDataSource` in a separate suite and without dependency on `MerklePatriciaTrie`? It is my understanding MPT depends on a DataSource, not vice versa. More importantly though, what if we find out that IODB doesn't meet our expectations? We shouldn't have to change MPT tests when we replace the DataSource implementation.

AlanVerbner: @rtkaczyk 

1. I do agree it's confusing. That being said, I don't like (at least in general terms) to push a broken test marked as `ignore`. All tests should be working or get fixed.
2. Lol. yes, we can create a separate config that's a good idea. The goal of some of them is to maintain certain level of compatibility against other implementations and allow others to compare our client against [ethereum wiki ones](https://github.com/ethereum/wiki/wiki/Benchmarks). 
3. No it's not too much atm, we can _uningnore_ it 
4. Yes we can, but, based on item (2) there is a benchmark test that requires a non ephemeral trie so it does make sense to use them there.

rtkaczyk: Good, then I would propose to name the configuration `benchmark`, and later we can configure Circle to run it when merging to master or by using [nightly builds](https://circleci.com/docs/nightly-builds/).

Regarding `IodbDataSource` it's fine if the tests utilise MPT when advantageous, especially in benchmark. However, a separate `IodbDataSourceSuite` testing `get/update` in isolation could be useful, even if it's really simple, wdyt?

AlanVerbner: @rtkaczyk Just pushed test partitioning. I used your the name you proposed at the beginning but it can be changed. You can compare it [here](https://github.com/input-output-hk/etc-client/compare/feature/patriciaTrie...feature/patriciaTrieTestPartitioning). If it's ok for you I will merge the changes to this branch.

What's is unclear to me is if you are proposing 3 types of tests:
1. Regular tests (the ones we already have)
2. Benchmark tests
3. Long running tests (which might not be for benchmarking purposes, just take some time to run)

Maybe we need to discuss this with the rest of the team and agree the best alternative for us.

About `IodbDataSource`..yes I do agree, I will add very simple tests to it in a separate file. My intention is not to test Iodb again but to be sure our `DataSource` impl doesn't break anything.

rtkaczyk: @AlanVerbner I guess using tags is OK, but `sbt test` will still run all the tests, won't it? I think that's slightly less convenient cause we'll have to filter them out explicitly.

I was actually thinking about a custom sbt configuration, like the built-in `it` (`sbt it:test`). That one is used for integration tests, so I though with a [custom configuration](http://stackoverflow.com/questions/18789477/define-custom-configuration-in-sbt) we could have a more appropriate name. If it's called `benchmark` then the tests reside in `src/benchmark/scala` and they're run with `sbt benchmark:test` (but I don't insist on that name).

At any rate, I would definitely like to see those tests in separate files. Especially in `MerklePatriciaTrieSuite.scala` I would like to have only the tests that use `HashMapDataSource`, so that we can replace the actual DB implementation more easily if needed.

rtkaczyk: Of course I'd be happy to see the rest of the team provide feedback here: @mcsherrylabs @adamsmo @whysoserious @LukasGasior1

AlanVerbner: @rtkaczyk no, `sbt test`won't run all tests because it's filtering long running ones by default `Tests.Argument("-l", "LongRunningTest")`. I really don't have a preferred way to do this (tags vs folders) so let's decide it together. I'm open to the one that we all like most and have no problem to update the code accordingly.

whysoserious: I fully agree with you Radek. Test which are not ran by CircleCI are just dead code.
Can we assume that in order to finish this PR CircleCI will have to successfully run an additional step - `sbt it:test`?

adamsmo: do we know how long it takes to run all tests?

AlanVerbner: @adamsmo in my computer (i7 8 cores 12gb ssd) the long running ones take between 70~90 sec

@whysoserious yes, we will definitely replace the `ignored` by some sort of test partitioning. I've created a branch using tags but @rtkaczyk proposed to use a custom configuration (benchmark for example). So lets try to define a way and I will code it during the following hour.

adamsmo: 90 sec does not looks scary to me ;)

whysoserious: > 90 sec does not looks scary to me ;)

So I'll have to wait 90 secs every time I save a file i a project ;)

whysoserious: @AlanVerbner let's use custom configs. Imho it's most convenient approach from user's point of view. Think of running `sbt it:test` in SBT console instead of dealing with `Tests.Argument("-l", "LongRunningTest")` settings.

AlanVerbner: @whysoserious deal. Will create a custom config called `benchmark` (as @rtkaczyk suggestion) and will move this tests over there.

rtkaczyk: > 90 sec does not looks scary to me ;)

It may be scary once we accumulate more of such tests. I wouldn't want to wait over an hour for a PR to go green.

> in order to finish this PR CircleCI will have to successfully run an additional step - sbt it:test

@whysoserious it looks that we need a proper `circle.yml` file first (I thought Circle was smart enough to pick up `travis.yml` but apparently it's just using some defaults for Scala). So I'd say that's a follow-up PR.

Anyway if those tests currently take just a few minutes then it's OK to require them for every PR, but _ultimately_ I'd like that we run them conditionally: either when merging to master or during nightly builds.

rtkaczyk:
One final remarks about the tests is that to putting those tests `it` config is not 100% accurate because they're not _integration tests_. I can imagine a scenario where we build a test net and test our client as a blackbox - that would constitute a true integration test (though using Hive is probably another valid option).

Anyway, for now I'm glad we have a separation of non-unit tests. Once we have more such tests and we can clearly categorise them we can think about further separation.

rtkaczyk:
LGTM 👍 

@adamsmo wanna give a second approval?

adamsmo:
LGTM as well 👍

AlanVerbner:
@rtkaczyk I definetly agree with your comment but what I do like most is

>  Once we have more such tests and we can clearly categorise them we can think about further separation.

PR #11, opened 2017-01-11
Request blockBodies was asking by parent hash instead of it's own hash

AlanVerbner:
## Description

When asking for block bodies, the request message was being created using each block parentHash. That was causing two problems: 
- 1st body corresponds to the parent of the first node (we won't be able to use it)
- It wasn't being requested the last header item body

## Solution

As done in EthereumJ, we are now sending the hash of each rlp encoded header returned by `GetBlockHeaders`


AlanVerbner:
@adamsmo I agree with you, will change it now

adamsmo:
@AlanVerbner I resolved conflicts with phase/2

AlanVerbner:
thanks @adamsmo .... is it ready to be merged?

adamsmo:
@AlanVerbner I think we can merge it

PR #12, opened 2017-01-12
Public key / address recovery

adamsmo:
I this PR I added versification of transaction signature, recovery of sender address
as of now we do not decided what to do with transactions that have sign set to 37/38
This implementation treats those transactions as invalid


src/main/scala/io/iohk/ethereum/network/TestSocketHandshaker.scala
```diff
@@ -64,7 +64,7 @@ object TestSocketHandshaker {
       msgs.collect {
         case m: Status =>
           sendMessage(m, frameCodec, out) //send same status message
-          sendMessage(GetBlockHeaders(Left(5000), maxHeaders = 20, skip = 0, reverse = 0), frameCodec, out) //ask for block further in chain to get some transactions
+          sendMessage(GetBlockHeaders(Left(1000000), maxHeaders = 1, skip = 0, reverse = 0), frameCodec, out) //ask for block further in chain to get some transactions
```
AlanVerbner: Do you want to change this or did it went through accidentally? it's not big deal though, just curious.

src/main/scala/io/iohk/ethereum/network/p2p/messages/CommonMessages.scala
```diff
@@ -96,6 +100,19 @@ object CommonMessages {
     signatureRandom: ByteString, //r
     signature: ByteString /*s*/) {
 
+    val bytesToSign: Array[Byte] = crypto.sha3(rlpEncode(RLPList(nonce, gasPrice, gasLimit,
```
AlanVerbner: As this seems to be an expensive operation, why not `lazy val`?

adamsmo: added

PR #13, opened 2017-01-12
Add basic actors structure (Akka IO)

LukasGasior1:
The actors that I've added are:
- ServerActor for handling incoming connections
- PeerManager for handling PeerActors
- PeerActor to handle communication with connected peer
- RLPxConnectionActor that handles underlying rlpx connection (separated out from PeerActor)

(this PR also includes Adam's versioning merged)

NOTE: this is still work in progress, creating PR to get early feedback on my approach


src/main/scala/io/iohk/ethereum/App.scala
```diff
@@ -0,0 +1,26 @@
...
+  def main(args: Array[String]): Unit = {
+
+    val actorSystem = ActorSystem("etc-client_system")
+
+    val peerManager = actorSystem.actorOf(PeerManagerActor.props(nodeKey))
+    val server = actorSystem.actorOf(ServerActor.props(peerManager))
+
+    server ! ServerActor.StartServer(new InetSocketAddress("127.0.0.1", 9076))
+
+    val peerUri = new URI("enode://8e5e9fcbd8e292de71052e6d075112a893bbc4993529098cd974474b4df14685dcfc4221d601a73fdcff398b4c6c4f5d8d7f1d8c06d23db9086d9c9796b9c671@192.168.3.154:30303")
```
AlanVerbner: `args(0)` ? :innocent:

src/main/scala/io/iohk/ethereum/network/AuthInitiateMessage.scala
```diff
@@ -6,6 +6,8 @@ import org.spongycastle.math.ec.ECPoint
 
 object AuthInitiateMessage {
 
+  val encodedLength = 65+32+64+32+1
+
   def decode(input: Array[Byte]) = {
     AuthInitiateMessage(
       signature = ECDSASignature.decode(input.take(65)),
```
AlanVerbner: maybe this `65` can be also taken as constant and be used in `encodedLength` ?

src/main/scala/io/iohk/ethereum/App.scala
LukasGasior1: Yeah :+1:

src/main/scala/io/iohk/ethereum/network/AuthInitiateMessage.scala
LukasGasior1: Good point, I could move it to `ECDSASignature.encodedLength`

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -0,0 +1,103 @@
...
+      case RLPxConnectionHandler.MessageReceived(message) => handleMessage(message)
+      case s: SendMessage[_] => sendMessage(s.message)(s.enc)
+    }
+
+    def sendMessage[M <: Message : RLPEncoder](message: M): Unit = {
+      rlpxConnection ! RLPxConnectionHandler.SendMessage(message)
+    }
+
+    def handleMessage(message: Message): Unit = message match {
+      case Ping() => sendMessage(Pong())
```
adamsmo: should we add Status automatic response as well?
also what about Disconnect message?

src/main/scala/io/iohk/ethereum/network/p2p/MessageCodec.scala
```diff
@@ -0,0 +1,38 @@
...
+    frames map { frame => Try(Message.decode(frame.`type`, frame.payload.toArray, Message.PV61)) }
+  }
+
+  def encodeMessage[M <: Message : RLPEncoder](message: M): ByteString = {
+    val encoded = rlp.encode(message)
+    val numFrames = Math.ceil(encoded.length / MaxFramePayloadSize.toDouble).toInt
+    val contextId = contextIdCounter.incrementAndGet()
+
+    val frames = (0 until numFrames) map { frameNo =>
+      val payload = encoded.drop(frameNo * MaxFramePayloadSize).take(MaxFramePayloadSize)
```
ntallar: Minor detail: isn't it more efficient to use the slice function instead of drop and take as with: `encoded.slice(frameNo * MaxFramePayloadSize, frameNo * MaxFramePayloadSize + MaxFramePayloadSize)`?

LukasGasior1: I don't know about efficiency, but for me drop and take is more readable in this case

ntallar: Regarding efficiency using drop and take requires two calls to slice. But I agree that using drop and take is more readable, so we can leave it as it is.

src/main/scala/io/iohk/ethereum/crypto/ECDSASignature.scala
```diff
@@ -10,8 +10,11 @@ import org.spongycastle.math.ec.{ECAlgorithms, ECPoint, ECCurve}
 import org.spongycastle.util.BigIntegers._
 
 object ECDSASignature {
+
   val RAndSLength = 32
```
adamsmo: what `RAndS` mean?

src/main/scala/io/iohk/ethereum/network/AuthInitiateMessage.scala
```diff
@@ -5,16 +5,16 @@ import io.iohk.ethereum.crypto._
 import org.spongycastle.math.ec.ECPoint
 
 object AuthInitiateMessage {
-  val SignatureSize = 65
+  val encodedLength = ECDSASignature.encodedLength + 32 + 64 + 32 + 1
```
adamsmo: maybe we can add names to those numbers?

src/main/scala/io/iohk/ethereum/network/AuthInitiateMessage.scala
```diff
@@ -5,16 +5,16 @@ import io.iohk.ethereum.crypto._
...
 
   def decode(input: Array[Byte]): AuthInitiateMessage = {
-    val publicKeyIndex = SignatureSize + 32
+    val publicKeyIndex = ECDSASignature.encodedLength + 32
```
adamsmo: same, here, it would be nicer to read if I know that 32 is for example package length or hash lengt

src/main/scala/io/iohk/ethereum/network/NodeInfo.scala
```diff
@@ -0,0 +1,11 @@
+package io.iohk.ethereum.network
+
+import java.net.InetSocketAddress
+
+import org.spongycastle.crypto.AsymmetricCipherKeyPair
+import org.spongycastle.crypto.params.ECPublicKeyParameters
+
+// TODO: what if listen port changes? or server is stopped?
```
adamsmo: should this to do be addressed?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -0,0 +1,116 @@
...
+import io.iohk.ethereum.network.RLPxConnectionHandler.MessageReceived
+import io.iohk.ethereum.network.p2p._
+
+class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
+
+  import PeerActor._
+  import context.{dispatcher, system}
+
+  val P2pVersion = 4
+  val EthProtocolVersion = 63.toByte
```
adamsmo: we have constant for this in `Message.scala`

src/main/scala/io/iohk/ethereum/crypto/ECDSASignature.scala
ntallar: I added that variable when adopting ScalaStyle, it is used to mean the length of both `r` and `s` from the `ECDSASignature` class. I agree that is not a very good name, maybe we could split it in 2 or come up with a better name.

src/main/scala/io/iohk/ethereum/network/NodeInfo.scala
LukasGasior1: yes, but I'm still thinking about good approach for that. It's basically a config that can change it's value over time of running app, any suggestions?

src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala
```diff
@@ -77,6 +78,21 @@ object WireProtocol {
       }
     }
 
+    object Reasons {
+      val DisconnectRequested = 0x00
```
ntallar: Minor detail, if we add variables for each of the codes for the reasons we should also remove them from the `MagicNumberChecker` ignore list from the scalastyle config file to keep it as small as possible.

src/main/scala/io/iohk/ethereum/crypto/ECDSASignature.scala
adamsmo: I think that 2 separate values will be better, I do not know how to correctly name value that is length of both

AlanVerbner:
@LukasGasior1 I'm trying to connect to the scala node using geth's `admin.addPeer` and i'm getting an error while doing handshake but using [EthereumJ](https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/net/rlpx/Handshaker.java#L29) seems to work. Have you tried to do so? Did you get the same error?

LukasGasior1:
It works from EthereumJ for me too. And I also cannot connect from geth (only if geth is initiator). I'm trying to debug that and it seems the packet it sends is not in the format we expect (in ECIESCoder.decrypt), which is strange. I'll try to track that down tomorrow.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -0,0 +1,116 @@
...
+      rlpxConnection ! RLPxConnectionHandler.SendMessage(hello)
+      val timeout = system.scheduler.scheduleOnce(3.seconds, self, ProtocolHandshakeTimeout)
+      context become waitingForHello(rlpxConnection, timeout)
+
+    case RLPxConnectionHandler.ConnectionFailed =>
+      context stop self
+  }
+
+  def waitingForHello(rlpxConnection: ActorRef, timeout: Cancellable): Receive = handleTerminated orElse {
+    case MessageReceived(hello: Hello) =>
```
AlanVerbner: @LukasGasior1 I think i'm receiving a `Disconnect(Too many peers) from peer` here and `case MessageReceived(d: Disconnect)` is missing. I haven't checked the protocol yet to figure out if a Disconnect is valid here (and not sure if it's a buggy peer).
The node I'm trying to connect to (in case you want to try it out) is `enode://65dfc3d6b1685a23c7801f89da0821f1ff6702f6380f5867c9918a1e0a34573da66b5e914ce91d719324beb6f32f2fcb98a858870dd9c34da72221d9ddf27793@77.72.144.197:30303`

LukasGasior1:
I've added support for eip-8 making this PR even larger. I still have to fix the tests for that and make sure it works both ways (connection to/from) with geth and parity.

LukasGasior1:
@AlanVerbner I can now successfully connect to/from geth, can you please check if that also works for you?

AlanVerbner:
sure @LukasGasior1 , I will try now as I need that for my task :smile: Thanks!

AlanVerbner:
@LukasGasior1 Now I can connect to both geth and parity :metal: 

There is still something weird i'm trying to figure out. Geth seems to send a status message as soon as you connect to it but parity seems not to...does that statement make sense to you? Have you experienced it?

adamsmo:
It may be that parity waits for your status message

AlanVerbner:
@adamsmo yes of course, will confirm in a couple of mins. I was wondering why there isn't a specific order to exchange those messages defined by the protocol

whysoserious:
:eyes:

LukasGasior1:
@AlanVerbner I'm looking at it now

AlanVerbner:
@LukasGasior1 It's a question regarding documentation...I don't think status message is part of this task.

src/main/scala/io/iohk/ethereum/App.scala
```diff
@@ -0,0 +1,32 @@
...
+import io.iohk.ethereum.crypto._
+import io.iohk.ethereum.network.{NodeInfo, ServerActor, PeerManagerActor}
+
+object App {
+
+  val nodeKey = generateKeyPair()
+
+  def main(args: Array[String]): Unit = {
+    val listenHostname = "127.0.0.1"
+    val listenPort = 9076
```
whysoserious: Feel free to ignore this comment but I would read these values from config file.  The sooner we create it the less time we'll have to spend on grepping code for all possible config values.

adamsmo:
from what I remember there is statement that no messages can be exchanged before sending status message
I think it does not specify who sends message first because of p2p nature of network

src/main/scala/io/iohk/ethereum/App.scala
ntallar: I have been working on providing a list of peers to connect to from an `application.conf` file so I could easily insert this values in that configuration file.

LukasGasior1:
Trace logging revealed when parity doesn't send Status. Here's what I get (after auth handshake): 
```
2017-01-20 15:04:53  IO Worker #1 TRACE network  Session read error: 17:Some(1ece2bb830ebc03c84505c3e3d1bd165e31b0c5500ad24d742d4f87c255ef9049d037920ff71fcfe5be0948ed1012039cf1b8bbf22e1f85e956b4875f2ac7cd8) (Ok(V4(127.0.0.1:9076))) Auth
```
investigating now

LukasGasior1:
@AlanVerbner @adamsmo I've fixed etc-cli <-> Parity communication (it was due to unsupported multi-frame packages, in fact even single frame packets with specified totalPayloadSize are not handled properly by Parity).
@whysoserious I've addressed most of your comments. Regarding config, I'd rather do it in separate PR, this one is already pretty broad.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -0,0 +1,120 @@
...
+import akka.util.ByteString
+import RLPxConnectionHandler.MessageReceived
+import io.iohk.ethereum.network.p2p._
+
+class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
+
+  import PeerActor._
+  import context.{dispatcher, system}
+
+  val P2pVersion = 4
```
adamsmo: isn't version directly coupled with implementation, so it should be here?

AlanVerbner:
@LukasGasior1 I was able to connecto to my parity instance :tada:

PR #15, opened 2017-01-13
ScalaStyle Integration

ntallar:
# Description

This branch integrates ScalaStyle and solves many of the issues it raised except `Cyclomatic complexity`. For this purpose the `scalastyle-config.xml` file used by the Scorex team was used. There were also made some syntactic simplifications to the code as recommended by IntelliJ's code analyzer tool.

There are still some warnings left when using the tool but it didn't made sense to fix them.


whysoserious:
`scalastyle` output looks ok:
```
> scalastyle
[info] scalastyle using config /Users/jan/Dev/etc-client/scalastyle-config.xml
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/crypto/EthereumIESEngine.scala:95:14: Method is longer than 50 lines
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/FrameCodec.scala:26:44: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/FrameCodec.scala:94:63: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/FrameCodec.scala:96:38: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/FrameCodec.scala:131:36: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/FrameCodec.scala:133:60: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/FrameCodec.scala:135:44: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/Message.scala:18:6: Cyclomatic complexity of 16 exceeds max of 10
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala:89:13: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala:90:13: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala:91:13: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala:92:13: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala:93:13: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala:94:13: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala:95:13: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala:96:13: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala:97:13: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala:98:13: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala:99:13: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala:100:13: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala:86:17: Cyclomatic complexity of 14 exceeds max of 10
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/TestSocketHandshaker.scala:40:4: Regular expression matched 'println'
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/TestSocketHandshaker.scala:45:4: Regular expression matched 'println'
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/TestSocketHandshaker.scala:49:4: Regular expression matched 'println'
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/TestSocketHandshaker.scala:62:4: Regular expression matched 'println'
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/TestSocketHandshaker.scala:64:4: Regular expression matched 'println'
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/TestSocketHandshaker.scala:68:26: Regular expression matched 'println'
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/TestSocketHandshaker.scala:85:4: Regular expression matched 'println'
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/TestSocketHandshaker.scala:100:25: Regular expression matched 'println'
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/network/TestSocketHandshaker.scala:26:6: Method is longer than 50 lines
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/rlp/RLP.scala:51:47: Magic Number
[warn] /Users/jan/Dev/etc-client/src/main/scala/io/iohk/ethereum/rlp/RLP.scala:51:52: Magic Number
[info] Processed 22 file(s)
[info] Found 0 errors
[info] Found 32 warnings
[info] Found 0 infos
[info] Finished in 197 ms
[success] created output: /Users/jan/Dev/etc-client/target
[success] Total time: 2 s, completed Jan 16, 2017 2:52:52 PM
```

src/main/scala/io/iohk/ethereum/network/AuthHandshaker.scala
```diff
@@ -192,4 +159,51 @@ case class AuthHandshaker(
...
+
+    val mac2 = new KeccakDigest(MacSize)
+    mac2.update(xor(macSecret, initiateMessage.nonce.toArray), 0, macSecret.length)
+    new KeccakDigest(mac2).doFinal(buf, 0)
+    mac2.update(responsePacket.toArray, 0, responsePacket.toArray.length)
+    new KeccakDigest(mac2).doFinal(buf, 0)
+
+    if (isInitiator) (mac1, mac2)
+    else (mac2, mac1)
+  }
```
rtkaczyk: Lots of new code. Is it relevant to the scalastyle?

src/main/scala/io/iohk/ethereum/crypto/EthereumIESEngine.scala
```diff
@@ -111,7 +101,7 @@ class EthereumIESEngine(kdf: Either[ConcatKDFBytesGenerator, MGF1BytesGeneratorE
         val derivedKey: ByteString = fillKDFunction(ECIESCoder.KeySize / 8 + ECIESCoder.KeySize / 8)
         val (firstPart, secondPart) = derivedKey.splitAt(ECIESCoder.KeySize / 8)
 
-        if (IV != null)
+        if (Option(IV).isDefined)
```
rtkaczyk: What's this? How is that better style / more readable? Maybe the warning is about using `eq` rather than `==` with `nulls`? If not, I'd vote we disable that rule.

rtkaczyk:
So, still a lot of warnings. Do we want to fix them or adjust the rules?

src/main/scala/io/iohk/ethereum/network/AuthHandshaker.scala
rtkaczyk: Ah, I can see it was just reorganised. Nvm

ntallar: It is as we currently we are currently using a rule that throws warnings if there is any method that has more than 50 lines. I think it is up to discussion whether or not we should use that rule as there are some cases where its not possible to reduce the number of lines of a method.

src/main/scala/io/iohk/ethereum/crypto/EthereumIESEngine.scala
ntallar: I just tried using `!(IV eq null)` and scalastyle continues to throw the warning of using `nulls`. So we could disable this rule and rollback to before this changes.

rtkaczyk: Yeah, I would prefer we disable that rule. I guess being dependent on Java libs we just have to deal with those `nulls`.

ntallar: I agree. I already disabled the rule and rollbacked those changes.

whysoserious: I'd revert this commit. Is there any reason why we could expect `null` in this place? Summoning @adamsmo to explain.

ntallar:
I think we could at least remove the `RegexChecker` for `println`, but the `MagicNumberChecker` is a necessary rule but it doesn't make sense to fix the current warnings. We could fix them but in some cases, as with `WireProtocol.scala` it would just duplicate the code length.

whysoserious:
My proposals:
- [ ] `Magic Number`: add all values found by the tool to the list of ignored values
- [ ] `Regular expression matched 'println'`: ban printlns unless it's a CLI code.
- [ ] `Method is longer than 50 lines`: set max to 50.
- [ ] `Cyclomatic complexity of 16 exceeds max of 10`: set max to 16.
- [ ] **Block merge if scalastyle returns warnings.**
- [ ] Document the rule above, probably in README.md.

src/main/scala/io/iohk/ethereum/crypto/EthereumIESEngine.scala
adamsmo: @whysoserious it can be null because IV vector may be not provided to 
https://github.com/input-output-hk/etc-client/blob/phase/2/txHashValidation/src/main/scala/io/iohk/ethereum/crypto/ECIESCoder.scala#L36
but as I see this method is used only in one place now so we can remove this check or replace it with Option

adamsmo: @whysoserious it can be null because IV vector may be not provided to 
https://github.com/input-output-hk/etc-client/blob/phase/2/txHashValidation/src/main/scala/io/iohk/ethereum/crypto/ECIESCoder.scala#L36
but as I see this method is used only in one place now so we can remove this check or replace it with Option
`IV: Option[Array[Byte]]`

src/main/scala/io/iohk/ethereum/crypto/EthereumIESEngine.scala
```diff
@@ -52,7 +52,7 @@ class EthereumIESEngine(kdf: Either[ConcatKDFBytesGenerator, MGF1BytesGeneratorE
         val derivedKey = fillKDFunction(ECIESCoder.KeySize / 8 + ECIESCoder.KeySize / 8)
         val (firstPart, secondPart) = derivedKey.splitAt(ECIESCoder.KeySize / 8)
 
-        if (IV != null)
+        if (Option(IV).isDefined)
```
whysoserious: Please convert `IV` to `Option[Array[Byte]]` type.

src/main/scala/io/iohk/ethereum/crypto/EthereumIESEngine.scala
```diff
@@ -111,7 +101,7 @@ class EthereumIESEngine(kdf: Either[ConcatKDFBytesGenerator, MGF1BytesGeneratorE
         val derivedKey: ByteString = fillKDFunction(ECIESCoder.KeySize / 8 + ECIESCoder.KeySize / 8)
         val (firstPart, secondPart) = derivedKey.splitAt(ECIESCoder.KeySize / 8)
 
-        if (IV != null)
```
whysoserious: Please convert `IV` to `Option[Array[Byte]]` type.

adamsmo:
@whysoserious I added better null handling
https://github.com/input-output-hk/etc-client/compare/feature/scalastyle...feature/scalastyleFixForNull
@ntallar if you approve I can merge those changes

adamsmo:
maybe as part of this PR we can add command for running scala style to readme file?

whysoserious:
Such command already exists: `sbt scalastyle` but it won't be used by anyone if it's not ran automatically.

adamsmo:
yes, suggestion was about adding it to readme for documentation

rtkaczyk:
I'm all for documenting such things, but just wanted to point out that it probably doesn't belong in the main `README.md` which should be about what our client is and how to use it. This information is in the genre of tips&tricks for the developers and justifies its own doc.

ntallar:
@adamsmo I agree with that way of handling `nulls` in `feature/scalastyleFixForNull`. I already merged that branch with the PR.

src/main/scala/io/iohk/ethereum/crypto/EthereumIESEngine.scala
ntallar: Fixed by merging `feature/scalastyleFixForNull` branch.

ntallar:
@whysoserious I followed your advice and fixed the problems related to `Magic Numbers` and `Cyclomatic complexity` (and will commit them soon). When you talk about method length and `set max to 50` do you mean setting the max to a number above the one in the current file for which the warning is thrown? I think that if we modify the max value of both `Cyclomatic complexity` and method length limit we should set a reasonable limit and modify our future code to follow it. I think that if we just change the rules when we are dealing with warnings like this, using both rules won't make sense.

whysoserious:
> do you mean setting the max to a number above the one in the current file for which the warning is thrown?

Yes, sorry for the mistake. I'd set it to 55 and then one `scalastyle` warning will disappear.

> I think that if we modify the max value of both Cyclomatic complexity and method length limit we should set a reasonable limit and modify our future code to follow it. I think that if we just change the rules when we are dealing with warnings like this, using both rules won't make sense.

I am not sure if I understand what you mean. All in all, these two rules are of negligible importance. Just make your decision whether we should increase the max values or turn these rules off. Whatever you decide you have my sword :) Let's just don't spend more time on this issue.

whysoserious:
@ntallar I created a branch which adds following changes:
- removes a few printlns from @adamsmo code. This was enough ton get rid of all scalastyle warnings.
- turns off scalastyle warnings for `println` (there's no easy way to enable this check only for selected files).
-converts all scalastyle warnings to errors.

https://github.com/input-output-hk/etc-client/tree/feature/scalastyle-cleanup-errors

Imho if you merge this branch into  `feature/scalastyle` then this PR will be ready.

Later we'll ask Alan to add one more build step in CircleCI, namely `sbt scalastyle`.

ntallar:
@whysoserious I agree with your fixes, I just merged them into this branch. With those fixes, there are no more warnings when using scalastyle.

whysoserious:
That's great. 

I am giving 👍   for a merge.

whysoserious:
@rtkaczyk or @adamsmo (whoever comes first) could you add second thumb up and hit the `merge` button? :)

rtkaczyk:
👍

adamsmo:
@whysoserious `TestSocketHandshaker.scala` was only for demo and I think we can remove it

whysoserious:
I'd keep it just as a reference on how to use it.

AlanVerbner:
@adamsmo @whysoserious Yesterday we had a briefly discussion with  @mcsherrylabs and we agreed that:
- It's not a good idea to keep them inside the repo as we don't want to be having compilation / update problems if we change the API. That doesn't mean we should update them if we can.
- This files should be kept and shared to allow others to have reference demos / implementation of stuff. 

An alternative we talked about is to upload this tests to gists and share them in a wiki page (for example GH one). What I don't like is that afaik gists cannot be attached to organizations. What do you think?

whysoserious:
This code has to compile and be up-to-date with our APIs. Otherwise these examples will lose any value.
That said, I'd prefer to keep them in our repo. Moving them anywhere else and checking manually whether they compile or not does not make sense - nobody is going to do that.

AlanVerbner:
@whysoserious you have an important point there. I guess what @mcsherrylabs tried to express is that is not the focus of our code and we don't want to block any release/merge/deploy whatever because of this demo snippets. Maybe he can explain it better :smile:

PR #18, opened 2017-01-18
[MPT] Test partitioning in order not to ignore long running tests

AlanVerbner:
## Description

In order to avoid long running tests to be ignored (or bother local executions) we decided to split them in two different configurations. Test can be run as:
- `sbt test` will run faster tests 
- `sbt it:test` will run benchmark and integration tests that take more time to run

`it` configuration was based on `Test` to be able to share `ObjectGenerators` and maybe any other test utils class. Please let me know if you don't like it.

I also created an `EphemDataSource` as after several discussions this will be needed besides tests. We can create a custom for testing purposes in the future. Will add tests for this and IodbDataSource once this PR is merged.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,14 +1,23 @@
-name := "etc-client"
 
-version := "0.1"
+val commonSettings = Seq(
+  name := "etc-client",
+  version := "0.1",
+  scalaVersion := "2.11.8"
+)
 
-scalaVersion := "2.11.8"
-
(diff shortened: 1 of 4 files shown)


AlanVerbner:
@whysoderious I rather to to that change in the main MOST branch but it can do it here if you want to

whysoserious:
No worries, feel free to merge it.

PR #19, opened 2017-01-20
Add messages for receipts download

adamsmo:
This PR is part of task to support fast download of block chain, it adds receipts message decoding


src/main/scala/io/iohk/ethereum/network/TestSocketHandshaker.scala
```diff
@@ -55,22 +54,45 @@ object TestSocketHandshaker {
...
 
     sendMessage(helloMsg, frameCodec, out)
-
     val remoteHello = readAtLeastOneMessage(frameCodec, inp).head.asInstanceOf[Hello]
 
     while (true) {
       val msgs = readAtLeastOneMessage(frameCodec, inp)
-      msgs.foreach { m => println("\n Received message: " + m) }
+      msgs.foreach { m =>
+        //println("\n Received message: " + m)
```
ntallar: Minor detail: Maybe you forgot about this comment, maybe this should be uncommented or deleted?

adamsmo: Thanks for spotting that, fixed

ntallar:
LGTM!

AlanVerbner:
@adamsmo i'm having problem while running this using `TestSocketHandshaker`. The error is `Invalid point encoding 0x1`. I think it's related with something that @LukasGasior1 has been working on (connecting through actors work) ... @LukasGasior1 any quick ideas? I can take a look if you want to

adamsmo:
@AlanVerbner yes I have them too it is because of merge ac80a09 it changed way of how connection works

AlanVerbner:
@adamsmo do you want me to try to fix `TestSocketHandshaker` or it's not worthy?

adamsmo:
@AlanVerbner I think it is not worth it, we should create new `TestApp` based on actors

AlanVerbner:
Ok, will try to test your code using your quick fix in order to approve this PR

AlanVerbner:
@adamsmo I think it will be nice to have encode/decode tests about this messages, what do you think?

adamsmo:
@AlanVerbner added unit tests, also did refactoring

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -25,6 +27,140 @@ object PV63 {
...
+      override def decode(rlp: RLPEncodeable): Receipts = rlp match {
+        case rlpList: RLPList => Receipts(rlpList.items.collect { case r: RLPList => r.items.map(rlpDecode[Receipt]) })
+        case _ => throw new RuntimeException("Cannot decode Receipts")
+      }
+    }
+
+    val code: Int = Message.SubProtocolOffset + 0x10
+  }
+
+  case class Receipts(receiptsForBlocks: Seq[Seq[Receipt]]) extends Message {
```
rtkaczyk: A question about `Seq[Seq[...]]`. Does the outer sequence index determine which block the inner receipts are for?

adamsmo: yes, you ask for receipts with list of hashes and in return you are getting lists of receipts corresponding to that block hashes

rtkaczyk: Perhaps that's ok, but have you considered `Map[BlockHash, Seq[Receipt]]` ?

src/main/scala/io/iohk/ethereum/rlp/package.scala
```diff
@@ -20,10 +20,14 @@ package object rlp {
 
   def encode[T](input: T)(implicit enc: RLPEncoder[T]): Array[Byte] = RLP.encode(enc.encode(input))
```
AlanVerbner: can we use in this lien the one defined next?

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
adamsmo: I think such map can be constructed by user of this interface, this is only logic for decoding/encoding protocol messages and hash is not part of response

src/main/scala/io/iohk/ethereum/rlp/package.scala
```diff
@@ -20,10 +20,14 @@ package object rlp {
...
 
+  def encodeToRlp[T](input: T)(implicit enc: RLPEncoder[T]): RLPEncodeable = enc.encode(input)
```
AlanVerbner: Is this only needed in tests? If so, why not calling directly `enc.encode(input)`

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
rtkaczyk: Ah yes, makes sense

src/main/scala/io/iohk/ethereum/rlp/package.scala
adamsmo: it is also used in PV63.scala but renamed to `rlpEncode`
for example https://github.com/input-output-hk/etc-client/pull/19/files#diff-048743a755e4bf01956ef907e6704010R90

src/main/scala/io/iohk/ethereum/rlp/package.scala
```diff
@@ -20,10 +20,14 @@ package object rlp {
...
+
   def encode(input: RLPEncodeable): Array[Byte] = RLP.encode(input)
 
   def decode[T](data: Array[Byte])(implicit dec: RLPDecoder[T]): T = dec.decode(RLP.rawDecode(data))
 
+  def decode[T](data: RLPEncodeable)(implicit dec: RLPDecoder[T]): T = dec.decode(data)
```
AlanVerbner: same here as ^

adamsmo: sam as previous ;) `rlpDecode`
https://github.com/input-output-hk/etc-client/pull/19/files#diff-048743a755e4bf01956ef907e6704010R95

src/main/scala/io/iohk/ethereum/network/p2p/Message.scala
```diff
@@ -26,19 +26,30 @@ object Message {
...
 
-    case (PV61, pv61.NewBlockHashes.code) => rlp.decode(payload)(pv61.NewBlockHashes.rlpEndDec)
-    case (PV61, pv61.BlockHashesFromNumber.code) => rlp.decode(payload)(pv61.BlockHashesFromNumber.rlpEndDec)
+    case (PV61, t) => handlePV61(t, payload)
 
     case (PV62 | PV63, pv62.NewBlockHashes.code) => rlp.decode(payload)(pv62.NewBlockHashes.rlpEndDec)
     case (PV62 | PV63, pv62.GetBlockHeaders.code) => rlp.decode(payload)(pv62.GetBlockHeaders.rlpEndDec)
     case (PV62 | PV63, pv62.BlockHeaders.code) => rlp.decode(payload)(pv62.BlockHeaders.rlpEndDec)
     case (PV62 | PV63, pv62.GetBlockBodies.code) => rlp.decode(payload)(pv62.GetBlockBodies.rlpEndDec)
     case (PV62 | PV63, pv62.BlockBodies.code) => rlp.decode(payload)(pv62.BlockBodies.rlpEndDec)
```
LukasGasior1: how about `handlePV62`?

adamsmo: but there is no message specific to PV62, they are shared between PV62 and PV63 so what that method should be called `handlePV62andPartOfPV63` ? for me it is better to leave this messages unnamed in decode method

src/main/scala/io/iohk/ethereum/rlp/package.scala
ntallar: I think that this function and the `encodeToRLP` function might get confused with the `encode` and `decode` that use the `RLP` internal versions. Maybe they should be renamed to something that more explicitly differentiates them.

adamsmo: 👍 
maybe
`encodeToRLP`
`decodeFromRLP`

src/main/scala/io/iohk/ethereum/rlp/package.scala
AlanVerbner: @adamsmo you are using to encode `ByteString` only? Was it because you were having trouble?  I think it's using the wrong implicit `ByteArray` -> `Seq` -> `io.iohk.ethereum.rlp.RLPImplicits#toEncodeableList`

adamsmo: yes it is, so I wanted to use it explicitly

AlanVerbner: Adding this to implicits before Seq one

 `implicit def byteStringFromEncodeable(b: ByteString)(implicit enc: RLPEncoder[ByteString]): RLPEncodeable = enc.encode(b)` 

Allowed me to do 

`RLPList(loggerAddress, logTopics: RLPList, data)`

adamsmo: :+1: thanks it works, will update code

src/main/scala/io/iohk/ethereum/rlp/package.scala
ntallar: Maybe something like `toRLPEncodeable` and `fromRLPEncodeable` to make it more explicit that we are transforming objects to and from `RLPEncodeable`. Either way, we might we able to not have these functions (or at least `encodeToRLP`) using what @AlanVerbner mentioned.

adamsmo: was able to remove `encodeToRlp`, only thing left is 
`def decode[T](data: RLPEncodeable)(implicit dec: RLPDecoder[T]): T` which has different argument type than other `decode` do you thing it is necessary to indicate that it decodes from RLP if it is type of input argument? 
Besides that it does the same thing as 
`def decode[T](data: Array[Byte])(implicit dec: RLPDecoder[T]): T`

PR #20, opened 2017-01-23
Add logback logging

LukasGasior1:
File + stdout in app, stdout only in tests

### DIFF
--- a/src/it/scala/io/iohk/ethereum/mpt/MerklePatriciaTreeIntegrationSuite.scala
+++ b/src/it/scala/io/iohk/ethereum/mpt/MerklePatriciaTreeIntegrationSuite.scala
@@ -4,15 +4,14 @@ import java.io.File
 import java.nio.ByteBuffer
 import java.security.MessageDigest
 
-import akka.util.ByteString
 import io.iohk.ethereum.ObjectGenerators
 import io.iohk.ethereum.crypto.sha3
 import io.iohk.ethereum.mpt.MerklePatriciaTrie.defaultByteArraySerializable
-import io.iohk.ethereum.rlp.{decode => decodeRLP, encode => encodeRLP}
 import io.iohk.iodb.LSMStore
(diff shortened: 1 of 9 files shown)


src/it/scala/io/iohk/ethereum/mpt/MerklePatriciaTreeIntegrationSuite.scala
```diff
@@ -4,22 +4,24 @@ import java.io.File
...
+import org.slf4j.LoggerFactory
 import org.spongycastle.util.encoders.Hex
 
 import scala.util.Random
 
 class MerklePatriciaTreeIntegrationSuite extends FunSuite
   with PropertyChecks
   with ObjectGenerators {
+
+  val log = LoggerFactory.getLogger(this.getClass)
```
adamsmo: Don't Repeat Yourself

LukasGasior1: Done

LukasGasior1:
@etc-builder I pushed something similar just before your comment :)

LukasGasior1:
@mcsherrylabs Does it make sense? In most cases we want the actual logging class name to show in log, not the Logger

whysoserious:
>  In most cases we want the actual logging class name to show in log, not the Logger 

I'd use it logger directly as well.

mcsherrylabs:
I think it's worth adding. The 'getLogger' allows the conscientious to pass a category. Mixing in a trait might not always be appropriate/possible. 

object Logger extends Logging {

  def getLogger(category: String): Logger = {
    LoggerFactory.getLogger(category)
  }

}

LukasGasior1:
I don't see much difference in:
`LoggerFactory.getLogger(getClass)` vs our `Logger.getLogger(getClass)` but ok, I'll add it

LukasGasior1:
@mcsherrylabs done

PR #21, opened 2017-01-23
Configuration file for initial peer list

ntallar:
Included initial peer list (bootstrap nodes) and server address in an `application.conf` file.


src/main/resources/application.conf
```diff
@@ -0,0 +1,10 @@
+serverAddress {
+  IP = "127.0.0.1"
```
LukasGasior1: I'd call it more generic `interface`

src/main/resources/application.conf
```diff
@@ -0,0 +1,10 @@
...
+  Port = 9076
```
LukasGasior1: lower case?

src/main/resources/application.conf
ntallar: Fixed!

src/main/resources/application.conf
ntallar: Fixed!

src/main/resources/application.conf
```diff
@@ -1,5 +1,16 @@
+serverAddress {
+  interface = "127.0.0.1"
+  port = 9076
+}
+
+discovery {
+  bootstrapNodes = [
+    "enode://08aa4a60712b17b4144cc6e97a6fbb254007d2e73e8eb83c0b2757cb696b3b8464a299ca504e9907769ac7ca398993282524c4c0afa3cd07d15250c179a875e0@127.0.0.1:60303?discport=0"
```
adamsmo: @ntallar I assume that this is your local node with specific public key, should this vale be in main repo?

ntallar: I agree, I already changed it to a node I obtained from https://www.epool.io/static-nodes.json. While testing the new node (and others) I sometimes got disconnected with the reason `Disconnect(Too many peers)`, but it seems to be something common when trying to connect with other nodes. I was able to connect successfully with this node many times I tried so that's why I selected it.

src/main/resources/application.conf
```diff
@@ -1,5 +1,16 @@
+serverAddress {
```
whysoserious: Not a blocker but I would change all names config file to hyphen-case (`server-address`) as it used eg. in Akka's reference.conf: https://github.com/akka/akka/blob/master/akka-actor/src/main/resources/reference.conf

ntallar: I agree, fixed!

PR #22, opened 2017-01-23
Tx validation

ntallar:
Included tx syntactic validation rules (Eq. 15 in yellow paper section 4.3 and EthereumJ: https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/core/Transaction.java#L207).


src/main/scala/io/iohk/ethereum/network/p2p/messages/CommonMessages.scala
```diff
@@ -111,10 +118,23 @@ object CommonMessages {
     )
 
     val FirstByteOfAddress = 12
-    val LastByteOfAddress = 32
+    val LastByteOfAddress = 12 + AddressLength
```
mcsherrylabs: Is LastByteOfAddress = FirstByteOfAddress + AdressLength? :)

src/main/scala/io/iohk/ethereum/network/p2p/messages/CommonMessages.scala
```diff
@@ -111,10 +118,23 @@ object CommonMessages {
     )
 
     val FirstByteOfAddress = 12
-    val LastByteOfAddress = 32
+    val LastByteOfAddress = FirstByteOfAddress + AddressLength
```
rtkaczyk: Should this be part of the `Transaction`? If at all I would say it goes better in the comp object. In the future I think it would useful to have an `Address` class.

adamsmo: this constants are here to get `recoveredAddress` which is part of transaction, but maybe they should be private or moved somewhere?

ntallar: Moved constants to `Transaction` companion object for now but I also think it will be useful to have an `Address` class.

PR #23, opened 2017-01-23
Add an implementation and tests for a DataWord

whysoserious:
Please read the documentation in code to see what a `DataWord` is. Ping me if anything is not clear enough.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/DataWord.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/DataWord.scala
@@ -2,8 +2,6 @@ package io.iohk.ethereum.vm
 
 import akka.util.ByteString
 
-import scala.math.ScalaNumericConversions
-
 
 object DataWord {
 
@@ -41,7 +39,7 @@ object DataWord {
 
 /** Stores 256 bit words and adds a few convenience methods on them.
  *  Internally a word is stored as a BigInt. */
-class DataWord private (private val n: BigInt) extends ScalaNumericConversions {
+class DataWord private (private val n: BigInt) {
 
   import DataWord._
 
@@ -76,18 +74,6 @@ class DataWord private (private val n: BigInt) extends ScalaNumericConversions {
   def *(that: DataWord): DataWord = DataWord(this.n * that.n)
 
   def /(that: DataWord): DataWord = DataWord(this.n / that.n)
-
-  def doubleValue(): Double = longValue().toDouble
-
-  def floatValue(): Float = longValue().toFloat
-
-  def intValue(): Int = value.foldLeft(0) { (n, b) => (n << 8) + (b & 0xff) }
-
-  def longValue(): Long = value.foldLeft(0l) { (n, b) => (n << 8) + (b & 0xff) }
-
-  def isWhole(): Boolean = true
-
-  def underlying(): AnyRef = value
 
   override def equals(that: Any): Boolean = {
     that match {
--- a/src/main/scala/io/iohk/ethereum/vm/DataWord.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/DataWord.scala
@@ -7,9 +7,11 @@ object DataWord {
 
   val MaxLength = 32
 
-  val MaxWord: BigInt = BigInt(2).pow(MaxLength * 8) - 1
+  private val Modulus: BigInt = BigInt(2).pow(MaxLength * 8)
 
-  val Zeros: ByteString = ByteString(Array.fill[Byte](MaxLength)(0))
+  val MaxWord = DataWord(Modulus - 1)
+
+  private val Zeros: ByteString = ByteString(Array.fill[Byte](MaxLength)(0))
 
   def apply(value: ByteString): DataWord = {
     require(value.length <= MaxLength, s"Input byte array cannot be longer than $MaxLength: ${value.length}")
@@ -25,13 +27,13 @@ object DataWord {
     apply(BigInt(num.toLong(n)))
   }
 
-  def fixBigInt(n: BigInt): BigInt = {
-    if (n == -MaxWord) {
+  private def fixBigInt(n: BigInt): BigInt = {
+    if (n == -Modulus) {
       0
     } else if (n < 0) {
-      n % MaxWord + MaxWord
+      n % Modulus + Modulus
     } else {
-      n % MaxWord
+      n % Modulus
     }
   }
 
@@ -46,7 +48,7 @@ class DataWord private (private val n: BigInt) {
   /** Converts a BigInt to a ByteString.
    *  Output ByteString is padded with 0's from the left side up to MaxLength bytes.
    */
-  lazy val value: ByteString = {
+  lazy val bytes: ByteString = {
     val bs: ByteString = ByteString(n.toByteArray)
     val padLength: Int = MaxLength - bs.length
     if (padLength > 0) {
@@ -56,7 +58,7 @@ class DataWord private (private val n: BigInt) {
     }
   }
 
-  require(n >= 0 && n <= MaxWord, s"Invalid word value: $n")
+  require(n >= 0 && n < Modulus, s"Invalid word value: $n")
 
   def &(that: DataWord): DataWord = DataWord(this.n & that.n)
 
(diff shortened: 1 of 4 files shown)


src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -0,0 +1,105 @@
...
+  def isWhole(): Boolean = true
+
+  def underlying(): AnyRef = value
+
+  override def hashCode(): Int = n.hashCode()
+
+  override def equals(that: Any): Boolean =
+    that match {
+      case that: DataWord =>
+        that.isInstanceOf[DataWord] && this.hashCode == that.hashCode
```
ntallar: Minor detail: Is `isInstanceOf[DataWord]`necessary in this case? I'm not sure if doing the pattern matching does the same type check.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  def *(that: DataWord): DataWord = DataWord(this.n * that.n)
+
+  def /(that: DataWord): DataWord = DataWord(this.n / that.n)
+
+  def doubleValue(): Double = longValue().toDouble
+
+  def floatValue(): Float = longValue().toFloat
+
+  def intValue(): Int = value.foldLeft(0) { (n, b) => (n << 8) + (b & 0xff) }
```
ntallar: Is there any reason why you obtain the int value from the `ByteString` `value` and not from the `BigInt` `n`? I'm not sure if maybe you intend to stop using the `n` BigInt in the future.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -0,0 +1,105 @@
...
+
+}
+
+/** Stores 256 bit words and adds a few convenience methods on them.
+ *  Internally a word is stored as a BigInt. */
+class DataWord private (private val n: BigInt) extends ScalaNumericConversions {
+
+  import DataWord._
+
+  // TODO: Consider changing internal representation of a DataWord to a ByteString.
```
adamsmo: should this TODO be addressed?

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
adamsmo: what about int overflow, do we care?

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
LukasGasior1: also shouldn't it be `this.n == that.n` afair two non equal objects can share the same hash code

whysoserious:
I addressed all your comments in the code:
- broken `.equals` method: I promise not to copy/paste from SO anymore
- `.intValue` etc.: this was required by `ScalaNumericConversions` trait which was implemented by `DataWord` which didn't make sense at all :)

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -0,0 +1,90 @@
...
+    } else {
+      n % MaxWord
+    }
+  }
+
+}
+
+/** Stores 256 bit words and adds a few convenience methods on them.
+ *  Internally a word is stored as a BigInt. */
+class DataWord private (private val n: BigInt) {
```
adamsmo: can this be case class and that way we can avoid adding our own `equals` and `hashCode` ?

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -0,0 +1,90 @@
...
+import akka.util.ByteString
+
+
+object DataWord {
+
+  val MaxLength = 32
+
+  val MaxWord: BigInt = BigInt(2).pow(MaxLength * 8) - 1
+
+  val Zeros: ByteString = ByteString(Array.fill[Byte](MaxLength)(0))
```
rtkaczyk: Let's make this private. Doesn't make much sense as part of public API

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -0,0 +1,90 @@
...
+  def apply(n: BigInt): DataWord = {
+    new DataWord(fixBigInt(n))
+  }
+
+  def apply[N: Integral](n: N): DataWord = {
+    val num = implicitly[Integral[N]]
+    apply(BigInt(num.toLong(n)))
+  }
+
+  def fixBigInt(n: BigInt): BigInt = {
```
rtkaczyk: Also private.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -0,0 +1,90 @@
...
+    if (n == -MaxWord) {
+      0
+    } else if (n < 0) {
+      n % MaxWord + MaxWord
+    } else {
+      n % MaxWord
```
rtkaczyk: This is actually incorrect (yes, I know I did that 😛 ). Let's rename `MaxWord` to `Modulus`, increment it and privatise (because it's a `BigInt` not a `DataWord`).

Also we can have it written more consisely: `(n % Modulus + Modulus) % Modulus), though the `if`/`else` version is more efficient - so just noting that. Kudos for spotting the `n == -MaxWord` case.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
rtkaczyk: No, it shouldn't be. We want to have our own `equals` and we don't want to expose the underlying data type - it may change in the future.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  require(n >= 0 && n <= MaxWord, s"Invalid word value: $n")
+
+  def &(that: DataWord): DataWord = DataWord(this.n & that.n)
+
+  def |(that: DataWord): DataWord = DataWord(this.n | that.n)
+
+  def ^(that: DataWord): DataWord = DataWord(this.n ^ that.n)
+
+  def unary_-(): DataWord = DataWord(-this.n)
```
rtkaczyk: Let's not use empty `()` for such functions. Empty parens suggest side effects.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -0,0 +1,90 @@
...
+  def -(that: DataWord): DataWord = DataWord(this.n - that.n)
+
+  def *(that: DataWord): DataWord = DataWord(this.n * that.n)
+
+  def /(that: DataWord): DataWord = DataWord(this.n / that.n)
+
+  override def equals(that: Any): Boolean = {
+    that match {
+      case that: DataWord => this.n.equals(that.n)
+      case _ => false
```
rtkaczyk: `case other => other == n`. This way, having: `val a = DataWord(0)` I can write: `a == 0`, which is more efficient convenient and readable than: `a == DataWord(0)`.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+
+  override def hashCode(): Int = n.hashCode()
```
rtkaczyk: Remove empty parens

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  override def toString(): String = {
```
rtkaczyk: We won't use `toString` for anything other than debugging, and for that I find this the most useful form:

```
DataWord(16).toString == "DataWord(0x10)"
```

The format would be: `f"DataWord(0x$n%02x)"`

rtkaczyk:
> .intValue etc.: this was required by ScalaNumericConversions trait which was implemented by DataWord which didn't make sense at all :)

Actually, we need `intValue` for multiple reasons. But that doesn't mean we necessarily need `ScalaNumericConversions`

rtkaczyk:
I have pushed the improvements I described in comments. Also gave the tests a closer look - they're very good!

LukasGasior1:
lgtm

rtkaczyk:
I would also give an LGTM, but since I've committed to it I shouldn't. @adamsmo @ntallar ?

PR #24, opened 2017-01-23
[EC-32] Classic Chain Check 

AlanVerbner:
## Description

The client will connect only to an ETC supporting client. In order to do so, after handshake Dao BlockHeaders is requested in order to compare it's hash with the ETC fork one.

If the other peer does not return the block we have a simplification about the one stated [here](https://github.com/ethereum/go-ethereum/pull/2795). We are not currently checking our blockchain but comparing DaoBlock total difficulty (TD) against other node TD as this is and indicator for the longest chain. That being said, this tweak should not break thingsbecause the node sync process will disconnect from that peer if DAO block is found.

Note1: our client node status is hardcoded and needs to be changed during the next sprints

Note: BlockHeader validation logic was encapsulated in order to be re-used by @adamsmo (not approved by him yet :stuck_out_tongue:) while block downloading / sync is in progress. Also, it's ment to be used in a some sort of _validators chain_ to apply several checks to different message responses. Feel free to leave any suggestions / changes on this.

### Test

You can check an ETH node running this `sbt 'run-main io.iohk.ethereum.App enode://a979fb575495b8d6db44f750317d0f4622bf4c2aa3365d6af7c284339968eef29b69ad0dce72a4d8db5ebb4968de0e3bec910127f134779fbcb0cb6d3331163c@52.16.188.185:30303'`

### DIFF
--- a/src/main/scala/io/iohk/ethereum/network/PeerActor.scala
+++ b/src/main/scala/io/iohk/ethereum/network/PeerActor.scala
@@ -122,7 +122,7 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
         log.warning("Peer is not running the ETC fork, disconnecting")
         disconnectFromPeer(rlpxConnection, Disconnect.Reasons.UselessPeer)
       }
-    case DaoHeaderReceiveTimeout =>
+    case MessageReceived(BlockHeaders(Nil)) =>
       // FIXME We need to do some checking related to our blockchain. If we haven't arrived to the DAO block we might
       // take advantage of this peer and grab as much blocks as we can until DAO.
       // ATM we will only check by DaoBlockTotalDifficulty
@@ -130,10 +130,7 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
         log.info("Peer probably hasn't arrived to DAO yet")
         context become new HandshakedHandler(rlpxConnection).receive
       }
-      else {
-        log.info("Peer probably has arrived to DAO, so we will disconnect")
-        disconnectFromPeer(rlpxConnection, Disconnect.Reasons.UselessPeer)
-      }
+    case DaoHeaderReceiveTimeout => disconnectFromPeer(rlpxConnection, Disconnect.Reasons.TimeoutOnReceivingAMessage)
   }
 
   private def disconnectFromPeer(rlpxConnection: ActorRef, reason: Int): Unit = {
--- a/src/main/scala/io/iohk/ethereum/network/PeerActor.scala
+++ b/src/main/scala/io/iohk/ethereum/network/PeerActor.scala
@@ -103,7 +103,7 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
       rlpxConnection ! RLPxConnectionHandler.SendMessage(
         GetBlockHeaders(Left(DaoBlockNumber), maxHeaders = 1, skip = 0, reverse = 0)
       )
-      val waitingForDaoTimeout = system.scheduler.scheduleOnce(waitForStatusInterval, self, NoDaoHeaderReceived)
+      val waitingForDaoTimeout = system.scheduler.scheduleOnce(waitForStatusInterval, self, DaoHeaderReceiveTimeout)
       context become waitingForChainForkCheck(rlpxConnection, status, waitingForDaoTimeout)
     case StatusReceiveTimeout =>
       log.warning("Timeout while waiting status")
@@ -128,7 +128,7 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
     case MessageReceived(d: Disconnect) =>
       log.info("Received {} from peer. Closing connection", d)
       disconnectFromPeer(rlpxConnection, Disconnect.Reasons.UselessPeer)
-    case NoDaoHeaderReceived =>
+    case DaoHeaderReceiveTimeout =>
       // FIXME We need to do some checking related to our blockchain. If we haven't arrived to the DAO block we might
       // take advantage of this peer and grab as much blocks as we can until DAO.
       // ATM we will only check by DaoBlockTotalDifficulty
@@ -143,7 +143,7 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
   }
 
   private def disconnectFromPeer(rlpxConnection: ActorRef, reason: Int): Unit = {
-    rlpxConnection ! SendMessage(Disconnect(Disconnect.Reasons.TimeoutOnReceivingAMessage))
+    rlpxConnection ! RLPxConnectionHandler.SendMessage(Disconnect(Disconnect.Reasons.TimeoutOnReceivingAMessage))
     context.system.scheduler.scheduleOnce(5.seconds, self, PoisonPill)
   }
 
@@ -204,7 +204,7 @@ object PeerActor {
 
   case class SendMessage[M <: Message](message: M)(implicit val enc: RLPEncoder[M])
 
-  private case object NoDaoHeaderReceived
+  private case object DaoHeaderReceiveTimeout
 
   private case object ProtocolHandshakeTimeout
 
--- a/src/main/scala/io/iohk/ethereum/network/PeerActor.scala
+++ b/src/main/scala/io/iohk/ethereum/network/PeerActor.scala
@@ -102,7 +102,7 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
       rlpxConnection ! RLPxConnectionHandler.SendMessage(
         GetBlockHeaders(Left(DaoBlockNumber), maxHeaders = 1, skip = 0, reverse = 0)
       )
-      val waitingForDaoTimeout = system.scheduler.scheduleOnce(waitForStatusInterval, self, DaoHeaderReceiveTimeout)
+      val waitingForDaoTimeout = system.scheduler.scheduleOnce(waitForChainCheck, self, DaoHeaderReceiveTimeout)
       context become waitingForChainForkCheck(rlpxConnection, status, waitingForDaoTimeout)
     case StatusReceiveTimeout =>
       log.warning("Timeout while waiting status")
--- a/src/main/scala/io/iohk/ethereum/network/PeerActor.scala
+++ b/src/main/scala/io/iohk/ethereum/network/PeerActor.scala
@@ -80,10 +80,10 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
 
   def waitingForHello(rlpxConnection: ActorRef, timeout: Cancellable): Receive = handleTerminated orElse {
     case MessageReceived(d: Disconnect) =>
-      log.info("Received {} from peer {}. Closing connection", d, peerId)
+      log.info("Received {}. Closing connection", d)
       context stop self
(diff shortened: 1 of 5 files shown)


src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -31,6 +32,14 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
 
   val peerId = self.path.name
 
+  val waitForStatusInterval = 30.seconds
+  val waitForChainCheck = 15.seconds
+
+  //FIXME move this to props
+  lazy val DaoBlockNumber = 1920000
+  lazy val DaoBlockTotalDifficulty = BigInt("39490964433395682584")
```
LukasGasior1: shouldn't these be in ForkValidator?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -71,29 +80,81 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
 
   def waitingForHello(rlpxConnection: ActorRef, timeout: Cancellable): Receive = handleTerminated orElse {
     case MessageReceived(d: Disconnect) =>
-      log.info("Received {}. Closing connection", d)
+      log.info("Received {} from peer {}. Closing connection", d, peerId)
```
LukasGasior1: it's redundant, actor id is also logged (which happens to be peer id)

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -71,29 +80,81 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
...
+        context become new HandshakedHandler(rlpxConnection).receive
       }
+      else {
+        log.info("Peer {} probably hasn't arrived to DAO yet, so we will wait")
+        disconnectFromPeer(rlpxConnection, Disconnect.Reasons.UselessPeer)
+      }
+  }
+
+  private def disconnectFromPeer(rlpxConnection: ActorRef, reason: Int): Unit = {
+    rlpxConnection ! SendMessage(Disconnect(Disconnect.Reasons.TimeoutOnReceivingAMessage))
```
LukasGasior1: I believe it should say `RLPxConnection.SendMessage`, ATM we have two distinct messages (one in PeerActor and one in RLPxConn) we could think about having just one such msg

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -71,29 +80,81 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
...
-            genesisHash = ByteString(Hex.decode("d4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3"))))
+  def waitingForNodeStatus(rlpxConnection: ActorRef, timeout: Cancellable): Receive = handleTerminated orElse
+    handlePing(rlpxConnection) orElse {
+    case MessageReceived(status: Status) =>
+      timeout.cancel()
+      log.info("Peer {} returned status ({})", peerId, status)
+      rlpxConnection ! RLPxConnectionHandler.SendMessage(
+        GetBlockHeaders(Left(DaoBlockNumber), maxHeaders = 1, skip = 0, reverse = 0)
+      )
+      val waitingForDaoTimeout = system.scheduler.scheduleOnce(waitForStatusInterval, self, NoDaoHeaderReceived)
```
LukasGasior1: scheduleOnce(waitForChainCheck, ...

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -71,29 +80,81 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
...
+        log.warning("Peer {} is running the ETC chain", peerId)
         context become new HandshakedHandler(rlpxConnection).receive
       } else {
-        log.info("Peer does not support eth {} protocol. Disconnecting.", Message.PV63.toByte)
-        rlpxConnection ! SendMessage(Disconnect(Disconnect.Reasons.IncompatibleP2pProtocolVersion))
-        context.system.scheduler.scheduleOnce(5.seconds, self, PoisonPill)
+        log.warning("Peer {} is not running the ETC fork, disconnecting", peerId)
+        disconnectFromPeer(rlpxConnection, Disconnect.Reasons.UselessPeer)
+      }
+    case MessageReceived(d: Disconnect) =>
```
LukasGasior1: this is used in quite a few places, maybe we could extract it to `handleDisconnect`. And maybe we could name such message handlers differently, like handleDisconnectMsg, handlePingMsg? Just a suggestion

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -71,29 +80,81 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
...
+      log.info("Received {} from peer {}. Closing connection", d, peerId)
+      disconnectFromPeer(rlpxConnection, Disconnect.Reasons.UselessPeer)
+    case NoDaoHeaderReceived =>
+      // FIXME We need to do some checking related to our blockchain. If we haven't arrived to the DAO block we might
+      // take advantage of this peer and grab as much blocks as we can until DAO.
+      // ATM we will only check by DaoBlockTotalDifficulty
+      if (status.totalDifficulty < DaoBlockTotalDifficulty) {
+        log.info("Peer {} probably hasn't arrived to DAO yet, so we will wait")
```
LukasGasior1: undefined placeholder `{}`

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -126,16 +191,23 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
...
   def props(nodeInfo: NodeInfo): Props =
     Props(new PeerActor(nodeInfo))
 
   case class HandleConnection(connection: ActorRef)
+
   case class ConnectTo(uri: URI)
 
   case class SendMessage[M <: Message](message: M)(implicit val enc: RLPEncoder[M])
 
+  private case object NoDaoHeaderReceived
```
LukasGasior1: let's stay consistent, how about `DaoHeaderReceiveTimeout`?

src/main/scala/io/iohk/ethereum/network/p2p/validators/ForkValidator.scala
```diff
@@ -0,0 +1,31 @@
...
+/**
+  * This Validator checks if a given `BlockHeaders` contains a `BlockHeader` having a specific block number and hash
+  *
+  * This will be usually used to check if we are dealing with DAO Forked block (number = 1920000 and hash distinct of
+  * 94365e3a8c0b35089c1d1195081fe7489b528a84b22199c916180db8b28ade7f)
+  *
+  * @param blockNumber BlockNumber to check
+  * @param blockHash Hash of the block whose number is `blockNumber`
+  */
+case class ForkValidator(blockNumber: BigInt, blockHash: Array[Byte]) extends MessageValidator[BlockHeaders, ForkValidatorError] {
```
LukasGasior1: as they are mutable byte arrays don't play well with case classes, how about ByteString?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
AlanVerbner: yes, it's a little error prone but I think we can deal with it.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
AlanVerbner: Yes, I do agree. I've done something similar with Ping. Will change disconnect too.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
AlanVerbner: @LukasGasior1 I'm not sure. My intentention was to create a ForkValidator not specific for Dao but for any fork we want to check so thought putting this values within `application.conf` would make more sense

AlanVerbner:
@LukasGasior1 I've fixed/commented all the things you pointed out

LukasGasior1:
LGTM, but we should also handle the situation when we're not yet at forked block. we probably should retry the chain test as soon as we get there

AlanVerbner:
@LukasGasior1 yes, we need to include this check while doing sync / receiving blocks too. There are open tasks for that on youtrack.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
AlanVerbner: **Note: This values should be moved to a config before merging**

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -69,31 +78,74 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
...
+    rlpxConnection ! RLPxConnectionHandler.SendMessage(Disconnect(Disconnect.Reasons.TimeoutOnReceivingAMessage))
+    context.system.scheduler.scheduleOnce(5.seconds, self, PoisonPill)
+  }
+
+  // FIXME This is a temporal function until we start keeping our status
+  private def ourStatus: Status = {
+    val totalDifficulty = 34351349760L
+    Status(
+      protocolVersion = Message.PV63, 1, totalDifficulty,
+      ByteString(Hex.decode("88e96d4537bea4d9c05d12549907b32561d3bf31f45aae734cdc119f13406cb6")),
```
adamsmo: minor:
if storage is empty should this be set to genesis block hash instead of first block hash?

adamsmo:
other then mentioned LGTM

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
rtkaczyk: How do you guys feel about putting links to documentation in comments? I for one would like to able to confirm these number and, even more so, would like to read up on the related information.

This also relates to comments in `application.conf` once the numbers get moved there.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
AlanVerbner: @adamsmo this is a super temp value until we decide to work with status. It should be taken from somewhere within the etc-client

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
AlanVerbner: @rtkaczyk I like puting links in comments (specially when grabbing snippets/gists) ... I will add it

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
adamsmo: 👍

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
AlanVerbner: @rtkaczyk done :smile:

LukasGasior1:
Merging as there are 2 LGTMs

PR #25, opened 2017-01-24
Add messages for MPT download

adamsmo:
This PR is about adding messages and conversion for downloading MPT from other nodes.
NodeData message is also used for downloading EVM code and storage for contracts.


src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -114,6 +115,15 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
     }
 
     def processMessage(message: Message): Unit = message match {
+      /*TODO delete test code*/
```
AlanVerbner: Don't forget about this

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -1,30 +1,284 @@
...
+        }
+      }
+
+      override def decode(rlp: RLPEncodeable): MptNode = rlp match {
+        case rlpList: RLPList if rlpList.items.length == 17 =>
+          MptBranch(rlpList.items.take(16).map(byteStringEncDec.decode), byteStringEncDec.decode(rlpList.items(16)))
+        case RLPList(hpEncoded, value) =>
+          hpDecode(hpEncoded: Array[Byte]) match {
+            case (decoded, true) =>
+              MptLeaf(ByteString(decoded), Account.rlpEndDec.decode(rawDecode(value)))
```
ntallar: Would it be the same to use `rlpDecode[Account](value)` (with rlpDecode being a renamed `rlp.decode[T]`) instead of `Account.rlpEndDec.decode(rawDecode(value))`?

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -1,30 +1,284 @@
...
+          }
+        case _ =>
+          throw new RuntimeException("Cannot decode NodeData")
+      }
+    }
+  }
+
+  trait MptNode
+
+  case class MptBranch(childHashes: Seq[ByteString], terminator: ByteString) extends MptNode {
```
AlanVerbner: Branch values might not be hashes but rlpEncoded nodes. Check out yellow paper equation `178`

screenshot from 2017-01-24 17-26-02

adamsmo: will add test for that and implant it

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -1,30 +1,284 @@
...
+
+      s"""NodeData{
+         |values: $v
+         |}
+       """.stripMargin
+    }
+  }
+
+  object MptNode {
+    implicit val rlpEndDec = new RLPEncoder[MptNode] with RLPDecoder[MptNode] {
```
AlanVerbner: The following lines look quite similar to:
- https://github.com/input-output-hk/etc-client/blob/phase/3/fastDownload/src/main/scala/io/iohk/ethereum/mpt/MerklePatriciaTrie.scala#L69
- https://github.com/input-output-hk/etc-client/blob/phase/3/fastDownload/src/main/scala/io/iohk/ethereum/mpt/MerklePatriciaTrie.scala#L104

I know they are not exactly the same but do you think it's worthy to take a second look and try to generalize it? I'm opening the discussion as I'm not convinced yet :stuck_out_tongue_closed_eyes:

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -1,30 +1,284 @@
...
+    override def toString: String = {
+      s"""MptBranch{
+         |childHashes: ${childHashes.map(e => Hex.toHexString(e.toArray[Byte])).mkString("(", ",\n", ")")}
+         |terminator: ${Hex.toHexString(terminator.toArray[Byte])}
+         |}
+       """.stripMargin
+    }
+  }
+
+  case class MptExtension(keyNibbles: ByteString, childHash: ByteString) extends MptNode {
```
AlanVerbner: Same as stated above for branch nodes

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -1,30 +1,284 @@
...
+    //todo do they exists in messages?
```
AlanVerbner: is this todo relevant?

src/test/scala/io/iohk/ethereum/network/p2p/messages/NodeDataSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+    Right(emptyStorageRoot)))
+
+  val encodedNodeData = RLPList(
+    RLPValue(encode(encodedLeafNode)),
+    RLPValue(encode(encodedBranchNode)),
+    RLPValue(encode(encodedExtensionNode)),
+    RLPValue(emptyEvmHash.toArray[Byte]),
+    RLPValue(emptyStorageRoot.toArray[Byte]))
+
+  "NodeData" should "be encoded properly" in {
```
AlanVerbner: Wouldn't it be better to test encode -> decode -> encode and check the data is still there?

Also, maybe it's a good idea to decode some known encoded data (taken from a real node or ethereumJ) and check everything is ok. This will prove that we don't have any compatibility issues.

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
adamsmo: I would like to keep it separate because it gives us freedom to change our internal implementation of MTP.
In my opinion, despite similarities those are 2 different things
one is content of received message and other is part of MPT that we store

src/test/scala/io/iohk/ethereum/network/p2p/messages/NodeDataSpec.scala
adamsmo: I will add test for encode -> decode -> encode but testing encode decode separate in relation to predefined output also makes sense because it can catch case where encde and decode makes same mistake
Actually reference value `encodedNodeData` was prepared base on messages that I got from geth node, I just put all cases in single list (which can happen)

AlanVerbner: @adamsmo :+1: .... I was talking about taking `encodedNodeData` a step further and define it directly as an encoded string like it's done in this test https://github.com/input-output-hk/etc-client/blob/phase/3/fastDownload/src/test/scala/io/iohk/ethereum/rlp/RLPSuite.scala#L486.

But as you have taken already defined data, I think ok too.

adamsmo: I think that encoding result as RLP makes tests also part of documentation (I see how message is encoded) where if I have byte array I would not have that. Also we have tests for RLP encoding so if RLP will be broken we will know.

AlanVerbner: @adamsmo I really like _encoding result as RLP makes tests also part of documentation_ . As long as we are 100% sure that the RLP has the exact same structure as the one generated by other client, i'm fine with it. I want to avoid problems, for example by say an extra RLPList both in our tests and in our decode function.

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
adamsmo: done

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -25,37 +28,174 @@ object PV63 {
...
+            import n._
+            RLPList(children.map(e => e.fold(_.hash, _.value)).map(e => byteStringEncDec.encode(e)) :+ (terminator: RLPEncodeable): _*)
+        }
+      }
+
+      override def decode(rlp: RLPEncodeable): MptNode = rlp match {
+        case rlpList: RLPList if rlpList.items.length == 17 =>
+          MptBranch(rlpList.items.take(16).map { bytes =>
+            val v = rlpDecode[ByteString](bytes)
+            if (v.nonEmpty && v.length < 32) {
```
ntallar: I'm not sure what is the purpose of calling `v.nonEmpty`, shouldn't `v` always be `nonEmpty`? Because the RLP encoding of anything is non empty and also all hashes have length 32.

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -25,37 +28,174 @@ object PV63 {
...
+      s"""MptExtension{
+         |key nibbles: $keyNibbles
+         |key nibbles length: ${keyNibbles.length}
+         |childHash: ${child.fold(a => s"Hash(${Hex.toHexString(a.hash.toArray[Byte])})", b => s"Value(${Hex.toHexString(b.value.toArray[Byte])})")}
+         |}
+       """.stripMargin
+    }
+  }
+
+  case class MptLeaf(keyNibbles: ByteString, value: Either[Account, ByteString]) extends MptNode {
```
ntallar: In which case would a `MptLeaf` node be storing something that's not an account? In this case the `ByteString` is used to store hashes of accounts maybe? (I'm not sure if that could be the case, at least from what I have seen in EthereumJ: https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/sync/FastSyncManager.java#L218)

adamsmo: for example for hash `72e55bd454f4d2caf57508b08c832d06d598eac31ee1d53b823a674bc1863452` geth node returns returns `MptBranch` with empty hashes, so it is this case
It means that those branches do not exist

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
adamsmo: for example for hash `72e55bd454f4d2caf57508b08c832d06d598eac31ee1d53b823a674bc1863452` geth node returns returns `MptBranch` with empty hashes, so it is this case
It means that those branches do not exist

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
adamsmo: in case of contract storage

adamsmo:
note:
before merge delete:  https://github.com/input-output-hk/etc-client/pull/25/files#diff-d88dc0e9076bb9b1c58875187ebcd4acR118

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -25,37 +28,174 @@ object PV63 {
...
+        }: _*)
       }
 
       override def decode(rlp: RLPEncodeable): NodeData = rlp match {
-        case rlpList: RLPList => NodeData(rlpList.items)
+        case rlpList: RLPList =>
+          NodeData(rlpList.items.map { e =>
+            Try {
+              val v = rawDecode(e: Array[Byte])
+              Left(MptNode.rlpEndDec.decode(v))
```
ntallar: Minor detail: Would it be the same to use `rlpDecode[MptNode](e: Array[Byte])` instead of `MptNode.rlpEndDec.decode(v)` and `rawDecode(e: Array[Byte])`? That way we could keep `rawDecode` as an internal function for now.

adamsmo: 👍  thanks
done

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -25,37 +28,173 @@ object PV63 {
...
+              Right(MptValue(v))
+            } else {
+              Left(MptHash(v))
+            }
+          }, byteStringEncDec.decode(rlpList.items(16)))
+        case RLPList(hpEncoded, value) =>
+          hpDecode(hpEncoded: Array[Byte]) match {
+            case (decoded, true) =>
+              MptLeaf(ByteString(decoded),
+                Try(Left(rlpDecode[Account](value: Array[Byte])))
```
AlanVerbner: @adamsmo both ethereumJ [1] and geth [2] work in a similar fashion. They keep track the requests they made (where they have what type of info they are asking either node or code) and once the response arrives, they manage to determine how to decode without the need of try and catch. Maybe we can think of working in a similar way, what do you think?

[1]: https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/sync/FastSyncManager.java#L349
[2]: https://github.com/ethereumproject/go-ethereum/blob/master/trie/sync.go#L154

adamsmo: doing it this way we need to provide additional information to `decode` method so we have to change that method and introduce coupling between send/receive and decode, 
it creates special case where 1 message will not be decoded same way as the rest

AlanVerbner: You can create a `NodeData` and `CodeData` and use different decoders. I might be saying something stupid and need to think this twice, but does that make sense?

adamsmo: What I was trying to say is that we have to have a way of determining when use which decoder, we can do it as it is done now by checking if byts string can be decoded to MptNode or by passing control to caller that keeps track of what should be where.
Instead of Try we can introduce if that will call some method to determine if ByteString can be decoded to MptNode, but that method underneath will do the same thing as Try does:
check if you are able to decode MptNode from given ByteString

AlanVerbner: I'm saying that instead of determining which type of node is it based on "try-catch" decoding we can have _something like a map_ where we store tuples that will might look like `(hash, codeRequest)` or `(hash, nodeRequest)` and once the response arrives, based on a lookup using hash as key we can call a decoder for a `codeRequestResponse` or one for a `nodeRequestResponse` 

(please read the stated code snippets as hints to make my idea more clear, not to say this is how should we actually implement it)

adamsmo: this map (or other data structure) has to be crated while sending request and than maintained and passed on receiving response, I am concerned about how to do that in our case where asynchronous actors are handling those tasks. Also currently signature of our decode method looks like this
`def decode(`type`: Int, payload: Array[Byte], protocolVersion: Version): Message`
so we have to introduce additional parameter (for example map) that will be used only to decode one type of messages - `NodeData` and will be unused in every other case

AlanVerbner: yes, you are right on that. We can return something like `case pv63.NodeData.code => RawNodeResponse(.....)` to be used after but this seems to start getting ugly. 

I'm still not convinced about the "try-catch" decoder either.

ntallar: An alternative to the `try-catch` approach can't be using pattern matching? Because in both cases where the `try-catch` is used one of the possibilities is an `RLPList` (`Account` or `MptNode`) and the other is an `RLPValue` (`ByteString`).

adamsmo: in our case all items in response are RLP values but some of them contains bytes that can be decoded to RLP list and some contains bytes the represents for example EVM code, so even in this  case if we try to decode all of the to RLP and that pattern match we will have exception for items with EVM code

adamsmo: maybe we can leave it as Bytes and add method that will try to produce account base on that bytes, this method can than return something like `Try[MptNode]`

AlanVerbner: @adamsmo if we leave them as bytes and keep track of requests, is the `Try` needed?

adamsmo: @AlanVerbner you mean in signature, nope it is not, I just added it there because conversion may fail, but signature of method can be
`def getMptNode(index: Int): MptNode`

AlanVerbner: ok..and will we call `getMptNode` if we are sure it's an mpt node then? Is it similar to https://github.com/input-output-hk/etc-client/pull/25#discussion_r98106916 ?

adamsmo: in some way yes, but we will still return `NodeData` but we will add functionality to that case class

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -25,37 +28,173 @@ object PV63 {
...
+            import n._
+            RLPList(RLPValue(hpEncode(keyNibbles.toArray[Byte], isLeaf = false)), child.fold(_.hash, _.value): RLPEncodeable)
+          case n: MptBranch =>
+            import n._
+            RLPList(children.map(e => e.fold(_.hash, _.value)).map(e => byteStringEncDec.encode(e)) :+ (terminator: RLPEncodeable): _*)
+        }
+      }
+
+      override def decode(rlp: RLPEncodeable): MptNode = rlp match {
+        case rlpList: RLPList if rlpList.items.length == 17 =>
```
LukasGasior1: can we make 17 a constant? maybe `BranchNodeChildLength`

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -25,37 +28,173 @@ object PV63 {
...
+
+              MptExtension(ByteString(decoded), child)
+          }
+        case _ =>
+          throw new RuntimeException("Cannot decode NodeData")
+      }
+    }
+  }
+
+  trait MptNode
```
LukasGasior1: how about sealed?

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
AlanVerbner: I think it's the best way without rethinking too much stuff.

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
adamsmo: 👍 
`BranchNodeChildLength` is 16
17th element is terminator element

LukasGasior1:
lgtm

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
AlanVerbner: @LukasGasior1 any comments regarding :arrow_up:  ?

LukasGasior1: How about keeping this message more low-level, i.e according to documentation https://github.com/ethereum/wiki/wiki/Ethereum-Wire-Protocol) it returns a list of Bytes (instead of Either[MptNode, ByteString]) - I think it's what Alan suggested as RawNodeResponse. Then we might have one place (actor) that requests NodeData and receives such responses - it can store a map of what was requested and decode accordingly without the need to modify the Message decode interface.

LukasGasior1: if we do it this way we can also move Mpt* classes there (because they're not really a part of protocol, are they?)

LukasGasior1: On the other hand try-catch approach seems much simpler (and doesn't really look bad for me).

mcsherrylabs: That particular Try smells pretty bad. We can talk on Monday about the alternatives even if that means some rework...

adamsmo: @AlanVerbner / @LukasGasior1 / @mcsherrylabs 
can you take a look at:
https://github.com/input-output-hk/etc-client/compare/feature/stateMptDownload...feature/stateMptDownloadTryFix

AlanVerbner: @adamsmo lgtm :metal:

PR #27, opened 2017-01-27
[EC-41] Volatile Memory and updates to DataWord

whysoserious:
Changes
* `Memory` (see comments in code for explanation)
* Bugfixes and changes in `DataWord`
* Dependency on `org.typelevel.cats`. We use it in EVM.


mcsherrylabs:
@whysoserious 
- please change the [EVM] to use the task number in youtrack.
- remove cats library dependency. Use of cats is still under review and until anything changes, no use of it please. Radek has already brought this up.

src/main/scala/io/iohk/ethereum/vm/Memory.scala
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  def load(addr: DataWord, size: DataWord): (ByteString, Memory) = doLoad(addr, size.intValue)
+
+  /** Returns a ByteString of a given size starting at the given address of the Memory.
+   * Underlying byte array is expanded and filled with 0's if addr + size exceeds size
+   * of the memory.
+   * This method may throw OOM.
+   */
+  private def doLoad(addr: DataWord, size: Int): (ByteString, Memory) = {
```
LukasGasior1: as there's a gas cost associated with memory expansion, does it make sense to expand it here? We could just load as much bytes as we have and pad to `size` with zeros (without changing underlying ByteString)

LukasGasior1: well I guess we can't just choose what to do but instead see what other implementations do (to have the same gas usage)

whysoserious:
> - please change the [EVM] to use the task number in youtrack.

Done.

> - remove cats library dependency. Use of cats is still under review and until anything changes, no use of it please. Radek has already brought this up.

This was our common decision to use `cats` in a very limited scope. We use functionality only from `cats.syntax.either` which makes `scala.Either` much more convenient to use. After upgrading to Scala 2.12 this dependency can be removed. We definitely don't want to use other features of this lib.
To remove that dependency we need Radek to agree on that and also a good reasons for doing that.

src/main/scala/io/iohk/ethereum/vm/Memory.scala
whysoserious: We'll add gas related functionalities in follow-up PRs

whysoserious:
@mcsherrylabs ^^

whysoserious:
Hmm, I just realized that we might completely get rid of `Either`s in EVM code. Let's leave it till Monday. Summoning @rtkaczyk to express his opinion.

src/main/scala/io/iohk/ethereum/vm/Memory.scala
rtkaczyk: @LukasGasior1 the semantics of memory access in EVM are such that when you access a previously unallocated address the memory gets expanded accordingly and the caller pays the proportionate gas. In other words, loading from memory has the same effect on its size (and gas consumption) as storing in it.

rtkaczyk:
@mcsherrylabs regarding cats: we're only using extensions to `scala.util.Either` that enable right-bias and some minor goodies. The right-bias has been added to `Either` in Scala 2.12 anyway - I guess cats could go when we upgrade. For now, I'd sooner make my own monad-like class for handling errors, rather than use bare `Either` from Scala 2.11 through left/right projections.

And regarding the usage `Either` altogether - I still don't have the full picture of all the errors that can occur in the EVM. However they are much less common than previously anticipated. For example, to our surprise, actions like division by zero or accessing unallocated memory _do not_ result in errors. 
  Because of the reduced number of erroneous situations, I think there might be an easier (though less generic) way to handle them than transformations on `Either`. I'm not ready to make that call yet though.

src/main/scala/io/iohk/ethereum/vm/Memory.scala
LukasGasior1: That makes sense, thanks

LukasGasior1:
LGTM

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -49,13 +44,12 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
    *  Output ByteString is padded with 0's from the left side up to MaxLength bytes.
    */
   lazy val bytes: ByteString = {
-    val bs: ByteString = ByteString(n.toByteArray)
+    val bs: ByteString = ByteString(n.toByteArray).takeRight(32)
```
ntallar: Minor detail: shouldn't this be `takeRight(MaxLength)`?

ntallar:
LGTM

src/main/scala/io/iohk/ethereum/vm/Memory.scala
```diff
@@ -0,0 +1,89 @@
+package io.iohk.ethereum.vm
+
+import akka.util.ByteString
+import cats.syntax.either._
```
rtkaczyk: Please take a note of my comment before approving this PR. `cats` is not needed here, and we're going to remove it in the main PR.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
whysoserious: Absolutely.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -78,9 +72,14 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
 
   def /(that: DataWord): DataWord = DataWord(this.n / that.n)
 
+  def **(that: DataWord): DataWord = DataWord(this.n.modPow(that.n, Modulus))
```
adamsmo: as I see this is not tested in `DataWordSpec`

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -78,9 +72,14 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
...
+
   def compare(that: DataWord): Int = this.n.compare(that.n)
 
-  def intValue: Int = n.intValue
+  /**
+    * @return an Int with MSB=0, thus a value in range [0, Int.MaxValue]
+    */
+  def intValue: Int = n.intValue & Int.MaxValue
```
adamsmo: is this necessary if we have `require(n >= 0 && n < Modulus, s"Invalid word value: $n")`
maybe just def intValue: Int = n.intValue
in case of integer overflow it will return unexpected values for example:
```
scala> BigInt("3000000000").intValue
res11: Int = -1294967296
```
```
scala> (Int.MaxValue & BigInt("3000000000").intValue)
res12: Int = 852516352
```
but this case is covered by `< Modulus` so do we need this `& Int.MaxValue` ?

rtkaczyk: `Modulus > Int.MaxValue` so what do you mean by "this case is covered"?

Anyway, the `intValue` is used for accessing the byte code (usages in `Memory` are irrelevant because of gas calculation beforehand). But accessing code at undefined position should return zeroes anyway, so perhaps we shouldn't worry about this going negative, @whysoserious ?

rtkaczyk: Not that I see anything wrong with restricting it to non-negatives.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
rtkaczyk: good catch

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
adamsmo: @rtkaczyk you are right it is not covered

whysoserious: Let's keep it as it is for now. We'll return to it in a follow-up PR when we'll have to implement division on both signed and unsigned ints.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
whysoserious: @adamsmo I added missing test.

PR #29, opened 2017-01-31
[EC-35] Add support for fast download

adamsmo:
This PR is about adding actor that will perform fast download and store downloaded data in IODB
This PR depends on https://github.com/input-output-hk/etc-client/pull/31


src/main/resources/application.conf
```diff
@@ -33,6 +33,11 @@ etc-client {
     dao-fork-block-hash = "94365e3a8c0b35089c1d1195081fe7489b528a84b22199c916180db8b28ade7f"
   }
 
+  fast-sync {
```
AlanVerbner: I know this is a WIP but take a look at this https://github.com/ethereumproject/go-ethereum/blob/master/eth/downloader/downloader.go#L43

```go
MaxHashFetch    = 512 // Amount of hashes to be fetched per retrieval request
MaxBlockFetch   = 128 // Amount of blocks to be fetched per retrieval request
MaxHeaderFetch  = 192 // Amount of block headers to be fetched per retrieval request
MaxSkeletonSize = 128 // Number of header fetches to need for a skeleton assembly
MaxBodyFetch    = 128 // Amount of block bodies to be fetched per retrieval request
MaxReceiptFetch = 256 // Amount of transaction receipts to allow fetching per request
MaxStateFetch   = 384 // Amount of node state values to allow fetching per request
```

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
```diff
@@ -0,0 +1,304 @@
...
+  import context.{dispatcher, system}
+
+  val GenesisBlockNumber = 0
+  val EmptyAccountStorageHash = ByteString(Hex.decode("56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421"))
+  val EmptyAccountEvmCodeHash = ByteString(Hex.decode("c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"))
+
+  def handleTerminated: Receive = {
+    case _: Terminated =>
+      log.info("FastSync actor terminated")
+      context stop self
```
LukasGasior1: as I have already started working with this code, I'll let myself add some comments even thought it's still WIP.

`Terminated` is received when one of the actors you're watching (with `contet.watch`) is stopped. I don't see FastSyncActor watching any other actor. Am I missing something?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -261,6 +270,19 @@ class PeerActor(
...
+        //todo get storage form somewhere?
+        val storage = FastSyncActor.Storage(
+          new BlockHeadersStorage(EphemDataSource()),
+          new BlockBodiesStorage(EphemDataSource()),
+          new ReceiptStorage(EphemDataSource()),
+          new MptNodeStorage(EphemDataSource()),
+          new EvmCodeStorage(EphemDataSource())
+        )
+        val fastSyncActor = context.actorOf(FastSyncActor.props(self, storage), UUID.randomUUID().toString)
+        context watch fastSyncActor
```
LukasGasior1: why does PeerActor need to watch fastSyncActor? `handleTerminated` should be updated to consider that (currently it assumes it's rlpxConnection that is being terminated).

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  override def receive: Receive = handleTerminated orElse {
+    case StartSync(targetBlockHash) =>
+      //TODO check DB if it is not empty delete everything
+      peerActor ! PeerActor.Subscribe(Set(NodeData.code, Receipts.code, BlockBodies.code, BlockHeaders.code))
```
LukasGasior1: just a note: we should make sure PeerActor is in `Handshaked` status before starting fast sync. otherwise we might receive not the block we expect in `waitForTargetBlockHeader` because PeerActor asks for dao block during the handshake (and that would be the block we'd also get, possibly before the one that was requested here)

alternatively we can check `targetBlockHash` in `waitForTargetBlockHeader`

adamsmo: starting fast sync is possible only in state that we handle messages
https://github.com/input-output-hk/etc-client/pull/29/files#diff-d88dc0e9076bb9b1c58875187ebcd4acR260
am I right?

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
```diff
@@ -0,0 +1,297 @@
...
+      peerActor ! PeerActor.SendMessage(GetBlockHeaders(Right(targetBlockHash), 1, 0, reverse = false))
+      context become waitForTargetBlockHeader(targetBlockHash)
+  }
+
+  def waitForTargetBlockHeader(targetBlockHash: ByteString): Receive = {
+    case MessageReceived(BlockHeaders(blockHeaders)) if blockHeaders.nonEmpty =>
+      peerActor ! PeerActor.SendMessage(GetNodeData(Seq(blockHeaders.head.stateRoot)))
```
AlanVerbner: I'm wondering if it's ok to download the state before getting close to the target hash (in case of an error + abort for example)

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
```diff
@@ -0,0 +1,297 @@
...
+      peerActor ! PeerActor.SendMessage(GetBlockHeaders(Left(GenesisBlockNumber + 1), FastSync.BlocksPerMessage, 0, reverse = false))
```
AlanVerbner: Where are we storing the Genesis block in the db?

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
```diff
@@ -0,0 +1,297 @@
...
+import io.iohk.ethereum.db.storage._
+import io.iohk.ethereum.domain.BlockHeader
+import io.iohk.ethereum.network.FastSyncActor._
+import io.iohk.ethereum.network.PeerActor.MessageReceived
+import io.iohk.ethereum.network.p2p.messages.PV62._
+import io.iohk.ethereum.network.p2p.messages.PV63._
+import io.iohk.ethereum.utils.Config.FastSync
+import org.spongycastle.util.encoders.Hex
+
+class FastSyncActor(
```
AlanVerbner: There is no timeout when sending messages from this actor too peers. I'm not saying in this version we should include that but maybe we need to consider this for future iterations, what do you think @adamsmo ?

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
```diff
@@ -40,3 +43,14 @@ case class BlockHeader(
...
     }
   }
+
+object BlockHeader {
+    /**
+      * calculates blockHash for given block header
+      * @param header - block header for which hash will be calculated
+      * @return - hash that can be used to get block bodies / receipts
+      */
+    def hash(header: BlockHeader): ByteString = {
```
ntallar: This is used a couple of times (and sometimes with the same headers), I'm wondering if maybe we could define it as a `lazy val` on the `BlockHeader` class?

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
adamsmo: I thin yes, because blockchain is easy to download from other pears and easy to resume download after failure.
This way we parallelize download and we are syncing faster

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
adamsmo: we do not store it now, but we can add it as constant, I think even in code because it is part of specification and will not change

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
adamsmo: I think it would be improvement, we can add ticket for that

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
adamsmo: I was talking with @whysoserious and he is suggesting that we keep domain as pure case classes, but we can talk about that

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
LukasGasior1: @adamsmo @AlanVerbner I've already started EC-64 based on this code. I'll add the timeouts there.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -261,6 +270,18 @@ class PeerActor(
 
       case GetStatus =>
         sender() ! StatusResponse(Handshaked)
+
+      case StartFastSync(targetHash) =>
+        //todo get storage form somewhere?
```
AlanVerbner: Are we going to change this in this PR?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -261,6 +270,18 @@ class PeerActor(
...
+        val storage = FastSyncActor.Storage(
+          new BlockHeadersStorage(EphemDataSource()),
```
AlanVerbner: Will you wait for the `LevelDB` PR or will this one be meged with Ephem?

adamsmo: I think I would leave it as Ephem because it is currently working version and changing it level db is easy

AlanVerbner: @adamsmo Actually I did a lot of _wiring_ work in order to have the `GetBlockByid` so it might be better not to change it. Will tell more during the VSU  :smile:

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
adamsmo: changed

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
adamsmo: It is still `EphemDataSource` in latest version

adamsmo: @AlanVerbner To show block dump we will need to change this to Level DB

AlanVerbner: @adamsmo yup...i've have everything almost ready to do so :smile:

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
AlanVerbner: @adamsmo Will you create the task for this? It doesn't need to be done in this PR imo but lets not forget about it

src/test/scala/io/iohk/ethereum/vm/DataWordSpec.scala
```diff
@@ -1,10 +1,10 @@
 package io.iohk.ethereum.vm
 
 import akka.util.ByteString
+import io.iohk.ethereum.vm.DataWord._
```
AlanVerbner: @adamsmo why has this changed?

adamsmo: when merging I optimized imports, and it was picked, will revert that

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
adamsmo: yes I will

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
```diff
@@ -0,0 +1,297 @@
+package io.iohk.ethereum.network
```
mcsherrylabs: FastSyncActor moves out to 'controller' layer in this pr?

ntallar:
LGTM

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
mcsherrylabs: We are fastDownloading from every peer... does this mean we don't need a strategy for picking a peer (EC-64) or this is a temporary measure?

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
```diff
@@ -0,0 +1,297 @@
...
+
+        case (EvmCodeHash(hash), idx) =>
+          val evmCode = m.values(idx)
+          val msg =
+            s"got EVM code: ${Hex.toHexString(evmCode.toArray[Byte])}"
+          log.info(msg)
+          evmStorage.put(hash, evmCode)
+        case (StorageRootHash(hash), idx) =>
+          val rootNode = m.getMptNode(idx)
+          val msg =
```
mcsherrylabs: Is there a reason for val msg to exist?

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
```diff
@@ -0,0 +1,297 @@
...
+  private def handleContractMptNode(hash: ByteString, mptNode: MptNode) = {
+    mptNode match {
+      case n: MptLeaf =>
+        log.info("Got contract leaf node: {}", n)
+        mptNodeStorage.put(n)
+
+      case n: MptBranch =>
+        log.info("Got contract branch node: {}", n)
+        val hashes = n.children.collect { case Left(MptHash(childHash)) => childHash }.filter(_.nonEmpty)
+        println(hashes)
```
mcsherrylabs: Ahem... did scalastyle allow this println? :)

adamsmo: looks like it does, will change that, maybe we should investigate this further?

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
adamsmo: nope

src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala
mcsherrylabs: I'll bring it up tomorrow at vsu.

PR #30, opened 2017-01-31
[EC-47] Add possibility to subscribe to messages from peer

LukasGasior1:


ntallar:
LGTM!

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -207,21 +209,44 @@ class PeerActor(nodeInfo: NodeInfo) extends Actor with ActorLogging {
...
+        rlpxConnection.sendMessage(s.message)(s.enc)
+
+      case GetStatus =>
+        sender() ! StatusResponse(Handshaked)
+    }
+
+    def notifySubscribers(message: Message): Unit = {
+      val subscribers = messageSubscribers.filter(_.messageCodes.contains(message.code))
+      val toNotify = subscribers.map(_.ref).toSet
+      toNotify.foreach { _ ! MessageReceived(message) }
```
mcsherrylabs: Do subscribers ever need to reply to the Message I wonder...? 
If we change this to 'forward' they can reply assuming the sender is the connection handler. 
I can't think of any scenario that requires reply yet...

PR #31, opened 2017-01-31
[EC-42] and [EC-46] Receipt and node storage

ntallar:
Includes:
- Refactoring of `DataSource` files and inclusion of `KeyValueStorage` class to store keys and values of different types.
- Removed `rootHash` as a parameter from `DataSource` as we are not currently dealing with different `IODB` versions.
- Use of `DataSource` with logical namespaces separated through a different prefix.
- Node `[EC-46]`, code and receipt storage `[EC-42]` (decided to merge the tasks into one PR as they were closely related).
- MerklePatriciaTree uses `NodeStorage` instead of `DataSource`.
- _Important_: Due to the different use of storage by the `MerklePatriciaTrie` and FastSync, two different node storages were included (`NodeStorage` and `MptNodeStorage`) one for each of the different `MPT` node classes.


ntallar:
So far using the same `DataSource` for different `KeyValueStorage` could make some of them work inappropriately, do to possible conflicts with the keys used. This can be solved by using different prefixes for the different types of objects as is done in Geth (https://github.com/ethereum/go-ethereum/blob/master/core/database_util.go#L43).

adamsmo:
@ntallar looks good, but what in case where I would like to store receipt and node that have same hash, or extending it to also store EVM code.
Another thing is using this to persist contract storages where we have multiple MPTs do you think it will fit?
I know that probability of collision is extremely low.

src/main/scala/io/iohk/ethereum/db/ReceiptStorage.scala
```diff
@@ -0,0 +1,17 @@
...
+object ReceiptStorage {
+  type BlockHash = Array[Byte]
+  type ReceiptStorage = KeyValueStorage[BlockHash, Seq[Receipt]]
+
+  def apply(dataSource: DataSource): ReceiptStorage = new ReceiptStorage(
+    dataSource = dataSource,
+    keySerializer = identity,
+    valueSerializer = rlpEncode(_)(seqEncDec[Receipt]),
+    valueDeserializer = rlpDecode[Seq[Receipt]](_)(seqEncDec[Receipt])
+  )
```
adamsmo: why no `put` method here?

ntallar: A `put` method similar to the one implemented in `NodeStorage`? The problem is that in that case we can't derive the `BlockHash` from the `Seq[Receipt]`. There is also a method `def put(key: K, value: V)` defined in the `KeyValueStorage` class that can be used in that case.

ntallar:
@adamsmo I'm still looking at how `EthereumJ` deals with that problem, but `Geth` deals with this problem by adding a different prefix to each object type  (https://github.com/ethereum/go-ethereum/blob/master/core/database_util.go#L43), so I think we should also do something like that to prevent possible conflicts.

> Looks good, but what about case when MPT node, receipt or EVM code (I assume that we will use same solution for that) has same hash?

adamsmo:
👍

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
```diff
@@ -0,0 +1,26 @@
+package io.iohk.ethereum.db.dataSource
+
+trait DataSource {
+
+  type Key = Array[Byte]
+  type Value = Array[Byte]
```
whysoserious: Could you add a few more words in the scaladoc explaining why a `Key`s and `Value`s are `Array[Byte]`s and not `ByteString`s or any other type?

whysoserious: We use `ByteString`s almost everywhere else in the system. Imagine a situation when a new storage would use `ByteString`s as well. Then we would have to convert whole data passing through a `DataSource` to `Array[Byte]` and back to `ByteString`.

src/main/scala/io/iohk/ethereum/db/storage/CodeStorage.scala
```diff
@@ -0,0 +1,22 @@
+package io.iohk.ethereum.db.storage
+
+import akka.util.ByteString
+import io.iohk.ethereum.db.dataSource.DataSource
+
+import CodeStorage._
+
+class CodeStorage(val dataSource: DataSource) extends KeyValueStorage[CodeHash, Code] {
```
whysoserious: Add some docs pls :)

src/main/scala/io/iohk/ethereum/db/storage/CodeStorage.scala
```diff
@@ -0,0 +1,22 @@
...
+  type T = CodeStorage
```
whysoserious: Is it used anywhere?

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
```diff
@@ -0,0 +1,26 @@
...
+
+  /**
+    * This function obtains the associated value to a key, if there exists one.
+    *
+    * @param namespace which will be searched for the key.
```
ntallar: The main reason is that `IODB` requires all the keys to be of the same size, so as a single `Byte` is enough (at least for now) I decided `namespace: Byte`. If it is preferable I can change it or at least add a few words on this explanation.

src/it/scala/io/iohk/ethereum/db/IodbDataSourceIntegrationSuite.scala
```diff
@@ -1,27 +1,29 @@
...
   def putMultiple(dataSource: DataSource, toInsert: Seq[(Array[Byte], Array[Byte])]): DataSource = {
     toInsert.foldLeft(dataSource){ case (recDB, keyValuePair) =>
-      recDB.update(DefaultRootHash, Seq(), Seq(keyValuePair))
+      recDB.update(OtherNamespace, Seq(), Seq(keyValuePair))
     }
   }
 
   test("IodbDataSource insert"){
-    forAll(seqByteArrayOfNItemsGen(KeySize)) { unFilteredKeyList: Seq[Array[Byte]] =>
+    forAll(seqByteArrayOfNItemsGen(KeySize - 1)) { unFilteredKeyList: Seq[Array[Byte]] =>
```
ntallar: You are right, thanks for spotting that :)

src/main/scala/io/iohk/ethereum/db/storage/CodeStorage.scala
ntallar: Yes, this is required by the `KeyValueStorage` trait as it is used in `def update(toRemove: Seq[K], toUpdate: Seq[(K, V)]): T` and other functions that return a modified `KeyValueStorage`. This is also done in `MptNodeStorage` and the other different `KeyValueStorage` implementations.

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
mcsherrylabs: (If you don't need ByteString in your type, you can use IndexedSeq[Byte] as ByteString is a IndexedSeq[Byte])

src/main/scala/io/iohk/ethereum/db/storage/CodeStorage.scala
ntallar: Done!

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
ntallar: Done! In the last commit I replaced the use of `Array[Byte]` for `IndexedSeq[Byte]` as the `Key` and `Value` of `DataSource`

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
whysoserious: > IODB requires all the keys to be of the same size

I'd take care of this contraint in `IodbDataSource`. Apart from that `Key`, being an `IndexedSeq[Byte]` does not enforce fixed length of its values.

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
```diff
@@ -0,0 +1,26 @@
...
+    * @return the value associated with the passed key.
+    */
+  def get(namespace: Byte, key: Key): Option[Value]
+
+  /**
+    * This function updates the DataSource by deleting and inserting new (key-value) pairs.
+    *
+    * @param namespace from which the (key-value) pairs will be removed and inserted.
+    * @param toRemove which includes all the keys to be removed from the DataSource.
+    * @param toUpdate which includes all the (key-value) pairs to be inserted into the DataSource.
```
whysoserious: I'd change the name of this val to `toInsert`.

src/main/scala/io/iohk/ethereum/db/dataSource/IodbDataSource.scala
```diff
@@ -0,0 +1,35 @@
...
+import io.iohk.iodb.{ByteArrayWrapper, LSMStore}
+
+class IodbDataSource(lSMStore: LSMStore) extends DataSource {
+
+  import IodbDataSource._
+
+  override def get(namespace: Byte, key: Key): Option[Value] =
+    lSMStore.get(ByteArrayWrapper((namespace +: key).toArray)).map(v => v.data.toIndexedSeq)
+
+  override def update(namespace: Byte, toRemove: Seq[Key], toUpdate: Seq[(Key, Value)]): DataSource = {
```
whysoserious: What will happen if `Key`s in `toUpdate` will have a varying length? Do we need to handle that?

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
```diff
@@ -0,0 +1,26 @@
+package io.iohk.ethereum.db.dataSource
+
+trait DataSource {
```
whysoserious: What's the reason for not implementing an `insert` method separately from `update`? Just asking, please do not implement that :)

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
whysoserious: So, if I am getting this design right then I'd add a new `type Namespace = IndexedSeq[Byte]` and make all changes in code resulting from that.

If it's going to take you more than one hour, then let's keep it as it and move forward :)

src/it/scala/io/iohk/ethereum/db/IodbDataSourceIntegrationSuite.scala
```diff
@@ -1,39 +1,43 @@
...
 
+import akka.util.ByteString
+import io.iohk.ethereum.db.dataSource.{DataSource, IodbDataSource}
+
 class IodbDataSourceIntegrationSuite extends FunSuite
   with PropertyChecks
   with ObjectGenerators {
 
-  val KeySize: Int = 32
+  val KeySizeWithoutPrefix: Int = 32
```
whysoserious: Scenarios to test:
* Insert many keys in one `recDB.update(..)` method. In `IodbDataSource insert` they are inserted separately.
* Update many keys in one `recDB.update(..)` method.
* Try to insert/update/get keys with invalid length.

Please don't implement these tests if we don't want to use IODB right now.

src/test/scala/io/iohk/ethereum/db/storage/KeyValueStorageSuite.scala
```diff
@@ -0,0 +1,62 @@
...
+import io.iohk.ethereum.db.dataSource.{DataSource, EphemDataSource}
+import io.iohk.ethereum.rlp.RLPImplicits._
+import io.iohk.ethereum.rlp.{decode => rlpDecode, encode => rlpEncode}
+import org.scalacheck.Gen
+import org.scalatest.FunSuite
+import org.scalatest.prop.PropertyChecks
+
+import scala.util.Random
+
+class KeyValueStorageSuite extends FunSuite with PropertyChecks with ObjectGenerators{
```
whysoserious: This spec doesn't test all public functions in a `KeyValueStorage`. Specs testing implementations of a `KeyValueStorage` are also not complete.

src/test/scala/io/iohk/ethereum/db/storage/KeyValueStorageSuite.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  test("Insert ints to KeyValueStorage") {
+    forAll(Gen.listOfN(iterationsNumber, Gen.listOf(intGen))) { listOfListOfInt =>
+
+      val keyValueStorage = listOfListOfInt.foldLeft(initialIntStorage){ case (recKeyValueStorage, intList) =>
+        recKeyValueStorage.update(Seq(), intList.zip(intList))
+      }
+
+      listOfListOfInt.flatten.foreach{ i =>
+        assert(keyValueStorage.get(i).contains(i))
```
whysoserious: Just a remark, `.get` method is untested and used in an assertion at the same time.

src/it/scala/io/iohk/ethereum/db/IodbDataSourceIntegrationSuite.scala
mcsherrylabs: (Fyi - We're using IODB, we're just adding a second data source)

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
```diff
@@ -0,0 +1,26 @@
...
+  type Value = Array[Byte]
+
+  /**
+    * This function obtains the associated value to a key, if there exists one.
+    *
+    * @param namespace which will be searched for the key.
+    * @param key
+    * @return the value associated with the passed key.
+    */
+  def get(namespace: Byte, key: Key): Option[Value]
```
mcsherrylabs: Do you want a `def apply(ns: Byte, key: Key) : Value` for cases where the value is known to exist...?

whysoserious: I would avoid adding one more place which potentially could throw NPEs. We have `Option.getOrElse` and `Option.get` for such scenarios.

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
ntallar: An `insert` function is not yet implemented as we didn't have a reason to do so. It can be very easily implemented if needed, I can add it if you think it will be needed.

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
mcsherrylabs: This only makes sense if the result is *always* expected to be optional. Which for a DataSource, I don't think we can say that's the case. (or can we?) It's fine to create an operation with the pre condition that a value must exist for the given key. E.g. Map.apply(k: K): V

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
whysoserious: Having two separate functions for inserts and updates instead of one seems to be more intuitive choice. However, I don't have a strong opinion about that so let's keep this API in current state.

AlanVerbner: @whysoserious we have found at least two data sources (iodb and levelDb) that allow to do batch modifications. Also, we found that we weren't using single items insertions. That being said, if we think it's needed in the future we can easily add it then, how does it sound?

mcsherrylabs: +1

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
ntallar: I'm not sure if that would also be a proper name as the keys from `toUpdate` could already be on the `DataSource`, resulting in the update of its value and not insertion. I will improve the description on that function to include that, maybe we could have a name like `toInsertOrUpdate` that will be more general?

mcsherrylabs: (I've seen `toUpsert' in places. Just adding that information... :) )

src/main/scala/io/iohk/ethereum/db/dataSource/IodbDataSource.scala
ntallar: I rechecked how `IODB` handles that problem, and although it throws an exception in that case, it only does so after some version updates and inserting the keys which have the correct length. I agree we should handle this but it will need some refactoring that's also necessary for `EC-60`. So what do you think on leaving it right now as it is, so as to not delay much more other PRs, and I can add the handling of this in my next PR for `EC-60` or maybe in another PR?

ntallar:
@whysoserious I have just pushed a commit that does some refactoring and adds tests that were missing (@mcsherrylabs I also added an apply function to `DataSource`). If there's anything missing please let me know :)

src/it/scala/io/iohk/ethereum/db/IodbDataSourceIntegrationSuite.scala
```diff
@@ -0,0 +1,148 @@
...
+
+  test("IodbDataSource insert/update with invalid length") {
+    forAll(seqByteStringOfNItemsGen(KeySizeWithoutPrefix)) { unFilteredKeyList: Seq[ByteString] =>
+      //create temporary dir
+      val dir = File.createTempFile("iodbInvalidLength", "iodbInvalidLength")
+      dir.delete()
+      dir.mkdir()
+
+      val keyList = unFilteredKeyList.take(KeyNumberLimit)
+      val (keysLeft, keysToInsert) = Random.shuffle(keyList).splitAt(Gen.choose(0, keyList.size/2).sample.get)
```
whysoserious: Why `Random.shuffle`? `unFilteredKeyList` is already randomized.
Apart from that, in case of failure this test will be hard to reproduce.

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
whysoserious: Yes, that absolutely makes sense.

PR #32, opened 2017-01-31
Feature/evm sketch

whysoserious:


rtkaczyk:
@whysoserious let's add the missing tests, then we can prepare a proper non-WIP PR

PR #33, opened 2017-01-31
Remove old fake types implementations

AlanVerbner:
## Description

- Some Test classes added to test RLP encoding were replaced by real implementations (made after some sprints)
- There are still missing ones like `Block`. 
- Test implementations were renamed to `TestXXXXX` to avoid confusions while coding
- `FIXME` comments were added so we can replace them in the future


whysoserious:
+88 −229 that's nice!

ntallar:
LGTM!

PR #34, opened 2017-02-01
[EC-41] EVM sketch

whysoserious:
In this PR we present a sizeable sketch of the EVM. Over 70% of all the opcodes were implemented, which enabled us to run simple programs compiled with Solidity. Please take a note of the following:

- all code execution consists of transformations of immutable state
- the `VM` can be viewed as a function of `ProgramContext => ProgramResult`
- each `OpCode` can be viewed as a function of `ProgramState => ProgramState`
- in future iterations `ProgramContext` will be extended to provide the necessary environmental information (account, block, transaction)
- the usages of `Either` in the OpCodes may be replaced with a more fine-tuned flow control (only an internal change to each `OpCode`)
- gas consumption will be added in a future iteration


src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -0,0 +1,507 @@
...
+  }
+}
+
+case object ADD extends BinaryOp(0x01)(_ + _)
+
+case object MUL extends BinaryOp(0x02)(_ * _)
+
+case object SUB extends BinaryOp(0x03)(_ - _)
+
+case object DIV extends BinaryOp(0x04)((a, b) => if (b != 0) a / b else DataWord(0))
```
AlanVerbner: `DataWord.Zero` can be used here ?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -0,0 +1,507 @@
...
+
+case object EXP extends BinaryOp(0x0a)(_ ** _)
+
+case object LT extends BinaryOp(0x10)((a, b) => DataWord(a < b))
+
+case object EQ extends BinaryOp(0x14)((a, b) => DataWord(a == b))
+
+case object ISZERO extends UnaryOp(0x15)(a => DataWord(a == 0))
```
AlanVerbner: Same here about `DataWord.Zero`

rtkaczyk:
I have added class documentation and removed WIP label. The documentation may of course still be improved if pointed out

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
rtkaczyk: So `a == 0` vs `a == DataWord.Zero` is a matter of conciseness/readability vs slight efficiency boost. I went for the former cause comparing `DataWord` to numbers feels natural to me.

Also with certain tweaks we could even write `UnaryOp(0x15)(_ == 0)` (or `UnaryOp(0x15)(_ == DataWord.Zero)`), but I'm not sure if it's worth the effort.

I think it boils down to a matter of taste, so I would appreciate a second opinion. Happy to change it if others feel the same way @LukasGasior1 @whysoserious

src/main/scala/io/iohk/ethereum/utils/EitherExtensions.scala
```diff
@@ -0,0 +1,33 @@
+package io.iohk.ethereum.utils
+
+import scala.util.{Either, Left, Right}
+
+object EitherExtensions {
+
+  implicit class RightBiasedEither[A, B](val either: Either[A, B]) extends AnyVal {
```
LukasGasior1: we can add a comment to remove it once we're on scala 2.12

src/main/scala/io/iohk/ethereum/vm/ProgramState.scala
```diff
@@ -0,0 +1,58 @@
...
+  * @param pc program counter - an index of the opcode in the program to be executed
+  * @param returnData data to be returned from the program execution
+  * @param halted a flag to indicate program termination
+  * @param error indicates whether the program terminated abnormally
+  */
+case class ProgramState(
+  context: ProgramContext,
+  stack: Stack = Stack.empty(),
+  memory: Memory = Memory.empty,
+  storage: Storage,
```
LukasGasior1: what is the difference between this and `context.storage`?

src/main/scala/io/iohk/ethereum/vm/Stack.scala
```diff
@@ -0,0 +1,95 @@
...
+      copy(updated).asRight
+    }
+  }
+
+  def size: Int = underlying.size
+
+  def toSeq: Seq[DataWord] = underlying.reverse
+
+  override def equals(that: Any): Boolean = that match {
+    case that: Stack => this.underlying == that.underlying && (22244 / 2 == 11122)
```
LukasGasior1: why is this needed: `&& (22244 / 2 == 11122)`?

src/main/scala/io/iohk/ethereum/vm/Stack.scala
```diff
@@ -0,0 +1,95 @@
...
+  def empty(maxSize: Int = DefaultMaxSize): Stack =
+    new Stack(Vector(), maxSize)
+}
+
+//TODO: consider a List with head being top of the stack (DUP,SWAP go at most the depth of 16)
+/**
+  * Stack for the EVM. Instruction pop their arguments from it and push their results to it.
+  * The Stack handles overflow and underflow errors.
+  */
+class Stack private(private val underlying: Vector[DataWord], val maxSize: Int) {
```
LukasGasior1: could it be a case class?

src/main/scala/io/iohk/ethereum/vm/Storage.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  val Empty: Storage = new Storage()
+}
+
+/** Persistent storage of a transaction. May be viewed as a map of 256 bit keys
+  * to 256 bit values.
+  * This is just a representation of storage changes occuring during program execution. The
+  * actual persistence is not this class's resposibility.
+  */
+class Storage private(private val underlying: Map[DataWord, DataWord] = Map()) {
```
LukasGasior1: could it be a case class?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -0,0 +1,507 @@
...
+      nextPos = if (cond != 0) pos.intValue else state.pc + 1
+    } yield state.withStack(stack1).goto(nextPos)
+
+    updatedState.valueOr(state.withError)
+  }
+}
+
+case object JUMPDEST extends OpCode(0x5b, 0, 0) {
+  def execute(state: ProgramState): ProgramState = {
+    //TODO: what is it for, really?
```
AlanVerbner: I think this is used to verify that we are doing JUMP to a valid destinations. Is that checked in our implementation? 

In parity and ethereumJ there is a specific error thrown in that case. Some refs:
- https://github.com/ethcore/parity/blob/master/ethcore/src/evm/interpreter/shared_cache.rs#L66
- https://github.com/ethcore/parity/blob/master/ethcore/src/evm/interpreter/mod.rs#L107
- https://github.com/ethcore/parity/blob/master/ethcore/src/evm/interpreter/mod.rs#L156

rtkaczyk: Yes, we actually already know what it is. This is just a bit inaccurate comment to remember to do the verification. But it will be a separate PR. For now I can update the comment.

src/main/scala/io/iohk/ethereum/vm/ProgramState.scala
LukasGasior1: I guess context.storage is the initial context supplied to VM, while this field changes throughout code execution, am I right?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
whysoserious: I am ok with any approach.

src/main/scala/io/iohk/ethereum/vm/ProgramState.scala
whysoserious: Yes. I'll update the documentation.

src/main/scala/io/iohk/ethereum/vm/Stack.scala
whysoserious: lol

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
AlanVerbner: @rtkaczyk works for me!

src/main/scala/io/iohk/ethereum/vm/Stack.scala
whysoserious: It's a matter of taste. I didn't have a strong opinion; Radek chose ~case~. Let's keep it as it is if you don't mind.

src/main/scala/io/iohk/ethereum/vm/Storage.scala
whysoserious: It's a matter of taste. I didn't have a strong opinion; Radek chose case. Let's keep it as it is if you don't mind.

src/main/scala/io/iohk/ethereum/vm/ProgramState.scala
whysoserious: Actually, I think that line 13. in the file explains that.

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
whysoserious: Leaving this line intact..

whysoserious:
Ok, this PR is ready for another review.

src/main/scala/io/iohk/ethereum/vm/Stack.scala
rtkaczyk: Haha, I was checking scalastyle behaviour and forgot to remove it

src/main/scala/io/iohk/ethereum/vm/Stack.scala
rtkaczyk: We decided we don't want classes like `Stack`, `Memory` or `Storage` to be case classes. The reason is that the implementation details of those classes are subject to change and therefore should be hidden. The rule of thumb is that case classes are used for data which is to be accessed directly from the class - here it's not the case. The few extra lines of code (equals, toString etc) required for the encapsulation are pretty standard and shouldn't be a concern. Does that sound reasonable?

src/main/scala/io/iohk/ethereum/vm/Storage.scala
rtkaczyk: See the comment about `Stack`

src/main/scala/io/iohk/ethereum/vm/Stack.scala
LukasGasior1: That's a good point.

LukasGasior1:
LGTM

PR #35, opened 2017-02-01
[EC-44] Send proper node status when handshaking

LukasGasior1:
I've added ~~new actor: `NodeStatusHolder`~~ an Agent which is responsible for storing node-related values that may change over time (current best block, server listen address, possibly other needed values).
This ~~actor~~ Agent should be updated with new values as soon as they change (for example, `ServerActor` updates listen address when binding is successful - the same should happen to blockchain info when we get new blocks).


src/main/scala/io/iohk/ethereum/NodeStatusHolder.scala
```diff
@@ -0,0 +1,72 @@
...
+import akka.actor.{ActorLogging, Props, Actor}
+import akka.util.ByteString
+import io.iohk.ethereum.network._
+import io.iohk.ethereum.utils.Config
+import org.spongycastle.crypto.AsymmetricCipherKeyPair
+import org.spongycastle.crypto.params.ECPublicKeyParameters
+import org.spongycastle.util.encoders.Hex
+
+class NodeStatusHolder(key: AsymmetricCipherKeyPair) extends Actor with ActorLogging {
+
```
LukasGasior1: It should work well for such a simple case, the questions is what if we'll need to extends NodeStatusHolder at some point with feature not supported by Agents?

mcsherrylabs: Can you think of an example? And would it still be NodeStatusHolder then...?

LukasGasior1: I can't think of any ATM. If you with I'll make it Agent

mcsherrylabs: It will remove a few lines of code, if/when we have to change it back later - you can remind me I asked you to change it. ;)

LukasGasior1: :+1: You're right, agents seem a good use here

LukasGasior1:
@mcsherrylabs done

whysoserious:
👍

PR #36, opened 2017-02-03
[EC-45] Add dao fork check

LukasGasior1:
When connection is initiated we try to check if peer is running ETC fork.
* it is - we keep connection
* it is not, but we're not at a fork yet - we can download some blocks until fork, thus we keep connection
* it is not at the fork yet - we keep the connection
* it is not running ETC (and we're after the fork) - disconnect

Then, when connected, we check if received block header is a fork header, if so, we validate it and if it's not ETC fork we disconnect. Hope that makes sense.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/network/PeerActor.scala
+++ b/src/main/scala/io/iohk/ethereum/network/PeerActor.scala
@@ -16,7 +16,6 @@ import io.iohk.ethereum.network.p2p.validators.ForkValidator
 import io.iohk.ethereum.network.rlpx.RLPxConnectionHandler
 import io.iohk.ethereum.rlp.RLPEncoder
 import io.iohk.ethereum.utils.{ServerStatus, NodeStatus, Config}
-import org.spongycastle.util.encoders.Hex
 
 /**
   * Peer actor is responsible for initiating and handling high-level connection with peer.
@@ -32,6 +31,7 @@ class PeerActor(
 
   import PeerActor._
   import Config.Network.Peer._
+  import Config.Blockchain._
   import context.{dispatcher, system}
 
   val P2pVersion = 4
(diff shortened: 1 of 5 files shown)


adamsmo:
LGTM

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -43,11 +43,7 @@ class PeerActor(
   val waitForStatusInterval = 30.seconds
   val waitForChainCheck = 15.seconds
 
-  //FIXME move this to props
-  // Doc: https://blog.ethereum.org/2016/07/20/hard-fork-completed/
-  lazy val DaoBlockNumber = 1920000
-  lazy val DaoBlockTotalDifficulty = BigInt("39490964433395682584")
-  lazy val daoForkValidator = ForkValidator(DaoBlockNumber, ByteString(Hex.decode("94365e3a8c0b35089c1d1195081fe7489b528a84b22199c916180db8b28ade7f")))
```
mcsherrylabs: Nice!

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -275,8 +280,18 @@ class PeerActor(
...
+
+        daoBlockHeaderOpt foreach { daoBlockHeader =>
+          log.info("Reached the ETC fork block header")
+          if (daoForkValidator.validate(msg).nonEmpty) {
+            log.info("Peer is not running the ETC fork, disconnecting")
+            disconnectFromPeer(rlpxConnection, Disconnect.Reasons.UselessPeer)
+          }
+        }
+
+      case _ => // nothing
```
mcsherrylabs: Can we add a log here so in case this line survives longer than we expect it to?

LukasGasior1: I've moved that log a little higher, the message is still logged

PR #38, opened 2017-02-06
[EC-60] Remove all data from DataSource

ntallar:
## Description

- Added `clear` function to DataSource, to remove all the data from it.
- Added `close` function to DataSource, to close the DataSource.
- Added `destroy` function to DataSource, that closes it and deletes all the data from it.
- Added checking of the keys length in the `get` and `update` functions of IODB (as discussed in #31 ).
- Added option to use a previously closed (but not destroyed) `IodbDataSource` when creating one. This is now done instead of deleting the folder by default.

## Design choices
- As IODB doesn't yet provide functionality to easily clear it, the DataSource is destroyed and created again.

### DIFF
--- a/src/it/scala/io/iohk/ethereum/db/IodbDataSourceIntegrationSuite.scala
+++ b/src/it/scala/io/iohk/ethereum/db/IodbDataSourceIntegrationSuite.scala
@@ -3,14 +3,12 @@ package io.iohk.ethereum.db
 import org.scalatest.FunSuite
 import org.scalatest.prop.PropertyChecks
 import io.iohk.ethereum.ObjectGenerators
-import io.iohk.iodb.LSMStore
-import java.io.File
-
 import akka.util.ByteString
 import io.iohk.ethereum.db.dataSource.{DataSource, IodbDataSource}
 import org.scalacheck.Gen
 
 import scala.util.{Random, Try}
+import java.io.File
 
 //FIXME: Add IodbDataSource delete tests (currently not implemented as they failed in previous IODB implementation)
 class IodbDataSourceIntegrationSuite extends FunSuite
@@ -32,44 +30,33 @@ class IodbDataSourceIntegrationSuite extends FunSuite
 
   test("IodbDataSource insert keys in separate updates"){
     forAll(seqByteStringOfNItemsGen(KeySizeWithoutPrefix)) { unFilteredKeyList: Seq[ByteString] =>
-      //create temporary dir
-      val dir = File.createTempFile("iodbInsert", "iodbInsert")
-      dir.delete()
-      dir.mkdir()
-
       val keyList = unFilteredKeyList.take(KeyNumberLimit)
       val db = updateInSeparateCalls(
-        dataSource = new IodbDataSource(new LSMStore(dir = dir, keySize = KeySize)),
+        dataSource = IodbDataSource(path = "/tmp/iodbInsert", keySize = KeySize),
         toUpsert = keyList.zip(keyList)
       )
       keyList.foreach { key => assert(db.get(OtherNamespace, key).contains(key)) }
+
+      db.close()
     }
   }
 
   test("IodbDataSource insert keys in a single update"){
     forAll(seqByteStringOfNItemsGen(KeySizeWithoutPrefix)) { unFilteredKeyList: Seq[ByteString] =>
-      //create temporary dir
-      val dir = File.createTempFile("iodbInsert", "iodbInsert")
-      dir.delete()
-      dir.mkdir()
-
       val keyList = unFilteredKeyList.take(KeyNumberLimit)
-      val db = new IodbDataSource(new LSMStore(dir = dir, keySize = KeySize))
+      val db = IodbDataSource(path = "/tmp/iodbInsert", keySize = KeySize)
         .update(OtherNamespace, Seq(), keyList.zip(keyList))
 
       keyList.foreach { key => assert(db.get(OtherNamespace, key).contains(key)) }
+
+      db.close()
     }
   }
 
   test("IodbDataSource update keys in separate updates"){
     forAll(seqByteStringOfNItemsGen(KeySizeWithoutPrefix)) { unFilteredKeyList: Seq[ByteString] =>
-      //create temporary dir
-      val dir = File.createTempFile("iodbUpdate", "iodbUpdate")
-      dir.delete()
-      dir.mkdir()
-
       val keyList = unFilteredKeyList.take(KeyNumberLimit)
-      val db = new IodbDataSource(new LSMStore(dir = dir, keySize = KeySize))
+      val db = IodbDataSource(path = "/tmp/iodbUpdate", keySize = KeySize)
         .update(OtherNamespace, Seq(), keyList.zip(keyList))
 
       val keyListWithExtraByte = keyList.map(1.toByte +: _)
@@ -77,18 +64,15 @@ class IodbDataSourceIntegrationSuite extends FunSuite
 
       keyList.zip(keyListWithExtraByte).foreach { case (key, value) =>
         assert(dbAfterUpdate.get(OtherNamespace, key).contains(value)) }
+
+      dbAfterUpdate.close()
     }
   }
 
   test("IodbDataSource update keys in a single update"){
     forAll(seqByteStringOfNItemsGen(KeySizeWithoutPrefix)) { unFilteredKeyList: Seq[ByteString] =>
-      //create temporary dir
-      val dir = File.createTempFile("iodbUpdate", "iodbUpdate")
-      dir.delete()
-      dir.mkdir()
-
       val keyList = unFilteredKeyList.take(KeyNumberLimit)
-      val db = new IodbDataSource(new LSMStore(dir = dir, keySize = KeySize))
+      val db = IodbDataSource(path = "/tmp/iodbUpdate", keySize = KeySize)
         .update(OtherNamespace, Seq(), keyList.zip(keyList))
 
       val keyListWithExtraByte = keyList.map(1.toByte +: _)
@@ -97,19 +81,16 @@ class IodbDataSourceIntegrationSuite extends FunSuite
       keyList.zip(keyListWithExtraByte).foreach { case (key, value) =>
         assert(dbAfterUpdate.get(OtherNamespace, key).contains(value))
       }
+
+      dbAfterUpdate.close()
     }
   }
 
   test("IodbDataSource insert/update with invalid length") {
     forAll(seqByteStringOfNItemsGen(KeySizeWithoutPrefix)) { unFilteredKeyList: Seq[ByteString] =>
-      //create temporary dir
-      val dir = File.createTempFile("iodbInvalidLength", "iodbInvalidLength")
-      dir.delete()
-      dir.mkdir()
-
       val keyList = unFilteredKeyList.take(KeyNumberLimit)
       val (keysLeft, keysToInsert) = Random.shuffle(keyList).splitAt(Gen.choose(0, keyList.size/2).sample.get)
-      val db = new IodbDataSource(new LSMStore(dir = dir, keySize = KeySize))
+      val db = IodbDataSource(path = "/tmp/iodbInvalidLength", keySize = KeySize)
         .update(OtherNamespace, Seq(), keysToInsert.zip(keysToInsert))
 
       val keyListWithExtraByte = keyList.map(1.toByte +: _)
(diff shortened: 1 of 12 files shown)


src/it/scala/io/iohk/ethereum/db/IodbDataSourceIntegrationSuite.scala
```diff
@@ -97,19 +81,16 @@ class IodbDataSourceIntegrationSuite extends FunSuite
...
     forAll(seqByteStringOfNItemsGen(KeySizeWithoutPrefix)) { unFilteredKeyList: Seq[ByteString] =>
-      //create temporary dir
-      val dir = File.createTempFile("iodbInvalidLength", "iodbInvalidLength")
-      dir.delete()
-      dir.mkdir()
-
       val keyList = unFilteredKeyList.take(KeyNumberLimit)
       val (keysLeft, keysToInsert) = Random.shuffle(keyList).splitAt(Gen.choose(0, keyList.size/2).sample.get)
-      val db = new IodbDataSource(new LSMStore(dir = dir, keySize = KeySize))
+      val db = IodbDataSource(path = "/tmp/iodbInvalidLength", keySize = KeySize)
```
whysoserious: Could you remove `Random.shuffle` from tests? `keyList` is random anyway. Introuducing randomness to the test makes it impossible to recreate a bug which occurs only for small set of input data.

whysoserious:
Apart from the above, LGTM!

src/main/scala/io/iohk/ethereum/db/dataSource/IodbDataSource.scala
```diff
@@ -32,4 +53,29 @@ object IodbDataSource {
   private def storageVersionGen(): Array[Byte] = {
     ByteBuffer.allocate(java.lang.Long.SIZE / java.lang.Byte.SIZE).putLong(updateCounter.incrementAndGet()).array()
   }
+
+  def apply(path: String, keySize: Int): IodbDataSource = {
+    //Delete directory if it existed and create a new one
+    val dir: File = new File(path)
+    val dirCreationSuccess = (!dir.exists() || deleteDirectory(dir)) && dir.mkdir()
```
LukasGasior1: how about `dir.mkdirs()`?

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
```diff
@@ -34,4 +34,16 @@ trait DataSource {
...
+
+  /**
+    * This function updates the DataSource by deleting all the (key-value) pairs in it.
+    *
+    * @return the new DataSource after all the data was removed.
+    */
+  def clear: DataSource
+
+  /**
+    * This function closes the DataSource, removing any allocated resources to it.
```
LukasGasior1: can we add a note to make it clear that it actually deletes the data? `allocated resources` sounds like it relates to memory or open file descriptors (at least that's how it sounds to me)

src/main/scala/io/iohk/ethereum/db/dataSource/IodbDataSource.scala
```diff
@@ -32,4 +53,29 @@ object IodbDataSource {
...
+    assert(dirCreationSuccess, "Iodb folder creation failed")
+
+    val lSMStore: LSMStore = new LSMStore(dir = dir, keySize = keySize, keepSingleVersion = true)
```
LukasGasior1: the doc says:
`  * @param keepSingleVersion  if true compaction will automatically delete older versions, it will not be possible to use rollback`
but as I see we use rollback in `clear`

am I missing something?

src/main/scala/io/iohk/ethereum/db/dataSource/EphemDataSource.scala
```diff
@@ -10,6 +10,10 @@ case class EphemDataSource(storage: Map[IndexedSeq[Byte], IndexedSeq[Byte]]) ext
       storage + ((namespace ++ toUpdate._1) -> toUpdate._2))
     EphemDataSource(afterUpdate)
   }
+
+  override def clear: DataSource = EphemDataSource(Map())
+
+  override def close(): Unit = Unit
```
LukasGasior1: since this method does not behave as documented, maybe it's better to use `???` here

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
AlanVerbner: @LukasGasior1 maybe it's better to split this in two methods. `close` and `destroy` (or something like that)

LukasGasior1: that would be even better

src/it/scala/io/iohk/ethereum/db/IodbDataSourceIntegrationSuite.scala
ntallar: Done! Removed `Random.shuffle` from all the tests for which it wasn't necessary, I only left them in a couple of tests for the MPT where it is needed (for example, to see that independently of the order of insertion the hash root of the MPT is the same).

src/main/scala/io/iohk/ethereum/db/dataSource/IodbDataSource.scala
ntallar: You are right, I changed the way in which the `clear` is implemented in the most recent commit.

ntallar:
@whysoserious @LukasGasior1 I recently made changes to fix the problems you pointed out. If you detect any other problem please let me know. I changed the PR description to reflect those changes and the fact that I included the option to continue using a previously closed (but not destroyed) `IodbDataSource`.

LukasGasior1:
lgtm

src/it/scala/io/iohk/ethereum/db/IodbDataSourceIntegrationSuite.scala
```diff
@@ -32,63 +29,49 @@ class IodbDataSourceIntegrationSuite extends FunSuite
...
     forAll(seqByteStringOfNItemsGen(KeySizeWithoutPrefix)) { unFilteredKeyList: Seq[ByteString] =>
-      //create temporary dir
-      val dir = File.createTempFile("iodbInsert", "iodbInsert")
-      dir.delete()
-      dir.mkdir()
-
       val keyList = unFilteredKeyList.take(KeyNumberLimit)
       val db = updateInSeparateCalls(
-        dataSource = new IodbDataSource(new LSMStore(dir = dir, keySize = KeySize)),
+        dataSource = IodbDataSource(path = "/tmp/iodbInsert", keySize = KeySize, recreate = true),
```
whysoserious: I'd suggest using random file for tests if possible. This won't work if two tests are going to be run at the same time. Same goes for other tests. 

Personally, I'd create a separate fixture for that which could be share among other tests.
http://www.scalatest.org/user_guide/sharing_fixtures

src/main/scala/io/iohk/ethereum/db/dataSource/EphemDataSource.scala
```diff
@@ -10,6 +10,12 @@ case class EphemDataSource(storage: Map[IndexedSeq[Byte], IndexedSeq[Byte]]) ext
...
+
+  override def destroy(): Unit = Unit
```
rtkaczyk: TLDR: it should be 

```
override def destroy(): Unit = ()
```

Explanation: funny thing... it will work just fine, but so would:

```
override def destroy(): Unit = "Unit"
```

Whatever you return will match the `Unit` type. But the sole expression `Unit` is not of type `Unit` but `Unit.type` 😁 A literal for the `Unit` type is `()`

src/main/scala/io/iohk/ethereum/db/dataSource/IodbDataSource.scala
```diff
@@ -1,23 +1,46 @@
...
+    destroy()
+    IodbDataSource(path, keySize, recreate = true)
+  }
+
+  override def close(): Unit = lSMStore.close()
+
+  override def destroy(): Unit = {
+    close()
+    val directoryDeletionSuccess = deleteDirectory(new File(path))
+    assert(directoryDeletionSuccess, "Iodb folder destruction failed")
```
rtkaczyk: Shouldn't that be in a `finally` clause?

src/main/scala/io/iohk/ethereum/db/dataSource/IodbDataSource.scala
```diff
@@ -1,23 +1,46 @@
...
+import java.io.File
 import java.nio.ByteBuffer
 import java.util.concurrent.atomic.AtomicLong
 
 import io.iohk.iodb.{ByteArrayWrapper, LSMStore}
 
-class IodbDataSource(lSMStore: LSMStore) extends DataSource {
+class IodbDataSource private (lSMStore: LSMStore,
+                              keySize: Int,
+                              path: String) extends DataSource {
```
rtkaczyk: I'm not requiring this change, but it is recommended to keep multiline parameters on their own lines. Main reason is that when you change the name you don't have re-indent. http://docs.scala-lang.org/style/declarations.html#classes

src/it/scala/io/iohk/ethereum/db/IodbDataSourceIntegrationSuite.scala
ntallar: Fixed!

src/main/scala/io/iohk/ethereum/db/dataSource/IodbDataSource.scala
AlanVerbner: actually @rtkaczyk i think it's a good comment. The problem I see it's probably around other places (did a quick search and found some)....mayte it's better to create a separate task for this, what do you think?

rtkaczyk: @AlanVerbner I think it would be useful to schedule such a house-keeping task, only not sure if we can afford it at the moment with so many radical changes going on.

src/main/scala/io/iohk/ethereum/db/dataSource/IodbDataSource.scala
```diff
@@ -32,4 +53,34 @@ object IodbDataSource {
...
+      else dir.exists() && dir.isDirectory
+    assert(dirSetupSuccess, "Iodb folder creation failed")
+
+    val lSMStore: LSMStore = new LSMStore(dir = dir, keySize = keySize, keepSingleVersion = true)
+    new IodbDataSource(lSMStore, keySize, path)
+  }
+
+  private def deleteDirectory(dir: File): Boolean = {
+    require(dir.exists() && dir.isDirectory, "Trying to delete a file thats not a folder")
+    val files = Option(dir.listFiles()).getOrElse(Array())
```
whysoserious: There's no need of using `Option` here.

src/main/scala/io/iohk/ethereum/db/dataSource/IodbDataSource.scala
AlanVerbner: @rtkaczyk I'm 100% with you on that

src/main/scala/io/iohk/ethereum/db/dataSource/IodbDataSource.scala
```diff
@@ -32,4 +53,34 @@ object IodbDataSource {
...
+    val filesDeletionSuccess: Boolean = files.map { f =>
```
whysoserious: Stop at first error:
```
val allFiles: Array[File] = dir.listFiles() :+ dir
all.find{ f => 
  val deleted = if(f.isDirectory) deleteDirectory(f) else f.delete()
  !deleted
}
```

src/it/scala/io/iohk/ethereum/db/IodbDataSourceIntegrationSuite.scala
```diff
@@ -32,117 +41,176 @@ class IodbDataSourceIntegrationSuite extends FunSuite
...
     }
   }
 
   test("IodbDataSource update keys in a single update"){
     forAll(seqByteStringOfNItemsGen(KeySizeWithoutPrefix)) { unFilteredKeyList: Seq[ByteString] =>
-      //create temporary dir
-      val dir = File.createTempFile("iodbUpdate", "iodbUpdate")
-      dir.delete()
-      dir.mkdir()
+      withDir { path =>
```
whysoserious: :+1:

src/main/scala/io/iohk/ethereum/db/dataSource/IodbDataSource.scala
ntallar: I agree with stopping at first error but instead of checking if the result of `allFiles.find` is empty I used `forall` which also stops at first error. I also kept separate `dir.delete()` because if not a `StackOverflowError` was thrown due to infinite recursion.

PR #40, opened 2017-02-06
[EC-48] Java LevelDB DataSource

AlanVerbner:
# Description

LevelDB data source was introduced in order to be used until IODB is ready. Pure java (no native interface) implementation was used.


src/main/scala/io/iohk/ethereum/db/dataSource/LevelDBDataSource.scala
```diff
@@ -0,0 +1,76 @@
...
+    *
+    * @param namespace from which the (key-value) pairs will be removed and inserted.
+    * @param toRemove  which includes all the keys to be removed from the DataSource.
+    * @param toUpsert  which includes all the (key-value) pairs to be inserted into the DataSource.
+    *                  If a key is already in the DataSource its value will be updated.
+    * @return the new DataSource after the removals and insertions were done.
+    */
+  override def update(namespace: Namespace, toRemove: Seq[Key], toUpsert: Seq[(Key, Value)]): DataSource = {
+    val batch = new WriteBatchImpl()
+    toRemove.foldLeft(batch) { (batch, key) => batch.delete(key.toArray[Byte]); batch }
```
LukasGasior1: isn't `foreach` sufficient?

src/main/scala/io/iohk/ethereum/db/dataSource/IodbDataSource.scala
```diff
@@ -30,7 +30,7 @@ class IodbDataSource private (lSMStore: LSMStore, keySize: Int, path: String) ex
 
   override def clear: DataSource = {
     destroy()
-    IodbDataSource(path, keySize, recreate = true)
```
LukasGasior1: is there a reason to remove this?

AlanVerbner: @LukasGasior1 yes. After discussing, `recreate` because It's not actually being used (at least what I've seen) and adds more complexity (that obviously needs to be tested). That being said, I thought it was better move to remove it until we need it. If that happens we can either look for the change in the git logs or maybe look for a better alternative to it.

src/main/scala/io/iohk/ethereum/db/dataSource/LevelDBDataSource.scala
AlanVerbner: fixed! thx :smile:

src/it/scala/io/iohk/ethereum/db/DataSourceIntegrationTestBehavior.scala
```diff
@@ -0,0 +1,157 @@
...
+          keyList.zip(keyListWithExtraByte).foreach { case (key, value) =>
+            assert(dbAfterUpdate.get(OtherNamespace, key).contains(value))
+          }
+
+          dbAfterUpdate.destroy()
+        }
+      }
+    }
+
+    it should "be able to update keys in a single update" in {
```
adamsmo: would it be worth to test concurrent updates, for example by adding few threads, same for inserts?

src/main/scala/io/iohk/ethereum/db/dataSource/LevelDBDataSource.scala
```diff
@@ -0,0 +1,76 @@
...
+    * @param toUpsert  which includes all the (key-value) pairs to be inserted into the DataSource.
+    *                  If a key is already in the DataSource its value will be updated.
+    * @return the new DataSource after the removals and insertions were done.
+    */
+  override def update(namespace: Namespace, toRemove: Seq[Key], toUpsert: Seq[(Key, Value)]): DataSource = {
+    val batch = new WriteBatchImpl()
+    toRemove.foreach { key => batch.delete(key.toArray[Byte]) }
+    toUpsert.foreach { item => db.put(item._1.toArray, item._2.toArray) }
+    db.write(batch, new WriteOptions())
+    new LevelDBDataSource(db, path)
```
adamsmo: is it efficient to create new data source on every update?

AlanVerbner: good catch. Just fixed it, thx!

src/it/scala/io/iohk/ethereum/db/DataSourceIntegrationTestBehavior.scala
AlanVerbner: @adamsmo the implementation we are using of levelDB seems to be thread safe [1](https://github.com/dain/leveldb/blob/master/leveldb/src/main/java/org/iq80/leveldb/impl/DbImpl.java#L682) and [2](https://github.com/dain/leveldb/blob/master/leveldb/src/main/java/org/iq80/leveldb/impl/DbImpl.java#L584) (havent checked jni one but I guess that too).

That being said it might be interesting to have this kind of tests, were you thinking these ones in terms of performance or thread safety?

adamsmo: I was thinking about thread safety, so I think we are ok

adamsmo:
LGTM

src/main/scala/io/iohk/ethereum/db/dataSource/LevelDBDataSource.scala
```diff
@@ -0,0 +1,76 @@
...
+
+object LevelDBDataSource {
+
+  def apply(path: String): LevelDBDataSource = {
+    // Configs available https://rawgit.com/google/leveldb/master/doc/index.html
+    val options = new Options()
+      .createIfMissing(true)
+      .paranoidChecks(true) // raise an error as soon as it detects an internal corruption
+      .verifyChecksums(true) // force checksum verification of all data that is read from the file system on behalf of a particular read
+      .cacheSize(0) // do not cache
```
whysoserious: Are we sure that end users will never want to use different configuration settings? I personally would create two `apply` methods:
* `def apply(config: Config): LevelDBDataSource = ??? // reads all config options from application.conf`
* `def apply(path: String, options: Options): LevelDBDataSource = ???`

mcsherrylabs: In this case you can either add as individual values but as that will probably obscure the interface consider adding a config trait that you can build 'Options' from ...

src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
```diff
@@ -437,4 +417,4 @@ class MerklePatriciaTrieSuite extends FunSuite
     val storage = EmptyTrie.put(key1, key1).put(key2, key2).put(key3, key3)
     assert(Hex.toHexString(storage.getRootHash) == "a3d0686205c7ed10a85c3bce4118d5d559bcda47ca39e4dd4f09719958a179f1")
   }
-}
\ No newline at end of file
+}
```
mcsherrylabs: LevelDBDataSource needs unit tests also, even if they are dead simple...

src/main/scala/io/iohk/ethereum/db/dataSource/LevelDBDataSource.scala
```diff
@@ -0,0 +1,76 @@
...
+  override def close(): Unit = db.close()
+
+  /**
+    * This function closes the DataSource, if it is not yet closed, and deletes all the files used by it.
+    */
+  override def destroy(): Unit = {
+    try {
+      close()
+    } finally {
+      Iq80DBFactory.factory.destroy(new File(path), null) // Options are not being used ¯\_(ツ)_/¯
```
AlanVerbner: @mcsherrylabs yup, unfortunately I cannot change it

src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
AlanVerbner: @mcsherrylabs I understood we will create a new task for this as Iodb doesn't have either, do you want me to include those tests in this PR?

mcsherrylabs: Yes please. Unit tests for the LevelDBDataSource. I'll add a ticket for the IODB unit tests....

src/main/scala/io/iohk/ethereum/db/dataSource/LevelDBDataSource.scala
```diff
@@ -0,0 +1,78 @@
...
+  }
+}
+
+trait LevelDbOptions {
+  // Configs available https://rawgit.com/google/leveldb/master/doc/index.html
+  def buildOptions(): Options
+}
+
+object LevelDBDataSource {
+  def apply(path: String, options: LevelDbOptions): LevelDBDataSource =
```
AlanVerbner: @mcsherrylabs please let me know if this is what you have requested

src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
AlanVerbner: @mcsherrylabs just pushed Super basic tests for `LevelDB` and `IODB`

src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -58,4 +58,16 @@ object Config {
     val daoForkBlockHash = ByteString(Hex.decode(blockchainConfig.getString("dao-fork-block-hash")))
   }
 
+  object Db {
+
+    private val dbConfig = config.getConfig("db")
+
+    object LevelDb {
```
mcsherrylabs: object LevelDbConfig extends LevelDbConfig

PR #43, opened 2017-02-07
[EC-51] Create a domain model

whysoserious:
This PR extracts a couple of domain models from networking layer of `etc-client` and moves them to `.domain` package. We need all types listed below in EVM or in MPT.

Changes:
- An `io.iohk.ethereum.domain` package with following members:
  - Account
  - Address
  - BlockHeader
  - Transaction
  - SignedTransaction
- All changes resulting from the above.
- Fixed typos.


src/main/scala/io/iohk/ethereum/vm/Runner.scala
```diff
@@ -1,85 +1,85 @@
-package io.iohk.ethereum.vm
+// package io.iohk.ethereum.vm
```
AlanVerbner: @whysoserious as all the file is commented, can we remove it?

src/it/scala/io/iohk/ethereum/rlp/RLPSpeedSuite.scala
```diff
@@ -1,8 +1,11 @@
 package io.iohk.ethereum.rlp
 
+import scala.language.implicitConversions
```
rtkaczyk: I don't think we need this. Implicit parameters and implicit classes are not considered dangerous and do not require this import.

src/main/scala/io/iohk/ethereum/domain/Account.scala
```diff
@@ -0,0 +1,16 @@
...
+case class Account(nonce: BigInt, balance: BigInt, storageRoot: ByteString, codeHash: ByteString) {
+  override def toString: String = {
+    s"""Account{
+         |nonce: $nonce
+         |balance: $balance wei
+         |storageRoot: ${Hex.toHexString(storageRoot.toArray[Byte])}
+         |codeHash: ${Hex.toHexString(codeHash.toArray[Byte])}
+         |}
+       """.stripMargin
+  }
```
rtkaczyk: Avoid curly braces for purely functional code: http://docs.scala-lang.org/style/control-structures#curly-braces

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -0,0 +1,72 @@
...
+  val FirstByteOfAddress = 12
+  val AddressLength = 20
+  val LastByteOfAddress = FirstByteOfAddress + AddressLength
+
+}
+
+case class SignedTransaction(tx: Transaction,
+                             pointSign: Byte, //v - 27 or 28 according to yellow paper, but it is 37 and 38 in ETH
+                             signatureRandom: ByteString, //r
+                             signature: ByteString /*s*/) {
```
rtkaczyk: For multi-line declarations it's best to have each parameter on it's own line, then we don't need to re-indent if the name changes: http://docs.scala-lang.org/style/declarations#classes

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  import SignedTransaction._
+
+  lazy val bytesToSign: Array[Byte] = crypto.sha3(
+    rlpEncode(RLPList(tx.nonce, tx.gasPrice, tx.gasLimit,
+                      tx.receivingAddress.toArray,
+                      tx.value,
+                      tx.payload)))
```
rtkaczyk: Similarly as with declarations, for multi-line invocations it's best to put arguments on their own lines to avoid future re-indents: http://docs.scala-lang.org/style/indentation#methods-with-numerous-arguments

src/main/scala/io/iohk/ethereum/vm/Runner.scala
whysoserious: I'll just revert it to the version from `phase/3/fastDownload`.

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -177,7 +177,7 @@ sealed abstract class BinaryOp(code: Int)(val f: (DataWord, DataWord) => DataWor
   def execute(state: ProgramState): ProgramState = {
     val updatedState = for {
       popped <- state.stack.pop(2)
-      (Seq(a, b), stack1) = popped
+                               (Seq(a, b), stack1) = popped
```
rtkaczyk: Oops, that's surely not intended

src/main/scala/io/iohk/ethereum/domain/Transaction.scala
```diff
@@ -0,0 +1,35 @@
...
+
+}
+
+case class Transaction(
+  nonce: BigInt,
+  gasPrice: BigInt,
+  gasLimit: BigInt,
+  receivingAddress: Address,
+  value: BigInt,
+  payload: ByteString) {
```
rtkaczyk: Yes!

whysoserious:
🚶 ➡️ 🍕 ➡️ ⌨️ ➡️ :review:

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -38,8 +40,8 @@ object PV63 {
     }
   }
 
-  object Account {
-    implicit val rlpEndDec = new RLPEncoder[Account] with RLPDecoder[Account] {
+  object AccountImplicits {
+    implicit val rlpEncDec = new RLPEncoder[Account] with RLPDecoder[Account] {
```
rtkaczyk: If you put these instances in the companion object of `Account` won't they be available without an explicit import? I believe that's a standard practice for typeclasses.

src/main/scala/io/iohk/ethereum/vm/VM.scala
```diff
@@ -13,9 +14,9 @@ object VM {
     * @param context context to be executed
     * @return result of the execution
     */
-  def run(context: ProgramContext): ProgramResult = {
-    val finalState = run(ProgramState(context))
-    ProgramResult(finalState.returnData, finalState.storage, finalState.error)
+  def run(env: ExecEnv, storage: Storage): ProgramResult = {
```
rtkaczyk: Why not put `Storage` in `ExecEnv`?

src/test/scala/io/iohk/ethereum/vm/Generators.scala
```diff
@@ -48,17 +49,18 @@ object Generators extends ObjectGenerators {
     memGen: Gen[Memory] = getMemoryGen(),
     storageGen: Gen[Storage] = getStorageGen(),
     codeGen: Gen[ByteString] = getByteStringGen(0, 0),
-    callDataGen: Gen[ByteString] = getByteStringGen(0, 0),
-    callValueGen: Gen[ByteString] = getByteStringGen(0, 32)
-  ) : Gen[ProgramState] =
+    inputDataGen: Gen[ByteString] = getByteStringGen(0, 0),
+    valueGen: Gen[BigInt] = getBigIntGen(0)
```
rtkaczyk: Why was this generator changed?

src/main/scala/io/iohk/ethereum/vm/ProgramError.scala
```diff
@@ -9,3 +9,5 @@ case class  InvalidOpCode(code: Byte) extends ProgramError
 sealed trait StackError extends ProgramError
 case object StackOverflow extends StackError
 case object StackUnderflow extends StackError
+
+sealed trait ContractCreationError extends ProgramError
```
rtkaczyk: A sealed trait without implementation? I understand it's an unfinished business but are you sure we need such an error? See YP 9.4.2 and the definition of `x` in `CREATE` instruction description.

src/test/scala/io/iohk/ethereum/vm/Generators.scala
rtkaczyk: Ahh, just noticed the type changed, but still I think we should use `2^256 - 1` as the max value.

src/main/scala/io/iohk/ethereum/domain/Account.scala
whysoserious: This syntax is used everywhere else and it's just a comment. Let's keep it as it is. Summoning the author - @adamsmo - learn about Radek's objections.

src/it/scala/io/iohk/ethereum/rlp/RLPSpeedSuite.scala
```diff
@@ -11,18 +14,16 @@ import org.scalatest.FunSuite
...
-
 /**
-  * Tests based on
-  *   - https://github.com/cryptape/ruby-rlp/blob/master/test/speed.rb
-  *   - https://github.com/ethereum/pyrlp/blob/develop/tests/speed.py
-  */
+ * Tests based on
+ *   - https://github.com/cryptape/ruby-rlp/blob/master/test/speed.rb
+ *   - https://github.com/ethereum/pyrlp/blob/develop/tests/speed.py
+ */
```
LukasGasior1: > A long time ago in a repository far, far away...

src/it/scala/io/iohk/ethereum/rlp/RLPSpeedSuite.scala
```diff
@@ -56,32 +57,33 @@ class RLPSpeedSuite extends FunSuite
     val ITERATIONS: Int = 10000000
     log.info("Starting " + ITERATIONS + " decoding iterations...")
     val start1: Long = System.currentTimeMillis
-    (1 to ITERATIONS).foreach { _ => RLP.rawDecode(payload); Unit }
+      (1 to ITERATIONS).foreach { _ => RLP.rawDecode(payload); Unit }
```
LukasGasior1: why the `; Unit` part?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
whysoserious: Emacs <3

src/it/scala/io/iohk/ethereum/rlp/RLPSpeedSuite.scala
LukasGasior1: you know, once you refactor the code you're blamed for the code you didn't write

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
whysoserious: I prefer to keep logic of encoding/decoding domain object to RLPs in `ethereum.network.p2p.messages` package. I am almost sure that we won't need that anywhere else.

src/main/scala/io/iohk/ethereum/vm/ExecEnv.scala
```diff
@@ -0,0 +1,15 @@
+package io.iohk.ethereum.vm
+
+import akka.util.ByteString
+import io.iohk.ethereum.domain.Address
+
+/** Execution environment constants of an EVM program */
```
LukasGasior1: can you explain the fields (for example senderAddr vs execAddr) or callDepth

src/main/scala/io/iohk/ethereum/vm/ExecEnv.scala
```diff
@@ -0,0 +1,15 @@
...
+case class ExecEnv(ownerAddr: Address,
+                   senderAddr: Address,
+                   gasPrice: BigInt,
+                   inputData: ByteString,
+                   execAddr: Address,
+                   value: BigInt,
+                   program: Program,
+                   blockHeader: Any,
```
LukasGasior1: why `Any`?

src/main/scala/io/iohk/ethereum/vm/ProgramResult.scala
```diff
@@ -1,12 +1,17 @@
...
+ *
+ * @param returnData bytes returned by the executed contract (set by [[RETURN]] opcode)
+ * @param storage represents changes to the storage of this account
+ * @param error defined when the program terminated abnormally
+ */
+case class ProgramResult(returnData: ByteString,
+                         storage: Storage,
+                         transactions: Seq[Transaction] = Seq(),
+                         addressesToDelete: Seq[Address] = Seq(),
+                         error: Option[ProgramError])
```
LukasGasior1: can the program return an error but still have other effects (like storage changes or return data)?
if not, maybe we can make it something like
`type ProgramResult = Either[Error, ProgramSuccessfulResult]`

src/main/scala/io/iohk/ethereum/vm/VM.scala
whysoserious: `ExecEnv` stores `val`s which are not supposed to be changed during execution of a program. It's an implementation of vals listen in section 9.3. of YP _"Execution environment"_.
I updated docs.

src/main/scala/io/iohk/ethereum/vm/VM.scala
```diff
@@ -13,9 +14,9 @@ object VM {
...
+    val finalState = run(ProgramState(env, storage))
+    ProgramResult(finalState.returnData, finalState.storage, Seq(), Seq(), finalState.error)
```
LukasGasior1: shouldn't it be `finalState.internalTransactions` instead of empty `Seq()` (same for acc to delete)

src/test/scala/io/iohk/ethereum/vm/Generators.scala
whysoserious: Yes, this is the default value :) see l. 18.

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -246,17 +246,21 @@ case object CALLDATALOAD extends OpCode(0x35, 1, 1) {
...
-
+      data = state.inputData.slice(i, i + 32).padTo(32, 0.toByte)
       stack2 <- stack1.push(DataWord(data))
     } yield state.withStack(stack2).step()
 
     updatedState.valueOr(state.withError)
   }
 }
 
+trait OpCode32 {
```
LukasGasior1: is it used?

src/main/scala/io/iohk/ethereum/vm/ProgramResult.scala
whysoserious: We may want to be able to have both an error and state of the program prior to this error.

whysoserious:
@rtkaczyk @AlanVerbner @LukasGasior1  the PR is ready for next round of reviews.

LukasGasior1:
@whysoserious currently we have some "model" classes in network (Receipt, TransactionLog) and some in `domain`. How about moving them all to `domain` and leaving only Message classes in `network`? rlp enc/dec for these we can put in the package object (currently naming is not consistent, for example we have `AccountImplicits` and `SignedTransactions` objects to hold these implicits). This is just a suggestion

whysoserious:
@LukasGasior1 That's a good idea. However, let's do that on a one-by-one basis. I would have to spend one more day refactoring it and I really want to avoid it.

Re. rlp enc/dec, this functionality deserves for another refactor. It's a mix of implicits conversions, explicit conversions, uses of discouraged package objects etc. However, let's keep it in a backburner as well.

src/main/scala/io/iohk/ethereum/domain/Account.scala
rtkaczyk: ok, it's probably the least of those concerns anyway

src/main/scala/io/iohk/ethereum/vm/ProgramResult.scala
rtkaczyk: Yes, it might be useful to have that state, but then again we could use the type to make sure the error does not get overlooked. I'd prefer we leave it as it is, and rethink it once we actually want to do something with the `ProgramResult`.

src/test/scala/io/iohk/ethereum/vm/Generators.scala
rtkaczyk: No, the default is `2^255 - 1`. BTW, those defaults were chosen arbitrarily...

src/it/scala/io/iohk/ethereum/rlp/RLPSpeedSuite.scala
AlanVerbner: @LukasGasior1 lol dont be mean :imp:

src/main/scala/io/iohk/ethereum/vm/VM.scala
rtkaczyk: Firstly, the `Storage` class is just a representation and the actual contract's storage will be updated after the program finishes (changes to `Storage` class will be "replayed" to actual storage)

Secondly, `Storage` is a function of contract's address, and therefore a function of `ExecEnv`. So it could be evaluated within `VM.run` or as a _body def_ within `ExecEnv`.

Finally, I would like to think `VM` as a simple function of `input => result`. And it's easier to reason when the input is a single entity. Otherwise you may find you'll need to keep adding stuff to `VM.run` to access other accounts' code, balance etc.

src/it/scala/io/iohk/ethereum/rlp/RLPSpeedSuite.scala
AlanVerbner: I'm not trying to :fire: but the previous formatting style is the one that IntelliJ Idea creates, lol

Besides my meaningless comment, formatting commits might lead to difficult to read PRs. So, if we are doing to make a lot of this, I suggest to keep them in separate PRs.

src/main/scala/io/iohk/ethereum/vm/ProgramState.scala
```diff
@@ -1,36 +1,35 @@
...
+ *
+ * @param context the context which initiates the program
+ * @param stack current stack
+ * @param memory current memory
+ * @param storage current storage
+ * @param pc program counter - an index of the opcode in the program to be executed
+ * @param returnData data to be returned from the program execution
+ * @param halted a flag to indicate program termination
+ * @param error indicates whether the program terminated abnormally
+ */
```
rtkaczyk: Please supplement with newly added parameters

src/main/scala/io/iohk/ethereum/vm/ProgramResult.scala
```diff
@@ -1,12 +1,17 @@
...
-  * @param storage represents changes to the storage of this account
-  * @param error defined when the program terminated abnormally
-  */
-case class ProgramResult(returnData: ByteString, storage: Storage, error: Option[ProgramError])
+ * Represenation of the result of execution of a contract
+ *
+ * @param returnData bytes returned by the executed contract (set by [[RETURN]] opcode)
+ * @param storage represents changes to the storage of this account
+ * @param error defined when the program terminated abnormally
+ */
```
rtkaczyk: Please supplement this

src/main/scala/io/iohk/ethereum/domain/Address.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  def apply(dw: DataWord): Address = Address(dw.bytes.takeRight(Length))
+
+  def apply(arr: Array[Byte]): Address = Address(ByteString(arr))
+
+  val empty: Address = Address(ByteString())
+
+}
+
+case class Address(bytes: ByteString) {
```
AlanVerbner: It has been pointed out about using `IndexedSeq[Byte]` ... @whysoserious do you think we can use it here too?

LukasGasior1:
LGTM

src/main/scala/io/iohk/ethereum/domain/Address.scala
whysoserious: Please note that a mutable `ArrayBuffer` is also a subclass of `IndexedSeq`. 
I'd prefer to be sure that we use ropes instead of plain byte arrays in our domain models. This makes sense especially in case of `Address`es. Thanks to `ByteString` implementation we avoid copying of memory eg. in line 10 in `Address.scala`.

src/main/scala/io/iohk/ethereum/vm/VM.scala
whysoserious: I am adding TODO. We'll fix that in an upcoming PR.

rtkaczyk:
LGTM :+1:

PR #44, opened 2017-02-08
Bugfix in Storage and updates of docs.

whysoserious:
If a `DataWord.Zero` is passed as a value to a `Storage` then a corresponding key should removed from the underlying hashmap.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
@@ -156,12 +156,12 @@ object OpCode {
 }
 
 /**
- * Base class for all the opcodes of the EVM
- *
- * @param code Opcode byte representation
- * @param delta number of words to be popped from stack
- * @param alpha number of words to be pushed to stack
- */
+  * Base class for all the opcodes of the EVM
+  *
+  * @param code Opcode byte representation
+  * @param delta number of words to be popped from stack
+  * @param alpha number of words to be pushed to stack
+  */
 sealed abstract class OpCode(val code: Byte, val delta: Int, val alpha: Int) {
   def this(code: Int, pop: Int, push: Int) = this(code.toByte, pop, push)
 
@@ -236,7 +236,7 @@ case object SHA3 extends OpCode(0x20, 2, 1) {
 
 case object CALLVALUE extends OpCode(0x34, 0, 1) {
   def execute(state: ProgramState): ProgramState =
-    state.stack.push(DataWord(state.env.value))
+    state.stack.push(DataWord(state.context.callValue))
       .map(state.withStack(_).step())
       .valueOr(state.withError)
 }
@@ -246,8 +246,10 @@ case object CALLDATALOAD extends OpCode(0x35, 1, 1) {
     val updatedState = for {
       popped <- state.stack.pop
       (offset, stack1) = popped
+
       i = offset.intValue
-      data = state.inputData.slice(i, i + 32).padTo(32, 0.toByte)
+      data = state.context.callData.slice(i, i + 32).padTo(32, 0.toByte)
+
       stack2 <- stack1.push(DataWord(data))
     } yield state.withStack(stack2).step()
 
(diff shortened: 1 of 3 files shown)


src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -365,6 +365,10 @@ case object JUMPI extends OpCode(0x57, 2, 0) {
 }
 
 case object JUMPDEST extends OpCode(0x5b, 0, 0) {
+  // TODO To correctly implement this opcode EVM has to scan whole bytecode of a program
+  // to find all valid JUMP destinations.
+  // We will also have to scan for SSTORE and SSLOAD opcodes and if they are not found
+  // then there's no sense in loading Storage of a transaction from MPT
```
rtkaczyk: 1) The comment is misplaced because JUMP and JUMPI have to be validated prior to reaching a JUMPDEST (and those instructions already have the comments).

2) SSTORE and SSLOAD wouldn't be enough: http://ethereum.stackexchange.com/questions/3667/difference-between-call-callcode-and-delegatecall 
Anyway, I would not force ourselves onto this optimisation yet. It may actually be uncalled for if we decide to fetch stored values on the fly.

src/main/scala/io/iohk/ethereum/vm/Storage.scala
```diff
@@ -10,15 +10,27 @@ object Storage {
...
-/** Persistent storage of a transaction. May be viewed as a map of 256 bit keys
-  * to 256 bit values.
-  * This is just a representation of storage changes occuring during program execution. The
-  * actual persistence is not this class's resposibility.
-  */
+/* Persistent storage of a transaction. May be viewed as a map of 256 bit keys
+ * to 256 bit values.
+ * This is just a representation of storage changes occuring during program execution. The
+ * actual persistence is not this class's resposibility.
+ */
```
rtkaczyk: Please stop doing that! 😁

src/main/scala/io/iohk/ethereum/vm/Storage.scala
```diff
@@ -10,15 +10,27 @@ object Storage {
...
 class Storage private(private val underlying: Map[DataWord, DataWord] = Map()) {
 
-  def store(addr: DataWord, value: DataWord): Storage = new Storage(underlying + (addr -> value))
+  /* Stores new value in an underlying hashmap.
```
rtkaczyk: At some point generating a scaladoc might be useful so please add a second *

src/main/scala/io/iohk/ethereum/vm/Storage.scala
```diff
@@ -10,15 +10,27 @@ object Storage {
...
+   */
+  def store(addr: DataWord, value: DataWord): Storage = {
+    val newUnderlying: Map[DataWord, DataWord] = value match {
+      case DataWord.Zero => underlying - addr
+      case dw => underlying + (addr -> value)
+    }
+    new Storage(newUnderlying)
+  }
 
+  /* Retrieves a value from an underlying hashmap for a given key.
```
rtkaczyk: Second *

src/main/scala/io/iohk/ethereum/vm/Storage.scala
rtkaczyk: Also second * is missing

LukasGasior1:
lgtm

src/main/scala/io/iohk/ethereum/vm/Storage.scala
```diff
@@ -17,8 +17,20 @@ object Storage {
   */
 class Storage private(private val underlying: Map[DataWord, DataWord] = Map()) {
 
-  def store(addr: DataWord, value: DataWord): Storage = new Storage(underlying + (addr -> value))
+  /** Stores new value in an underlying hashmap.
+    * If DataWord.Zero is passed as a value then a corresponding key is removed from the hashmap
+    */
+  def store(addr: DataWord, value: DataWord): Storage = {
+    val newUnderlying: Map[DataWord, DataWord] = value match {
+      case DataWord.Zero => underlying - addr
```
rtkaczyk: I'm not sure this will work actually... I will have a look

rtkaczyk: Nevermind, I thought it might be using reference comparison but it's using `equals`, so that's good 👍

rtkaczyk:
LGTM

PR #46, opened 2017-02-13
[EC-52][EC-50] EVM gas calculation + refactoring

rtkaczyk:
Note on refactoring:
- stack doesn't handle errors now (no `Either`), operations that exceed the stack bounds return the stack unmodified
- `StackOverflow` and `StackUnderflow` (as well as `OutOfGas`) are now handled in `OpCode#execute`
- the reason for those changes is that there are much fewer errors possible in the EVM than previously expected. We only need to implement one more error (illegal jump) and we don't need `Either` for that. Getting rid off `Either` made implementation of individual opcodes more straight-forward and readable.


src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -98,4 +110,9 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
     f"DataWord(0x$n%02x)" //would be even better to add a leading zero if odd number of digits
 
   def toBigInt: BigInt = n
+
+  /**
+    * @return Size in bytes excluding the leading 0 bytes
+    */
+  def byteSize: Int = if (isZero) 0 else (n.bitLength - 1) / 8 + 1
```
adamsmo: why n.bitLength - 1 ? to get integer result?

rtkaczyk: It's just maths 😁 
For positive values it is equivalent to `1 + floor(log_256(n))`. See YP (222) (EXP gas cost) and the [unit tests](https://github.com/input-output-hk/etc-client/pull/46/files#diff-d7ae98f2e6380f5568a3e55bcbbd40a7R180)

rtkaczyk: ... and I just noticed that test would fail for `x == 0`. Fixed

adamsmo: 222 is about this: https://github.com/input-output-hk/etc-client/pull/46/files#diff-f692200418f9d3438ebb71f6200c7046R33
but in line above I was talking about this `(n.bitLength - 1) / 8 + 1`
why we subtracting 1 from length?

rtkaczyk: Sorry, I meant the formula 220.

rtkaczyk: If you're still not convinced, here's the proof:

```
def:
  n.bitLength = floor(log_2(n))  + 1 // not going to prove that, consider some examples

then: 
  n.bitLength - 1 = floor(log_2(n))
  1 + floor((n.bitLength - 1) / 8) = 1 + floor(log_2(n) / 8) = 1 + floor(log_2(n) / log_2(256))

thus:
  1 + floor((n.bitLength - 1) / 8) = 1 + floor(log_256(n))

QED
```

Is that convincing enough? 😁

adamsmo: yes it is :) thanks

adamsmo: maybe we can add
`1 + floor((n.bitLength - 1) / 8) = 1 + floor(log_256(n))`
as comment, because it was not obvious for me at first glance

rtkaczyk: @adamsmo done

adamsmo: thx

PR #47, opened 2017-02-13
[EC-72] change signing scheme for chain specific transactions

adamsmo:
This PR adds support for new signing scheme for chain specific transactions added by 
this EIP
https://github.com/ethereum/EIPs/issues/155

fork block number is not yet set but we are receiving transactions that are signed according to that proposal.

`0x3d` is blockchain id for ethereum classic

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -36,6 +36,8 @@ etc-client {
     dao-fork-block-number = "1920000"
     dao-fork-block-total-difficulty = "39490964433395682584"
     dao-fork-block-hash = "94365e3a8c0b35089c1d1195081fe7489b528a84b22199c916180db8b28ade7f"
+
+    chain-id = "3d"
   }
 
 }
--- a/src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
@@ -61,16 +61,17 @@ case class SignedTransaction(
     recoveredPublicKey.map(key => crypto.sha3(key).slice(FirstByteOfAddress, LastByteOfAddress))
 
   /**
-    * formula for calculating pointt sing new way
+    * new formula for calculating point sign post EIP 155 adoption
     * v = CHAIN_ID * 2 + 35 or v = CHAIN_ID * 2 + 36
     */
-  lazy val recoveredPointSign: Option[Int] = pointSign match {
-    case p if p == negativePointSign || p == (Blockchain.chainId * 2 + newNegativePointSign).toByte =>
+  lazy val recoveredPointSign: Option[Int] =
+    if (pointSign == negativePointSign || pointSign == (Blockchain.chainId * 2 + newNegativePointSign).toByte) {
       Some(negativePointSign)
-    case p if p == positivePointSign || p == (Blockchain.chainId * 2 + newPositivePointSign).toByte =>
+    } else if (pointSign == positivePointSign || pointSign == (Blockchain.chainId * 2 + newPositivePointSign).toByte) {
       Some(positivePointSign)
-    case _ => None
-  }
+    } else {
+      None
+    }
 
   lazy val recoveredPublicKey: Option[Array[Byte]] = recoveredPointSign.flatMap { p =>
     ECDSASignature.recoverPubBytes(
--- a/src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
@@ -64,22 +64,21 @@ case class SignedTransaction(
     * formula for calculating pointt sing new way
     * v = CHAIN_ID * 2 + 35 or v = CHAIN_ID * 2 + 36
     */
-  lazy val recoveredPublicKey: Option[Array[Byte]] = pointSign match {
+  lazy val recoveredPointSign: Option[Int] = pointSign match {
     case p if p == negativePointSign || p == (Blockchain.chainId * 2 + newNegativePointSign).toByte =>
-      ECDSASignature.recoverPubBytes(
-        new BigInteger(1, signatureRandom.toArray[Byte]),
-        new BigInteger(1, signature.toArray[Byte]),
-        ECDSASignature.recIdFromSignatureV(negativePointSign),
-        bytesToSign
-      )
+      Some(negativePointSign)
     case p if p == positivePointSign || p == (Blockchain.chainId * 2 + newPositivePointSign).toByte =>
-      ECDSASignature.recoverPubBytes(
-        new BigInteger(1, signatureRandom.toArray[Byte]),
-        new BigInteger(1, signature.toArray[Byte]),
-        ECDSASignature.recIdFromSignatureV(positivePointSign),
-        bytesToSign
-      )
+      Some(positivePointSign)
     case _ => None
+  }
+
+  lazy val recoveredPublicKey: Option[Array[Byte]] = recoveredPointSign.flatMap { p =>
+    ECDSASignature.recoverPubBytes(
+      new BigInteger(1, signatureRandom.toArray[Byte]),
+      new BigInteger(1, signature.toArray[Byte]),
+      ECDSASignature.recIdFromSignatureV(p),
+      bytesToSign
+    )
   }
 
   lazy val syntacticValidity: Boolean = {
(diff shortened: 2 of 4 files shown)


src/main/resources/application.conf
```diff
@@ -36,6 +36,8 @@ etc-client {
     dao-fork-block-number = "1920000"
     dao-fork-block-total-difficulty = "39490964433395682584"
     dao-fork-block-hash = "94365e3a8c0b35089c1d1195081fe7489b528a84b22199c916180db8b28ade7f"
+
+    chain-id = "3d"
```
rtkaczyk: Where does this number come from?

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -17,42 +19,61 @@ object SignedTransaction {
...
-    None
+  lazy val recoveredPublicKey: Option[Array[Byte]] = pointSign match {
+    case p if p == negativePointSign || p == (Blockchain.chainId * 2 + 35).toByte =>
+      ECDSASignature.recoverPubBytes(
+        new BigInteger(1, signatureRandom.toArray[Byte]),
+        new BigInteger(1, signature.toArray[Byte]),
+        ECDSASignature.recIdFromSignatureV(negativePointSign),
+        bytesToSign
+      )
+    case p if p == positivePointSign || p == (Blockchain.chainId * 2 + 36).toByte =>
```
rtkaczyk: Somehow scalastyle doesn't qualify that as a magic number. I still think it would be nice to have those numbers as vals and/or provide comments where they came from.

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -17,42 +19,61 @@ object SignedTransaction {
...
+      ECDSASignature.recoverPubBytes(
+        new BigInteger(1, signatureRandom.toArray[Byte]),
+        new BigInteger(1, signature.toArray[Byte]),
+        ECDSASignature.recIdFromSignatureV(positivePointSign),
+        bytesToSign
```
rtkaczyk: This function call is the same for both cases except for `pointSign` arg. So how about:

```
val maybeP = pointSign match {
  case ... => ...
  case ... => ...
  case _ => None
}

maybeP.map { p => ECDSASignature.recoverPubBytes(..., ..., f(p), ...) }
```

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -17,42 +19,61 @@ object SignedTransaction {
...
+    crypto.sha3(
+      rlpEncode(RLPList(
+        tx.nonce,
+        tx.gasPrice,
+        tx.gasLimit,
+        tx.receivingAddress.toArray,
+        tx.value,
+        tx.payload)))
+  } else {
+    //chain specific transaction
```
rtkaczyk: What's the difference between global and chain-specific transactions? (Just a question for my own benefit)

adamsmo: global is transaction that do not have chain specific point sign / signature which means that it can be part of block in any chain (and can be replayed on any chain), chain-specific is transaction that is signed in a way that includes chain information so it can be part of block only on single chain
(terminology is from parity client)

src/main/resources/application.conf
adamsmo: from here
https://github.com/ethcore/parity/blob/85e9091b29a9d9b54d2f6243489a73bb968a6485/ethcore/res/ethereum/classic.json#L30

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
rtkaczyk: Thx

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
adamsmo: added something similar

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -16,43 +18,67 @@ object SignedTransaction {
...
-                  tx.value,
-                  tx.payload)))
+        tx.nonce,
+        tx.gasPrice,
+        tx.gasLimit,
+        tx.receivingAddress.toArray,
+        tx.value,
+        tx.payload,
+        Config.Blockchain.chainId,
+        0,
```
mcsherrylabs: It feels wrong to have this hard coded and potentially a source of error to load it from Config.whatever.Transaction.defaultR (or S)..... can we run it by the group?

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -16,43 +18,67 @@ object SignedTransaction {
...
+  }
 
 
   lazy val recoveredAddress: Option[Array[Byte]] =
     recoveredPublicKey.map(key => crypto.sha3(key).slice(FirstByteOfAddress, LastByteOfAddress))
 
-  //todo implement EC-72
-  lazy val recoveredPublicKey: Option[Array[Byte]] = if (allowedSignValues.contains(pointSign)) {
+  /**
+    * formula for calculating pointt sing new way
```
mcsherrylabs: "new formula for calculating point sign post EIP 155 adoption"?

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -16,43 +18,67 @@ object SignedTransaction {
...
+    * v = CHAIN_ID * 2 + 35 or v = CHAIN_ID * 2 + 36
+    */
+  lazy val recoveredPointSign: Option[Int] = pointSign match {
```
mcsherrylabs: Could save few key strokes with if/else here. We don't leverage 'match', might even be clearer, although it's clear right now.

src/test/scala/io/iohk/ethereum/network/p2p/messages/TransactionSpec.scala
```diff
@@ -28,12 +28,22 @@ class TransactionSpec extends FlatSpec with Matchers {
     signatureRandom = ByteString(Hex.decode("cfe3ad31d6612f8d787c45f115cc5b43fb22bcc210b62ae71dc7cbf0a6bea8df")),
     signature = ByteString(Hex.decode("57db8998114fae3c337e99dbd8573d4085691880f4576c6c1f6c5bbfe67d6cf0")))
 
+  val newInvalidStx = SignedTransaction(
```
mcsherrylabs: I know I'm wrong here but I don't know how - isn't newInvalidStx valid? And so the new test "should be(Some(publicKey))".

adamsmo: maybe test name is wrong, it should file when we trying to get global transaction and just by changing point sign encoding change it to chain specific transaction
chain specific transaction have to include chain id as part of signed message

mcsherrylabs: Can we make the signature a val and describe what it is (old style sig minus chain id) , and add a test to recover sender public key for new sign encoding schema with chain_id in signed data?

adamsmo: signature is combination of 3 fields `pointSign` `signatureRandom` and `signature` so we kind of have it separated as `validStx` and `newInvalidStx`, maybe names could be better, will try to change them, and will add test for valid signature with new signing scheme

adamsmo:
@mcsherrylabs added some changes, could you take a look?

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -104,7 +106,7 @@ case class SignedTransaction(
          |tx: $tx
          |pointSign: $pointSign
          |signatureRandom: ${Hex.toHexString(signatureRandom.toArray[Byte])}
-         |signature: ${Hex.toHexString(signatureRandom.toArray[Byte])}
+         |signature: ${Hex.toHexString(signature.toArray[Byte])}
```
mcsherrylabs: Catch!

PR #48, opened 2017-02-14
[EC-76] EVM Stop, Arithmetic, Comparison and Bitwise opcodes

whysoserious:
Introduces all missing `0`s and `10`s opcodes plus a few minor formatting changes.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/DataWord.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/DataWord.scala
@@ -9,15 +9,15 @@ object DataWord {
   /**
     * DataWord size in bytes
     */
-  val Size = 32
+  val Size: Int = 32
 
-  private val Modulus: BigInt = BigInt(2).pow(Size * 8)
+  val Modulus: BigInt = BigInt(2).pow(Size * 8)
 
-  val MaxValue = DataWord(Modulus - 1)
+  val MaxSignedValue: BigInt = BigInt(2).pow(Size * 8 - 1) - 1
 
-  val Zero = DataWord(0)
+  val MaxValue: DataWord = DataWord(Modulus - 1)
 
-  private val Zeros: ByteString = ByteString(Array.fill[Byte](Size)(0))
+  val Zero: DataWord = DataWord(0)
 
   def apply(value: ByteString): DataWord = {
     require(value.length <= Size, s"Input byte array cannot be longer than $Size: ${value.length}")
@@ -38,13 +38,21 @@ object DataWord {
     apply(BigInt(num.toLong(n)))
   }
 
+  implicit def dataWord2BigInt(dw: DataWord): BigInt = dw.toBigInt
+
+  private val Zeros: ByteString = ByteString(Array.fill[Byte](Size)(0))
+
   private def fixBigInt(n: BigInt): BigInt = (n % Modulus + Modulus) % Modulus
 
-  implicit def dataWord2BigInt(dw: DataWord): BigInt = dw.toBigInt
+  private def toSignedBigInt(n: BigInt): BigInt = if (n > MaxSignedValue) n - Modulus else n
+
+  private def toUnsignedBigInt(n: BigInt): BigInt = if (n < 0) n + Modulus else n
+
 }
 
 /** Stores 256 bit words and adds a few convenience methods on them.
- *  Internally a word is stored as a BigInt. */
+ *  Internally a word is stored as an unsinged BigInt. */
+// TODO: consider moving to domain as Uint256, which follows Scala numeric conventions, and is used across the system for P_256 numbers (see YP 4.3)
 class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
 
   import DataWord._
@@ -60,6 +68,8 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
     else
       bs
   }
+
+  private lazy val signedN: BigInt = toSignedBigInt(n)
 
   require(n >= 0 && n < Modulus, s"Invalid word value: $n")
 
@@ -79,9 +89,52 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
 
   def *(that: DataWord): DataWord = DataWord(this.n * that.n)
 
-  def /(that: DataWord): DataWord = DataWord(this.n / that.n)
+  def div(that: DataWord): DataWord = zeroCheck(that) { DataWord(this.n / that.n) }
+
+  def sdiv(that: DataWord): DataWord = zeroCheck(that) {
+    DataWord(toUnsignedBigInt(this.signedN / that.signedN))
+  }
+
+  def mod(that: DataWord): DataWord = zeroCheck(that) { DataWord(this.n mod that.n) }
+
+  def smod(that: DataWord): DataWord = zeroCheck(that) {
+    DataWord(this.signedN.signum * (this.signedN.abs mod that.signedN.abs))
+  }
+
+  def addmod(that: DataWord, modulus: DataWord): DataWord = zeroCheck(modulus) {
+    DataWord((this.n + that.n) mod modulus.n)
+  }
+
+  def mulmod(that: DataWord, modulus: DataWord): DataWord = zeroCheck(modulus) {
+    DataWord((this.n * that.n) mod modulus.n)
+  }
 
   def **(that: DataWord): DataWord = DataWord(this.n.modPow(that.n, Modulus))
+
+  def signExtend(that: DataWord): DataWord = {
+    if (that.n < 0 || that.n > 31) {
+      this
+    } else {
+      val idx = that.n.toByte
+      val mask: Int = if (this.signedN.testBit((idx * 8) + 7)) 0xFF else 0x00
+      val leftFill: Array[Byte] = Array.fill(Size - idx - 1)(mask.toByte)
+      val bs = ByteString(leftFill) ++ this.bytes.takeRight(idx + 1)
+      DataWord(bs)
+    }
+  }
+
+  def slt(that: DataWord): DataWord = DataWord(this.signedN < that.signedN)
+
+  def sgt(that: DataWord): DataWord = DataWord(this.signedN > that.signedN)
+
+  def getByte(that: DataWord): DataWord =
+    if (that.n > 31)
+      Zero
+    else
+      DataWord(Zeros.take(Size - 1) :+ this.bytes(that.n.toInt))
+
+  private def zeroCheck(dw: DataWord)(result: =>DataWord): DataWord =
+    if (dw == Zero) Zero else result
 
   def compare(that: DataWord): Int = this.n.compare(that.n)
 
--- a/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
@@ -14,25 +14,25 @@ object OpCode {
     MUL,
     SUB,
     DIV,
-    //SDIV,
-    //MOD,
-    //SMOD,
-    //ADDMOD,
-    //MULMOD,
+    SDIV,
+    MOD,
+    SMOD,
+    ADDMOD,
+    MULMOD,
     EXP,
-    //SIGNEXTEND,
+    SIGNEXTEND,
 
     LT,
-    //GT,
-    //SLT,
-    //SGT,
+    GT,
+    SLT,
+    SGT,
     EQ,
     ISZERO,
     AND,
-    //OR,
-    //XOR,
+    OR,
+    XOR,
     NOT,
-    //BYTE,
+    BYTE,
 
     SHA3,
 
@@ -193,6 +193,15 @@ case object STOP extends OpCode(0x00, 0, 0, G_zero) with ConstGas {
     state.halt
 }
 
+sealed abstract class UnaryOp(code: Int, constGas: BigInt)(val f: DataWord => DataWord) extends OpCode(code, 1, 1, constGas) with ConstGas {
+  protected def exec(state: ProgramState): ProgramState = {
+    val (a, stack1) = state.stack.pop
+    val res = f(a)
+    val stack2 = stack1.push(res)
+    state.withStack(stack2).step()
+  }
+}
+
 sealed abstract class BinaryOp(code: Int, constGas: BigInt)(val f: (DataWord, DataWord) => DataWord)
   extends OpCode(code.toByte, 2, 1, constGas) {
 
@@ -204,10 +213,12 @@ sealed abstract class BinaryOp(code: Int, constGas: BigInt)(val f: (DataWord, Da
   }
 }
 
-sealed abstract class UnaryOp(code: Int, constGas: BigInt)(val f: DataWord => DataWord) extends OpCode(code, 1, 1, constGas) {
+sealed abstract class TernaryOp(code: Int, constGas: BigInt)(val f: (DataWord, DataWord, DataWord) => DataWord)
+    extends OpCode(code.toByte, 3, 1, constGas) {
+
   protected def exec(state: ProgramState): ProgramState = {
-    val (a, stack1) = state.stack.pop
-    val res = f(a)
+    val (Seq(a, b, c), stack1) = state.stack.pop(3)
+    val res = f(a, b, c)
     val stack2 = stack1.push(res)
     state.withStack(stack2).step()
   }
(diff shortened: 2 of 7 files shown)


src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -228,16 +249,30 @@ case object EXP extends BinaryOp(0x0a, G_exp)(_ ** _) {
...
+
 case object EQ extends BinaryOp(0x14, G_verylow)((a, b) => DataWord(a == b)) with ConstGas
 
 case object ISZERO extends UnaryOp(0x15, G_verylow)(a => DataWord(a.isZero)) with ConstGas
 
 case object AND extends BinaryOp(0x16, G_verylow)(_ & _) with ConstGas
 
+case object OR extends BinaryOp(0x17, G_verylow)(_ & _) with ConstGas
+
+case object XOR extends BinaryOp(0x18, G_verylow)(_ & _) with ConstGas
```
ntallar: Shouldn't the `XOR` function be `(_ ^ _)`? Same with `OR` and `|`.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -79,10 +88,56 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
...
+
+  def sgt(that: DataWord): DataWord = DataWord(this.signedN > that.signedN)
+
+  def getByte(that: DataWord): DataWord = {
+    val cnt: Byte = that.n.toByte
+    if (cnt < 0 || cnt > 31) {
+      Zero
+    } else {
+      DataWord(Zeros.take(Size - 1) :+ this.bytes(cnt))
+    }
```
ntallar: While trying to undertand the BYTE opcode, I came up with the test: `DataWord(1).getByte(DataWord(287))` that currently returns `DataWord(1)`. As `287>=32`, shouldn't this test return `DataWord(0)` according to the paper? Maybe I'm thinking something wrong.

rtkaczyk: Yes, the `.toByte` conversion is wrong.

src/test/scala/io/iohk/ethereum/vm/DataWordSpec.scala
```diff
@@ -114,6 +168,57 @@ class DataWordSpec extends FunSuite with PropertyChecks {
     }
   }
 
+  test("signExtend") {
+    assert(DataWord(42).signExtend(DataWord(3)) == DataWord(42))
```
rtkaczyk: How about using a `Table` for such tests?

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -79,10 +88,56 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
...
+      this
+    } else {
+      val mask: Int = if (this.signedN.testBit((cnt * 8) + 7)) 0xFF else 0x00
+      val leftFill: Array[Byte] = Array.fill(Size - cnt - 1)(mask.toByte)
+      val bs = ByteString(leftFill) ++ this.bytes.takeRight(cnt + 1)
+      DataWord(bs)
+    }
+  }
+
+  def slt(that: DataWord): DataWord = DataWord(this.signedN < that.signedN)
```
rtkaczyk: I'm not fond of the inconsistency in that normal comparators (`<`, `>`) return a `Boolean` whereas these (`sgt`, `slt`) return a `DataWord`.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -79,10 +88,56 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
 
   def *(that: DataWord): DataWord = DataWord(this.n * that.n)
 
-  def /(that: DataWord): DataWord = DataWord(this.n / that.n)
+  def /(that: DataWord): DataWord = zeroCheck(that) { DataWord(this.n / that.n) }
```
rtkaczyk: When you see a `/` method on a numeric type you would expect it to mean the usual division. It's not the case so it would make sense add a comment. But I think it's even better to rename it to `div`.

src/test/scala/io/iohk/ethereum/vm/DataWordSpec.scala
```diff
@@ -114,6 +168,57 @@ class DataWordSpec extends FunSuite with PropertyChecks {
...
+    assert(DataWord(42).signExtend(DataWord(-1)) == DataWord(42))
+    assert(DataWord(42).signExtend(DataWord(0)) == DataWord(42))
+    assert(DataWord(42).signExtend(DataWord(Size)) == DataWord(42))
+    assert(DataWord(42).signExtend(DataWord(Size + 1)) == DataWord(42))
+    assert(DataWord(-42).signExtend(DataWord(Size)) == DataWord(-42))
+    assert(DataWord(-42).signExtend(DataWord(Size + 1)) == DataWord(-42))
+    assert(DataWord(-42).signExtend(DataWord(-11)) == DataWord(-42))
+    assert(DataWord(-1).signExtend(DataWord(1)) == DataWord(-1))
+    assert(DataWord(-1).signExtend(DataWord(1)) == DataWord(-1))
+    // 1737215.signExtend(1) == 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff81ff
```
rtkaczyk: Commented out code

rtkaczyk:
Please add the new opcodes to `constGasFees` in `OpCodeGasSpec`. That should be used in line 104 (my mistake). Could you please also change that line to: 

```
verifyGas(constGasFees(op), stateIn, stateOut)
```

?

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -79,10 +88,56 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
...
+    DataWord((this.n + that.n) mod modulus.n)
+  }
+
+  def mulmod(that: DataWord, modulus: DataWord): DataWord = zeroCheck(modulus) {
+    DataWord((this.n * that.n) mod modulus.n)
+  }
 
   def **(that: DataWord): DataWord = DataWord(this.n.modPow(that.n, Modulus))
 
+  def signExtend(that: DataWord): DataWord = {
```
ntallar: Regarding @rtkaczyk said about the `.toByte` conversion being wrong in `getByte`, isn't it also the case here? Geth also

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
rtkaczyk: Also no need to perform bytes gymnastics or for `{}` in the conditional: 

```scala
    if (cnt > 31) Zero else DataWord(bytes(cnt))
```

rtkaczyk: Also, no need for the conditional :grin::

```scala
def getByte(that: DataWord): DataWord =
  DataWord(n >> (256 - 8 * (that.n + 1)) & 0xff)
```

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
whysoserious: Note to myself: never write code during the meetings ;)

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -79,10 +88,56 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
...
+  }
+
+  def mod(that: DataWord): DataWord = zeroCheck(that) { DataWord(this.n mod that.n) }
+
+  def smod(that: DataWord): DataWord = zeroCheck(that) {
+    DataWord(this.signedN.signum * (this.signedN.abs mod that.signedN.abs))
+  }
+
+  def addmod(that: DataWord, modulus: DataWord): DataWord = zeroCheck(modulus) {
+    DataWord((this.n + that.n) mod modulus.n)
```
rtkaczyk: It would be a bit faster to use `new DataWord(...)` instead of `DataWord(...)`. Since the resulting integer must be within bounds `[0, MaxValue]` there's no need to call `fixBigInt`.

This also pertains to other methods such as: / (div), sdiv, mod, smod, addmod, mulmod, **, signExtend, getByte, sgt, slt, and, or, xor

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -79,10 +88,56 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
...
+  def /(that: DataWord): DataWord = zeroCheck(that) { DataWord(this.n / that.n) }
+
+  def sdiv(that: DataWord): DataWord = zeroCheck(that) {
+    DataWord(toUnsignedBigInt(this.signedN / that.signedN))
+  }
+
+  def mod(that: DataWord): DataWord = zeroCheck(that) { DataWord(this.n mod that.n) }
+
+  def smod(that: DataWord): DataWord = zeroCheck(that) {
+    DataWord(this.signedN.signum * (this.signedN.abs mod that.signedN.abs))
```
rtkaczyk: Isn't that equivalent to: `DataWord(this.signedN % that.signedN)` ?

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -79,10 +88,56 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
...
   def *(that: DataWord): DataWord = DataWord(this.n * that.n)
 
-  def /(that: DataWord): DataWord = DataWord(this.n / that.n)
+  def /(that: DataWord): DataWord = zeroCheck(that) { DataWord(this.n / that.n) }
+
+  def sdiv(that: DataWord): DataWord = zeroCheck(that) {
+    DataWord(toUnsignedBigInt(this.signedN / that.signedN))
+  }
+
+  def mod(that: DataWord): DataWord = zeroCheck(that) { DataWord(this.n mod that.n) }
```
rtkaczyk: `%` and `mod` are the same for non-negative numbers but `%` is slightly faster.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -79,10 +88,56 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
...
 
+  def signExtend(that: DataWord): DataWord = {
+    val cnt: Byte = that.n.toByte
+    if (cnt < 0 || cnt > 31) {
+      this
+    } else {
+      val mask: Int = if (this.signedN.testBit((cnt * 8) + 7)) 0xFF else 0x00
+      val leftFill: Array[Byte] = Array.fill(Size - cnt - 1)(mask.toByte)
+      val bs = ByteString(leftFill) ++ this.bytes.takeRight(cnt + 1)
+      DataWord(bs)
```
rtkaczyk: Nothing wrong with this, but I would I take a different approach:

```scala
val negative = n.testBit(cnt * 8 + 7)
val mask = (BigInt(1) << ((cnt + 1) * 8)) - 1 // == BigInt(2).pow((cnt + 1) * 8) - 1
val newN = if (negative) n | (MaxValue ^ mask) else n & mask
new DataWord(newN)
```

I would expect this to be a tad faster (though not 100% sure)

src/test/scala/io/iohk/ethereum/vm/DataWordSpec.scala
rtkaczyk: Also we could use more numbers like that, with 42 the result is always either 42 or 0.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -79,10 +88,56 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
...
+
+  def mulmod(that: DataWord, modulus: DataWord): DataWord = zeroCheck(modulus) {
+    DataWord((this.n * that.n) mod modulus.n)
+  }
 
   def **(that: DataWord): DataWord = DataWord(this.n.modPow(that.n, Modulus))
 
+  def signExtend(that: DataWord): DataWord = {
+    val cnt: Byte = that.n.toByte
+    if (cnt < 0 || cnt > 31) {
```
rtkaczyk: I don't like the name `cnt`. It suggests we're counting something, but we're not, it's a constant index. I would propose `i` or `idx` or `index`. Same goes for `getByte`.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
```diff
@@ -79,10 +88,56 @@ class DataWord private (private val n: BigInt) extends Ordered[DataWord] {
...
+      Zero
+    } else {
+      DataWord(Zeros.take(Size - 1) :+ this.bytes(cnt))
+    }
+  }
+
+  private def zeroCheck(dw: DataWord)(result: =>DataWord): DataWord =
+    if (dw == Zero) Zero else result
```
rtkaczyk: Neat :+1:

src/test/scala/io/iohk/ethereum/vm/DataWordSpec.scala
```diff
@@ -100,9 +106,57 @@ class DataWordSpec extends FunSuite with PropertyChecks {
...
+      }
     }
+    assert((DataWord(-1) mod DataWord(MaxValue.toBigInt)) == Zero)
+    assert((DataWord(1) mod Zero) == Zero)
+  }
+
+  test("smod") {
+    assert((DataWord(Modulus - 1) smod DataWord(3)) == DataWord(Modulus - 1))
+    assert((DataWord(-1) smod DataWord(MaxValue.toBigInt)) == Zero)
+    assert((DataWord(1) smod Zero) == Zero)
```
rtkaczyk: `DataWord(Modulus - 1` and `DataWord(MaxValue.toBigInt)` are simply equivalent to `MaxValue`.
Also this could test more cases.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
rtkaczyk: If comparisons return `DataWord` it will be more convenient for the opcodes' implementation.

On the other hand, I was thinking that `vm.DataWord` could evolve into `domain.Uint256` and `domain.Int256`, given that such numbers are used across the whole system. For that reason `Uint256` should fully comply with Scala conventions for numeric types.

@adamsmo you researched the numbers a bit, would such numeric types be useful for us?
@mcsherrylabs any thoughts on this?

whysoserious: > On the other hand, I was thinking that vm.DataWord could evolve into domain.Uint256 and domain.Int256

I would keep it as it is now. Most important information for the rest of the system is that a `DataWord` is a 32 bytes long array.

adamsmo: @rtkaczyk I think yes it would be helpful to have something like this, but we have to provide conversion to bytes for that type (for RLP encoding)

rtkaczyk: > Most important information for the rest of the system is that a DataWord is a 32 bytes long array.

I disagree about that. It should be regarded as a number `P_256`, that can also be viewed as a byte array. Our usages in the vm code reflect that, in that it is much more often treated as a number, rather than as a byte array.

whysoserious: Absoulely true. Now I agree with you that this is a good idea.

rtkaczyk: Also parity uses something like `U256` all around. Anyway, I think changing `DataWord` in this manner is outside the scope of this PR, so maybe let's just leave it as is, and provide a comment like:

```
// TODO: consider moving to domain as Uint256, which follows Scala numeric conventions, and is used across the system for P_256 numbers (see YP 4.3)
```

whysoserious:
@ntallar, @rtkaczyk I think that this PR is ready for review again.

I didn't address all of your comments re. `DataWord` as I am planning to split it up into two new types: `UInt256` and `SInt256`. I assume that `DataWord.scala` and `DataWordSpec.scala` will be completely rewritten from scratch in an upcoming PR.

rtkaczyk:
> I assume that DataWord.scala and DataWordSpec.scala will be completely rewritten from scratch

Only if you are going to change the underlying implementation. I wouldn't recommend that as the first step. I'd rather we use those new types right away. And in that case we can reuse most of this code.

Also, I wouldn't call it SInt. I believe all the languages that differentiate signed and unsigned would call them Int and Uint.

src/main/scala/io/iohk/ethereum/vm/DataWord.scala
rtkaczyk: We should still at least achieve correctness. That means `that.n.toByte` has to go.

whysoserious: Argh, right. This somehow slipped through my own review. Fixed.

PR #50, opened 2017-02-15
[EC-67] Provide Block header, Block receipts and Block body download

adamsmo:
Allow connected clients to download bodies, headers and receipts


src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersNumbersStorage.scala
```diff
@@ -0,0 +1,17 @@
+package io.iohk.ethereum.db.storage
+
+import akka.util.ByteString
+import io.iohk.ethereum.db.dataSource.DataSource
+
+class BlockHeadersNumbersStorage(val dataSource: DataSource) extends KeyValueStorage[BigInt, ByteString] {
```
AlanVerbner: I've already have made this in https://github.com/input-output-hk/etc-client/pull/51/files#diff-675d2b192a8da56fd973c3416cb2b4b0 :cry:

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
```diff
@@ -24,7 +25,23 @@ class BlockHeadersStorage(val dataSource: DataSource) extends KeyValueStorage[Bl
   override def valueDeserializer: (IndexedSeq[Byte]) => BlockHeader =
     (encodedBlockHeader: IndexedSeq[Byte]) => rlpDecode[BlockHeader](encodedBlockHeader.toArray)
 
-  override protected def apply(dataSource: DataSource): BlockHeadersStorage = new BlockHeadersStorage(dataSource)
+  override protected def apply(dataSource: DataSource): BlockHeadersStorage =
+    new BlockHeadersStorage(dataSource, blockHeadersNumbersStorage)
+
+  def get(blockNumber: BigInt): Option[BlockHeader] = if (blockNumber == 0) {
+    //todo get block from config, or we will put 0 block to storage at app startup if it is not there?
```
AlanVerbner: good point....If i'm not misstaken in my branch i'm able to download genesis block but yes, this need to be defined

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
```diff
@@ -24,7 +25,23 @@ class BlockHeadersStorage(val dataSource: DataSource) extends KeyValueStorage[Bl
...
+    None
+  } else {
+    blockHeadersNumbersStorage.get(blockNumber).flatMap(key => super.get(key))
```
AlanVerbner: I've found this while coding. What about keeping chain forks? THe mapping is 1:n beetween blockNumber and blockHashes

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
```diff
@@ -24,7 +25,23 @@ class BlockHeadersStorage(val dataSource: DataSource) extends KeyValueStorage[Bl
...
+  }
+
+  override def update(toRemove: Seq[BlockHeaderHash], toUpsert: Seq[(BlockHeaderHash, BlockHeader)]): BlockHeadersStorage = {
```
AlanVerbner: what do you think about this approach https://github.com/input-output-hk/etc-client/pull/51/files#diff-161154b2be64312e90f9db14e8195d1c ?

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
adamsmo: The mapping is 1:1
I think that when we encounter fork we have to decide which branch to keep and store only main chain

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
adamsmo: I see that you are not saving mapping when saving blocks, separate operation is required and you have to remember to perform them

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersNumbersStorage.scala
adamsmo: no problem it was really easy to make ;)

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
AlanVerbner: Actually, I am. Checkout: https://github.com/input-output-hk/etc-client/pull/51/files#diff-d6b2cb3b669672a66dec1646db4121b0R48

AlanVerbner: Sorry, this is a better reference https://github.com/input-output-hk/etc-client/blob/feature/dumpSpecificBlock/src/main/scala/io/iohk/ethereum/blockchain/BlocksRepository.scala#L44

adamsmo: great, missed that, so you added another layer (repository) instead of using storages directly, look good to me.
Is there API to get blocks by number?

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
AlanVerbner: I need to check this again, but take a loot at this EthereumJ method: https://github.com/ethereum/ethereumj/blob/deb6f976392b2b5a7c6229670b07275495683c89/ethereumj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java#L149

You can get to it if you navigate starting from https://github.com/ethereum/ethereumj/blob/deb6f976392b2b5a7c6229670b07275495683c89/ethereumj-core/src/main/java/org/ethereum/jsonrpc/JsonRpcImpl.java#L712

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
AlanVerbner: yes, https://github.com/input-output-hk/etc-client/blob/feature/dumpSpecificBlock/src/main/scala/io/iohk/ethereum/blockchain/BlocksRepository.scala#L33

adamsmo: * I mean in wire protocol API

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
adamsmo: But what if we get `n` forks at a time how we will track which block is for which fork
I see think it is most efficient to just keep main chain and if we are downloading fork then calculate if it will be main chain and then replace it in storage

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
AlanVerbner: No, but in Json-RPC yes.

what about the following taken from [PV61](https://github.com/ethereum/wiki/wiki/Ethereum-Wire-Protocol#pv61-specific) 

> BlockHashesFromNumber [+0x08: P, number: P, maxBlocks: P] Requires peer to reply with a BlockHashes message. Message should contain block with that of number number on the canonical chain. Should also be followed by subsequent blocks, on the same chain, detailing a number of the first block hash and a total of hashes to be sent. Returned hash list must be ordered by block number in ascending order.

adamsmo: 👍

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -229,11 +231,46 @@ class PeerActor(
...
+
+      blockNumber match {
+        case Some(startBlockNumber) if startBlockNumber >= 0 && request.maxHeaders >= 0 && request.skip >= 0 =>
+
+          val headersCount: BigInt = if (maxBlocksHeadersPerMessage < request.maxHeaders) maxBlocksHeadersPerMessage else request.maxHeaders
+
+          val range = if (request.reverse) {
+            startBlockNumber to (startBlockNumber - (request.skip + 1) * headersCount + 1) by -(request.skip + 1)
+          } else {
+            startBlockNumber to (startBlockNumber + (request.skip + 1) * headersCount) by (request.skip + 1)
```
ntallar: While running the test `return block headers by block number` and looking at the value of `range` I noticed it was set to `NumericRange(3,4,5)` while the `headersCount` is set to 2, couldn't there be any problem with this calculation?

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -75,13 +75,17 @@ object PV63 {
...
+          MptBranch(rlpList.items.take(BranchNodeChildLength).map {
+            case bytes: RLPValue =>
+              val v = rlpDecode[ByteString](bytes)
+              if (v.nonEmpty && v.length < 32) {
+                Right(MptValue(v))
+              } else {
+                Left(MptHash(v))
+              }
+
+            case value:RLPEncodeable =>
```
ntallar: Shouldn't this be `case value: RLPList`? Is there any case when the `RLPEncodeable` matching is necessary?

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -69,19 +69,23 @@ object PV63 {
             RLPList(RLPValue(hpEncode(keyNibbles.toArray[Byte], isLeaf = false)), child.fold(_.hash, _.value): RLPEncodeable)
           case n: MptBranch =>
             import n._
-            RLPList(children.map(e => e.fold(_.hash, _.value)).map(e => byteStringEncDec.encode(e)) :+ (terminator: RLPEncodeable): _*)
+            RLPList(children.map(e => e.fold(a => byteStringEncDec.encode(a.hash), b => rawDecode(b.value.toArray[Byte]))) :+ (terminator: RLPEncodeable): _*)
```
ntallar: :+1:

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
adamsmo: waiting for merge of your storage implementation

AlanVerbner: @adamsmo ok .... please review it so it can be merged faster :smile:

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
adamsmo: only 2 classes that implements trait RLPEncodeable are RLP list and RLP value, so no, but I used RLPEncodeable to say `all other cases` but all other cases = only RLPList
can change that

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -69,19 +69,23 @@ object PV63 {
...
-            if (v.nonEmpty && v.length < 32) {
-              Right(MptValue(v))
-            } else {
-              Left(MptHash(v))
-            }
+          MptBranch(rlpList.items.take(BranchNodeChildLength).map {
+            case bytes: RLPValue =>
+              val v = rlpDecode[ByteString](bytes)
+              if (v.nonEmpty && v.length < 32) {
+                Right(MptValue(ByteString(rlpEncode(bytes))))
```
ntallar: I'm wondering when this would be the case, because the `MptBranch` children would only be `RLPValue`s when they are hashes of nodes, if they are encoded nodes they should fall into the `RLPList` case. I agree that we have to check for this case but as far as I know we should only raise an error.

adamsmo: what about case of contract storing for example byte `ff` in contract storage?

ntallar: In that case the `ff` byte wouldn't be inside a `MptLeaf`?

adamsmo: but why create opt leaf if it is < 32 bytes and can be in branch of met branch, envelope would be bigger then information stored

ntallar: I agree with you, but it seems the only way to store information in the MPT is through the leaf nodes

From https://github.com/ethereum/wiki/wiki/Patricia-Tree:
>  So to store a value in one item of v0...v15, we should instead store a kv node, where k is the result of compact encoding an empty nibbles list with terminator.

adamsmo: 👍 thx
so only case for branch in apt branch node to not be value is to be 2 item list which o one is nibbles and second is value

ntallar: That's right :+1:

adamsmo: @ntallar I was reading further and I see that it may not be the case
look at example there
```
ROOT: [ '\x16', A ]
A: [ '', '', '', '', B, '', '', '', C, '', '', '', '', '', '', '', '' ]
B: [ '\x00\x6f', D ]
D: [ '', '', '', '', '', '', E, '', '', '', '', '', '', '', '', '', 'verb' ]
E: [ '\x17', F ]
F: [ '', '', '', '', '', '', G, '', '', '', '', '', '', '', '', '', 'puppy' ]
G: [ '\x35', 'coin' ]
C: [ '\x20\x6f\x72\x73\x65', 'stallion' ]
```

node F is branch node and it has only value, without key

adamsmo: but it does not match our case because we have branch that is RLP encoded is if it was byte array then it should be decoded as RLP value but it is decoded as RLP list
I prepared changes in separate branch
https://github.com/input-output-hk/etc-client/compare/feature/peerFastDownloadBlocks...feature/peerFastDownloadBlocks_keyVal_inBranch
also I changed that extension node can only have hash as child, because otherwise they should be leaf node

adamsmo: FYI @mcsherrylabs ^

ntallar: In that example, isn't the value stored in the terminator of the branch node, not in the children? (As it is on the 17th position of the array) Although I agree, I forgot to mention that information can also be stored in that position of the Branch nodes.
> F: [ '', '', '', '', '', '', G, '', '', '', '', '', '', '', '', '', 'puppy' ]

adamsmo: 👍 thx for pointing that, ok so correct implementation is in `feature/peerFastDownloadBlocks_keyVal_inBranch`
I was thinking about cases of storing values instead of hashes in nodes that are not leafs, and  think that only case in state tree is storing values in branches of branch nodes because all keys have same length (32 bytes because they are kec256 hashes)

ntallar: I'm not sure about that, regarding the extension nodes having values, shouldn't it have a `MptValue` for the case where the `child` is a branch node that's small enough? As far as I know if that branch node, encoded, is smaller than 32 bytes it should be stored encoded in the extension node.

I searched for that in geth (`trie/triedbmut.rs`) and branch node children and extension node child have the same format:
```
enum Node {
	...
	Extension(NodeKey, NodeHandle),
	Branch(Box<[Option<NodeHandle>; 16]>, Option<DBValue>)	
}
```

adamsmo: yes it can be branch node so RLP list of 17 elements in which branches have to be values if node is less then 32 bytes, so it is also possible for branch node to have another branch node in on of its branches.

So to implement it correctly we have to handle 3 cases for branch nde:
* child is RLP value with hash
* child is 2 elements RLP list
* child is 17 elements RLP list

and 2 cases for extension node:
* child is RLP value with hash
* child is 17 elements RLP list

ntallar: That's right :+1:
And you are right about the child of a branch node being an extension or leaf if it is a RLP list of two elements.
I was also referring to that parity code but I mistakenly mentioned Geth :stuck_out_tongue_winking_eye:

adamsmo: IMO it would be simpler to have only hashes as children of node, I do not think that this optimisation has big impact on tree size, but it deffinitly impacts complexity

ntallar: I agree with that, it does make the MPT much more complicated. The problem is that changing the way the nodes are stored also would change the hash of them (and of the hash of the root of the MPT) and that isn't compatible with how other clients store the nodes or the `Yellow Paper (Appendix D, Eq. 178)`.

adamsmo:
@ntallar I pushed updated version of nodes children encoding/decoding

ntallar:
LGTM the node encoding/decoding!

adamsmo:
@AlanVerbner / @ntallar merged all the things, could you take a look?

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersNumbersStorage.scala
ntallar: Is the storage `BlockNumberMappingStorage` used for the same purpose? Maybe we should keep only one of them.

AlanVerbner: something else. This is already done in `Blockchain` ... is there any missing query you need to perform?

adamsmo: 👍 done

src/main/scala/io/iohk/ethereum/db/storage/KeyValueStorage.scala
```diff
@@ -62,6 +62,7 @@ private[storage] trait KeyValueStorage[K, V] {
 object Namespaces {
   val ReceiptsNamespace: IndexedSeq[Byte] = IndexedSeq[Byte]('r'.toByte)
   val HeaderNamespace: IndexedSeq[Byte] = IndexedSeq[Byte]('h'.toByte)
+  val HeaderNumberNamespace: IndexedSeq[Byte] = IndexedSeq[Byte]('n'.toByte)
```
whysoserious: We already have a `n` namespace.

adamsmo: 👍

adamsmo: it is not needed anymore because of storage changes

PR #51, opened 2017-02-15
[EC-37] Dump a specific Block from Storage

AlanVerbner:
# Description

Take a block at a given height, decode it and print out in a human readable format including any uncles or other pertinent data.

## Notes

- This PR does not support Json-RPC queries. This will be included in other PR. JSON output is comparable with JSON-RPC one though
- This uses LevelDB as DataSource
- A `Blockchain` entity was created in order to provide a common interface to do quieries against several related storages (block ones to be more specific)
- Some wiring code was made in order to be able to configure blockchain with the ability to change DataSources and Storages
- **As this was made when FastSync was only created, we will need to discuss the future changes in order to support handling multiple blockchain forks. This means that, when storing blockNumber->hash mapping we are storing tuples like `BlockNumber -> Hash` and not `BlockNumber->Seq[Hash]`**
- Tests for a DataSource handling inserts/removals while having multiple namespaces were added `[EC-93]`.


src/main/scala/io/iohk/ethereum/App.scala
```diff
@@ -19,6 +20,12 @@ object App {
 
   val nodeKey = generateKeyPair()
 
+  // FIXME We need to close DataSources before app ends
```
adamsmo: do we need to address this?

src/main/scala/io/iohk/ethereum/blockchain/BlocksRepository.scala
```diff
@@ -0,0 +1,68 @@
...
+  def blockBodiesStorage: BlockBodiesStorage
+  def blockNumberMappingStorage: BlockNumberMappingStorage
+
+  /**
+    * Allows to query for a block based on it's hash
+    *
+    * @param hash Keccak hash of the block that's being searched
+    * @return Block if found
+    */
+  def getBlockByHash(hash: ByteString): Option[Block] =
```
adamsmo: protocol messages allow only to ask for block header / block body I think we should add API for loading only header/body

src/main/scala/io/iohk/ethereum/db/storage/KeyValueStorage.scala
```diff
@@ -63,6 +63,7 @@ object Namespaces {
   val ReceiptsNamespace: IndexedSeq[Byte] = IndexedSeq[Byte]('r'.toByte)
   val HeaderNamespace: IndexedSeq[Byte] = IndexedSeq[Byte]('h'.toByte)
   val BodyNamespace: IndexedSeq[Byte] = IndexedSeq[Byte]('b'.toByte)
-  val NodeNamespace: IndexedSeq[Byte] = IndexedSeq[Byte]('n'.toByte)
+  val NodeNamespace: IndexedSeq[Byte] = IndexedSeq[Byte]('n'.toByte )
```
adamsmo: do we need whitespace here ;) ?

src/main/scala/io/iohk/ethereum/domain/Block.scala
```diff
@@ -0,0 +1,33 @@
...
+case class Block(header: BlockHeader, body: BlockBody) {
+  override def toString: String = {
+    s"""BlockHeader {
+       | header: $header
+       | body: $body
+     """.stripMargin
+  }
+}
+
+object Block {
```
adamsmo: should this be in domain, other implicits are in `io.iohk.ethereum.network.p2p.messages`
in this case maybe `io.iohk.ethereum.rpc` ?

src/main/scala/io/iohk/ethereum/rpc/BlockController.scala
```diff
@@ -0,0 +1,85 @@
...
+                              unixTimestamp: Long,
+                              transactions: Seq[ByteString],
+                              transactionsRoot: ByteString,
+                              uncles: Seq[ByteString],
+                              size: Long
+                            )
+
+  implicit val itemFormat = jsonFormat19(BlockView)
+
+  // FIXME This route is temporal until we switch to JSON-RPC
```
adamsmo: do we need to address it?

src/main/scala/io/iohk/ethereum/App.scala
AlanVerbner: this is already done, i forgot to remove the FIXME message

src/main/scala/io/iohk/ethereum/blockchain/BlocksRepository.scala
AlanVerbner: Perfect, will add it in a while

src/main/scala/io/iohk/ethereum/db/storage/KeyValueStorage.scala
AlanVerbner: lol, no

src/main/scala/io/iohk/ethereum/domain/Block.scala
AlanVerbner: I thought that too but it doesn't make sense to me to put this in network package....your proposal is to put it under `io.iohk.ethereum.rpc`?

src/main/scala/io/iohk/ethereum/rpc/BlockController.scala
AlanVerbner: Not yet, this is going to be made in a diffferent package.

src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -46,6 +46,14 @@ object Config {
       val waitForChainCheckTimeout = peerConfig.getDuration("wait-for-chain-check-timeout").toMillis.millis
     }
 
+    object Rpc {
```
mcsherrylabs: This should probably extend a trait RpcConfig ... and have JsonRpcServer take RpcConfig as a constructor param...

src/main/scala/io/iohk/ethereum/domain/Block.scala
adamsmo: I am not sure about correct placement but I think it should be placed near it`s usage 
If it is used for RPC then maybe it is good place

mcsherrylabs: Ideally Block should be under the 'blockchain' component (package). It's part of that interface along with BlockHeader and BlockBody. And the serialisation of the Block is also the responsibility of the Block component as it must know the internals of a Block (package blockchain.rlp?) with a dep on the ethereum.rlp package as a serialisation service.

src/main/scala/io/iohk/ethereum/db/components/StoragesComp.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import io.iohk.ethereum.db.storage._
+
+trait StoragesComp {
+
+  def storages: Storages
+
+  trait Storages {
+
+    def blockHeadersStorage: BlockHeadersStorage
```
mcsherrylabs: You can use 'val' in a trait, it's val-id.

src/main/scala/io/iohk/ethereum/blockchain/BlocksRepository.scala
AlanVerbner: Done

src/main/scala/io/iohk/ethereum/utils/Config.scala
AlanVerbner: Done

src/main/scala/io/iohk/ethereum/domain/Block.scala
AlanVerbner: well... BlockHeader and BlockBody are part of network package atm but I like your proposal @mcsherrylabs ... what do you think @adamsmo ?

src/main/scala/io/iohk/ethereum/db/components/StoragesComp.scala
AlanVerbner: I know I know. why do you think methods here are not a good choice?

src/main/scala/io/iohk/ethereum/blockchain/BlocksRepository.scala
```diff
@@ -0,0 +1,83 @@
...
+import akka.util.ByteString
+import io.iohk.ethereum.db.storage.{BlockBodiesStorage, BlockHeadersStorage, BlockNumberMappingStorage}
+import io.iohk.ethereum.domain.{Block, BlockHeader}
+import io.iohk.ethereum.network.p2p.messages.PV62.BlockBody
+
+/**
+  * This interface presents an API to persist and query blocks within the blockchain.
+  */
+trait BlocksRepository {
+  def blockHeadersStorage: BlockHeadersStorage
```
adamsmo: should this be protected so I am unable to directly access storage?

AlanVerbner: Well...I was accessing the storage but I think your comment is correct. Is fixed now.

src/main/scala/io/iohk/ethereum/db/components/StoragesComp.scala
mcsherrylabs: You probably wouldn't write 
`val i = 5`
`def j = 5`
blockHeadersStorage is a reference to one instance of a hopefully immutable Bloc...Storage. (<- if this is not true, then ignore the rest) It's not a Block...Storage generator. You are depending on implementation of the trait to override it with a val to make it correct. If you leave it as a def *and* implement that with another trait that is also a 'def' when it's trying to express a val you could end up with an instance per reference... I suggest the intent is  clearer using `val`...

AlanVerbner: Well...that's the point. I wanted to leave this open in order not to be too strict regarding if we are using a single instance. But as you suggest this might lead to confusion, I can change it to `vals` and then refactor this in case we no longer need a single reference, how does it sound?

AlanVerbner: done! :smile:

src/it/scala/io/iohk/ethereum/db/LevelDBDataSourceIntegrationSuite.scala
```diff
@@ -5,11 +5,12 @@ import org.scalatest.FlatSpec
 
 class LevelDBDataSourceIntegrationSuite extends FlatSpec with DataSourceIntegrationTestBehavior {
 
-  private def createDataSource(path: String) = LevelDBDataSource(path, new LevelDbConfig {
+  private def createDataSource(path: String) = LevelDBDataSource(new LevelDbConfig {
     override val cacheSize: Int = 0
     override val verifyChecksums: Boolean = true
     override val paranoidChecks: Boolean = true
     override val createIfMissing: Boolean = true
+    override val path: String = "/tmp/test/leveldb/"
```
ntallar: Shouldn't this `val path` use the `path` parameter? Because if not I think we might not we using the randomly created directories from the tests.

AlanVerbner: good catch! Fixed!

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -0,0 +1,156 @@
...
+  */
+trait Blockchain {
+
+  /**
+    * Allows to query a blockHeader by block hash
+    *
+    * @param hash of the block that's being searched
+    * @return [[BlockHeader]] if found
+    */
+  def getBlockHeaderByHash(hash: ByteString): Option[BlockHeader]
```
adamsmo: canwe add getBlockHeaderByNumber?

AlanVerbner: yes....once i finish the merge I will add it

AlanVerbner: @adamsmo done!

PR #49, opened 2017-02-15
[EC-64] Sync from multiple peers

LukasGasior1:
Added: FastSyncController + child actors that handle all four kinds of requests: block headers, block bodies, receipts and state nodes.

Controller tries to send requests to as many peers as possible (but within the `max-concurrent-requests` limit). Each child worker sends the request and waits for the response from peer. Syncing is done when:
there are no more queued blocks/state nodes, there are no requests waiting for the response and current block number == target block number.

When peer disappoints us (for example it does not respond or the response is empty) we mark it as blacklisted (for some time) and don't send any more requests to it.

@adamsmo as it strongly bases on your code from SyncActor, please take a closer look if I haven't missed anything :)

TODO: tests, improve performance

PR #52, opened 2017-02-16
[EC-49] Upgrade to Scala 2.12

ntallar:
## Description

- Upgrade to Scala 2.12, which includes upgrading to IODB 0.2.0 and removing the use of `EitherExtensions` and now deprecated methods.
- Added check for `println` in scalastyle.
- Now using `sbt test:scalastyle` the scalastyle rules are checked for both the tests and the integrations tests with a new configuration file `scalastyle-test-config.xml`.
- CircleCI fails due to the use of the Scala version in the CircleCI commands, this will be changed once the PR is merged.


src/test/scala/io/iohk/ethereum/rlp/RLPSuite.scala
```diff
@@ -387,14 +387,14 @@ class RLPSuite extends FunSuite
     val expected = "c88363617483646f67"
     val data = RLP.encode(RLPList("cat", "dog"))
     assert(expected == Hex.toHexString(data))
-    val dataObtained = decode[Seq[String]](data)
+    val dataObtained = decode[Seq[String]](data)(stringSeqEncDec)
```
rtkaczyk: Interesting...
It also works if we move the implicit decoder to the top of the class. Just saying.. this is OK, too.

src/test/scala/io/iohk/ethereum/network/p2p/FastSyncActorSpec.scala
```diff
@@ -140,13 +140,13 @@ class FastSyncActorSpec extends FlatSpec with Matchers {
     storage.mptNodeStorage.get(stateRootHash) shouldBe Some(NodeData(Seq(mptBranchWithTwoChild)).getMptNode(0))
 
     storage.mptNodeStorage.get(NodeData(Seq(mptBranchWithTwoChild)).getMptNode(0)
-      .asInstanceOf[MptBranch].children(0).asInstanceOf[Left[MptHash, MptValue]].a.hash) shouldBe Some(NodeData(Seq(mptExtension)).getMptNode(0))
+      .asInstanceOf[MptBranch].children(0).left.get.hash) shouldBe Some(NodeData(Seq(mptExtension)).getMptNode(0))
```
rtkaczyk: Yes!

whysoserious: Yes!

whysoserious:
Could we upgrade Akka to `2.4.17` as well?

scalastyle-test-config.xml
```diff
@@ -0,0 +1,99 @@
+<scalastyle>
```
whysoserious: Lot's of duplicated XML and afaik scalastyle doesn't support `include` directive. Could you add a comment at the top of the file describing differtences between this file and `scalastyle-config.xml` ?

rtkaczyk:
I have verified that the tests are passing. 
The only nitpick I have is the upgrade resulted in a new warning:

```
[warn] .../etc-client/src/main/scala/io/iohk/ethereum/network/FastSyncActor.scala:125: match may not be exhaustive.
[warn] It would fail on the following input: (_, _)
[warn]       state.requestedNodes.zipWithIndex.foreach {
[warn]                                                 ^
[warn] one warning found
```

Could we fix that? Pinging @adamsmo as he should be most familiar with that code.

So far our code has been warning-free. It would be nice if we could keep it that way.

ntallar:
@rtkaczyk Will fix that warning, I pointed out in #49 a way to solve this problem and will incorporate that to this PR, we only have to have the `trait BlockHash` sealed.

ntallar:
@rtkaczyk I recently fixed the warning and the problems pointed out by @whysoserious. If there's anything else missing please let me know :)

PR #53, opened 2017-02-21
call opcodes WIP

whysoserious:


src/main/scala/io/iohk/ethereum/domain/Address.scala
```diff
@@ -19,10 +20,14 @@ case class Address(bytes: ByteString) {
 
   import Address.Length
 
+  //FIXME: can we get rid off it and simply make sure that address is always 20 bytes, by either truncating or extending
```
whysoserious: Yes, if we keep only this one constructor:
```
def apply(dw: DataWord): Address
```
which is a good idea imho.

PR #55, opened 2017-02-21
[EC-51] EVM System Operations - SELFDESTRUCT

whysoserious:
~~Depends on https://github.com/input-output-hk/etc-client/pull/54~~
Minor changes: 
- Set log level in tests to `WARN`
- Create `ExecEnv` Generator


src/main/scala/io/iohk/ethereum/vm/VM.scala
```diff
@@ -17,7 +17,7 @@ object VM {
     ProgramResult(
       finalState.returnData,
       finalState.storage,
-      finalState.internalTransactions,
+      finalState.internalTransfers.reverse,
```
adamsmo: why reverse?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -558,3 +558,32 @@ case object RETURN extends OpCode(0xf3, 2, 0, G_zero) {
...
   }
 }
+
+case object SELFDESTRUCT extends OpCode(0xff, 1, 0, G_selfdestruct) {
+  protected def exec(state: ProgramState): ProgramState = {
+    val (refundDW, stack1) = state.stack.pop
+    val refundAddr: Address = Address(refundDW)
+    val ownerAccount: Account = state.context.account
+    val ownerAddr: Address = state.context.env.ownerAddr
+    val gasRefund: BigInt = if (state.addressesToDelete contains ownerAddr) 0 else R_selfdestruct
```
rtkaczyk: It only deals with the refund. There's no mechanism to prevent a subsequent message call to this account from executing its current code.

rtkaczyk: Actually doing according to YP the code and storage should remain intact until transaction finalisation. But there are 2 other problems for subsequent calls to this account within the same transaction:

- account's balance update will not be seen on subsequent call
- a subsequent self-destruction crediting a previously inexistent account will cost more gas than it should

PR #56, opened 2017-02-21
[EC-77] NewBlock message and TotalDifficulty storage

ntallar:
## Description
Includes `NewBlock` message used for propagating blocks to peers and `TotalDifficultyStorage` used to store the total difficulty associated with new blocks. Also includes use of TotalDifficultyStorage during the `FastSync` (having a `TotalDifficultyStorage` updated will be needed for block broadcast)


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncBlockHeadersRequestHandler.scala
```diff
@@ -22,6 +24,9 @@ class FastSyncBlockHeadersRequestHandler(
 
     (blockHashes zip blockHeaders.headers).foreach { case (hash, header) =>
       blockHeadersStorage.put(hash, header)
+      val parentTotalDifficulty = totalDifficultyStorage.get(header.parentHash)
+        .getOrElse(throw new Exception(s"Block ${Hex.toHexString(header.parentHash.toArray)} total difficulty not found"))
```
whysoserious: How this `Exception` is handled by the rest of the system?

ntallar: You were right, I had forgotten to properly handle it. In my last commit I changed this to handle properly not finding the total difficulty of the parent of a block.

PR #57, opened 2017-02-22
[EC-66] Provide support for download of the state trie (as the host)

adamsmo:
This PR add sending of MPT / contract storage / evm code when doing fast download


src/main/scala/io/iohk/ethereum/db/storage/BlockBodiesStorage.scala
```diff
@@ -23,6 +24,12 @@ class BlockBodiesStorage(val dataSource: DataSource) extends KeyValueStorage[Blo
   override def valueDeserializer: (IndexedSeq[Byte]) => BlockBody =
     (encodedBlockBody: IndexedSeq[Byte]) => rlpDecode[BlockBody](encodedBlockBody.toArray)
 
+  override def get(key: BlockBodyHash): Option[BlockBody] = if (key == Config.Blockchain.genesisBlockHeader.hash) {
```
ntallar: Minor detail that I just noticed: the `BlockBodyHash` is the type used for the hash of the block right? It seems a little confusing as it seems to show that it is some sort of hash of the block body.

adamsmo: storage will be changed when @AlanVerbner PR (https://github.com/input-output-hk/etc-client/pull/51) is merged

ntallar:
Apart from some of the things mentioned on #50 and that maybe we should have some tests for `MptNode` downloading, LGTM!

adamsmo:
Will add test for that

śr., 22.02.2017, 16:07 użytkownik Nicolás Tallar <notifications@github.com>
napisał:

adamsmo:
I changed `encodeUncles`
because I was getting warning
```
[warn] /Users/adam/iohk/etc-client/src/main/scala/io/iohk/ethereum/network/p2p/validators/BlockValidator.scala:37: reflective access of structural type member method encodeUncles should be enabled
[warn] by making the implicit value scala.language.reflectiveCalls visible.
[warn] This can be achieved by adding the import clause 'import scala.language.reflectiveCalls'
[warn] or by setting the compiler option -language:reflectiveCalls.
[warn] See the Scaladoc for value scala.language.reflectiveCalls for a discussion
[warn] why the feature should be explicitly enabled.
[warn]     val encodedOmmers = encode(BlockBody.rlpEncDec.encodeUncles(block.body.uncleNodesList))
[warn]                                                    ^
[warn] one warning found
```

src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -39,11 +41,16 @@ object Config {
...
-      val disconnectPoisonPillTimeout = peerConfig.getDuration("disconnect-poison-pill-timeout").toMillis.millis
-      val waitForStatusTimeout = peerConfig.getDuration("wait-for-status-timeout").toMillis.millis
-      val waitForChainCheckTimeout = peerConfig.getDuration("wait-for-chain-check-timeout").toMillis.millis
+      val connectRetryDelay: FiniteDuration = peerConfig.getDuration("connect-retry-delay").toMillis.millis
+      val connectMaxRetries: Int = peerConfig.getInt("connect-max-retries")
+      val disconnectPoisonPillTimeout: FiniteDuration = peerConfig.getDuration("disconnect-poison-pill-timeout").toMillis.millis
+      val waitForStatusTimeout: FiniteDuration = peerConfig.getDuration("wait-for-status-timeout").toMillis.millis
+      val waitForChainCheckTimeout: FiniteDuration = peerConfig.getDuration("wait-for-chain-check-timeout").toMillis.millis
+
+      val maxBlocksHeadersPerMessage: Int = peerConfig.getInt("max-blocks-headers-per-message")
```
AlanVerbner: @adamsmo Can we create a trait called `InfoPerMessage` (it's not a good name, lol) and an object here to implement it?

I'm talking about something like https://github.com/input-output-hk/etc-client/blob/phase/3/blockSync/src/main/scala/io/iohk/ethereum/utils/Config.scala#L87

adamsmo: changed

src/main/scala/io/iohk/ethereum/db/storage/KeyValueStorage.scala
```diff
@@ -62,6 +62,7 @@ private[storage] trait KeyValueStorage[K, V] {
 object Namespaces {
   val ReceiptsNamespace: IndexedSeq[Byte] = IndexedSeq[Byte]('r'.toByte)
   val HeaderNamespace: IndexedSeq[Byte] = IndexedSeq[Byte]('h'.toByte)
+  val HeaderNumberNamespace: IndexedSeq[Byte] = IndexedSeq[Byte]('n'.toByte)
```
AlanVerbner: there should be another one already defined for this, `HeightsNamespace` see https://github.com/input-output-hk/etc-client/blob/phase/3/blockSync/src/main/scala/io/iohk/ethereum/db/storage/KeyValueStorage.scala#L68

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -94,6 +103,8 @@ trait Blockchain {
 
   def save(hash: ByteString, evmCode: ByteString): Unit
 
+  def save(node: MptNode): Unit
```
AlanVerbner: I left this one outside blockchain intentionally. There are two main reasons to do so:
1. There are _2 implementations_ of node entities, the ones used while receiving them over the wire and the ones used by the MPT
2. Imo state isn't part of the blockchain. I like to think about state as a different part as it does not deal with blocks or how are those connected  :thinking:

AlanVerbner: ps: maybe EVM code is misplaced too

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -226,11 +233,62 @@ class PeerActor(
     context become waitingForConnectionResult(newConnection, noRetries)
   }
 
+  def handleEvmMptFastDownload(rlpxConnection: RLPxConnection): Receive = {
```
AlanVerbner: @adamsmo i'm wondering...should this code be performed by one (or many) separate actors that would be in charge of retrieving info for other nodes? This is an open question in order at least to think about this.

As pros to do so I can think of:

- Avoid accesing the storages within the main Peer actor (this can be done within a Future for example)
- Encapsulating the functionality
- To be able to use multiple actor for this (as it's done in FastSyncController with Handlers). This is not a concrete need though

On the other side, we might be adding more complexity that might not be needed.

ps: this is also related to https://github.com/input-output-hk/etc-client/pull/50

summoning @LukasGasior1

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
adamsmo: I think the state is part of blockchain because it is directly linked with blockchain, any change in blockchain will affect the state in some way, either ether transfer or contract creation/call, so I see no point in splitting those two things

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
adamsmo: I done it this way because we have one PeerActor per every peer so every peer can ask for fast download elements and requests will be handled in parallel. Also, all logic is encapsulated in a separate method.
About using futures for loading data: this functionality will be used only when a peer is doing fast download so for peer it is not important to get other information before he finishes fast download, but if we decide that it is required It should be easy to add.

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
AlanVerbner: Ok...i'm not still sure but I don't think it's a blocker for merging. We can always refactor this.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
AlanVerbner: ok..... @LukasGasior1 do you have any other opinion? If not it's ok for me....it can also be refactored in the future.

LukasGasior1: Yeah I do....forgot to submit my review! I hate this github feature!

src/main/resources/application.conf
```diff
@@ -29,6 +29,7 @@ etc-client {
       max-blocks-headers-per-message = 200
       max-blocks-bodies-per-message = 200
       max-receipts-per-message = 200
+      max-mpt-components-per-message = 400
```
adamsmo: I placed it here because they are used in peer actor, not in fsat sync actor

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
adamsmo: @AlanVerbner / @LukasGasior1 adding another actor will increase communication overhead and potentially will make application harder to debug, maybe we can just separate logic to different class like here:https://github.com/input-output-hk/etc-client/compare/feature/peerFastDownloadMpt...feature/peerFastDownloadMpt_separateUpload

LukasGasior1: :+1:

src/main/resources/application.conf
LukasGasior1: it depends how we want to organize configs. based on functionality (fast-sync) or the place where this config is used

src/main/scala/io/iohk/ethereum/network/FastUpload.scala
```diff
@@ -0,0 +1,63 @@
...
+import io.iohk.ethereum.network.p2p.messages.PV62.{BlockBodies, BlockHeaders, GetBlockBodies, GetBlockHeaders}
+import io.iohk.ethereum.network.p2p.messages.PV63._
+import io.iohk.ethereum.network.rlpx.RLPxConnectionHandler
+
+class FastUpload(fastSyncHostConfiguration: FastSyncHostConfiguration, storage: Blockchain) {
+  def handleEvmMptFastDownload(rlpxConnection: RLPxConnection): Actor.Receive = {
+    case RLPxConnectionHandler.MessageReceived(request: GetNodeData) =>
+      import io.iohk.ethereum.rlp.encode
+
+      val result: Seq[ByteString] = request.mptElementsHashes.slice(0, fastSyncHostConfiguration.maxMptComponentsPerMessage).flatMap { hash =>
```
LukasGasior1: isn't it just `.take(fastSyncHostConfiguration.maxMptComponentsPerMessage)`?

adamsmo: 👍 yes it is

LukasGasior1:
LGTM

PR #58, opened 2017-02-23
[EC-89] Accurately name sha3 functions in crypto

adamsmo:
changed name of sha3 to kec256 and sha512 to kec512


adamsmo:
I think this build failure is example that we have to run all tests locally (it:test)

ntallar:
LGTM! In relation with this, there is an opcode called `SHA3` should we also rename it?

whysoserious:
> LGTM! In relation with this, there is an EVM opcode called SHA3, should we also rename it?

Good point but let's keep names of the opcodes the same as in YP.

PR #59, opened 2017-02-23
[EC-90] Persistent node keys

ntallar:
## Description

Includes use of `nodeId.keys` file to persist the public and private keys (as plain text for now).


src/main/resources/application.conf
```diff
@@ -2,6 +2,8 @@ etc-client {
 
   client-id = "etc-client"
 
+  keys-file = "nodeId.keys"
```
whysoserious: Could you add some docs explaining syntax of a `keys-file`? This is not a blocker.

src/main/scala/io/iohk/ethereum/network/package.scala
```diff
@@ -19,4 +25,38 @@ package object network {
       Hex.decode(nodeId)
     curve.getCurve.decodePoint(bytes)
   }
+
+  object AsymmetricCipherKeyPairSerializable {
+
+    def toBytes(keysValuePair: AsymmetricCipherKeyPair): (String, String) = {
```
rtkaczyk: `toBytes` -> `toHexStrings` ?

src/main/scala/io/iohk/ethereum/network/package.scala
```diff
@@ -19,4 +25,38 @@ package object network {
...
+      val pubKey = keysValuePair.getPublic.asInstanceOf[ECPublicKeyParameters].toNodeId
+      val privKey = keysValuePair.getPrivate.asInstanceOf[ECPrivateKeyParameters].getD.toByteArray
+      (Hex.toHexString(pubKey), Hex.toHexString(privKey))
+    }
+
+    def fromBytes(publicKey: String, privateKey: String): AsymmetricCipherKeyPair = {
```
rtkaczyk: `fromBytes` -> `fromHexStrings` ?

src/main/scala/io/iohk/ethereum/network/package.scala
```diff
@@ -19,4 +25,38 @@ package object network {
       Hex.decode(nodeId)
     curve.getCurve.decodePoint(bytes)
   }
+
```
mcsherrylabs: This probably deserves it's own object in ti's own file.And I think it's orthogonal to network. Possibly 'crypto' or 'utils'

rtkaczyk:
Looks good, but I would change those names. I see the word `bytes`, but the type is `String` and I get confused 😉

src/main/scala/io/iohk/ethereum/network/package.scala
```diff
@@ -19,4 +25,38 @@ package object network {
...
+  }
+
+  def loadAsymmetricCipherKeyPair(filePath: String): AsymmetricCipherKeyPair = {
+    if(!new File(filePath).exists()){
+      val keysValuePair = generateKeyPair()
+
+      //Write keys to file
+      val (pub, priv) = AsymmetricCipherKeyPairSerializable.toBytes(keysValuePair)
+      val file = new PrintWriter(filePath)
+      file.write(pub ++ "\n" ++ priv)
```
mcsherrylabs: Can we load this as a system property? (I believe so)

src/test/scala/io/iohk/ethereum/network/AsymmetricCipherKeyPairSerializableSpec.scala
```diff
@@ -0,0 +1,34 @@
...
+    privateKeyParam.getD shouldBe privateKeyParamObtained.getD
+    privateKeyParam.getParameters shouldBe privateKeyParamObtained.getParameters
+    privateKeyParam.isPrivate shouldBe privateKeyParamObtained.isPrivate
+
+    //Serializing
+    val (pub2, priv2) = AsymmetricCipherKeyPairSerializable.toBytes(keysValuePairObtained)
+    ByteString(pub2) shouldBe ByteString(pub)
+    ByteString(priv2) shouldBe ByteString(priv)
+  }
+}
```
mcsherrylabs: :+1:

src/main/scala/io/iohk/ethereum/network/package.scala
```diff
@@ -19,4 +25,38 @@ package object network {
...
+      file.close()
+
```
mcsherrylabs: Should file.close happen in a 'finally'?

src/main/scala/io/iohk/ethereum/network/package.scala
adamsmo: my concern is that we are storing private key as plain text, but I guess there is no other option

mcsherrylabs: We *could* encrypt it with a passphrase. But not today.... :)

AlanVerbner: @adamsmo me too but I guess this task was only for storing this in plain text. That being said +1 to address this in the future as @mcsherrylabs said :smile:

build.sbt
```diff
@@ -38,4 +38,7 @@ val root = project.in(file("."))
 scalacOptions := Seq("-unchecked", "-deprecation", "-feature")
 
 (scalastyleConfig in Test) := baseDirectory.value / "scalastyle-test-config.xml"
-scalastyleSources in Test ++= {(unmanagedSourceDirectories in Integration).value}
\ No newline at end of file
+scalastyleSources in Test ++= {(unmanagedSourceDirectories in Integration).value}
+
+fork := true
+javaOptions += "-Dkey-file-format=%s\n%s"
```
adamsmo: minor: new line at end of the file

src/main/scala/io/iohk/ethereum/network/package.scala
```diff
@@ -19,4 +25,26 @@ package object network {
       Hex.decode(nodeId)
     curve.getCurve.decodePoint(bytes)
   }
+
+  def loadAsymmetricCipherKeyPair(filePath: String): AsymmetricCipherKeyPair = {
```
adamsmo: maybe I am missing something but this method name is `loadAsymmetricCipherKeyPair` but this method writes keys to file `file.write(keyFileFormat.format(pub, priv))` should we change the name?

adamsmo: ignore this comment

src/main/scala/io/iohk/ethereum/network/package.scala
```diff
@@ -19,4 +25,26 @@ package object network {
...
+      try {
+        val keyFileFormat: String = System.getProperty("key-file-format")
+        file.write(keyFileFormat.format(pub, priv))
+      } finally {
+        file.close()
+      }
+
+      keysValuePair
+    } else {
+      val List(pub, priv) = Source.fromFile(filePath).getLines().toList
```
adamsmo: what will happen if I change file format from `-Dkey-file-format=%s\n%s` for example to -Dkey-file-format=%s %s

build.sbt
whysoserious: We keep all our configuration in `application.conf` Why did we decide to read this particular setting from `javaOptions`?

src/main/scala/io/iohk/ethereum/network/package.scala
whysoserious: ^^

build.sbt
whysoserious: Also, what's the usecase for making this format configurable?

ntallar:
@whysoserious Having the key format configurable was required by @mcsherrylabs in:

> Can we load this as a system property? (I believe so)

src/main/scala/io/iohk/ethereum/network/package.scala
mcsherrylabs: As the guys have said I don't think we need a configurable format for this file. If I gave you a bad steer on this at some stage - apologies.

ntallar: No problem, I can change it back quickly :+1:

mcsherrylabs: @ntallar I meant the CR/LF character could be loaded from system properties! I'll get an example...

whysoserious:
LGTM

PR #61, opened 2017-02-27
[EC-83] UInt256 / Int256

whysoserious:


src/main/scala/io/iohk/ethereum/vm/Int256.scala
```diff
@@ -0,0 +1,201 @@
+package io.iohk.ethereum.vm
```
rtkaczyk: Can we move it to `util`?

mcsherrylabs: As discussed Radek, leave it in the VM for now.

rtkaczyk: Putting it in a different package doesn't mean we'll be using it elsewhere. With the package structure as is we'll have a cyclic dependency between `domain` and `vm`. Just saying.

src/main/scala/io/iohk/ethereum/vm/Int256.scala
```diff
@@ -0,0 +1,201 @@
...
+
+  def apply(bytes: IndexedSeq[Byte]): UInt256 = {
+    require(bytes.length <= Size, s"Input byte array cannot be longer than $Size: ${bytes.length}")
+    UInt256(bytes.foldLeft(BigInt(0)) { (n, b) => (n << 8) + (b & 0xff) })
+  }
+
+  def apply(int256: Int256): UInt256 = new UInt256(toUnsignedBigInt(int256.n))
+
+  implicit def numericToUInt256[N: Integral](n: N): UInt256 =
+    apply(BigInt(implicitly[Integral[N]].toLong(n)))
```
rtkaczyk: We probably don't want to convert BigInts automatically as that can conversion can be lossy

src/main/scala/io/iohk/ethereum/vm/Int256.scala
```diff
@@ -0,0 +1,201 @@
...
+
+  val Size: Int = 32
+
+  val Zeros: ByteString = ByteString(Array.fill[Byte](Size)(0))
+
+  val Modulus: BigInt = BigInt(2).pow(Size * 8)
+
+}
+
+trait Int256Like[T <: Int256Like[T]] extends ScalaNumber with Ordered[T] with ScalaNumericConversions {
```
rtkaczyk: What do we need `ScalaNumericConversions` for?

ntallar:
Apart from @rtkaczyk comments, LGTM!

src/main/scala/io/iohk/ethereum/vm/Int256.scala
rtkaczyk: @mcsherrylabs @whysoserious on the branch for EC-85, that @AlanVerbner is working on, `DataWord` escapes the `vm` package and is used in transaction execution. That's one more reason to put it in `util` package.

src/main/scala/io/iohk/ethereum/vm/Int256.scala
whysoserious: Yes, but let's fix that in next iteration.

src/main/scala/io/iohk/ethereum/vm/Int256.scala
whysoserious: I just used conventions used in other numeric types.

PR #62, opened 2017-02-27
[EC-94] [EC-80] Managing known blocks to peers

adamsmo:
Keep track of latest block number for peer


src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -30,6 +31,7 @@ import scala.concurrent.duration._
 class PeerActor(
     nodeStatusHolder: Agent[NodeStatus],
     rlpxConnectionFactory: ActorContext => ActorRef,
+    supervisor: ActorRef,
```
LukasGasior1: this name may be confusing, supervisor in Akka usually refers to parent actor

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -68,16 +70,17 @@ class PeerManagerActor(
...
 
 object PeerManagerActor {
   def props(nodeStatusHolder: Agent[NodeStatus], storage: Blockchain): Props =
-    Props(new PeerManagerActor(nodeStatusHolder, peerFactory(nodeStatusHolder, storage)))
-
-  def peerFactory(nodeStatusHolder: Agent[NodeStatus], storage: Blockchain): (ActorContext, InetSocketAddress) => ActorRef = { (ctx, addr) =>
-    val id = addr.toString.filterNot(_ == '/')
-    ctx.actorOf(PeerActor.props(nodeStatusHolder, storage), id)
-  }
+    Props(new PeerManagerActor(nodeStatusHolder) {
```
LukasGasior1: any reason for this change?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
LukasGasior1: I now see it is just `context.parent` can we use that instead?

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -48,6 +47,9 @@ class PeerManagerActor(
     case Terminated(ref) =>
       peers -= ref.path.name
 
+    case broadcast: PeerActor.Broadcast[_] =>
+      peers.values.foreach(_.ref ! broadcast)
```
LukasGasior1: we can just send `SendMessage(...)` to peers here. That way we don't have to handle `Broadcast` in `PeerActor`.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -360,6 +384,8 @@ object PeerActor {
 
   case class SendMessage[M <: Message](message: M)(implicit val enc: RLPEncoder[M])
 
+  case class Broadcast[M <: Message](block: M)
```
LukasGasior1: can we move it to PeerManager?

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
LukasGasior1: if so, maybe we can do the same with PeerActor? (just to stay consistent)

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -295,29 +297,51 @@ class PeerActor(
...
 
+    var currentPeerMaxBlock: BigInt = 0
+
     def receive: Receive =
       handleSubscriptions orElse handleTerminated(rlpxConnection) orElse
       handlePeerChainCheck(rlpxConnection) orElse handlePingMsg(rlpxConnection) orElse
       handleBlockFastDownload(rlpxConnection) orElse {
       case RLPxConnectionHandler.MessageReceived(message) =>
         log.debug("Received message: {}", message)
+        updateMaxBlock(message)
```
LukasGasior1: can we also update max block in `waitingForChainForkCheck`? we receive block headers there

LukasGasior1:
This is just a suggestion: maybe we can have separate actor that handles new block broadcast? It could subscribe to each Peers BlockHeaders and NewBlock and manage max blocks. Then if new block arrives it could get the peers from PeerManager (`GetPeers`) and send the message for those that are behind. It could reduce the complexity in (already quite big) PeerActor.

If you think it doesn't make sense then disregard this comment :)

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
adamsmo: nope, it is hard to mock parent in tests, but can change name to `peerManager`

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
adamsmo: I was thinking about that, but it would create circular dependency also peer manager only passes Broadcast to other peers

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
LukasGasior1: Not that hard actually, see `FastSyncBlockBodiesRequestHandlerSpec` for example

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -295,29 +297,51 @@ class PeerActor(
...
+      case m: BlockHeaders =>
+        val maxBlockNumber = m.headers.map(_.number).fold(0: BigInt) { case (a, b) => if (a > b) a else b }
+        if (maxBlockNumber > currentPeerMaxBlock) {
+          currentPeerMaxBlock = maxBlockNumber
+        }
+      case m: NewBlock =>
+        if (m.block.header.number > currentPeerMaxBlock)
+          currentPeerMaxBlock = m.block.header.number
+          supervisor ! Broadcast(m)
+      case _ =>
```
ntallar: I'm wondering if we should also handle the case of a `NewBlockHashes` message. Some of the new blocks are sent using that message (for example, in EthereumJ: https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/net/server/ChannelManager.java#L313)

adamsmo: when we receive new block hash then we will ask for the header by this hash so we will update max block number

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
adamsmo: to be able to construct peer and pass peer manager to it, will change peer actor

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
adamsmo: yes, I did it this way to indicate that it is different case then getting message from other peer

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
adamsmo: thx 👍

adamsmo:
I was looking at https://github.com/input-output-hk/etc-client/pull/65 and adjusted this PR accordingly

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
adamsmo: added

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -177,19 +180,19 @@ class PeerActor(
       daoBlockHeaderOpt match {
         case Some(_) if daoForkValidator.validate(msg).isEmpty =>
           log.info("Peer is running the ETC chain")
-          context become new HandshakedHandler(rlpxConnection, remoteStatus).receive
+          context become new HandshakedHandler(rlpxConnection, remoteStatus, daoForkBlockNumber).receive
 
         case Some(_) if nodeStatusHolder().blockchainStatus.totalDifficulty < daoForkBlockTotalDifficulty =>
           log.info("Peer is not running the ETC fork, but we're not there yet. Keeping the connection until then.")
-          context become new HandshakedHandler(rlpxConnection, remoteStatus).receive
+          context become new HandshakedHandler(rlpxConnection, remoteStatus, daoForkBlockNumber).receive
```
ntallar: I'm not sure if `currentPeerMaxBlock` should be `daoForkBlockNumber` in this case. Shouldn't it be `0` as the peer is not running the ETC fork?

adamsmo: but it responded with block header that has number `daoForkBlockNumber` so we should send to it only blocks that have number grater then `daoForkBlockNumber`
@LukasGasior1 shouldnt we disconnect from the peer that is not running on ETC chain?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -293,31 +296,54 @@ class PeerActor(
...
         log.debug("Received message: {}", message)
+        updateMaxBlock(message)
         notifySubscribers(message)
         processMessage(message)
 
       case s: SendMessage[_] =>
+        updateMaxBlock(s.message)
         rlpxConnection.sendMessage(s.message)(s.enc)
 
+      case GetMaxBlockNumber(actor) => actor ! MaxBlockNumber(currentPeerMaxBlock)
```
ntallar: Minor detail: couldn't we also do the same as with `GetStatus`, using `sender()` instead of having to pass an actor with `MaxBlockNumber`?

adamsmo: we can do it with the sender, I added explicit reference to ensure that it is specified because non-actor can also send message to actor

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
LukasGasior1: if we're not at the fork yet we can make use of that peer until reaching the fork

adamsmo: @LukasGasior1 do we have mechanism that prevent us from starting fast sync from this peer?

LukasGasior1: nice catch @adamsmo I think the only protection we currently have is to have at least X peers before starting the sync (and hope at least one of them is etc chain after the fork). We definately need a ticket to investigate/fix (if needed) that.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -270,23 +270,46 @@ class PeerActor(
...
-    def notifySubscribers(message: Message): Unit = {
+    private def updateMaxBlock(message: Message) = {
+      message match {
+        case m: BlockHeaders =>
+          update(m.headers.map(_.number))
+        case m: NewBlock =>
+          update(Seq(m.block.header.number))
+        case m: NewBlockHashes =>
+          update(m.hashes.map(_.number))
+        case _ =>
```
adamsmo: I recall that we agreed to support the latest version of protocol

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -270,23 +270,46 @@ class PeerActor(
...
+      }
+
+      def update(ns: Seq[BigInt]) = {
+        val maxBlockNumber = ns.fold(0: BigInt) { case (a, b) => if (a > b) a else b }
+        if (maxBlockNumber > currentPeerMaxBlock) {
+          context become new HandshakedHandler(rlpxConnection, initialStatus, maxBlockNumber).receive
```
LukasGasior1: can we have var for that instead? (maybe pass initialPeerMaxBlock?) if, at some point, someone calls `context.become` before or after `updateMaxBlock` it won't work as expected

LukasGasior1:
other than my one comment, LGTM

PR #64, opened 2017-02-28
[EC-54] call opcodes

rtkaczyk:
Changes:
- added instructions: CALL, CALLCODE, DELEGATECALL
- introduced `WorldStateProxy` 
- added several other instructions that were needed for testing: ADDRESS, ORIGIN, CALLER, CALLDATASIZE, CALLDATACOPY, EXTCODESIZE, EXTCODECOPY, GAS, INVALID


src/main/scala/io/iohk/ethereum/vm/AccountRetriever.scala
```diff
@@ -0,0 +1,9 @@
+package io.iohk.ethereum.vm
+
+import io.iohk.ethereum.domain.{Account, Address}
+
+trait AccountRetriever {
```
whysoserious: Do we still need that? :)

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -546,6 +591,127 @@ case object LOG3 extends LogOp(0xa3)
...
+      val context = state.context.copy(
+        env = env,
+        startGas = startGas,
+        world = world1)
+
+      VM.run(context)
+    }
+
+    val validCall =
+      state.env.callDepth < OpCode.MaxCallDepth &&
```
whysoserious: Just a remark. We need to test that JVM doesn't throw `StackOverflowExeption` and provide guidelines for setting `-Xss` option.

src/main/scala/io/iohk/ethereum/vm/AccountRetriever.scala
rtkaczyk: Nope :)

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
rtkaczyk: Yes, we could, although with MaxCallDepth == 1024 that's highly unlikely. Such test could be done after EC-86

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -59,8 +58,12 @@ case class SignedTransaction(
   }
 
 
-  lazy val recoveredAddress: Option[Array[Byte]] =
-    recoveredPublicKey.map(key => crypto.kec256(key).slice(FirstByteOfAddress, LastByteOfAddress))
+  lazy val recoveredSenderAddress: Option[Address] =
+    for {
+      key <- recoveredPublicKey
+      addrBytes = crypto.kec256(key).slice(FirstByteOfAddress, LastByteOfAddress)
+      if addrBytes.length == Address.Length
```
LukasGasior1: in which transactions can addBytes.length be less than Address.Length?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -291,23 +306,41 @@ case object SHA3 extends OpCode(0x20, 2, 1, G_sha3) {
...
+
+case object CALLDATACOPY extends OpCode(0x37, 3, 0, G_verylow) {
+  protected def exec(state: ProgramState): ProgramState = {
+    val (Seq(memOffset, dataOffset, size), stack1) = state.stack.pop(3)
+    val data = OpCode.sliceBytes(state.inputData, dataOffset.intValue, size.intValue)
+    val mem1 = state.memory.store(memOffset, data)
+    state.withStack(stack1).withMemory(mem1).step()
+  }
+
+  protected def varGas(state: ProgramState): BigInt = {
```
LukasGasior1: why don't we use `override` keyword?

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
rtkaczyk: I don't know, but see the remark I have added in the `Address` class

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
rtkaczyk: because we don't have to 😁

src/main/scala/io/iohk/ethereum/vm/Transfer.scala
```diff
@@ -0,0 +1,5 @@
+package io.iohk.ethereum.vm
+
+import io.iohk.ethereum.domain.Address
+
+case class Transfer(from: Address, to: Address, value: BigInt)
```
LukasGasior1: is this used?

src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
```diff
@@ -0,0 +1,42 @@
...
+    getAccount(address).get
+
+  def accountExists(address: Address): Boolean =
+    getAccount(address).isDefined
+
+
+  def transfer(from: Address, to: Address, value: BigInt): WS = {
+    val debited = getGuaranteedAccount(from).updateBalance(-value)
+    val credited = getAccount(to).getOrElse(Account.Empty).updateBalance(value)
+    saveAccount(from, debited).saveAccount(to, credited).asInstanceOf[WS]
```
LukasGasior1: `.asInstanceOf[WS]` is this needed?

src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  def newAddress(creatorAddr: Address): (Address, WS) = {
+    val creator = getGuaranteedAccount(creatorAddr)
+    val hash = kec256(rlp.encode(RLPList(creatorAddr.bytes, creator.nonce)))
+    val addr = Address(hash.takeRight(Address.Length))
+    addr -> saveAccount(creatorAddr, creator.increaseNonce)
```
LukasGasior1: shouldn't we also `saveAccount` here?

src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
rtkaczyk: Unfortunately yes, because it's a path dependent type. I believe it's a flaw in Scala

src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
rtkaczyk: No, why? We just want an address for an account that will be created next. We would be saving it twice.

src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
```diff
@@ -1,13 +1,51 @@
...
 
-import akka.util.ByteString
+import io.iohk.ethereum.domain._
 
+
+object ProgramContext {
+  def apply(stx: SignedTransaction, blockHeader: BlockHeader, world: WorldStateProxy): ProgramContext = {
+    import stx.tx
+
+    val senderAddress = stx.recoveredSenderAddress.get // FIXME: get, it should be validated but...
```
rtkaczyk: 2 questions that have appeared while implementing this:
- What is the use case for an invalid `SignedTransaction`?
- Do we still need the `Transaction` / `SignedTransaction` separation?

src/main/scala/io/iohk/ethereum/domain/Account.scala
```diff
@@ -1,21 +1,26 @@
 package io.iohk.ethereum.domain
 
 import akka.util.ByteString
-import org.spongycastle.util.encoders.Hex
+
+object Account {
+  val Empty = Account(0, 0, ByteString.empty, ByteString.empty)
```
adamsmo: there are hashes for empty storage root and empty evm code
https://github.com/input-output-hk/etc-client/blob/phase/3/blockSync/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncNodesRequestHandler.scala#L126

src/main/scala/io/iohk/ethereum/domain/Account.scala
```diff
@@ -1,21 +1,26 @@
...
-       """.stripMargin
-  }
+  def updateBalance(value: BigInt): Account =
+    copy(balance = balance + value)
+
+  def increaseNonce: Account =
+    copy(nonce = nonce + 1)
+
+  def withCode(codeHash: ByteString): Account =
+    copy(codeHash = codeHash)
```
adamsmo: just comment - EVM code is assigned when account is created and it cannot change

src/main/scala/io/iohk/ethereum/domain/Address.scala
```diff
@@ -19,10 +22,14 @@ case class Address(bytes: ByteString) {
 
   import Address.Length
 
+  //FIXME: can we get rid off it and simply make sure that address is always 20 bytes, by either truncating or extending
```
adamsmo: should we change this now?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -546,6 +592,127 @@ case object LOG3 extends LogOp(0xa3)
...
+      state
+        .withStack(stack2)
+        .withMemory(mem1)
+        .spendGas(gasAdjustment)
+        .step()
+
+    } else {
+      val stack2 = stack1.push(DataWord(1))
+      val output = result.returnData.take(outSize.intValue)
+      val mem2 = mem1.store(outAddr, result.returnData)
```
adamsmo: Why we are using `mem1` `mem2`, `stack1` `stack2`? it is hard to read
can we use some name with more meaning
for example: `stack2` => `stackAfterCall`

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -546,6 +592,127 @@ case object LOG3 extends LogOp(0xa3)
...
+
+  protected def varGas(state: ProgramState): BigInt = {
+    val (Seq(gas, to, callValue, inAddr, inSize, outAddr, outSize), _) = getParams(state)
+    val endowment = if (this == DELEGATECALL) BigInt(0) else callValue.toBigInt
+
+    val memCostIn = calcMemCost(state.memory.size, inAddr, inSize)
+    val memCostOut = calcMemCost(state.memory.size, outAddr, outSize)
+    val memCost = memCostIn max memCostOut
+
+    // FIXME: these are calculated twice (for gas and exec), especially account existence. Can we do better?
```
adamsmo: should we check now?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -546,6 +592,127 @@ case object LOG3 extends LogOp(0xa3)
...
+    val gExtra = gasExtra(state, endowment, Address(to))
+    val gCap = gasCap(state, gas, gExtra)
+    memCost + gCap + gExtra
+  }
+
+  private def getParams(state: ProgramState): (Seq[DataWord], Stack) = {
+    val (Seq(gas, to), stack1) = state.stack.pop(2)
+    val (value, stack2) = if (this == DELEGATECALL) DataWord(state.env.value) -> stack1 else stack1.pop
+    val (Seq(inAddr, inSize, outAddr, outSize), stack3) = stack2.pop(4)
```
adamsmo: what is `stack1`, what is `stack2`, what is `stack3` ?

src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
```diff
@@ -1,13 +1,51 @@
...
+
+  private def callOrCreate(world: WorldStateProxy, tx: Transaction, senderAddress: Address): (WorldStateProxy, Address, Program) = {
+    if (tx.receivingAddress.isEmpty) {
+      // contract create
+      val (address, world1) = world.newAddress(senderAddress)
+      val world2 = world1.saveAccount(address, Account.Empty)
+      val world3 = world2.transfer(senderAddress, address, tx.value)
+      val code = tx.payload
+
+      (world3, address, Program(code))
```
adamsmo: `worldAfterCall` ?

src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
```diff
@@ -1,13 +1,51 @@
...
+      tx.value, program, blockHeader, callDepth = 0)
+
+    ProgramContext(env, tx.gasLimit, world1)
+  }
+
+  private def callOrCreate(world: WorldStateProxy, tx: Transaction, senderAddress: Address): (WorldStateProxy, Address, Program) = {
+    if (tx.receivingAddress.isEmpty) {
+      // contract create
+      val (address, world1) = world.newAddress(senderAddress)
+      val world2 = world1.saveAccount(address, Account.Empty)
```
adamsmo: `worldWithAccount` ?

src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
```diff
@@ -1,13 +1,51 @@
...
+    val env = ExecEnv(recipientAddress, senderAddress, senderAddress, tx.gasPrice, tx.payload,
+      tx.value, program, blockHeader, callDepth = 0)
+
+    ProgramContext(env, tx.gasLimit, world1)
+  }
+
+  private def callOrCreate(world: WorldStateProxy, tx: Transaction, senderAddress: Address): (WorldStateProxy, Address, Program) = {
+    if (tx.receivingAddress.isEmpty) {
+      // contract create
+      val (address, world1) = world.newAddress(senderAddress)
```
adamsmo: `val (address, worldBeforeCall)` ?

src/main/scala/io/iohk/ethereum/domain/Account.scala
rtkaczyk: Should they be moved to `Account` companion object so that we can reuse them?

adamsmo: I think yes

src/main/scala/io/iohk/ethereum/domain/Account.scala
rtkaczyk: That's not 100% correct. On contract creation, first an account with balance and no code gets created. Then the code that was passed in the message call gets executed in the new account's context. The return data of that execution becomes the new account's code.

After that there is no direct way to modify an account's code, but that's a direct consequence of the set of instructions that we have available, so I don't think there's anything to enforce here. Would you agree?

Food for thought: what happens when 
1. contract A calls contract B, and as a result B self-destructs
2. after that contract A creates a new account with the same address as B

Following Yellow Paper by the letter that would mean B is not recreated, but A pays the gas and value cost for its creation, and consequently some amount Ether vanishes. @whysoserious

src/main/scala/io/iohk/ethereum/domain/Address.scala
whysoserious: I'd do that right now. It would simplify code in `SignedTransaction`. However it's up to @rtkaczyk :)

rtkaczyk: The comment is a question, I'm asking you 😄

adamsmo: yes we can 😄

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
rtkaczyk: I already explained it here: https://github.com/input-output-hk/etc-client/pull/60#commitcomment-21038366

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
rtkaczyk: Got any ideas?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
rtkaczyk: I already explained it here: https://github.com/input-output-hk/etc-client/pull/60#commitcomment-21038366

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
adamsmo: ok, but it is hard to read, also we can describe what was changed in `val` name so when I am reading I do not need to remember what state43 was for

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
adamsmo: nope

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
rtkaczyk: You don't need to remember, you need to go that line and see what happened there. This would be the same `stackAfterCall` which is almost meaningless - what call, how was it changed on that call? 

Here's a descriptive name: `val stackWithTheResultOfSuccessfulExternalCodeExecution =`. Is that something you'd really like to see? 😁 Actually you still don't know from the name what value was pushed onto the stack 😉 

Using this pseudo-mathematical notation it's at least clear `state_(n+1)` should result from modification of `state_n`, and that the final state is the one with highest `n`. That cannot be said for those "descriptive" names.

src/main/scala/io/iohk/ethereum/domain/Address.scala
rtkaczyk: I'll change that then.

src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
rtkaczyk: @adamsmo @whysoserious these question are adressesed mostly to you guys

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
adamsmo: yes I picked wrong name, mainly because I have no idea what `stack1.push(DataWord(1))` means, I am not for accumulating actions in names but rather describing them in names.
`you need to go that line and see what happened there` <- exactly, I would like to just read val name and know what is happening there.
I think this is why we name things rather then chaining calls like
```
      state
        .spendGas(-result.gasRemaining)
        .withStack(getParams(state)._2.push(DataWord(1)))
        .withMemory(state.memory.load(inAddr, inSize)._2.store(outAddr, result.returnData))
        .withWorld(result.world)
        .withAddressesToDelete(result.addressesToDelete)
        .step()
```

rtkaczyk: > I would like to just read val name and know what is happening there

I already addressed the practicality of this in my previous comments, and 2 other peers agreed my reasoning. If you're still not convinced, I would suggest for you to try what you propose in a separate branch. I will then be able to point out to you precisely why those names are no good :wink:

adamsmo: ok, LGTM then

src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
LukasGasior1: isn't the state that is returned in a tuple invalid? the nonce has been increased but new account is not there

LukasGasior1: btw, I prefer to use proper tuple notation `(addr, saveAccount(...))` - it does not need implicit conv (`->` does)

rtkaczyk: > isn't the state that is returned in a tuple invalid?

You are right about that. However, note that this proxy allows for plenty of ways to produce invalid state. Simplest example might be to save an account with 2^123 Wei, thus producing Ether out of thin air. It is the responsibility of the OpCodes to produce a valid state.

I agree about the tuple. `->` is most suitable for mappings. I will change that.

LukasGasior1: :+1:

src/main/scala/io/iohk/ethereum/domain/Account.scala
rtkaczyk: @adamsmo @LukasGasior1 

```
private val EmptyAccountStorageHash = ByteString(Hex.decode("56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421"))
private val EmptyAccountEvmCodeHash = ByteString(Hex.decode("c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"))
```

I know that `EmptyAccountEvmCodeHash` is simply `kec256(ByteString())` (and that's how I'm going to write it) and it's direct consequence of YP section 4.1 (p.3).

Were did we get the number for `EmptyAccountStorageHash`? How is it calculated?

adamsmo: I got it from downloading met state tree for accounts that had empty storage (not contracts) I was getting this hash, I assume that it is hash of empty mpt node

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -546,6 +592,127 @@ case object LOG3 extends LogOp(0xa3)
...
+
+      state
+        .withStack(stack2)
+        .withMemory(mem1)
+        .spendGas(gasAdjustment)
+        .step()
+
+    } else {
+      val stack2 = stack1.push(DataWord(1))
+      val output = result.returnData.take(outSize.intValue)
```
whysoserious: `output` is not used anywhere.

src/main/scala/io/iohk/ethereum/domain/Account.scala
LukasGasior1: @adamsmo @rtkaczyk It's `kec256(rlp(ByteString(""))`

adamsmo: I think we should replace constants with those calculated values

src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
whysoserious: > What is the use case for an invalid SignedTransaction?

I have no idea. @adamsmo ?

> Do we still need the Transaction / SignedTransaction separation?

Yes, we need to be able to create a `Transaction` without a signature.

rtkaczyk: > Yes, we need to be able to create a Transaction without a signature.

But why? Wasn't that meant for an _internal transaction_ which we're not going to use?

adamsmo: for transaction that we issue and is not yet signed?

rtkaczyk: Why can we not sign it while issuing?

src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
```diff
@@ -1,13 +1,51 @@
...
   * @param env set of constants for the execution
   * @param startGas initial gas for the execution
-  * @param storage representation of the storage accociated with the contract to be executed
+  * @param world provides interactions with world state
   */
-case class ProgramContext(env: ExecEnv, startGas: BigInt, storage: Storage)
+case class ProgramContext(
+  env: ExecEnv,
+  startGas: BigInt, //TODO: should we move it to ExecEnv
+  world: WorldStateProxy)
```
rtkaczyk: Now, this is quite important. Note `ProgramContext`, and consequently `ProgramState` and `ProgramResult`, are not type parameterised. This means that users of a `WorldStateProxy` implementation may need to cast it (`asInstanceOf`) to access methods which are not part of the interface, e.g. [in evm-runner](https://github.com/input-output-hk/etc-client/blob/demo/evmCallOpcodes/src/main/scala/io/iohk/ethereum/vmrunner/State.scala#L66)

However, type parameterising those classes also means we have to add type parameters to all the methods that use `ProgramState` in [OpCodes implementations](https://github.com/input-output-hk/etc-client/blob/feature/evmCallOpcodes/src/main/scala/io/iohk/ethereum/vm/OpCode.scala#L371), e.g:

```scala
protected def exec[WS <: WorldStateProxy](state: ProgramState[WS]): ProgramState[WS]
```

which produces some clutter.

Another way would be for the implementors to add new required methods to `WorldStateProxy`. But then that interface would no longer be unidirectional.

I ended up preferring casting (first time ever I think), which is quite safe in this case. But I'm interested in other opinions, especially @AlanVerbner 's since he's working on a implementation of `WorldStateProxy`.

AlanVerbner: @rtkaczyk can you point me out some examples (methods) of:

>  to add new required methods to WorldStateProxy

rtkaczyk: Example 1, return the representation of state changes:

```
def accounts: Map[Address, Account]
def codeRepo: Map[...]
def storages: Map[...]
```

Example 2, save changes to the DB directly from the implementation:

```
def commit(): Unit
```

Both those examples have a common flaw in that those methods should not be called from within the VM.

AlanVerbner: If I didn't misunderstand, that's correct. Those methods are called within the transaction executor once everything is ok so you shouldn't be worried about that. Actually, i'm keeping track of all the inner proxies I create in order to `commit` ( @whysoserious suggested to name it `persist`).once everything is ok.
The problem I see if that we hide the actual implementation I would need to _cast_ this in the `TransactionExecutor` but I really don't want tho do so.

src/test/scala/io/iohk/ethereum/vm/OpCodeFunSpec.scala
```diff
@@ -103,64 +120,48 @@ class OpCodeFunSpec extends FunSuite with OpCodeTesting with Matchers with Prope
     }
   }
 
-  ignore("ADDRESS") {
-    // to be implemented
-  }
-
-  ignore("BALANCE") {
```
adamsmo: should this be removed, I see no test for BALANCE op code?

whysoserious: This suite won't pass if we remove that. See how `verifyAllOpCodesRegistered()` for details.

LukasGasior1:
LGTM

src/test/scala/io/iohk/ethereum/vm/OpCodeFunSpec.scala
adamsmo: ok, the suit is passing so LGTM

PR #65, opened 2017-03-01
[EC-68] Block broadcast

ntallar:
## Description

Includes a `BlockBroadcastActor` for broadcasting the received new blocks from the peers.

## Notes

There are some missing features which will be included in next PRs:
- ~~Managing of known blocks to peers (#62)~~
- Evaluating BlockBroadcast propogation algorithm `[EC-87]`
- Handling of timeouts for the block header and block body requests `[EC-107]`


src/main/scala/io/iohk/ethereum/App.scala
```diff
@@ -1,59 +1,101 @@
...
+        FastSyncController.props(
+          peerManager,
+          nodeStatusHolder,
+          blockchain,
+          storagesInstance.storages.mptNodeStorage),
+        "fast-sync-controller")
+
+      fastSyncController ! FastSyncController.StartFastSync
+
+      Runtime.getRuntime.addShutdownHook(new Thread() {
```
whysoserious: Could we move it to `object App`?

src/main/scala/io/iohk/ethereum/App.scala
```diff
@@ -1,59 +1,101 @@
...
-        blockchainStatus = BlockchainStatus(Config.Blockchain.genesisDifficulty, Config.Blockchain.genesisHash, 0))
+    val appActor = actorSystem.actorOf(AppActor.props(nodeKey, blockchain, storagesInstance, actorSystem), "app")
+
+    appActor ! AppActor.StartApp
+  }
+
+}
+
+class AppActor(nodeKey: AsymmetricCipherKeyPair,
+               blockchain: Blockchain,
```
whysoserious: There is no need of passing `blockchain` as a constructor argument. It can be created inside of an `AppActor`.

src/main/scala/io/iohk/ethereum/App.scala
```diff
@@ -1,59 +1,101 @@
...
+               storagesInstance: DataSourcesComponent with StoragesComponent,
```
whysoserious: Same here. It's not used outside of the `AppActor`.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -284,6 +284,13 @@ class PeerActor(
 
       case GetStatus =>
         sender() ! StatusResponse(Handshaked(initialStatus))
+
+      case PeerActor.StartBlockBroadcast(blockchain) =>
+        val blockBroadcastActor = context.actorOf(
```
adamsmo: do we limit somehow number of broadcast actors per connection?
what will happen if actor will die, should we restart him?

src/main/scala/io/iohk/ethereum/network/BlockBroadcastActor.scala
```diff
@@ -0,0 +1,173 @@
...
+            context become processMessages(state.copy(unprocessedBlocks = state.unprocessedBlocks.tail))
+
+          case (Left(_), _) =>
+            log.info("Block {} not valid", Hex.toHexString(blockToProcess.header.hash.toArray))
+            context become processMessages(state.copy(unprocessedBlocks = state.unprocessedBlocks.tail))
+        }
+
+      case PeerManagerActor.PeersResponse(peers) if state.toBroadcastBlocks.nonEmpty =>
+        state.toBroadcastBlocks.foreach{ newBlock =>
+          sendNewBlockMsgToPeers(peers, newBlock)
```
adamsmo: should we first check `GetMaxBlockNumber` for given peer, or is this logic part of another ticket?

src/main/scala/io/iohk/ethereum/network/BlockBroadcastActor.scala
```diff
@@ -0,0 +1,173 @@
...
+            blockchain: Blockchain): Props = {
+    Props(new BlockBroadcastActor(nodeStatusHolder, peer, peerManagerActor, blockchain))
+  }
+
+  case object StartBlockBroadcast
+  case object ProcessNewBlocks
+
+  case class ProcessingState(unprocessedBlocks: Seq[Block],
+                             toBroadcastBlocks: Seq[Block],
+                             fetchedBlockHeaders: Seq[BlockHash],
```
adamsmo: should it be renamed to `fetchedBlockHashes`

src/test/scala/io/iohk/ethereum/network/p2p/BlockBroadcastActorSpec.scala
```diff
@@ -0,0 +1,295 @@
...
+
+  val PeersNumber = 5
+
+  "BlockBroadcastActor" should "subscribe for messages" in new TestSetup {
+    blockBroadcast ! BlockBroadcastActor.StartBlockBroadcast
+
+    peerProbe.expectMsg(PeerActor.Subscribe(Set(NewBlock.code, PV62.NewBlockHashes.code, BlockHeaders.code, BlockBodies.code)))
+  }
+
+  it should "broadcast new blocks received (but not to the sending peer)" in new TestSetup {
```
whysoserious: Add 
```scala
expectNoMsg(sendingPeer)
```

src/main/scala/io/iohk/ethereum/network/BlockBroadcastActor.scala
```diff
@@ -0,0 +1,173 @@
...
+
+  private def blockInStorage(hash: BlockHash): Boolean =
+    blockchain.getBlockByHash(hash).isDefined
+
+  private def blockToProcess(hash: BlockHash, state: ProcessingState): Boolean = !blockInProgress(hash, state) && !blockInStorage(hash)
+
+  //FIXME: Decide block propagation algorithm (for now we send block to every peer except the sender) [EC-87]
+  private def sendNewBlockMsgToPeers(peers: Seq[Peer], newBlock: Block) = {
+    val blockTd = blockchain.getTotalDifficultyByHash(newBlock.header.hash)
+      .getOrElse(throw new Exception(s"Block ${Hex.toHexString(newBlock.header.hash.toArray)} total difficulty is not on storage"))
```
adamsmo: where do we handle this exception?

src/main/scala/io/iohk/ethereum/App.scala
ntallar: Not if we move `storagesInstance` inside the `AppActor`. As we don't use `storagesInstance` outside of `AppActor` what do you think of keeping it here and creating the `storagesInstance` here as well?

src/main/scala/io/iohk/ethereum/App.scala
mcsherrylabs: I cant complain about an effort to encapsulate but please leave all these as arguments. And leave the shutdownhook also, have a look here https://github.com/input-output-hk/etc-client/pull/68/files#diff-4238cdc00fed700564f7ead54ac6eb08R45 at how we'll wire these things together. At the top level we want the flexibility to wire the implementations together in various ways.

src/main/scala/io/iohk/ethereum/network/BlockBroadcastActor.scala
ntallar: I thought of incorporating that in a future PR with `Handling of timeouts for the block header and block body requests`. Do you think it would be better to add it here?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
ntallar: We currently don't limit the number of broadcast actors per connection, I think we should add a separate task to look into that. I'll push a commit that monitors the BlockBroadcastActor termination,  you were right about that

src/main/scala/io/iohk/ethereum/network/BlockBroadcastActor.scala
adamsmo: I think it can be done there

src/main/scala/io/iohk/ethereum/network/BlockBroadcastActor.scala
```diff
@@ -0,0 +1,183 @@
...
+
+  override def receive: Receive = idle
+
+  private val msgsToSubscribe = Set(NewBlock.code, PV62.NewBlockHashes.code, BlockHeaders.code, BlockBodies.code)
+
+  def idle: Receive = {
+    case StartBlockBroadcast =>
+      peer ! PeerActor.Subscribe(msgsToSubscribe)
+      context watch sender()
+      context become processMessages(ProcessingState(Seq(), Seq(), Seq(), Seq()))
```
adamsmo: `Seq()` => `Seq.empty` ?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -300,11 +304,29 @@ class PeerActor(
...
         val maxBlockNumber = ns.fold(0: BigInt) { case (a, b) => if (a > b) a else b }
         if (maxBlockNumber > currentMaxBlockNumber) {
-          context become new HandshakedHandler(rlpxConnection, initialStatus, maxBlockNumber).receive
+          context become new HandshakedHandler(rlpxConnection, initialStatus, maxBlockNumber, blockBroadcastActor).receive
         }
       }
     }
 
+    def handleBlockBroadcastActorTerminated: Receive = {
+      case Terminated(peer) if blockBroadcastActor.exists(_.compareTo(peer) == 0) =>
```
adamsmo: why `compareTo` ?

src/main/scala/io/iohk/ethereum/network/BlockBroadcastActor.scala
LukasGasior1: Nil?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
ntallar: We are also watching the termination of the `rlpxConnection` so this is done to differentiate the `Terminated` messages got in both cases. I realized that we weren't doing that on `handleTerminated` so I'll add that check.

adamsmo: can we use `==` instead of `compareTo` ?

ntallar:
@adamsmo I have just added the handling of the known blocks to peers to Block broadcast, please take a look at it and let me know if any changes are needed

src/main/scala/io/iohk/ethereum/network/BlockBroadcastActor.scala
```diff
@@ -0,0 +1,172 @@
...
+
+        (blockHeader, blockParentTotalDifficulty) match {
+          case (Right(_), Some(parentTD)) =>
+            val blockToProcessDifficulty = parentTD + blockToProcess.header.difficulty
+            importBlockToBlockchain(blockToProcess, blockToProcessDifficulty)
+
+            peerManagerActor ! GetPeers
+            val newState = state.copy(
+              unprocessedBlocks = state.unprocessedBlocks.tail,
+              toBroadcastBlocks = state.toBroadcastBlocks :+ NewBlock(blockToProcess, blockToProcessDifficulty)
```
adamsmo: as I see we do not broadcast `toBroadcastBlocks`, this is intended because it will be handle in different ticket?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -300,11 +304,29 @@ class PeerActor(
...
+      blockBroadcastActor.foreach{ previousBlockBroadcastActor =>
+        context unwatch previousBlockBroadcastActor
+      }
+      val newBlockBroadcastActor = context.actorOf(
+        BlockBroadcastActor.props(nodeStatusHolder, self, context.parent, storage),
+        "blockbroadcast"
+      )
+      context watch newBlockBroadcastActor
+      newBlockBroadcastActor ! BlockBroadcastActor.StartBlockBroadcast
+      context become new HandshakedHandler(rlpxConnection, initialStatus, currentMaxBlockNumber, Some(newBlockBroadcastActor)).receive
```
adamsmo: is this correct behavior?
`context become new HandshakedHandler` 
`BlockBroadcastActor` termination does not mean that we are disconnected from the peer, I think changing context is not necessary in this case.
Summoning @LukasGasior1

adamsmo:
@ntallar looks like code will do what it should but solution is complicated - we collect incoming blockchain elements and then `PeerActor` is asking `PeerManager` for list of `peers` when `PeerManager` responds then `PeerActor` creates `BlockBroadcastMaxBlockRequestHandler` who asks `PeerActor` for MaxBlockNumber and after receiving response `BlockBroadcastMaxBlockRequestHandler` sends message to `PeerActor` to send actual blocks to peer (if I understand this correctly)

What I am trying to say is that it should work but maybe we can refactor that to be simpler and easier to maintain

side not: class `HandshakedHandler`has wrong name because it is responsible more for handling actual p2p communication then handshaked

ntallar:
As discussed with @adamsmo (and as other PR depends on this one) I think that the `BlockBroadcastActor` refactoring should be left for a later PR. If any other reviewer thinks it's a better to idea to fix it in this PR it can be done, but if not a new task has to be created to do the refactoring.

src/main/resources/application.conf
```diff
@@ -73,10 +73,10 @@ etc-client {
 
   db {
     iodb {
-      path = "/tmp/eth-client/iodb/"
+      path = "/tmp/etc-client/iodb/"
```
whysoserious: :)

PR #66, opened 2017-03-01
[EC-91] Support for sbt dist

ntallar:
## Description

Added support for running `sbt dist` to obtain a zip file with structure:
- bin: with Bash script
- lib: with the project jar and the jar for the dependencies
- conf: with the configuration files (the `application.conf` file)

Also some `scaladoc` warnings were fixed.


build.sbt
```diff
@@ -30,12 +35,18 @@ val dep = {
...
+  .settings(inConfig(Integration)(Defaults.testSettings) : _*)
 
 scalacOptions := Seq("-unchecked", "-deprecation", "-feature")
 
 (scalastyleConfig in Test) := baseDirectory.value / "scalastyle-test-config.xml"
-scalastyleSources in Test ++= {(unmanagedSourceDirectories in Integration).value}
\ No newline at end of file
+scalastyleSources in Test ++= {(unmanagedSourceDirectories in Integration).value}
+
+mappings in Universal ++= (resourceDirectory in Compile).value.listFiles()
```
mcsherrylabs: Files in a src/universal/resources/ should get dropped into the conf folder automagically?
(Have a look at https://github.com/mcsherrylabs/sss.asado/tree/master/sss.asado-nobu/src/universal/conf)

ntallar: I just changed so that the conf folder is detected automatically, however I don't like so much that for that it seems that the conf file must be duplicated (they also appear in https://github.com/mcsherrylabs/sss.asado/tree/master/sss.asado-nobu/src/main/resources).

mcsherrylabs: It's true, it sucks to have to remember to put new config in both places. The theory is that production settings will be different and more stable than development settings. Furthermore if there's a `resources.conf` that config file will end up embedded in the jar and be used first. So we could have common config in a single `resources.conf` and only have config that is different between production and development in `application.conf`. Might be worth investigating in a later task.

src/main/scala/io/iohk/ethereum/crypto/ConcatKDFBytesGenerator.scala
```diff
@@ -16,7 +16,7 @@ class ConcatKDFBytesGenerator(digest: Digest) {
     * @param outputLength length of output that will be produced by this method,
     *                     maximum value is (digest output size in bits) * (2^32 - 1) but it should not be a problem
     *                     because we are using Int
-    * @throws IllegalArgumentException ("Output length too large") when outputLength is greater than (digest output size in bits) * (2^32 - 1)
+    * @throws scala.IllegalArgumentException ("Output length too large") when outputLength is greater than (digest output size in bits) * (2^32 - 1)
```
mcsherrylabs: Is there a reason to alter these? In general, formatting should be left as is, it generates 'noise' in the PR.

ntallar: Yes, there were some warnings on the generation of the zip file related to the Scala Docs, without those I think there were some problems with the linking between them. So the changes on the comments were all related to that.

PR #67, opened 2017-03-02
[EC-56] EVM Block Information opcodes

adamsmo:
This PR adds op codes for block info
~It depends on https://github.com/input-output-hk/etc-client/pull/64~


src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -384,6 +384,56 @@ case object EXTCODECOPY extends OpCode(0x3c, 4, 0, G_extcode) {
...
+    state.withStack(stack1).step()
+  }
+}
+
+case object GASLIMIT extends OpCode(0x45, 0, 1, G_base) with ConstGas {
+  override protected def exec[W <: WorldStateProxy[W, S], S <: Storage[S]](state: ProgramState[W, S]): ProgramState[W, S] = {
+    val stack1 = state.stack.push(DataWord(state.context.env.blockHeader.gasLimit))
+    state.withStack(stack1).step()
+  }
+}
```
rtkaczyk: I think each of those new ops, except for BLOCKHASH, should be a `ConstOp`, no?

adamsmo: 👍

src/test/scala/io/iohk/ethereum/vm/OpCodeFunSpec.scala
```diff
@@ -251,28 +253,38 @@ class OpCodeFunSpec extends FunSuite with OpCodeTesting with Matchers with Prope
...
-    // to be implemented
+      withStackVerification(op, stateIn, stateOut) {
+        val (Seq(blockHeaderNumber), _) = stateIn.stack.pop(1)
+        val (Seq(blockHeaderHash), _) = stateOut.stack.pop(1)
+
+        if (stateIn.context.env.blockHeader.number - blockHeaderNumber.toBigInt <= 256) {
+          val expectedHash: DataWord = stateIn.world.getBlockHeader(blockHeaderNumber).map(bh => DataWord(bh.hash)).getOrElse(DataWord(0))
+          blockHeaderHash shouldBe expectedHash
+        } else {
+          blockHeaderHash shouldBe DataWord(0)
```
rtkaczyk: Please verify all of the state, rather than only the value pushed to the stack.

adamsmo: 👍  good point, will change that

src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
```diff
@@ -29,6 +29,7 @@ trait WorldStateProxy[WS <: WorldStateProxy[WS, S], S <: Storage[S]] {
 
   def getCode(address: Address): ByteString
   def getStorage(address: Address): S
+  def getBlockHeader(number: BigInt): Option[BlockHeader]
```
rtkaczyk: @AlanVerbner are you aware of this. Will that fit in the implementation of WPS?

Also @adamsmo shouldn't we simply return the hash? We don't have a use for the whole `BlockHeader`.

src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
```diff
@@ -29,6 +29,7 @@ trait WorldStateProxy[WS <: WorldStateProxy[WS, S], S <: Storage[S]] {
 
   def getCode(address: Address): ByteString
   def getStorage(address: Address): S
+  def getBlockHash(number: BigInt): Option[ByteString]
```
rtkaczyk: @AlanVerbner are you aware of this? Will that fit in the implementation of WPS?

AlanVerbner: @rtkaczyk yes....I think it will, actually I have the block info and i'm not using it within the WorldState

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -384,6 +384,31 @@ case object EXTCODECOPY extends OpCode(0x3c, 4, 0, G_extcode) {
...
+    val (blockNumber, stack1) = state.stack.pop
+
+    if (state.context.env.blockHeader.number - blockNumber > 256 || blockNumber > state.context.env.blockHeader.number) {
+      val stack2 = stack1.push(DataWord(0))
+      state.withStack(stack2).step()
+    } else {
+      val result = state.world.getBlockHash(blockNumber).map(DataWord(_)).getOrElse(DataWord(0))
+      val stack2 = stack1.push(result)
+      state.withStack(stack2).step()
+    }
```
rtkaczyk: This could be slightly improved so that we don't repeat the stack and state updates, in psuedo:

```
val hash = if (conditionsMet) getBlockhash(number).getOrElse(DataWord.Zero) else DataWord.Zero
val stack2 = stack1.push(result)
state.withStack(stack2).step()
```

src/test/scala/io/iohk/ethereum/vm/MockWorldState.scala
```diff
@@ -1,7 +1,7 @@
 package io.iohk.ethereum.vm
 
 import akka.util.ByteString
-import io.iohk.ethereum.domain.{Account, Address}
+import io.iohk.ethereum.domain.{Account, Address, BlockHeader}
```
rtkaczyk: This import is now unused

src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
```diff
@@ -1,7 +1,7 @@
 package io.iohk.ethereum.vm
 
 import akka.util.ByteString
-import io.iohk.ethereum.domain.{Account, Address}
+import io.iohk.ethereum.domain.{Account, Address, BlockHeader}
```
rtkaczyk: This import is now unused

adamsmo: 👍

src/test/scala/io/iohk/ethereum/vm/MockWorldState.scala
adamsmo: it is used by headers: `Map[BigInt, BlockHeader] = Map()`, but this case go as well

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -384,6 +384,32 @@ case object EXTCODECOPY extends OpCode(0x3c, 4, 0, G_extcode) {
   }
 }
 
+case object BLOCKHASH extends OpCode(0x40, 1, 1, G_blockhash) with ConstGas {
+  protected def exec[W <: WorldStateProxy[W, S], S <: Storage[S]](state: ProgramState[W, S]): ProgramState[W, S] = {
+    val (blockNumber, stack1) = state.stack.pop
+
+    val hash = if (state.context.env.blockHeader.number - blockNumber > 256 || blockNumber > state.context.env.blockHeader.number) {
```
ntallar: While looking at how other clients implement this opcode, I think I found a possible border case, if `blockNumber == state.context.env.blockHeader.number` we currently fetch the block hash but in EthereumJ it returns `DataWord(0)` in this case (https://github.com/ethereum/ethereumj/blob/master/ethereumj-core/src/main/java/org/ethereum/vm/program/Program.java#L654). Maybe the condition should be `blockNumber >= state.context.env.blockHeader.number`?

rtkaczyk: Makes sense, we can't have a hash of a currently mined block, can we?

src/test/scala/io/iohk/ethereum/vm/Generators.scala
```diff
@@ -82,11 +83,29 @@ object Generators extends ObjectGenerators {
...
+        receiptsRoot = ByteString(Hex.decode("56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")),
+        logsBloom = ByteString(Hex.decode("00" * 256)),
+        difficulty = BigInt("989772"),
+        number = blockNumber,
+        gasLimit = 131620495,
+        gasUsed = 0,
+        unixTimestamp = 1486752441,
+        extraData = ByteString(Hex.decode("d783010507846765746887676f312e372e33856c696e7578")),
+        mixHash = ByteString(Hex.decode("6bc729364c9b682cfa923ba9480367ebdfa2a9bca2a652fe975e8d5958f696dd")),
+        nonce = ByteString(Hex.decode("797a8f3a494f937b")))
```
rtkaczyk: Do we need all that?

src/test/scala/io/iohk/ethereum/vm/MockWorldState.scala
```diff
@@ -12,7 +12,8 @@ object MockWorldState {
 case class MockWorldState(
   accounts: Map[Address, Account] = Map(),
   codeRepo: Map[Address, ByteString] = Map(),
-  storages: Map[Address, MockStorage] = Map()
+  storages: Map[Address, MockStorage] = Map(),
+  headers: Map[BigInt, ByteString] = Map()
```
rtkaczyk: Name is inaccurate, but why do this anyway? `getBlockhash` could simply return a deterministic hash based on its arg, e.g.:

```
def getBlockHash(number: BigInt): Option[ByteString] = 
  Some(ByteString(kec256(number.toByteArray)))
```

src/test/scala/io/iohk/ethereum/vm/OpCodeFunSpec.scala
```diff
@@ -251,28 +253,40 @@ class OpCodeFunSpec extends FunSuite with OpCodeTesting with Matchers with Prope
...
-  }
+      doSave <- Gen.oneOf(false, true, true)
 
-  ignore("NUMBER") {
-    // to be implemented
-  }
+      addr = Address(stateIn.stack.pop._1)
+      hash = kec256(extCode)
+      world = if (doSave) stateIn.world.saveAccount(addr, Account.Empty.copy(codeHash = hash)) else stateIn.world
+    } yield stateIn.withWorld(world)
```
rtkaczyk: What are all those generators for? Did you just copy/paste? 😉

src/test/scala/io/iohk/ethereum/vm/OpCodeFunSpec.scala
```diff
@@ -251,28 +253,40 @@ class OpCodeFunSpec extends FunSuite with OpCodeTesting with Matchers with Prope
...
-  ignore("DIFFICULTY") {
-    // to be implemented
-  }
+    forAll(stateGen) { stateIn =>
+      val stateOut = executeOp(op, stateIn)
 
-  ignore("GASLIMIT") {
-    // to be implemented
+      withStackVerification(op, stateIn, stateOut) {
+        val (Seq(blockHeaderNumber), stack1) = stateIn.stack.pop(1)
```
rtkaczyk: val (blockHeaderNumber, stack1) = stateIn.stack.pop(1)

src/test/scala/io/iohk/ethereum/vm/OpCodeFunSpec.scala
```diff
@@ -251,28 +253,40 @@ class OpCodeFunSpec extends FunSuite with OpCodeTesting with Matchers with Prope
...
+        val (Seq(blockHeaderHash), _) = stateOut.stack.pop(1)
```
rtkaczyk: This is unused

src/test/scala/io/iohk/ethereum/vm/OpCodeFunSpec.scala
```diff
@@ -251,28 +253,40 @@ class OpCodeFunSpec extends FunSuite with OpCodeTesting with Matchers with Prope
...
+
+        if (stateIn.context.env.blockHeader.number - blockHeaderNumber.toBigInt <= 256) {
+          val expectedHash: DataWord = stateIn.world.getBlockHash(blockHeaderNumber).map(DataWord(_)).getOrElse(DataWord(0))
+          val expectedState = stateIn.withStack(stack1.push(expectedHash)).step()
+          stateOut shouldBe expectedState
+        } else {
+          val expectedState = stateIn.withStack(stack1.push(DataWord(0))).step()
+          stateOut shouldBe expectedState
```
rtkaczyk: Don't repeat yourself ;)

src/test/scala/io/iohk/ethereum/vm/OpCodeFunSpec.scala
```diff
@@ -7,6 +7,8 @@ import org.scalacheck.Gen
 import org.scalatest.{FunSuite, Matchers}
 import org.scalatest.prop.PropertyChecks
 
+import scala.collection.generic.SeqFactory
```
rtkaczyk: Unused?

src/test/scala/io/iohk/ethereum/vm/OpCodeFunSpec.scala
adamsmo: To generate various program states  with various world states 😉 
Since it is generative testing I think it is good to have many cases to test and since you approved this PR I assumed that this is correct approach, to check op code when other parts of the system are changing, not only stack and block hashes

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
adamsmo: actually definition form yellow paper says:
```
the hash of a block of a particular number, up to a maximum age.
0 is left on the stack if the looked for block number is greater than the current block number or more than 256 blocks behind the current block.
```
so in the case of block verification, it is possible to ask for the hash of current block but it would lead to inconsistency in block mining and block verification.
Good catch, I will change it.

src/test/scala/io/iohk/ethereum/vm/Generators.scala
adamsmo: to construct `BlockHeader` - yes
I can replace those values with nulls but it would be ugly

src/test/scala/io/iohk/ethereum/vm/MockWorldState.scala
adamsmo: to have limited number of hashes in storage, to be consistent with `state.context.env.blockHeader.number` which is the latest block and to test case when I ask for block header hash that does not exist in storage yet.
Will change 
`headers: Map[BigInt, ByteString] = Map()` => `numberOfHashes: Int`

src/test/scala/io/iohk/ethereum/vm/OpCodeFunSpec.scala
adamsmo: yep

src/test/scala/io/iohk/ethereum/vm/MockWorldState.scala
rtkaczyk: But why? It's not the world's responsibility to limit those hashes, it should be consequence of the opcode implementation. And in fact you cannot test it properly if the world is already limiting them.

src/test/scala/io/iohk/ethereum/vm/Generators.scala
rtkaczyk: I would use nulls for the values we don't need. Anyway, I guess it mostly bothers me because it is inside that function. Can we move it outside as `defaultBlockHeader` or something, please?

src/test/scala/io/iohk/ethereum/vm/MockWorldState.scala
adamsmo: if world would not limit them I am unable to test case where current block is last block of chain (we are mining)

src/test/scala/io/iohk/ethereum/vm/OpCodeFunSpec.scala
rtkaczyk: Those that do make sense. But code, memory and external accounts have nothing to do with functionality of `BLOCKHASH`. You could however add a generator for the block number, even if you're just going to repeat the defualts, to make it explicit.

src/test/scala/io/iohk/ethereum/vm/MockWorldState.scala
rtkaczyk: > I am unable to test case where current block is last block of chain (we are mining)

Huh? `state.context.env.blockHeader.number` is the last block.

adamsmo: not in all cases, when we are verifying blocks it may not be the last block in the storage

rtkaczyk: From the point of view of the executing contract it always is! This contract cannot access state modified in subsequent blocks, can it now? 

Anyway, why would that even matter? You just need to check whether the number is not greater than the current number.

adamsmo: no, it can not, but it can ask for block "from the future" and we should test that it will not return their hashes even if they are in storage

adamsmo: example of asking for block hash form the future is asking for hash of current block

rtkaczyk: Yes, and it should be the responsibility of the opcode to return 0 in that case. The state shouldn't be concerned with that. And in fact if the state returns 0, then that kind of invalidates the test, because if you change the comparison by mistake, the will not catch it.

adamsmo: yes, true
What I want to have here is both cases when current block (CB) is actually inside blockchain:
B0 B1 B2 CB B3 ...
and case when current block is at the end of block chain
B0 B1 B2 B3 ... CB
What you propose does not allow second case

adamsmo: please check https://github.com/input-output-hk/etc-client/pull/67/files#diff-432cbbf50ae41043b4807c983b793987R105

rtkaczyk: How does it not allow it? All you should care about is whether the requested number is greater or equal to the current block number. You don't need to reason any chain for that. What is `blockPlacement` for?

rtkaczyk: ok, I forgot the API returns an `Option`

src/test/scala/io/iohk/ethereum/vm/OpCodeFunSpec.scala
```diff
@@ -251,28 +251,33 @@ class OpCodeFunSpec extends FunSuite with OpCodeTesting with Matchers with Prope
...
-
-  ignore("TIMESTAMP") {
-    // to be implemented
-  }
+  test(BLOCKHASH) { op =>
+    val stateGen: Gen[PS] = for {
+      stateIn <- getProgramStateGen(
+        stackGen = getStackGen(maxWord = DataWord(256)),
+        codeGen = getByteStringGen(0, 256),
+        blockNumberGen = getBigIntGen(0, 256)
```
rtkaczyk: With `stackGen` and `blockNumberGen` we will never have a case of the block being older than 256. Maybe use 512 as max for both (although that will result in a requested number out of bounds more often than not).

`codeGen` is not needed.

adamsmo: changed

src/test/scala/io/iohk/ethereum/vm/MockWorldState.scala
```diff
@@ -27,6 +29,12 @@ case class MockWorldState(
   def getStorage(address: Address): MockStorage =
     storages.getOrElse(address, MockStorage.Empty)
 
+  def getBlockHash(number: BigInt): Option[ByteString] = if (numberOfHashes >= number && number >= 0) {
+    Some(ByteString(kec256(number.toByteArray)))
+  } else {
+    None
+  }
```
rtkaczyk: Minor. [Scala style guide](http://docs.scala-lang.org/style/control-structures.html#curly-braces) suggest avoiding curly braces in such simple expressions.

adamsmo: 👌

rtkaczyk:
```
[info] - should disconnect from non-ETC peer *** FAILED *** (3 seconds, 273 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsg while waiting for SendMessage(Disconnect(Useless peer))
```

3 seconds... just taking a note of that.

PR #68, opened 2017-03-02
Attempt to modularise components and minimise dependencies.

mcsherrylabs:
Start by adding the cake pattern to the startup process. 
This facilitates mix, match and replace  of components at startup.


adamsmo:
wrong target branch?

PR #69, opened 2017-03-02
redesign WorldStateProxy + fix for casting

rtkaczyk:
This refactors the `WorldStateProxy` in 2 ways:

1. Code and storage are now saved and accessed by `Address`. This simplifies the API, and the VM doesn't need to know anything about MPT hashes (besides it would have been impossible to implement WPS that way).

2. Everything has been type parameterised to avoid type casting (`asInstanceOf`).

Conclusion: I believe the result of type-parameterisation is absolutely dreadful. IMO, it actually shows that type safety is sometimes not worth its cost. If you agree with me, I'll take a bit more time to refactor this - remove type parametrization from `OpCode` and add casting in an isolated place (find a middle ground).


src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
```diff
@@ -4,23 +4,23 @@ import io.iohk.ethereum.domain._
...
 
     val senderAddress = stx.recoveredSenderAddress.get // FIXME: get, it should be validated but...
-    val (world1, recipientAddress, program) = callOrCreate(world, tx, senderAddress)
+    val (world1, recipientAddress, program) = callOrCreate[W, S](world, tx, senderAddress)
 
     val env = ExecEnv(recipientAddress, senderAddress, senderAddress, tx.gasPrice, tx.payload,
       tx.value, program, blockHeader, callDepth = 0)
 
-    ProgramContext(env, tx.gasLimit, world1)
+    ProgramContext(env, tx.gasLimit, world1.asInstanceOf[W])
```
whysoserious: This code does compile without `.asInstanceOf[W]` Hooray!

whysoserious:
Looks awesome.

> Everything has been type parameterised to avoid type casting (asInstanceOf).

I like that approach

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -224,16 +225,16 @@ sealed abstract class BinaryOp(code: Int, constGas: BigInt)(val f: (DataWord, Da
...
     val (Seq(a, b, c), stack1) = state.stack.pop(3)
     val res = f(a, b, c)
     val stack2 = stack1.push(res)
     state.withStack(stack2).step()
   }
 }
 
-sealed abstract class ConstOp(code: Int)(val f: ProgramState => DataWord) extends OpCode(code, 0, 1, G_base) with ConstGas {
-  protected def exec(state: ProgramState): ProgramState = {
+sealed abstract class ConstOp(code: Int)(val f: ProgramState[_ <: WorldStateProxy[_, _ <: Storage[_]], _ <: Storage[_]] => DataWord) extends OpCode(code, 0, 1, G_base) with ConstGas {
```
adamsmo: scala style error
```
[error] /Users/adam/iohk/etc-client/src/main/scala/io/iohk/ethereum/vm/OpCode.scala:236: File line length exceeds 160 characters
```

src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
AlanVerbner: I confirm that

adamsmo: yes it does

AlanVerbner:
@rtkaczyk based on:

> Conclusion: I believe the result of type-parameterisation is absolutely dreadful. IMO, it actually shows that type safety is sometimes not worth its cost. If you agree with me, I'll take a bit more time to refactor this - remove type parametrization from OpCode and add casting in an isolated place (find a middle ground).

The code looks good but yes, the type-parametrization is all around, so...are you saying you don't like the result?

rtkaczyk:
> The code looks good but yes, the type-parametrization is all around, so...are you saying you don't like the result?

Yes, I think it's horrible. The amount of boilerplate added just to prevent 2 or 3 occurrences of casting is appalling. And I'm confident we could isolate those so that casting would be 99.999% safe. True safety is overrated, especially when it comes at big cost.

But if I'm alone in that opinion, I am willing to accept yours.

src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
rtkaczyk: :+1: I was quick with this, and trying several things... There may be more little omissions like this

AlanVerbner:
Well...your last comment makes a lot of sense to me to be honest and I like that point of view so I would like to see other opinions too

whysoserious:
I am equally fine with _boilerplate_ and _casting_ solutions. Pick whatever you like more

adamsmo:
boilerplate is not prettiest thing in the world but I am ok with it if it, at least type-parametrization is present only in a few files, but pick whatever you like

src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
```diff
@@ -13,35 +13,39 @@ import io.iohk.ethereum.crypto.kec256
...
-  def getCode(codeHash: ByteString): ByteString
-  def getStorage(storageRoot: ByteString): Storage
-
-  def saveAccount(address: Address, account: Account): WS
-  def saveCode(codeHash: ByteString, code: ByteString): WS
-  def saveStorage(storageRoot: ByteString, storage: Storage): WS
+trait WorldStateProxy[WS <: WorldStateProxy[WS, S], S <: Storage[S]] {
 
+  protected def getAccount(address: Address): Option[Account]
+  protected def saveAccount(address: Address, account: Account): WS
```
LukasGasior1: actually, what is the reason why these methods can't just return `WorldStateProxy`? (and `Storage` in getStorage)?

rtkaczyk:
@whysoserious @adamsmo @AlanVerbner I gave it one more look I cannot quickly find a satisfiable middle ground solution. I don't want to take too much time on this, especially that you're OK with this approach. So let's merge it when tests pass.

BTW, if only Scala had a way to alias an expression like `[W <: WorldStateProxy[W, S], S <: Storage[S]]` then I wouldn't complain so much.

LukasGasior1:
> BTW, if only Scala had a way to alias an expression like [W <: WorldStateProxy[W, S], S <: Storage[S]] then I wouldn't complain so much.

macros :P but by no means I suggest to use it

src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
rtkaczyk: Because we need to access implementation-specific methods and values that @AlanVerbner is working on

rtkaczyk:
> macros :P

It did cross my mind. Fortunately it flew off quickly :P

src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
rtkaczyk: See, even [this simple thing](https://github.com/input-output-hk/etc-client/blob/demo/evmCallOpcodes/src/main/scala/io/iohk/ethereum/vmrunner/State.scala#L66) requires implementation specific details.

LukasGasior1: isn't the user of WSP supposed to rely on the interface instead of specific implementation?

LukasGasior1: if implementator needs these methods to return concrete types he can always override them such way (i.e specify subtype as a return type). or am I missing something obvious here?

LukasGasior1: why isn't `deleteAccount` part of WSP interface then? (cause I guess that's what that casting is for)

rtkaczyk: Yes, the user is the EVM, and it modifies the state using using the minimal interface defined by the WSP. However the provider of the WSP, which is the _transaction executor_ needs to operate on the implementation in order to save the state modifications to DB, et. al.

LukasGasior1: Alright, that makes sense now

rtkaczyk: This is another reason (not just boilerplate), why I wanted the opcodes to operate simply on `WorldStateProxy` and `Storage` instead of `[W <: WorldStateProxy[W, S], S <: Storage[S]]` - they don't even need to know that there are some types that implement these traits.

But that requires type-casting and I can't think of a way to do this quickly and elegantly.

LukasGasior1: I see the point now. Casting in layer above vs parametrized types all over the place

PR #70, opened 2017-03-03
[EC-104] evm-runner

rtkaczyk:
Need to decide how we're going to build the jar, given #66 

Edit: we now have a separate script for running it, this is good to go.


LukasGasior1:
To be honest there's quite a few things I'd change in the code (for example, make it stateless). But I realize this piece is really just a tool for us and not the priority. Besides, I have already started to use this codebase in EC-86. LGTM

src/main/scala/io/iohk/ethereum/vmrunner/State.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  def createAccount(name: String, balance: BigInt, gas: BigInt, code: ByteString, abis: Seq[ABI]): (Address, PR) = {
+    val (newAddress, _) = world.newAddress(creatorAddress)
+    val tx = MockVmInput.transaction(creatorAddress, code, balance, gas)
+    val bh = MockVmInput.blockHeader
+
+    val context: PC = ProgramContext(tx, bh, world)
+    val intermediateResult: PR = VM.run(context)
+
+    val result: PR = if (intermediateResult.error.isDefined) intermediateResult else {
```
AlanVerbner: I've also needed to add this logic. If you are not going to call `Ledger`(former TransactionExecutor) maybe it's a good idea to put this somewhere shared to be used from several places, what do you think?

rtkaczyk: If it won't be too much trouble then sure 👍 
(don't worry about it if it requires more effort)

src/universal/bin/evm-runner
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+java -cp "lib/*" io.iohk.ethereum.vmrunner.EvmRunner
```
rtkaczyk: @ntallar so I ended up doing this instead of a separate SBT task. Seems like the easiest solution. We can change it later if it's not universal enough.

LukasGasior1:
@rtkaczyk needs updating

rtkaczyk:
@LukasGasior1 I'll update as soon as this one is merged: #96

LukasGasior1:
@rtkaczyk you want us to review #96 ? :D 51 merged files

rtkaczyk:
@LukasGasior1 I think it's enough to have a look at the commits to verify to that the description is telling the truth 😉

build.sbt
```diff
@@ -26,6 +27,15 @@ val dep = {
     "ch.qos.logback" % "logback-classic" % "1.1.9",
     "org.scalatest" %% "scalatest" % "3.0.1" % "it,test",
     "org.scalacheck" %% "scalacheck" % "1.13.4" % "it,test",
+    "org.scalacheck" %% "scalacheck" % "1.13.4" % "it,test",
+    "org.scorexfoundation" %% "iodb" % "0.2.0",
+    "ch.qos.logback" % "logback-classic" % "1.1.9",
+    "org.jline" % "jline" % "3.1.2",
+    "org.scala-lang.modules" %% "scala-parser-combinators" % "1.0.5",
+    "io.circe" %% "circe-core" % circeVersion,
```
AlanVerbner: Rest block controller uses `akka-http-spray-json` it might be a good idea (not in this PR) to change it to Circe @mcsherrylabs

rtkaczyk: I started with Circe before the other JSON dependency was added. I agree we should have a single lib for JSON. I'm OK with changing it in evm-runner if we don't decide on Circe (separate PR of course).

mcsherrylabs: We're all agreed on a single dep. I thought this came up before (in Poland) and we decided on akka-http-spray-json? Circe is new and popular. I'll bring this up at VSU.

AlanVerbner: I don't remember about that discussion so it might be my fault :disappointed:

rtkaczyk: I remember we had a brief discussion in Warsaw, but I don't recall a decision being made.

AlanVerbner:
@rtkaczyk small comment. I would add `.evm-runner_history` to `.gitignore`

PR #71, opened 2017-03-06
Demo

whysoserious:


build.sbt
```diff
@@ -39,3 +39,5 @@ scalacOptions := Seq("-unchecked", "-deprecation", "-feature")
 
 (scalastyleConfig in Test) := baseDirectory.value / "scalastyle-test-config.xml"
 scalastyleSources in Test ++= {(unmanagedSourceDirectories in Integration).value}
+
+mainClass in assembly := Some("io.iohk.ethereum.App")
```
rtkaczyk: Is this just for feedback or do we want to merge this?
If the latter please remove `assembly`. We're using SBT Native Packager now.

PR #74, opened 2017-03-06
Establish cake type startup pattern.

mcsherrylabs:
Break up functionality into traits and declare that functionality's dependencies as other traits. These traits can then be mixed. matched and replaced to form full, partial or test nodes. The builder traits also offer a place to encapsulate a startup sequence that might vary across 'cakes'. 

In the beginning there was encapsulation and data hiding and that was hard to test, then there was Spring and DI and that was easier to test but data hiding and encapsulation was broken. Components were coupled with their internals through accessors. Then a compromise was reached, the internals would be exposed through the constructor, but the instance would only be used through it's interface, allowing very flexible wiring but not coupling the interface with the implementation details. (Then Spring ate it's own tail.) 

The constructors and builder traits may conspire as necessary to produce a top level interface that is then available to the rest of the system ....


LukasGasior1:
Is there any particular reason we want to go with cake pattern for DI? Have you considered other implementations, like Guice or MacWire?

rtkaczyk:
Why do people dislike the simple and clean constructor dependency injection? 😞

mcsherrylabs:
So this is using constructor dependency injection. They are just organised
into traits with deps.  Its not the full blown DI component pattern as you
might find online, my previous comment was to provide some background as we
discussed. We don't need a DI solution like guice or macwire. Most of our
interfaces are ActorRefs which are naturally isolated from their
construction anyway. Barring any killer reason to abandon it I'm happy with
this approach. Let me know if anyone spots any bugs...

src/main/scala/io/iohk/ethereum/App.scala
```diff
@@ -1,58 +1,35 @@
...
+      }
 
-    val fastSyncController = actorSystem.actorOf(
-      FastSyncController.props(
-        peerManager,
-        nodeStatusHolder,
-        blockchain,
-        storagesInstance.storages.mptNodeStorage),
-      "fast-sync-controller")
+      server ! ServerActor.StartServer(networkConfig.Server.listenAddress)
```
rtkaczyk: Won't this throw NPE?

mcsherrylabs: ... well I ran it and it didn't, if you switch to the branch and it does fail, please let me know. Is it 'server' you'd expect to be null? `server` is available through `Node`

rtkaczyk: Yes, but you're right, it's OK. I keep getting confused about Scala initialisation order which is why I don't like having it split over multiple traits.

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
```diff
@@ -0,0 +1,126 @@
...
+trait Node extends NodeKeyBuilder
+  with ActorSystemBuilder
+  with StorageBuilder
+  with BlockChainBuilder
+  with NodeStatusBuilder
+  with PeerManagerActorBuilder
+  with ServerActorBuilder
+  with FastSyncControllerBuilder
+  with JSONRpcServerBuilder
+  with ShutdownHookBuilder
```
rtkaczyk: How do we compose it, if for example we'd like to replace a Storage implementation for tests?

What's the advantage of mixing traits like that over something like: 

```
new Node(
  actorSystem,
  storage,
  blockChain,
  nodeStatus,
  ...
)
```

?

mcsherrylabs: Either 
```
trait TestStorageBuilder {
      lazy val storagesInstance =  new SharedLevelDBDataSources with Storages.DefaultStorages
    }
    
    new Node with Logger with TestStorageBuilder {
```
or 
```
new Node with Logger {

      override lazy val storagesInstance =  new SharedLevelDBDataSources with Storages.DefaultStorages
```
With the `new Node(` you have to build actorSystem, storage, blockChain  etc.  for each new node type rather than mixing in and or overriding the builder trait.

rtkaczyk: > With the new Node( you have to build actorSystem, storage, blockChain etc. for each new node type

Not necessarily, I can imagine several ways to avoid repetition (factory methods with default args, directly extending the class, ...)

Not that there's anything inherently wrong with the trait design, but I think for our needs something simpler would suffice. 

That being said, unless I'm missing something, this may simply be a case of personal preference.

mcsherrylabs: @rtkaczyk @LukasGasior1 Guys, can you finish your reviews...?

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
```diff
@@ -0,0 +1,133 @@
...
+      storagesInstance.storages.mptNodeStorage,
+      storagesInstance.storages.fastSyncStateStorage,
+      BlockValidator.validateHeaderAndBody),
+    "fast-sync-controller")
+
+}
+
+trait ShutdownHookBuilder {
+
+  def shutdown()
```
rtkaczyk: I think this falls under [this paragraph](http://docs.scala-lang.org/style/declarations.html#procedure-syntax) in that an implicit `Unit` type is discouraged. (non-blocking comment)

src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -22,6 +22,8 @@ object Config {
 
   val keysFile: String = config.getString("keys-file")
 
+  val shutdownTimeout: Duration = config.getDuration("shutdown-timeout").toMillis.millis
```
rtkaczyk: Is this added just for purity or do you think it might for example prevent those NPEs that we sometimes see when  closing the app?

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
```diff
@@ -91,16 +95,19 @@ trait FastSyncControllerBuilder {
     PeerManagerActorBuilder with
     StorageBuilder =>
 
-
   lazy val fastSyncController = actorSystem.actorOf(
```
rtkaczyk: Shouldn't this be named `syncController` now?

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
```diff
@@ -91,16 +95,19 @@ trait FastSyncControllerBuilder {
...
+    SyncController.props(
       peerManager,
       nodeStatusHolder,
+      storagesInstance.storages.appStateStorage,
       blockchain,
-      storagesInstance.storages.mptNodeStorage),
+      storagesInstance.storages.mptNodeStorage,
+      storagesInstance.storages.fastSyncStateStorage,
+      BlockValidator.validateHeaderAndBody),
     "fast-sync-controller")
```
rtkaczyk: And likewise, the actor name -> `"sync-controller"`

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
```diff
@@ -0,0 +1,133 @@
...
+trait FastSyncControllerBuilder {
+
+  self: ActorSystemBuilder with
+    ServerActorBuilder with
+    BlockChainBuilder with
+    NodeStatusBuilder with
+    PeerManagerActorBuilder with
+    StorageBuilder =>
+
+  lazy val fastSyncController = actorSystem.actorOf(
```
rtkaczyk: (don't understand why my comments on this became outdated)

Shouldn't it be called `syncController` now? And likewise the actor name below: `"sync-controlller"`?

src/main/scala/io/iohk/ethereum/utils/Config.scala
mcsherrylabs: Previously we were not shutting down the actorsystem on exit, so once that was added a 'wait' for the shutdown to complete was necessary and so a `shutdownTimeout` was born.

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
mcsherrylabs: No idea why comments became outdated ... but I've but some comment beside all the ones I've seen to show I've seen them.

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
mcsherrylabs: Changed.

mcsherrylabs:
@LukasGasior1 You can do it Lukasz!

LukasGasior1:
:eyes:

PR #75, opened 2017-03-06
[EC-96] Jump destination validation

ntallar:
## Description

Includes the validation of the destination of the `JUMP` and `JUMPI` instructions, as described in section 9.4.3 of the YP.


src/main/scala/io/iohk/ethereum/vm/Program.scala
```diff
@@ -16,6 +18,30 @@ case class Program(code: ByteString) {
   def getBytes(from: Int, size: Int): ByteString =
     code.slice(from, from + size).padTo(size, 0.toByte)
 
+  lazy val validJumpDestinations: Seq[Int] = validJumpDestinationsAfterPosition(0)
```
rtkaczyk: If anything that should be a `Set`. But what is the advantage of keeping those in a separate collection?

The `code` is already an `IndexedSeq`, so verifying whether a position is a valid JUMPDEST should be as easy as:

```
def isValidJumpDest(pos: Int): Boolean =
  getByte(pos).contains(JUMPDEST.code)
```

and just as efficient (if not more) as looking it up in precomputed `Set`.

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -446,17 +446,22 @@ case object SSTORE extends OpCode(0x55, 2, 0, G_zero) {
...
   protected def exec[W <: WorldStateProxy[W, S], S <: Storage[S]](state: ProgramState[W, S]): ProgramState[W, S] = {
     val (Seq(pos, cond), stack1) = state.stack.pop(2)
     val nextPos = if (!cond.isZero) pos.intValue else state.pc + 1
-    //TODO: JUMPDEST validation
-    state.withStack(stack1).goto(nextPos)
+    val isValidJump = !cond.isZero && state.program.validJumpDestinations.contains(nextPos)
+    if(isValidJump || cond.isZero)
+      state.withStack(stack1).goto(nextPos)
+    else
+      state.withError(InvalidJump)
```
rtkaczyk: I found this quite confusing and for a long time I couldn't figure out whether it's correct. The reason is the repeated `cond.isZero` as a factor of `isValidJump` and then in the `if`.

Now before I offer a solution there's an important question to be asked here. If we don't jump, should the next instruction be JUMPDEST? In other words, should JUMPDEST always follow JUMPI?

Using evm-runner I saw this in one of my test contracts (it's the code that determines what function is to be called):

```
evm-runner> Caller.code
  0000:    PUSH1 60
  0002:    PUSH1 40
  0004:    MSTORE
  0005:    PUSH4 ff ff ff ff
  000a:    PUSH1 e0
  000c:    PUSH1 02
  000e:    EXP
  000f:    PUSH1 00
  0011:    CALLDATALOAD
  0012:    DIV
  0013:    AND
  0014:    PUSH4 38 1b aa bc
  0019:    DUP2
  001a:    EQ
  001b:    PUSH2 00 2c
  001e:    JUMPI
  001f:    DUP1
  0020:    PUSH4 f5 53 32 ab
  0025:    EQ
  0026:    PUSH2 00 4e
  0029:    JUMPI
  002a:    JUMPDEST
  002b:    INVALID
  002c:    JUMPDEST
```

As you can see after the first JUMPI (offset 001f) there is no JUMPDEST. And that code is fine because we haven't matched a function signature and we're looking at another one. Ergo, `isValidJump` is wrong.

I would suggest the following rewrite:

```scala
    val (Seq(pos, cond), stack1) = state.stack.pop(2)
    
    if(cond.isZero)
      state.withStack(stack1).step()
    else if (state.program.isValidJumpDest(pos.intValue))
      state.withStack(stack1).goto(pos.intValue)
    else
      state.withError(InvalidJump)
```

src/main/scala/io/iohk/ethereum/vm/ProgramError.scala
```diff
@@ -9,6 +9,7 @@ case class  InvalidOpCode(code: Byte) extends ProgramError {
     f"${getClass.getSimpleName}(0x${code.toInt & 0xff}%02x)"
 }
 case object OutOfGas extends ProgramError
+case object InvalidJump extends ProgramError
```
rtkaczyk: It may be convenient for experimenting and debugging to add jump target as a parameter. Take a look at `InvalidOpCode`. Valid code offset will be at most 4 hex digits long, so print it zero-padded (`f"0x$dest%04x"`).

src/main/scala/io/iohk/ethereum/vm/Program.scala
rtkaczyk: Ah, I see why. Because of `PUSHx`...

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
ntallar: You are right, a JUMPDEST opcode is not necessary for when the jump is not done (so we step to the next instruction). Although using `isValidJump` worked in the same way, I agree that it was too confusing and I'll replace it with your suggestion

src/main/scala/io/iohk/ethereum/vm/Program.scala
ntallar: Yes, thats why a precomputed set is used :+1:
I will replace the `validJumpDestinations` with a `Set[Int]`

src/main/scala/io/iohk/ethereum/vm/Program.scala
```diff
@@ -16,6 +18,30 @@ case class Program(code: ByteString) {
...
+    * @param accum with the previously obtained valid jump destinations.
+    */
+  @tailrec
+  private def validJumpDestinationsAfterPosition(pos: Int, accum: Set[Int] = Set()): Set[Int] = {
+    if(pos < 0 || pos >= code.length) accum
+    else {
+      val byte = code(pos)
+      val opCode = OpCode.byteToOpCode.get(byte)
+      opCode match {
+        case Some(pushOp: PushOp) => validJumpDestinationsAfterPosition(pos + pushOp.code - PUSH1.code + 2, accum)
```
rtkaczyk: If you like you can use `pushOp.i` instead of `pushOp.code - PUSH1.code`

PR #77, opened 2017-03-08
[EC-118] Add regular download of blocks

adamsmo:
I created a separate actor because when FastSync finishes we have only 1000 blocks to download so parallelizing as per FastDownload is unnecessary.


src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,181 @@
...
+      //we did not get previous blocks, there is no way to resolve, blacklist peer and continue download
+      resumeWithDifferentPeer()
+    }
+
+  private def handleDownload(message: BlockHeaders, log: LoggingAdapter) = if (message.headers.nonEmpty) {
+    headers = message.headers.sortBy(_.number)
+    processBlockHeaders(headers, context.sender(), log)
+  } else {
+    //no new headers to process, schedule to ask again in future, we are at the top of chain
+    broadcasting = true
```
mcsherrylabs: Exciting times! :dancing_women:

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,181 @@
...
+        }
+      case None =>
+        log.warning("got header that does not have parent")
+        resumeWithDifferentPeer()
+    }
+  }
+
+  private def handleBlockBodies(m: BlockBodies, log: LoggingAdapter) = {
+    if (m.bodies.nonEmpty) {
+      val result = headers.zip(m.bodies).map { case (h, b) => blockValidator(h, b) }
```
mcsherrylabs: In a previous version I think the `bodies` length was checked against the 'headers` length. Is it necessary to check?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,181 @@
...
+              (blockHashToDelete, NewBlock(b, currentTd))
+            }
+
+            newBlocks.collect { case (Some(hash), _) => hash }.foreach(deleteOldBlock)
+            context.self ! BroadcastBlocks(newBlocks.map(_._2))
+          case None =>
+            log.error("no total difficulty for latest block")
+        }
+
+        headers = headers.drop(result.length)
```
mcsherrylabs: Is this going to assume the headers we get are definitely the ones we asked for? That's safe, right?!

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,181 @@
...
+  private var headers: Seq[BlockHeader] = Seq.empty
+//  private var totalDifficulty: Option[BigInt] = None
+  private var resolvingBranch = false
+  private var broadcasting = false
+
+  import Config.FastSync._
+
+  private implicit val sender: ActorRef = context.self
+
+  def regularSync(log: LoggingAdapter): Actor.Receive = {
```
mcsherrylabs: (If I do that a test fails consistently on a timeout. But the `self: ActorLoggig` seems to work...)

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -91,6 +89,30 @@ object PeerManagerActor {
...
+  }
+
+  trait FastSyncHostConfiguration {
+    val maxBlocksHeadersPerMessage: Int
+    val maxBlocksBodiesPerMessage: Int
+    val maxReceiptsPerMessage: Int
+    val maxMptComponentsPerMessage: Int
+  }
+
+  trait Storage {
```
mcsherrylabs: `Storage` is unused far as I can tell...

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,181 @@
...
+        resumeWithDifferentPeer()
+      }
+    } else {
+      //we got empty response for bodies from peer but we got block headers earlier
+      resumeWithDifferentPeer()
+    }
+  }
+
+  private def deleteOldBlock(blockHash: ByteString) = {
+    blockchain.removeBlockHeader(blockHash)
```
AlanVerbner: wouldn't it be better to create a single method called `removeBlockByHash` in blockchain that will group all this calls?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  private def scheduleResume = {
+    headers = Seq.empty
+    scheduler.scheduleOnce(checkForNewBlockInterval, context.self, ResumeRegularSync)
+  }
+
+  private def resumeWithDifferentPeer() = {
+    blacklist(context.sender(), blacklistDuration)
+    headers == Seq.empty
```
ntallar: Maybe you meant `headers = Seq.empty`?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,181 @@
...
+  val blockValidator: (BlockHeader, BlockBody) => Either[BlockError, Block]
+
+  def blacklist(peer: ActorRef, duration: FiniteDuration)
+
+  def peersToDownloadFrom: Map[ActorRef, Status]
+
+  def scheduler: Scheduler
+
+  private var headers: Seq[BlockHeader] = Seq.empty
+//  private var totalDifficulty: Option[BigInt] = None
```
ntallar: Minor detail: should we remove this if it's commented or it needs to be used somewhere?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,181 @@
...
+        //blacklist for errors in blocks
+        resumeWithDifferentPeer()
+      }
+    } else {
+      //we got empty response for bodies from peer but we got block headers earlier
+      resumeWithDifferentPeer()
+    }
+  }
+
+  private def deleteOldBlock(blockHash: ByteString) = {
```
AlanVerbner: what about receipts?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,181 @@
...
+        val blocks = result.collect { case Right(b) => b }
+
+        blockchain.getBlockHeaderByHash(blocks.head.header.parentHash)
+          .flatMap(b => blockchain.getTotalDifficultyByHash(b.hash)) match {
+          case Some(td) =>
+            var currentTd = td
+            val newBlocks = blocks.map { b =>
+              val blockHashToDelete = blockchain.getBlockHeaderByNumber(b.header.number).map(_.hash)
+              blockchain.save(b)
+              appStateStorage.putBestBlockNumber(b.header.number)
```
ntallar: Minor detail: Would it be the same to only store the best block number (with `putBestBlockNumber`) for the last block inserted instead for all of them?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,181 @@
...
+      case None =>
+        log.warning("no peers to download from")
+        scheduleResume
+    }
+  }
+
+  private def handleBlockBranchResolution(message: BlockHeaders, log: LoggingAdapter) =
+    //todo limit max branch depth?
+    if (message.headers.nonEmpty && message.headers.head.hash == headers.head.parentHash) {
+      headers = message.headers.reverse ++ headers
```
ntallar: I'm wondering, should we also sort the received `BlockHeaders` as in `handleDownload`:

> headers = message.headers.sortBy(_.number)

Could it also be the case here that the received `BlockHeaders` are not in order?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,181 @@
...
+
+    case MessageReceived(m: BlockBodies) =>
+      handleBlockBodies(m, log)
+
+    case m: BroadcastBlocks if broadcasting =>
+      peersToDownloadFrom.keys.foreach(_ ! m)
+
+    case ResolveBranch(peer) =>
+      resolvingBranch = true
+      peer ! SendMessage(GetBlockHeaders(Right(headers.head.parentHash), blockResolveDepth, skip = 0, reverse = true))
```
ntallar: Can't it be the case that while going back in the blockchain we ask for a block that we already have?
The problem with that is that in `handleBlockBodies` we first store the received blocks and then we delete the ones we had before, wouldn't we be deleting recently stored blocks?

For example if (with `B1..B4` being blocks):
```
Current blockchain: B1 B2 B3
Blocks received: B3 B4
```
Wouldn't we be inserting `B3 B4` and then deleting `B3`, ending up with our stored blocks being `B1 B2 B4`?

adamsmo: good point, we should delete B3 only if it is different then what we are getting from peer

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
adamsmo: actualy it should not be the case in first sorting, I added it there because I forgot that It is required by protocoll to return headers in order, I think instead of sorting I will and check and if they are not in order then we will bkacklist peer and start downloading from different one

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
adamsmo: 👍 will add delete for receipts

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
adamsmo: yes I did, thx

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
adamsmo: 👍

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
adamsmo: added check
https://github.com/input-output-hk/etc-client/pull/77/files#diff-6d6971fddeafb58d207e230d258a58c0R169

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
adamsmo: in this version I am just zipping headers with bodies and then perform validation, if block body will not match validation will fail, also zip will automatically pick same number of headers as we got bodies
also it is correct state if we have more headers then bodies, it means that we should ask for remaining bodies in next request

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
adamsmo: I add self, it turn out that this was the fix for tests
https://github.com/input-output-hk/etc-client/pull/77/commits/a22aaaddfa0cf617694033ad2ed7117381712434

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
ntallar: I agree, we could also not insert the block to the blockchain if we already have it (that is if `blockHashToDelete.isEmpty`), right? As this is a minor optimization there's no need to include it right now.

adamsmo: if it is empty it may indicate that we are inserting new block that is not in chain yet

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncController.scala
```diff
@@ -56,6 +65,9 @@ class FastSyncController(
         log.warning("Cannot start fast sync, not enough peers to download from. Scheduling retry in {}", startRetryInterval)
         scheduleStartFastSync(startRetryInterval)
       }
+    case StartRegularSync =>
+      context become (handlePeerUpdates orElse regularSync())
+      self ! StartRegularSync
```
LukasGasior1: can we just call a method here (`askForHeaders`?) ? It looks a little bit strange to resume the same message here

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncController.scala
```diff
@@ -5,29 +5,38 @@ import akka.actor._
...
 import io.iohk.ethereum.network.p2p.messages.PV62.{BlockHeaders, GetBlockHeaders}
 import io.iohk.ethereum.network.{PeerActor, PeerManagerActor}
 import io.iohk.ethereum.utils.{Config, NodeStatus}
 
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+
 import scala.concurrent.duration._
 
 class FastSyncController(
```
LukasGasior1: we should probably rename it to `SyncController`, it's no longer `Fast` only

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncController.scala
```diff
@@ -271,7 +285,8 @@ class FastSyncController(
       fastSyncStateStorage.purge()
       log.info("Fast sync finished")
       context.parent ! FastSyncDone
-      context stop self
+      context become idle
+      self ! StartRegularSync
```
LukasGasior1: should we rename the `finish` method? `finishFastSync` maybe?

LukasGasior1: or maybe even better, move this lines to separate method (`startRegularSync` ?)

and then update `processSyncing` to something like:
```
if (fastDownloadFullySynced) {
  finishFastSync()
  startRegularSync()
```

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+import io.iohk.ethereum.network.p2p.messages.CommonMessages.{NewBlock, Status}
+import io.iohk.ethereum.network.p2p.messages.PV62._
+import io.iohk.ethereum.network.p2p.validators.BlockValidator.BlockError
+import io.iohk.ethereum.utils.Config
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.util.Try
+
+trait RegularSyncController {
+  self: Actor with ActorLogging with BlacklistSupport =>
```
LukasGasior1: (if you rename FastSyncController) maybe just `with SyncController`?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncController.scala
LukasGasior1: in fact, we could try splitting the logic even further into separate classes and have something like
`class SyncController(...) extends Actor with FastSync with RegularSync`

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  val appStateStorage: AppStateStorage
+  val blockchain: Blockchain
+  val blockValidator: (BlockHeader, BlockBody) => Either[BlockError, Block]
+  def peersToDownloadFrom: Map[ActorRef, Status]
```
LukasGasior1: these won't be needed if you define selftype as above

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  private var headers: Seq[BlockHeader] = Seq.empty
```
LukasGasior1: can we name that more precisely? are these all headers we have? all headers downloaded during regular sync? headers downloaded in some time period?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+  private def resumeWithDifferentPeer() = {
+    blacklist(sender(), blacklistDuration)
+    headers = Seq.empty
+    context.self ! ResumeRegularSync
+  }
+
+  private def checkHeaders(headers: Seq[BlockHeader]): Boolean =
+    headers.zip(headers.tail).forall { case (parent, child) => parent.hash == child.parentHash && parent.number + 1 == child.number }
+
+  private def bestPeer: Option[ActorRef] = Try(peersToDownloadFrom.maxBy { case (_, status) => status.totalDifficulty }._1).toOption
```
LukasGasior1: just a note: this status is the initial status we received in `Status` message, it does not necessarily mean this peer is still the best one.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+
+    bestPeer match {
+      case Some(peer) =>
+        val blockNumber = appStateStorage.getBestBlockNumber()
+
+        peer ! Subscribe(Set(BlockHeaders.code, BlockBodies.code))
+        peer ! SendMessage(GetBlockHeaders(Left(blockNumber + 1), blockHeadersPerRequest, skip = 0, reverse = false))
+      case None =>
+        log.warning("no peers to download from")
+        scheduleResume
```
LukasGasior1: let's use `()` for side-effecting functions

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+      processBlockHeaders(headers, sender())
+    } else {
+      //we did not get previous blocks, there is no way to resolve, blacklist peer and continue download
+      resumeWithDifferentPeer()
+    }
+
+  private def handleDownload(message: BlockHeaders) = if (message.headers.nonEmpty) {
+    headers = message.headers
+    processBlockHeaders(headers, sender())
+  } else {
```
LukasGasior1: maybe we should try asking another peer here?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  private def askForHeaders() = {
+
+    bestPeer match {
+      case Some(peer) =>
+        val blockNumber = appStateStorage.getBestBlockNumber()
+
+        peer ! Subscribe(Set(BlockHeaders.code, BlockBodies.code))
+        peer ! SendMessage(GetBlockHeaders(Left(blockNumber + 1), blockHeadersPerRequest, skip = 0, reverse = false))
```
LukasGasior1: what happens if this peer does not send a response (for example, connections gets lost)? does regular sync stop in such situation?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+    } else {
+      //we did not get previous blocks, there is no way to resolve, blacklist peer and continue download
+      resumeWithDifferentPeer()
+    }
+
+  private def handleDownload(message: BlockHeaders) = if (message.headers.nonEmpty) {
+    headers = message.headers
+    processBlockHeaders(headers, sender())
+  } else {
+    //no new headers to process, schedule to ask again in future, we are at the top of chain
```
LukasGasior1: only according to one peer we asked

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+  private def processBlockHeaders(headers: Seq[BlockHeader], peer: ActorRef) = {
+    val parentByNumber = blockchain.getBlockHeaderByNumber(headers.head.number - 1)
+
+    parentByNumber match {
+      case Some(parent) if checkHeaders(headers) =>
+        //we have same chain
+        if (parent.hash == headers.head.parentHash) {
+          peer ! SendMessage(GetBlockBodies(headers.take(blockBodiesPerRequest).map(_.hash)))
+        } else {
+          context.self ! ResolveBranch(sender())
```
LukasGasior1: let's use `peer` instead of `sender()` since we already pass it. (alternatively remove this parameter)

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  private def scheduleResume = {
+    headers = Seq.empty
+    scheduler.scheduleOnce(checkForNewBlockInterval, context.self, ResumeRegularSync)
+  }
+
+  private def resumeWithDifferentPeer() = {
+    blacklist(sender(), blacklistDuration)
+    headers = Seq.empty
+    context.self ! ResumeRegularSync
```
LukasGasior1: I think you can just use `self`

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+  } else {
+    //no new headers to process, schedule to ask again in future, we are at the top of chain
+    broadcasting = true
+    scheduleResume
+  }
+
+  private def processBlockHeaders(headers: Seq[BlockHeader], peer: ActorRef) = {
+    val parentByNumber = blockchain.getBlockHeaderByNumber(headers.head.number - 1)
+
+    parentByNumber match {
```
LukasGasior1: doesn't it produce a warning? you're missing `case Some(_)` (when !checkHeaders)

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+      case Some(parent) if checkHeaders(headers) =>
+        //we have same chain
+        if (parent.hash == headers.head.parentHash) {
+          peer ! SendMessage(GetBlockBodies(headers.take(blockBodiesPerRequest).map(_.hash)))
```
LukasGasior1: again, what if we never receive a response?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+      resumeWithDifferentPeer()
+    }
+  }
+
+  private def scheduleResume = {
+    headers = Seq.empty
+    scheduler.scheduleOnce(checkForNewBlockInterval, context.self, ResumeRegularSync)
+  }
+
+  private def resumeWithDifferentPeer() = {
```
LukasGasior1: maybe pass the `peer` param just so that we are sure we blacklist the actual peer (and not for example, `self` if someone was to call this function somewhere else)

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        headers = headers.drop(result.length)
+        if (headers.nonEmpty) {
+          sender() ! SendMessage(GetBlockBodies(headers.take(blockBodiesPerRequest).map(_.hash)))
+        } else {
+          context.self ! ResumeRegularSync
+        }
+      } else {
+        //blacklist for errors in blocks
```
LukasGasior1: how about a log here?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+
+    case m: BroadcastBlocks if broadcasting =>
+      peersToDownloadFrom.keys.foreach(_ ! m)
+
+    case ResolveBranch(peer) =>
+      requestHeadersForNewBranch(peer)
+  }
+
+  private def requestHeadersForNewBranch(peer: ActorRef) = {
+    resolvingBranch = true
```
ntallar: Should we set `resolvingBranch` to `false` after resolving the branch fork?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+        val blocks = result.collect { case Right(b) => b }
+
+        blockchain.getBlockHeaderByHash(blocks.head.header.parentHash)
+          .flatMap(b => blockchain.getTotalDifficultyByHash(b.hash)) match {
+          case Some(td) =>
+            var currentTd = td
+            val newBlocks = blocks.map { b =>
+              val blockHashToDelete = blockchain.getBlockHeaderByNumber(b.header.number).map(_.hash).filter(_ != b.header.hash)
+              blockchain.save(b)
+              appStateStorage.putBestBlockNumber(b.header.number)
```
LukasGasior1: this can be done only once after the loop (or can't it?)

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
adamsmo: absolutely yes, missed that, thx

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,171 @@
...
+              currentTd += b.header.difficulty
+              blockchain.save(b.header.hash, currentTd)
+              (blockHashToDelete, NewBlock(b, currentTd))
+            }
+
+            newBlocks.collect { case (Some(hash), _) => hash }.foreach(blockchain.removeBlock)
```
LukasGasior1: since code inside `map` is already mutable maybe we can do that there?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncController.scala
LukasGasior1: nvm we can leave it as is

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncController.scala
adamsmo: I was trying to do it (split to FastSync and RegularSync), but it required to much rework

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
adamsmo: we kind of are with `scheduleResume()` if we get new block from different peer we will switch to that peer

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
adamsmo: yes according to best peer we asked but with another run we may ask different peer

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,193 @@
...
+
+    case m: BroadcastBlocks if broadcasting =>
+      peersToDownloadFrom.keys.foreach(_ ! m)
+
+    case ResolveBranch(peer) =>
+      requestHeadersForNewBranch(peer)
+
+    case PeerTimeOut(peer) =>
+      blacklist(peer, blacklistDuration)
+      askForHeaders()
```
ntallar: Should we set `headersQueue` to `Nil` here?
What happens if this timeout happens while resolving the branch, should we also set `resolvingBranch` to `false`?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,193 @@
...
+    timeout.foreach(_.cancel())
+    timeout = None
+  }
+
+  private def scheduleResume() = {
+    headersQueue = Seq.empty
+    scheduler.scheduleOnce(checkForNewBlockInterval, self, ResumeRegularSync)
+  }
+
+  private def resumeWithDifferentPeer() = {
```
ntallar: Maybe we need set `resolvingBranch` to false here as well. In the scenario that we are resolving a branch, we get an unexpected response and we have to call `resumeWithDifferentPeer` in `handleBlockBranchResolution`, shouldn't `resolvingBranch` be set to false so that the next `BlockHeaders` are handled properly?

src/main/resources/application.conf
```diff
@@ -73,6 +73,9 @@ etc-client {
     nodes-per-request = 300
     min-peers-to-choose-target-block = 2
     target-block-offset = 500
+
+    check-for-new-block-interval = 1.seconds
```
AlanVerbner: Dont forget to add this ones in `src/universal/conf/application.conf`

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
adamsmo: we should set `resolvingBranch` to `false`

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
```diff
@@ -5,30 +5,37 @@ import akka.actor._
 import akka.agent.Agent
```
whysoserious: Organize imports

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,195 @@
...
+      askForHeaders()
+
+    case ResumeRegularSync =>
+      askForHeaders()
+
+    case MessageReceived(m: BlockHeaders) if !resolvingBranch =>
+      cancelTimeout()
+      handleDownload(m)
+
+    case MessageReceived(m: BlockHeaders) if resolvingBranch =>
```
whysoserious: Just making sure, are `BlockHeaders` always sorted wrt. block number?

adamsmo: they should, and we are checking that:
  https://github.com/input-output-hk/etc-client/pull/77/files#diff-6d6971fddeafb58d207e230d258a58c0R185

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,195 @@
...
+        peer ! Subscribe(Set(BlockHeaders.code, BlockBodies.code))
+        sendWithTimeout(peer, SendMessage(GetBlockHeaders(Left(blockNumber + 1), blockHeadersPerRequest, skip = 0, reverse = false)))
+      case None =>
+        log.warning("no peers to download from")
+        scheduleResume()
+    }
+  }
+
+  private def handleBlockBranchResolution(message: BlockHeaders) =
+    //todo limit max branch depth?
```
whysoserious: Can we blacklist a peer which would require to drop too many blocks from currently processed branch?

I'd do that in a separate PR (and YT ticket).

adamsmo: this is what this question is about 😉

whysoserious: I'll add a ticket.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -0,0 +1,195 @@
...
+          .flatMap(b => blockchain.getTotalDifficultyByHash(b.hash)) match {
+          case Some(td) =>
+            var currentTd = td
+            val newBlocks = blocks.map { b =>
+              val blockHashToDelete = blockchain.getBlockHeaderByNumber(b.header.number).map(_.hash).filter(_ != b.header.hash)
+              blockchain.save(b)
+              appStateStorage.putBestBlockNumber(b.header.number)
+              currentTd += b.header.difficulty
+              blockchain.save(b.header.hash, currentTd)
+              blockHashToDelete match {
```
LukasGasior1: this can just be `blockHashToDelete.foreach(blockchain.removeBlock)`

LukasGasior1: also shouldn't we remove this block before we insert? or it doesn't matter?

adamsmo: it does not matter because they have different hash

PR #78, opened 2017-03-08
[EC-101] using uint256 - refactoring approach

rtkaczyk:
This is a refactoring PR, no functionality has been changed. Some interfaces were changed wrt the numeric type.

I took a bit different approach than expected previously. I removed `Int256Like` and `Int256` because:
- we simply don't need those, apart from intermediate arithmetic results there would be absolutely no use for the signed version
- maintaining those different versions is complicated
- I thought about privatising `Int256` and having `signed` method on `UInt256` that would change the behaviour of comparison, division, etc. E.g. `x.signed > y` but I think that could be confusing (the result of `signed` should never be kept anywhere, only one `signed`) and I decided `s`-prefixed methods are simpler (`x sgt y`).

Following changes were made:
- all occurrences of `BigInt` were removed from `vm` package (apart from the implementation of UInt256 of course)
- implicit conversions from compatible numeric types added
- methods in UInt256 grouped by category
- multiple variable names changed to reflect the type change
- variable names like `addr` were changed to `offset` when referring to memory or storage (to avoid confusion with account adresses)
- several other minor refactorings


src/main/scala/io/iohk/ethereum/vm/UInt256.scala
```diff
@@ -0,0 +1,196 @@
...
+
+  def apply(n: BigInt): UInt256 =
+    new UInt256(boundBigInt(n))
+
+  def apply(b: Boolean): UInt256 =
+    if (b) One else Zero
+
+  def apply[N: Integral](n: N): UInt256 = {
+    val num = implicitly[Integral[N]]
+    apply(BigInt(num.toLong(n)))
```
whysoserious: What about creating `apply` methods for all `Integral` types except `BigInt`?

rtkaczyk: Not sure, we actually don't even need this one, as all the other standard numeric types are smaller than `BigInt` and will be implicitly converted. 

It could make a difference when we move away from BigInt-based implementation. But we could still change this to `def apply(n: Long): UInt256` and all smaller types (`Int`, `Short`, `Byte`) will be handled. I think I'd go with this one. What do you think?

PR #79, opened 2017-03-08
Added nodeId.keys to .gitignore

AlanVerbner:
Just that, nodeId.keys should not be pushed


LukasGasior1:
actually, shouldn't we by default store it in `datadir`? maybe it's worth updating config by the occasion

mcsherrylabs:
... I think that makes sense... arguments against?

AlanVerbner:
@LukasGasior1 @mcsherrylabs just added `datadir` to config and updated all the keys to use it

LukasGasior1:
LGTM!

PR #80, opened 2017-03-08
Ensure ETC peer is used as fast sync target

LukasGasior1:
Use only ETC peers to choose fast sync target block. We may still make use of non-etc (or unknown chain) peers, until reaching the fork.


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncController.scala
```diff
@@ -12,6 +14,7 @@ import io.iohk.ethereum.domain.{BlockHeader, Blockchain}
 import io.iohk.ethereum.network.p2p.messages.PV62.{BlockHeaders, GetBlockHeaders}
 import io.iohk.ethereum.network.{PeerActor, PeerManagerActor}
 import io.iohk.ethereum.network.p2p.messages.CommonMessages.Status
```
AlanVerbner: Intellij says it's unused, can we remove it?

PR #81, opened 2017-03-08
[EC-117] Resume fast sync after restart

whysoserious:


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncController.scala
```diff
@@ -1,26 +1,26 @@
...
-    peerManager: ActorRef,
-    nodeStatusHolder: Agent[NodeStatus],
-    appStateStorage: AppStateStorage,
-    blockchain: Blockchain,
-    mptNodeStorage: MptNodeStorage,
-    externalSchedulerOpt: Option[Scheduler] = None)
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+
+class FastSyncController(peerManager: ActorRef,
```
LukasGasior1: let's stick to one formatting style

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncController.scala
```diff
@@ -114,7 +114,8 @@ class FastSyncController(
           log.info("Received target block from peer, starting fast sync")
 
           scheduler.schedule(0.seconds, printStatusInterval, self, PrintStatus)
-          context become new SyncingHandler(targetBlockHeader).receive
+
+          context become new SyncingHandler(targetBlockHeader, fastSyncStateStorage).receive
```
LukasGasior1: you don't have to pass `fastSyncStateStorage` there. it's already accessible in whole `FastSyncController`

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncController.scala
```diff
@@ -175,22 +176,38 @@ class FastSyncController(
...
-    private var downloadedNodesCount: Int = 0
+    private var mptNodesQueue: Set[HashType] = initialSyncState.mptNodesQueue
+    private var nonMptNodesQueue: Set[HashType] = initialSyncState.nonMptNodesQueue
+    private var blockBodiesQueue: Set[ByteString] = initialSyncState.blockBodiesQueue
+    private var receiptsQueue: Set[ByteString] = initialSyncState.receiptsQueue
+    private var downloadedNodesCount: Int = initialSyncState.downloadedNodesCount
 
     private var assignedHandlers: Map[ActorRef, ActorRef] = Map.empty
-
     private var bestBlockHeaderNumber: BigInt = appStateStorage.getBestBlockNumber()
```
LukasGasior1: we should also keep that in sync state (as discussed on slack)

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncController.scala
```diff
@@ -175,22 +176,38 @@ class FastSyncController(
...
 
+    val hashTypeSetPersist: Option[Cancellable] = if (continueAfterRestart) {
```
LukasGasior1: the name of this val seems no longer valid, how about `syncStatePersistCancellable`?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncController.scala
```diff
@@ -175,22 +176,38 @@ class FastSyncController(
...
+      val syncStateStorageActor: ActorRef = context.actorOf(Props[FastSyncStateActor])
```
LukasGasior1: let's add a name for this child actor `actorOf(..., "state-storage")` maybe?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+  * Persists current state of fast sync to a storage. Can save only one state at a time.
+  * If during persisting new state is received then it will be saved immediately after current state
+  * was persisted.
+  * If during persisting more than one new state is received then only the last state will be kept in queue.
+  */
+class FastSyncStateActor extends Actor with ActorLogging {
+
+  override val supervisorStrategy: SupervisorStrategy = OneForOneStrategy(){ case _ => Escalate }
```
LukasGasior1: this actor does not have any children, this is useless. If you want the logic for exception in this actor, add it in it's parent (FastSyncController)

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  def receive: Receive = {
+    // after initialization send a valid Storage reference
+    case storage: FastSyncStateStorage => context become idle(storage)
+  }
+
+  def idle(storage: FastSyncStateStorage): Receive = {
+    // begin saving of the state to the storage and become busy
+    case state: SyncState => persistState(storage, state)
+    //
```
LukasGasior1: unnecessary comment

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
```diff
@@ -0,0 +1,72 @@
...
+    case GetStorage => sender() ! storage
+  }
+
+  def busy(storage: FastSyncStateStorage, state: Option[SyncState]): Receive = {
+    // update state waiting to be persisted later. we only keep newest state
+    case state: SyncState => context become busy(storage, Some(state))
+    // exception was thrown during persisting of a state. push
+    case Failure(e) => throw e
```
LukasGasior1: just a suggestion: can we think of more "graceful" way of doing that? for example informing the parent with separate message

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
```diff
@@ -0,0 +1,72 @@
...
+    // state was saved in the storage. become idle
+    case Success(s: FastSyncStateStorage) if state.isEmpty => context become idle(s)
+    // state was saved in the storage but new state is already waiting to be saved.
+    case Success(s: FastSyncStateStorage) if state.isDefined => state.map(persistState(s, _))
```
LukasGasior1: also 
become busy(s, __None__)  ?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
```diff
@@ -0,0 +1,72 @@
...
+    //
```
LukasGasior1: please remove

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
LukasGasior1: nvm I now see it already happens in persistState

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
```diff
@@ -0,0 +1,72 @@
...
+    case GetStorage => sender() ! storage
+  }
+
+  private def persistState(storage: FastSyncStateStorage, syncState: SyncState): Unit = {
+    import context.dispatcher
```
LukasGasior1: we can consider using separate thread pool for doing io in the future

src/main/scala/io/iohk/ethereum/db/storage/FastSyncStateStorage.scala
```diff
@@ -0,0 +1,48 @@
...
+  val syncStateKey: String = "fast-sync-state"
+
+}
+
+class FastSyncStateStorage(val dataSource: DataSource) extends KeyValueStorage[String, SyncState] {
+  type T = FastSyncStateStorage
+
+  import FastSyncStateStorage._
+
+  override val namespace: IndexedSeq[Byte] = Namespaces.NodeHashSetNamespace
```
LukasGasior1: `Namespaces.FastSyncStateNamespace`?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncController.scala
```diff
@@ -175,22 +176,38 @@ class FastSyncController(
...
+      syncStateStorageActor ! syncStateStorage
+      val hashTypeSetPersist: Cancellable = scheduler.schedule(
```
LukasGasior1: you reuse the same name as above here, maybe just wrap with `Some { ... }`

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
```diff
@@ -0,0 +1,72 @@
...
+    }
+    persistingQueues pipeTo self
+    context become busy(storage, None)
+  }
+
+}
+
+object FastSyncStateActor {
+
+  case object GetStorage
```
AlanVerbner: Is this used only in tests? Why isn't `SyncState` returned instead?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncController.scala
whysoserious: Needs more investigation..

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
whysoserious: Parent actor wouldn't know what to do with this message as well. Benefit of throwing this exception is that it will be handled by a supervisor somewhere in the actor hierarchy.

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
whysoserious: Let's wait half a year before we go back to that :)

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
whysoserious: > Is this used only in tests?

Yes.

> Why isn't SyncState returned instead?

Good point.

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
```diff
@@ -0,0 +1,72 @@
...
+    case Success(s: FastSyncStateStorage) if state.isDefined => state.map(persistState(s, _))
+    //
+    case GetStorage => sender() ! storage
+  }
+
+  private def persistState(storage: FastSyncStateStorage, syncState: SyncState): Unit = {
+    import context.dispatcher
+    val persistingQueues: Future[Try[FastSyncStateStorage]] = Future {
+      lazy val result = Try { storage.putSyncState(syncState) }
+      if (log.isDebugEnabled) {
```
whysoserious: > Do you think it will help us to detect, for example, a bottleneck there?

Absolutely. Core reason why I started working on that PR is that I ran some benchmarks first :)

> Wouldn't be a good idea to create something to be reused in some other places? 

Good idea. I would do that in a separate mini PR. Let's chat tomorrow.

whysoserious:
@AlanVerbner, @LukasGasior1 I addressed all your remarks. This PR is read for a second round of reviews :)

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
AlanVerbner: :metal:

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
```diff
@@ -0,0 +1,72 @@
...
+  * was persisted.
+  * If during persisting more than one new state is received then only the last state will be kept in queue.
+  */
+class FastSyncStateActor extends Actor with ActorLogging {
+
+  override val supervisorStrategy: SupervisorStrategy = OneForOneStrategy(){ case _ => Escalate }
+
+  def receive: Receive = {
+    // after initialization send a valid Storage reference
+    case storage: FastSyncStateStorage => context become idle(storage)
```
whysoserious: `Storage` is not static. Methods `.putSyncState` and `.purge` return new copy of a `Storage`.

It's not put into constructor in order to avoid storing reference to an outdated copy of `Storage`.

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  def idle(storage: FastSyncStateStorage): Receive = {
+    // begin saving of the state to the storage and become busy
+    case state: SyncState => persistState(storage, state)
+    //
+    case GetStorage => sender() ! storage
```
whysoserious: Good point, however see my other comment.

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
mcsherrylabs: Got it. (doh)

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
LukasGasior1: that's the kind of future I was thinking about :)

LukasGasior1:
Apart from one (possible) issue it's LGTM.

Consider following scenario:
receipts queue = [4, 5, 6]
bodies queue = [2, 4, 5]
best block header = 1


you just receive a portion of receipts, say [2, 3].
turns out there's already a body for block 3, so we update best block in `app state`
(we still don't have full block 2, but it's not a problem because block body 2 is in the queue, it won't get lost)

now the app restarts.

on startup, we take the recent persisted sync state for the queues (we don't know when that was! since `appState` and `syncState` are not updated at the same time, so this can be state from a minute before app stopped) and the most recent block from `appStateStorage` as `bestBlockHeaderNumber`, which is 3

the result is, we think best block is 3, but since we don't know when did we last dump sync state before app stopped, we **cannot** be sure that we won't loose anything that was queued (it could be, for example, block body 2 if it was added to the queue few seconds before we received the aforementioned receipts)

hope that makes sense

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncStateActor.scala
```diff
@@ -0,0 +1,70 @@
...
+    case Success(s: FastSyncStateStorage) if stateToPersist.isEmpty => context become idle(s)
+    // state was saved in the storage but new state is already waiting to be saved.
+    case Success(s: FastSyncStateStorage) if stateToPersist.isDefined => stateToPersist.map(persistState(s, _))
+
+    case GetStorage => sender() ! storage.getSyncState()
+  }
+
+  private def persistState(storage: FastSyncStateStorage, syncState: SyncState): Unit = {
+    import context.dispatcher
+    val persistingQueues: Future[Try[FastSyncStateStorage]] = Future {
```
LukasGasior1: won't this Future block on one of Akka's threads (since you're using `context.dispatcher`)? shouldn't we use implicit global ec here instead?

whysoserious: Let's keep it as it is. We need to do more research which EC we should use. I am adding new ticket in YT.

LukasGasior1:
LGTM

PR #82, opened 2017-03-08
Fallback to reference.conf if configuration key is not found in user's config

whysoserious:
This will allow us to support a following scenario:
* Create a custom configuration eg.:
```
etc-client.network.server-address.port = 9077
```
* Run `etc-client` like this:
```
java -jar -Dconfig.file="custom.conf" etc-client.jar
```
Prior to that commit user would have to add one more line in a config:
```
include "application"
```
(and know somehow that our config is stored in an `application.conf`)


src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -16,7 +16,7 @@ import scala.concurrent.duration._
 
 object Config {
 
-  private val config = ConfigFactory.load().getConfig("etc-client")
+  private val config = ConfigFactory.load().withFallback(ConfigFactory.defaultOverrides()).getConfig("etc-client")
```
mcsherrylabs: Is this change (specific line) necessary? The reference.conf in the jar should be picked up by 
`ConfigFactory.load().getConfig("etc-client")`. Was it tried without the 'withFallback'?

whysoserious: > Can I presume it was tried without the 'withFallback'?

Yes and config is not parsed properly without it.

mcsherrylabs: Got it!

rtkaczyk: What do you mean? I just tried it without that extra method and it works fine?

rtkaczyk:
```
$ git diff
diff --git a/src/main/resources/reference.conf b/src/main/resources/reference.conf
index a1152c6..5292d39 100644
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -40,7 +40,7 @@ etc-client {
     rpc {
       enabled: true
       interface: "127.0.0.1"
-      port: 8546
+      port: 9999
     }
   }

diff --git a/src/main/scala/io/iohk/ethereum/utils/Config.scala b/src/main/scala/io/iohk/ethereum/utils/Config.scala
index 823eb92..2dcaa90 100644
--- a/src/main/scala/io/iohk/ethereum/utils/Config.scala
+++ b/src/main/scala/io/iohk/ethereum/utils/Config.scala
@@ -16,7 +16,7 @@ import scala.concurrent.duration._

 object Config {

-  private val config = ConfigFactory.load().withFallback(ConfigFactory.defaultOverrides()).getConfig("etc-client")
+  private val config = ConfigFactory.load().getConfig("etc-client")

   val clientId: String = config.getString("client-id")
```

```
15:20:20.406 akka.io.TcpListener - Successfully bound to /127.0.0.1:9999
```

whysoserious: let me double check this..

PR #83, opened 2017-03-08
[EC-109] Fix log messages from peers

AlanVerbner:
## Description

This task involves cleaning up the logging categories and configuration such that a clear, sensible log appears on the console by default and everything else goes into the log file.

## Notes 

- A sample of how end users will see this running on console can be found [here](https://asciinema.org/a/106318)
- Some logging levels were changed, if you see someone incorrect or missing please let me know :smile: 
- **After discussing with @mcsherrylabs, the main idea is to keep the end user console clean in order to show important information and only display that the app is _doing something_ without dumping too much info that won't be useful. That being said, the possibility to view more detailed logs should be available too**
- It would be great to stick to the rules stated [here](https://svn.apache.org/repos/asf/commons/proper/logging/tags/LOGGING_1_0_3/usersguide.html#Best%20Practices). 
- Ideally I would like to log with `TRACE` the heavy info within the lower layer (like network messages, for example) so we can put in debug connection life cycle data for. Unfortunately akka [seems not to provide it](https://groups.google.com/forum/#!topic/akka-user/XHVOqQD8E4k) so it was skipped from console intentionally. If you think this is confusing we can spend some more time creating a new `LoggingAdapter` or even considering another alternative.


src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -178,23 +178,23 @@ class PeerActor(
 
       daoBlockHeaderOpt match {
         case Some(_) if daoForkValidator.validate(msg).isEmpty =>
-          log.info("Peer is running the ETC chain")
+          log.warning("Peer is running the ETC chain")
```
mcsherrylabs: Is this really a warning? Sounds like info?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -178,23 +178,23 @@ class PeerActor(
...
           context become new HandshakedHandler(rlpxConnection, remoteStatus, daoForkBlockNumber, None).receive
 
         case Some(_) if nodeStatusHolder().blockchainStatus.totalDifficulty < daoForkBlockTotalDifficulty =>
-          log.info("Peer is not running the ETC fork, but we're not there yet. Keeping the connection until then.")
+          log.warning("Peer is not running the ETC fork, but we're not there yet. Keeping the connection until then.")
```
mcsherrylabs: This also sounds like something that could happen in the normal flow of events ... info?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -178,23 +178,23 @@ class PeerActor(
...
           context become new HandshakedHandler(rlpxConnection, remoteStatus, daoForkBlockNumber, None).receive
 
         case Some(_) =>
-          log.info("Peer is not running the ETC fork, disconnecting")
+          log.warning("Peer is not running the ETC fork, disconnecting")
           disconnectFromPeer(rlpxConnection, Disconnect.Reasons.UselessPeer)
 
         case None if remoteStatus.totalDifficulty < daoForkBlockTotalDifficulty =>
-          log.info("Peer is not at ETC fork yet. Keeping the connection until then.")
+          log.warning("Peer is not at ETC fork yet. Keeping the connection until then.")
```
mcsherrylabs: If the above changes to info then this would probably should too.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -253,7 +253,7 @@ class PeerActor(
 
   def handlePeerChainCheck(rlpxConnection: RLPxConnection): Receive = {
     case RLPxConnectionHandler.MessageReceived(message@GetBlockHeaders(Left(number), _, _, _)) if number == 1920000 =>
-      log.info("Received message: {}", message)
+      log.debug("Received message: {}", message)
```
mcsherrylabs: :+1:

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
AlanVerbner: you are right, this one is wrong

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
AlanVerbner: Well...i tried to apply _other runtime situations that are undesirable or unexpected, but not necessarily "wrong"_. I can change it tho.

PR #84, opened 2017-03-08
[EC-116] Refactor block broadcast

ntallar:
## Description

Includes simplifying the `BlockBroadcastActor` by creating two traits:
- *GetPeersHandler*: for handling the obtaining of the currently connected peers. 
- *MaxBlockNumberRequest*: for handling the obtaining of the MaxBlockNumber of the peers and sending the blocks to them if they are unknown to them.


src/main/scala/io/iohk/ethereum/network/BlockBroadcastActor.scala
```diff
@@ -169,6 +152,21 @@ object BlockBroadcastActor {
...
+
+    def withToBroadcastBlocks(newToBroadcastBlocks: Seq[NewBlock]): ProcessingState =
+      this.copy(toBroadcastBlocks = newToBroadcastBlocks)
+
+    def withUnprocessBlocks(newUnprocessedBlocks: Seq[Block]): ProcessingState =
+      this.copy(unprocessedBlocks = newUnprocessedBlocks)
+  }
+
+  object ProcessingState {
+    def empty: ProcessingState = ProcessingState(Nil, Nil, Map(), Nil, Nil)
```
mcsherrylabs: This can be a `val`. It never changes.

src/main/scala/io/iohk/ethereum/network/MaxBlockNumberRequestHandler.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+    * This function handles the RequestMaxBlockNumber message:
+    *   - It requests the MaxBlockNumber from each peer that didn't send the blocks
+    *   - Modifies the state to keep the blocks that need to could sent to each peer (if they don't already know of them)
+    */
+  def handleRequestMaxBlockNumber(state: ProcessingState): Receive = {
+    case RequestMaxBlockNumber(peers, blocksToSend) =>
+      if(blocksToSend.nonEmpty) {
+        val newState = peers.foldLeft(state) { case (recState, peerToSendReq) =>
```
mcsherrylabs: I think you can remove `case`?

src/main/scala/io/iohk/ethereum/network/MaxBlockNumberRequestHandler.scala
```diff
@@ -0,0 +1,65 @@
...
+          sender() ! PeerActor.SendMessage(newBlockMsg)
+        }
+      }
+      stopWatchingPeer(sender(), state)
+
+    case Terminated(terminatedPeer) => stopWatchingPeer(terminatedPeer, state)
+  }
+
+  private def stopWatchingPeer(terminatedPeer: ActorRef, state: ProcessingState): Unit = {
+    context unwatch terminatedPeer
```
mcsherrylabs: Afaik You don't need to `unwatch` a terminated, it's removed automatically.

src/main/scala/io/iohk/ethereum/network/BlockBroadcastActor.scala
```diff
@@ -169,6 +152,21 @@ object BlockBroadcastActor {
 
   case class ProcessingState(unprocessedBlocks: Seq[Block],
                              toBroadcastBlocks: Seq[NewBlock],
+                             toBroadcastBlocksToEachPeer: Map[ActorRef, Seq[NewBlock]],
                              fetchedBlockHashes: Seq[BlockHash],
-                             blockHeaders: Seq[BlockHeader])
+                             blockHeaders: Seq[BlockHeader]) {
+
+    def withToBroadcastBlocksToEachPeer(newToBroadcastBlocksToPeers: Map[ActorRef, Seq[NewBlock]]): ProcessingState =
+      this.copy(toBroadcastBlocksToEachPeer = newToBroadcastBlocksToPeers)
```
adamsmo: is `this` needed here?

adamsmo:
https://github.com/input-output-hk/etc-client/pull/84/files#diff-173ebd65502d06f780eeb3e2a3d28cb4R132
there is also block chain modification in regular download
saving blockchain in 2 actors may lead to inconsistent block chain
scenario:
Actor 1 checks that block header is valid
Actor 2 checks that block header is valid
Actor 2 saves block
Actor 1 saves block

ntallar:
@adamsmo In that scenario both blocks would be different forks on the blockchain right?
You are right, the function `importBlockToBlockchain` was auxiliary, we should properly consider a possible blockchain forking, will add a `FIXME` to remind us of that.

src/main/scala/io/iohk/ethereum/network/MaxBlockNumberRequestHandler.scala
```diff
@@ -0,0 +1,64 @@
+package io.iohk.ethereum.network
+
+import akka.actor.{ActorRef, Terminated}
+import io.iohk.ethereum.network.BlockBroadcastActor.ProcessingState
+import io.iohk.ethereum.network.MaxBlockNumberRequestHandler.{PeerTerminated, RequestMaxBlockNumber}
+import io.iohk.ethereum.network.PeerActor.{GetMaxBlockNumber, MaxBlockNumber}
+import io.iohk.ethereum.network.p2p.messages.CommonMessages.NewBlock
+
+//FIXME: Decide block propagation algorithm (for now we send block to every peer except the sender) [EC-87]
+trait MaxBlockNumberRequestHandler { this: BlockBroadcastActor =>
```
adamsmo: I think name is miss leading, I see that some of methods are handling MaxBlockNumber message, but this trait is responsible for sending blocks to peers

ntallar: You are right, it was misleading. I just changed this trait's name to `BlockSendingHandler`

PR #87, opened 2017-03-13
Added configuration to universal/conf/application.conf

ntallar:
## Description

Moves configuration added to `resources/application.conf` to `universal/conf/application.conf` that caused the zipped project not to work. Also changed some comments to remove warnings when using `sbt dist`.


ntallar:
Yes, there was a warning thrown when using `sbt dist` due to the linking to private functions in scaladoc.
> I'm going to assume there's a good reason you're altering a comment.

PR #88, opened 2017-03-13
Proposition of change that may improve mpt download performance

adamsmo:


src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
```diff
@@ -198,7 +198,7 @@ class SyncController(
       else
         (SyncState.empty, fastSyncStateStorage.purge())
 
-    private var mptNodesQueue: Set[HashType] = initialSyncState.mptNodesQueue
+    private var mptNodesQueue: Seq[HashType] = initialSyncState.mptNodesQueue
     private var nonMptNodesQueue: Set[HashType] = initialSyncState.nonMptNodesQueue
     private var blockBodiesQueue: Set[ByteString] = initialSyncState.blockBodiesQueue
     private var receiptsQueue: Set[ByteString] = initialSyncState.receiptsQueue
```
rtkaczyk: Why not change all of them?

rtkaczyk: Also I would suggest being explicit and using `List` instead of `Seq`

mcsherrylabs: Why would you use a more specific type?

rtkaczyk: Well, this code bit is fully isolated, there is no interface, there are no callers. So we can explicitly choose and enforce the implementation that has the most fitting performance characteristics. 

And we shouldn't rely on the fact that a default `Seq` is a `List` because it may change in the future Scala releases. It used to be a `Vector`.

Now, it does make some theoretical sense to use a supertype in a parameter position like in line 390. Probably less so in practice though, personally I would also change it to `List` (if we wanted to be really permissive we should probably use something like `Traversable`).

adamsmo: > Why not change all of them?

order is not required in case of storage root and evm code as evm code does not expand to new hashes and mpt root expands to branches that are later added to `Seq`

rtkaczyk: > order is not required

It's not about whether we need order or not. Prepending to a `List` is much faster than adding to a `Set`, especially when we're talking about sizeable `ByteStrings`!

adamsmo: yes it is true, that is why it was Seq when it was first implemented (all queues)

rtkaczyk: ok, so why the Sets then?

adamsmo: it was changed to `Set`, now I am changing it back
so I am not sure why it is Set 😉

rtkaczyk: Ah, I thought you were disagreeing 😉

mcsherrylabs: It was changed to `Seq` in a review I think. 
Constructing List internally (toList?) is fine, but lets keep the types flexible.

adamsmo:
yes we can, I think it will perform better because there will be no hashcode calculation
in fact I did change them

PR #89, opened 2017-03-13
Pad small keys in IODB

ntallar:
## Description

Includes allowing the use of smaller keys to access into IODB, by padding them, which enables the use of IODB instead of LevelDB. Also we begin to use IODB version `0.3.0`.


src/main/scala/io/iohk/ethereum/db/dataSource/IodbDataSource.scala
```diff
@@ -44,9 +47,8 @@ class IodbDataSource private (lSMStore: LSMStore, keySize: Int, path: String) ex
     }
   }
 
-  private def asStorables(namespace: Namespace,
-                          keyValues: Seq[(Key, Value)]): Seq[(ByteArrayWrapper, ByteArrayWrapper)] =
-    keyValues.map(kv => ByteArrayWrapper((namespace ++ kv._1).toArray) -> ByteArrayWrapper(kv._2.toArray))
+  private def asStorables(keyValues: Seq[(Key, Value)]): Seq[(ByteArrayWrapper, ByteArrayWrapper)] =
+    keyValues.map(kv => ByteArrayWrapper(kv._1.toArray) -> ByteArrayWrapper(kv._2.toArray))
```
adamsmo: maybe
`keyValues.map{case (key, val) => ByteArrayWrapper(key.toArray) -> ByteArrayWrapper(val.toArray)}` ?

PR #90, opened 2017-03-14
EC-130 Resume regular download if fast sync is done

adamsmo:
Add check to prevent fast sync to be executed more then once
This PR should be merged after https://github.com/input-output-hk/etc-client/pull/88 which is based on.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
@@ -36,8 +36,6 @@ class SyncController(
       case _ => Stop
     }
 
-  import context.dispatcher
-
   var handshakedPeers: Map[ActorRef, PeerStatus.Handshaked] = Map.empty
 
   scheduler.schedule(0.seconds, peersScanInterval, peerManager, PeerManagerActor.GetPeers)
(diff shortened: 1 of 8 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
```diff
@@ -47,7 +45,19 @@ class SyncController(
   override def receive: Receive = idle
 
   def idle: Receive = handlePeerUpdates orElse {
-    case StartFastSync =>
+
+    case StartSync =>
+      (appStateStorage.isFastSyncDone(), doFastSync) match {
+        case (false, true) =>
+          self ! StartFastSync
```
ntallar: Should we clear the database before starting fast sync? I got a strange behavior while first setting `doFastSync=false` and then setting it to `true` so that first regular sync is done and then fast sync. This results in:
```
11:15:15.691 i.i.e.blockchain.sync.SyncController - Block: 18816/3359983. Peers: 2/2 (0 blacklisted). State: 17/17 known nodes.
11:15:16.147 i.i.e.b.s.FastSyncReceiptsRequestHandler - Received 64 receipts in 635 ms
11:15:16.677 i.i.e.b.s.FastSyncNodesRequestHandler - Received 256 state nodes in 2452 ms
11:15:17.175 i.i.e.b.s.FastSyncBlockBodiesRequestHandler - Received 128 block bodies in 1014 ms
11:15:17.700 i.i.e.blockchain.sync.SyncController - Block: 192/3359983. Peers: 2/2 (0 blacklisted). State: 273/4369 known nodes.
```
It seems that the blocks obtained from regular sync still remain in storage and will not be used. I don't think this should be an impediment to merging if its too difficult to fix

ntallar:
Apart from that comment LGTM!

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
adamsmo: actually behavior seems to be correct - there was no fast sync state to resume from so when fast sync started it started downloading blocks from genesis
but we can add check if there is any saved state to resume (`fastSyncStateStorage.getSyncState().isEmpty`) and if there is no state then wipe db before starting fast sync
@mcsherrylabs what do you think?

whysoserious: We need to sort that out. However, I'd merge this PR and and work on Nicolas' issue in a separate branch.

mcsherrylabs: Is there any benefit to wiping the db?

adamsmo: @ntallar I changed it so it is no longer possible to start fast sync after regular sync
https://github.com/input-output-hk/etc-client/pull/90/commits/d58dc856cbc103234c4a30980d09b996404c5290

PR #92, opened 2017-03-14
Fix target block on fast sync restart

LukasGasior1:


src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
```diff
@@ -267,27 +273,30 @@ class SyncController(
...
       syncStatePersistCancellable.cancel()
+      syncStateStorageActor ! PoisonPill
+
       fastSyncStateStorage.purge()
-      log.info("Fast sync finished")
       appStateStorage.fastSyncDone()
-      context.parent ! FastSyncDone
+
+      log.info("Fast sync finished")
+      context.parent ! FastSyncDone // TODO: is this still needed (seems to be used for BlockBroadcastActor)
```
AlanVerbner: I guess that actor is going to be removed, isn't it?

ntallar: Yes, this was removed in https://github.com/input-output-hk/etc-client/pull/91 as they are no longer used

rtkaczyk:
Could you provide some description? I don't fully understand the problem

LukasGasior1:
Sure @rtkaczyk. When resuming fast sync after app restart we didn't use the same target block as before, instead we were choosing another one (which lead to invalid mpt state download).

PR #93, opened 2017-03-15
[EC-134] Investigate if blocks.head in RegularSyncController is 'safe'.

adamsmo:
It is not safe because we can get empty response from peer


adamsmo:
this fix is already in block sync

PR #94, opened 2017-03-15
[EC-135] Fix best peer in regular sync

LukasGasior1:


src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSyncController.scala
```diff
@@ -180,9 +180,12 @@ trait RegularSyncController {
...
 
-  private def bestPeer: Option[ActorRef] = Try(peersToDownloadFrom.maxBy {
-    case (_, Handshaked(_, Chain.ETC, totalDifficulty)) => totalDifficulty
-  }._1).toOption
+  private def bestPeer: Option[ActorRef] = Try {
+    peersToDownloadFrom
+      .collect { case (ref, Handshaked(_, Chain.ETC, totalDifficulty)) => (ref, totalDifficulty) }
+      .maxBy { case (_, td) => td }
+      ._1
+  }.toOption
```
rtkaczyk: What can throw an exception here?

LukasGasior1: `maxBy` if called on empty list

rtkaczyk: Yeah, I would try to avoid it by either doing a simple `if` or:

```
Option(peersToDownloadFrom.collect(...).filter(_.nonEmpty).map(_.maxBy(...)._1)
```

but it's your call.

LukasGasior1:
@rtkaczyk updated

PR #97, opened 2017-03-16
[EC-86] Add evm tests

LukasGasior1:


build.sbt
```diff
@@ -40,14 +42,43 @@ val dep = {
   )
 }
 
+lazy val solidityCompileTask = TaskKey[Unit]("solidityCompile", "Compiles solidity contracts")
+
+solidityCompileTask := {
```
rtkaczyk: Nice :+1:, though at some point I think this could be moved to the evm-runner itself

LukasGasior1: could be, but it relies on `solc` installed in system, for me it kind of makes sense to have it on sbt level

src/evmTest/resources/solidity/MinimumViableToken.sol
```diff
@@ -0,0 +1,19 @@
...
+    function MinimumViableToken(uint256 initialSupply) {
+        balanceOf[msg.sender] = initialSupply;              // Give the creator all initial tokens
+    }
+
+    /* Send coins */
+    function transfer(address _to, uint256 _value) {
+        if (balanceOf[msg.sender] < _value) throw;           // Check if the sender has enough
+        if (balanceOf[_to] + _value < balanceOf[_to]) throw; // Check for overflows
+        balanceOf[msg.sender] -= _value;                     // Subtract from the sender
+        balanceOf[_to] += _value;                            // Add the same to the recipient
```
rtkaczyk: I find this function interesting for multiple reasons:
* I wonder what the balance updates compile to, because a single line `balanceOf[_to] += _value; ` would suggest that you could possibly add value to the recipient's balance without deducting from sender's balance, which should not be allowed by EVM
* the check for sufficient funds should be redundant. The resulting message call in the EVM would simply return 0 indicating failure, but Solidity compiler should add a check for that resulting in an `InvalidOpCode` or `InvalidJump`.
* the check for overflow is interesting, because EVM doesn't have mechanism for overflow detection (although [it's been proposed](https://github.com/ethereum/EIPs/issues/159)). Is a standard practice to add such checks in Solidity contracts?

LukasGasior1: @rtkaczyk This contract is straight from ethereum webpage: https://www.ethereum.org/token#minimum-viable-token ;)

LukasGasior1: Regarding your first point
> I wonder what the balance updates compile to, because a single line balanceOf[_to] += _value; would suggest that you could possibly add value to the recipient's balance without deducting from sender's balance, which should not be allowed by EVM

balanceOf is just a map, you can update it however you want, can't you?

src/evmTest/resources/solidity/MutualRecursion.sol
```diff
@@ -0,0 +1,18 @@
...
+            return true;
+        else
+            return isOdd(n - 1);
+    }
+
+    function isOdd(uint n) returns (bool) {
+        if (n == 0)
+            return false;
+        else
+            return isEven(n - 1);
```
rtkaczyk: I have no reason to doubt it would work. Mutual recursion within a single contract will simply be a series of jumps, and as for testing for StackOverflow standard recursion would suffice.

I believe that a more interesting test would mutual recursion between 2 different contracts. One of the aspects would be checking for call-depth overflow (though possibly that could be checked with a single contract calling itself)

But I'm just making notes here. This contract is simple enough so no need to change it, and we can add the test with 2 different contracts later.

src/evmTest/resources/solidity/MinimumViableToken.sol
rtkaczyk: LOL, sorry I didn't look at what `balanceOf` is. I thought you were actually transferring Ether 😄

src/evmTest/scala/io/iohk/ethereum/vm/ThrowSpec.scala
```diff
@@ -0,0 +1,20 @@
...
+class ThrowSpec extends FreeSpec with Matchers {
+
+  "EVM running Throw contract" - {
+
+    "should handle throwing" in new EvmTestEnv {
+      val (_, contract) = deployContract("Throw")
+
+      val result = contract.justThrow().call()
+
+      result.error shouldBe Some(InvalidOpCode) // TODO: change to invalid jump destination
```
rtkaczyk: I thought #75 was merged already. We need to do that before merging this one.

But this also raises a question: once this gets merged should we enable EVM tests on Circle CI? pinging @mcsherrylabs

src/test/scala/io/iohk/ethereum/vm/StackSpec.scala
```diff
@@ -9,8 +9,8 @@ class StackSpec extends FunSuite with Matchers with PropertyChecks {
   val maxStackSize = 32
   val stackGen = Generators.getStackGen(maxSize = maxStackSize)
   val intGen = Gen.choose(0, maxStackSize).filter(_ >= 0)
-  val dataWordGen = Generators.getUInt256Gen()
-  val dataWordListGen = Generators.getListGen(0, 16, dataWordGen)
+  val uint256Gen = Generators.getUInt256Gen()
+  val uint256ListGen = Generators.getListGen(0, 16, uint256Gen)
```
rtkaczyk: Thank you 😃

src/evmTest/scala/io/iohk/ethereum/vm/ThrowSpec.scala
LukasGasior1: Yes as soon as this PR is merged I'll update Circle to run these tests

project/SolidityPlugin.scala
```diff
@@ -0,0 +1,41 @@
+import java.io.PrintWriter
+
+import sbt.TaskKey
+import sbt._
+import Keys._
+
+object SolidityPlugin extends AutoPlugin {
```
rtkaczyk: Nice. Apart from code separation are other other advantages to this approach?

LukasGasior1: Not really at the moment. This plugin can possibly be reused in the future

src/evmTest/scala/io/iohk/ethereum/vm/utils/EvmTestEnv.scala
```diff
@@ -0,0 +1,142 @@
...
+    case b: BigInt => UInt256(b).bytes
+    case a: Array[Byte] => UInt256(a).bytes
+    case i: Int => UInt256(i).bytes
+    case a: Address => UInt256(a.bytes).bytes
+    case other => throw new RuntimeException("Invalid call argument")
+  }
+
+  class Contract(val name: String, val address: Address) extends Dynamic {
+
+    def applyDynamic(methodName: String)(args: Any*): ContractMethodCall = {
```
rtkaczyk: ❤️

src/evmTest/scala/io/iohk/ethereum/vm/utils/EvmTestEnv.scala
```diff
@@ -0,0 +1,142 @@
...
+      callMethod(methodName)(args: _*)
+    }
+
+    def callMethod(methodName: String)(args: Any*): ContractMethodCall = {
+      if (methodName.contains("("))
+        callWithSignature(methodName)(args: _*)
```
rtkaczyk: Is this calling mode used/useful?

LukasGasior1: Yes, by default it finds ABI by parameter count, however if there are two overloaded methods with same parameter count one need to specify full signature to disambiguate. see: https://github.com/input-output-hk/etc-client/pull/97/files/8083de05eedc45c18a3e58b27328797d597b034b#diff-1c9763f46649c5ebc0ad7b9444ef8aa4R111

rtkaczyk:
@LukasGasior1 please update, the dependency PR has been, so this should be good now.

LukasGasior1:
@rtkaczyk updated

src/evmTest/scala/io/iohk/ethereum/vm/ThrowSpec.scala
```diff
@@ -13,7 +13,7 @@ class ThrowSpec extends FreeSpec with Matchers {
 
       val result = contract.justThrow().call()
 
-      result.error shouldBe Some(InvalidOpCode) // TODO: change to invalid jump destination
+      result.error shouldBe Some(InvalidOpCode(0xfd.toByte))
```
rtkaczyk: Wait, IIRC we've been over this and an explicit `throw` should result in an invalid jump (as the previous TODO comment suggests) ?

src/evmTest/scala/io/iohk/ethereum/vm/MinimumViableTokenSpec.scala
```diff
@@ -31,7 +31,7 @@ class MinimumViableTokenSpec extends FreeSpec with Matchers {
       val (_, contract) = deployContract("MinimumViableToken", creatorAddress = sender, constructorArgs = Seq(100))
 
       val transferRes = contract.transfer(receiver, 200).call(sender = sender)
-      transferRes.error shouldBe Some(InvalidOpCode) // TODO: change to invalid jump destination
+      transferRes.error shouldBe Some(InvalidOpCode(0xfd.toByte))
```
rtkaczyk: `InvalidOpCode(INVALID.code)`

src/evmTest/scala/io/iohk/ethereum/vm/ThrowSpec.scala
rtkaczyk: Ah, I forget those tests are not run on Circle :wink:. I'm pretty sure you need to change this.

LukasGasior1: Not really in solidity,
> Internally, Solidity performs a revert operation (instruction 0xfd) when a user-provided exception is thrown or the condition of a require call is not met. In contrast, it performs an invalid operation (instruction 0xfe) if a runtime exception is encountered or the condition of an assert call is not met.

src/evmTest/scala/io/iohk/ethereum/vm/MinimumViableTokenSpec.scala
rtkaczyk: Hmm, no. This is a result of an explicit throw, right? So that should also be a bad jump. Implicit error-checking implemented by Solidity (such as passing value to non-payable function) would result in INVALID opcode.

LukasGasior1: well that's what solidity generates, and seems we don't have a constant for 0xfd (revert op?)

src/evmTest/scala/io/iohk/ethereum/vm/ThrowSpec.scala
rtkaczyk: Well, OK, that makes sense, maybe expect for the part that there is no revert instruction (0xfd) in the latest YP. 

But we've been over this? Don't you remember? We had a `StackOverflow` because jump destination were not implemented.

So what changed? The solidity version is the same as before, right?

rtkaczyk: Ekhm, did you run those tests?

```
[info] EVM running MinimumViableToken contract
[info] - should init the balance and allow token transfer (586 milliseconds)
[info] - should return an error when attempted to transfer more tokens than owned *** FAILED *** (86 milliseconds)
[info]   Some(InvalidJump(0x0000)) was not equal to Some(InvalidOpCode(0xfd)) (MinimumViableTokenSpec.scala:34)
```

```
[info] EVM running Throw contract
[info] - should handle throwing *** FAILED *** (45 milliseconds)
[info]   Some(InvalidJump(0x0000)) was not equal to Some(InvalidOpCode(0xfd)) (ThrowSpec.scala:16)
```

LukasGasior1: Yes I remember checking that in doc, and I'm pretty sure it was InvalidJump (SO at the time) for user throws and something else for "system" throws. Maybe it was always that way and the doc was outdated?

LukasGasior1: wait, no. It doesn't make sense :D I'm pretty sure we were getting StackOverflow so now we should get InvalidJump for sure!

LukasGasior1: "Ekhm, did you run those tests?" what the heck. it works for me

LukasGasior1: @rtkaczyk `solc --version` please. Maybe it's not properly using the version from `pragma solidity ^0.4.9;`

rtkaczyk: Yes, maybe what you quoted above is the outdated doc? Anyway, just run the tests and see for yourself 😁

LukasGasior1: well I have just run them and getting the opposite result:
`Some(InvalidOpCode(0xfd)) was not equal to Some(InvalidJump) (MinimumViableTokenSpec.scala:34)`

rtkaczyk:
```
solc --version
solc, the solidity compiler commandline interface
Version: 0.4.9+commit.364da425.Darwin.appleclang
```

rtkaczyk: WTF?

LukasGasior1: solc, the solidity compiler commandline interface
Version: 0.4.10+commit.59099908.Linux.g++

I think I know what's going on...`^` in `pragma solidity ^0.4.9;` probably means "use at least this version or newer"

LukasGasior1: (btw mine seems newer :P)

rtkaczyk: Well, yeah, but `pragma solidity ^0.4.9;`....
Anyway, let me try upgrading.

rtkaczyk: Ahh... `^`, ok

LukasGasior1: I'll update it to 0.4.10 (as it seems to be the recent version) and remove `^` so that we don't get unexpected test failures

LukasGasior1: funny thing is, solidity 0.4.10 is unable to compile as if it was 0.4.9. After removing `^`:
> Source file requires different compiler version (current compiler is 0.4.10+commit.59099908.Linux.g++ - note that nightly builds are considered to be strictly less than the released version
pragma solidity 0.4.9;

rtkaczyk: Do you have a pre-release version? 0.4.10 was only relasesed on March 15th.

LukasGasior1: honestly I don't remember. It may as well be that I just compiled the master some time ago

rtkaczyk:
There's a minor nuisance:

```
> evm:test
Refusing to overwrite existing file "/Users/radek/scalac/iohk/etc-client/target/contracts/Callee.bin" (use --overwrite to force).
Refusing to overwrite existing file "/Users/radek/scalac/iohk/etc-client/target/contracts/Caller.bin" (use --overwrite to force).
```

We cannot easily re-run these tests

src/evmTest/scala/io/iohk/ethereum/vm/ThrowSpec.scala
rtkaczyk: I just installed a new version and I have the same thing. It seems we have 2 options:

* require a specific solc version - fine for CI, not so much for developers
* bit a bullet and deal with those test becoming broken - hopefully that's not going to happen very often

LukasGasior1: let's stick with minimum version then, I'll set it to `^0.4.10`

rtkaczyk: Just for the record: this sucks

LukasGasior1:
@rtkaczyk updated. `sbt evm:test` should now work the second time

src/main/scala/io/iohk/ethereum/vmrunner/Utils.scala
```diff
@@ -0,0 +1,26 @@
...
+
+object Utils {
+
+  def loadContractCodeFromFile(file: File): ByteString = {
+    val raw = Source.fromFile(file).mkString
+    ByteString(raw.trim.grouped(2).map(Integer.parseInt(_, 16).toByte).toArray)
+  }
+
+  def loadContractAbiFromFile(file: File): Either[Error, List[ABI]] = {
+    val raw = Source.fromFile(file).mkString
```
ntallar: Minor detail: Should we close this `Source` and the one from the `loadContractCodeFromFile` function?

src/evmTest/scala/io/iohk/ethereum/vm/utils/EvmTestEnv.scala
```diff
@@ -0,0 +1,142 @@
...
+
+    val payload = constructorArgs.map(parseArg).foldLeft(contractInitCode)(_ ++ _)
+
+    val tx = MockVmInput.transaction(creatorAddress, payload, value, gasLimit, gasPrice)
+    val bh = MockVmInput.blockHeader
+
+    val context = ProgramContext[MockWorldState, MockStorage](tx, bh, world)
+    val result = VM.run(context)
+
+    contractsAbis += (name -> contractAbi.right.get)
```
ntallar: Minor detail: I now that this doesn't fail in our current tests but should we handle the case when this could fail in case `loadContractAbiFromFile` where to return an error?

src/evmTest/scala/io/iohk/ethereum/vm/ContractCallingItselfSpec.scala
```diff
@@ -0,0 +1,23 @@
...
+class ContractCallingItselfSpec extends FreeSpec with Matchers {
+
+  "EVM running ContractCallingItself contract" - {
+
+    "should handle a call to itself" in new EvmTestEnv {
+      val (_, contract) = deployContract("ContractCallingItself")
+
+      contract.getSomeVar().call().returnData shouldBe UInt256(10).bytes
+
+      val result = contract.doubleSomeVar().call()
```
ntallar: Minor detail: As the contract is called `ContractCallingItself` shouldn't this test call here `callSelf` to test it?

src/evmTest/scala/io/iohk/ethereum/vm/utils/EvmTestEnv.scala
LukasGasior1: Ce could handle that but the only result we can produce is a failed test, exactly the same as with failed `.get`

src/evmTest/scala/io/iohk/ethereum/vm/ContractCallingItselfSpec.scala
LukasGasior1: Nice, good catch :+1:

LukasGasior1:
@ntallar I've addressed your comments

ntallar:
@LukasGasior1 LGTM with the latest changes!

PR #99, opened 2017-03-16
[EC-131] Update log messages

whysoserious:
This PR updates log messages:
- Reason of blacklisting a fastsync peer
- Warning if a target block for a fast sync is very old
- Warning if fastsync has stalled

configures `RollingFileAppender`

and adds some formatting changes in `application.conf`.


src/universal/conf/application.conf
```diff
@@ -18,7 +18,14 @@ etc-client {
...
+        "enode://18a551bee469c2e02de660ab01dede06503c986f6b8520cb5a65ad122df88b17b285e3fef09a40a0d44f99e014f8616cf1ebc2e094f96c6e09e2f390f5d34857@47.90.36.129:30303",
```
rtkaczyk: I think that as a (near) future improvement, we could have the `dist` task copy `reference.conf` instead of having to maintain both these files.

mcsherrylabs: 'reference.conf' should end up in the '.jar' and thus in the lib folder and so only production overrides should need to go in universal 'application.conf'. That's the theory...

whysoserious: Yes, I wanted to raise this issue soon as well.

rtkaczyk: > only production overrides should need to go in universal 'application.conf'

I would still like to see an `application.conf` in the distribution just as a documentation. Ideally all config settings should be explained with defaults shown. I think [something like this](https://github.com/antirez/redis/blob/unstable/redis.conf) became the industry standard.

We could still have that done automatically by prepending `#` to each setting line. I am of course thinking about a post-MVR improvement.

mcsherrylabs: Makes sense. 
Jan, can we confirm it works as we think when using the 'dist' version and then delete  all out of `universal/application.conf`?

whysoserious: Could we do that in a separate PR?

mcsherrylabs: Yep.

PR #100, opened 2017-03-16
Refactor sync controller into separate traits

LukasGasior1:


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -0,0 +1,350 @@
...
+    }
+  }
+
+  private def startingFastSync: Receive = handlePeerUpdates orElse {
+    case RetryStart => startFastSync()
+  }
+
+  private def startFastSync(syncState: SyncState): Unit = {
+    scheduler.schedule(0.seconds, printStatusInterval, self, PrintStatus)
+    context become new SyncingHandler(syncState).receive
```
adamsmo: should context be combined with
`handlePeerUpdates orElse` ?

adamsmo:
other than question LGTM

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -0,0 +1,350 @@
...
+      peersUsedToChooseTarget.foreach { case (peer, Handshaked(status, _, _)) =>
+        peer ! PeerActor.Subscribe(Set(BlockHeaders.code))
+        peer ! PeerActor.SendMessage(GetBlockHeaders(Right(status.bestHash), 1, 0, reverse = false))
+      }
+      log.info("Asking {} peers for block headers", peersUsedToChooseTarget.size)
+      val timeout = scheduler.scheduleOnce(peerResponseTimeout, self, BlockHeadersTimeout)
+      context become waitingForBlockHeaders(peersUsedToChooseTarget.keySet, Map.empty, timeout)
+    } else {
+      log.warning("Cannot start fast sync, not enough peers to download from. Scheduling retry in {}", startRetryInterval)
+      scheduleStartRetry(startRetryInterval)
```
ntallar: Should we do here `context become startingFastSync`? I'm not sure whether we can handle the `RetryStart` message if that's not the case, I wasn't able to start our node without this line.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -23,11 +21,13 @@ trait RegularSyncController {
 
   import Config.FastSync._
 
-  def regularSync(): Receive = {
-
-    case StartSyncing =>
-      askForHeaders()
+  def startRegularSync(): Unit = {
```
ntallar: Should we do here `appStateStorage.fastSyncDone()` as we previously did when handling the `StartRegularSync` message? If that's not the case we are able to run first the regular sync and then the fast sync, which I think we agreed that we shouldn't allow in #90.

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
LukasGasior1: It's already in `SyncingHandlerer.receive`

LukasGasior1:
Thanks @ntallar for pointing that out. Fixed now.

adamsmo:
I tested this branch and it looks like fast sync does not work when running sbt run with empty db

PR #104, opened 2017-03-20
[EC-115] Validate transaction signature

ntallar:
## Description

Includes validations in `SignedTransactionValidator` from Appendix F in the Yellow Paper to check that the transaction signature is valid before executing it.


src/main/scala/io/iohk/ethereum/network/p2p/validators/SignedTransactionValidator.scala
```diff
@@ -0,0 +1,77 @@
...
+    *                             before the [[Config.Blockchain.HomesteadBlock]]
+    * @return Either the validated transaction or TransactionSignatureError if an error was detected
+    */
+  private def validateSignature(stx: SignedTransaction, fromBeforeHomestead: Boolean): Either[SignedTransactionError, SignedTransaction] = {
+    val r: BigInt = BigInt(new BigInteger(1, stx.signatureRandom.toArray[Byte]))
+    val s: BigInt = BigInt(new BigInteger(1, stx.signature.toArray[Byte]))
+
+    val validR = r > 0 && r < secp256k1n
+    val validS = s > 0 && s < (if(fromBeforeHomestead) secp256k1n else secp256k1n / 2)
+    val validV = stx.recoveredPointSign.isDefined
```
rtkaczyk: Since we do validation here, does it still make sense to keep `recoveredPointSign` as an `Option` (as well as the dependent fields) on the `SignedTransaction`? Pinging @adamsmo

adamsmo: I we will use this pice in rap decoder then we can remove old version and produce only signed transactions with recovered address and public key

rtkaczyk: @adamsmo could you elaborate? Can we do this now?

adamsmo: I mean we should sender public key and address in transaction if we delete logic from transaction we should use this logic for example here
https://github.com/input-output-hk/etc-client/blob/phase/3/txExecution/src/main/scala/io/iohk/ethereum/network/p2p/messages/CommonMessages.scala#L59
to preform recovery of public key / address

rtkaczyk: > I mean we should sender public key and address in transaction if we delete logic from transaction we should use this logic for example here

Could you rephrase/correct that sentence please? It's hard to understand.

The code you pointed to simply decodes a SignedTransaction from RLP. What does it have to do with validating the signature or recovering the address?

The way I see is that we currently have the validation logic split into `SignedTransactionValidator` and `SignedTransaction` itself. I would prefer to have it one place.

adamsmo: > I mean we should have sender's public key and address in transaction if we delete logic from transaction we should use this logic for example here
that link was example where we can recover public key and put it to transaction so we know that it is valid

rtkaczyk: Yeah, I actually raised this issue a couple of weeks ago. We only need to deal with invalid transaction/block signatures right upon downloading them.

ntallar: @adamsmo Then you are suggesting that that the `SignedTransaction` should have something like a sender field and that we should obtain it during the decoding of them?

adamsmo: if the logic is separated, than yes

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -23,98 +23,87 @@ object SignedTransaction {
...
   }
+}
+
+case class SignedTransaction (
+  tx: Transaction,
+  pointSign: Byte, //v
+  signatureRandom: ByteString, //r
+  signature: ByteString, //s
+  senderAddress: Address,
+  recoveredPointSign: Int) {
```
adamsmo: where do we use `recoveredPointSign`?
Is it necessary if we already recovered public key and address?

ntallar: You are right :+1: Removed!

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -23,98 +23,94 @@ object SignedTransaction {
...
-    recoveredSenderAddress.isDefined
+  def apply(tx: Transaction, pointSign: Byte, signatureRandom: BigInt, signature: BigInt): Option[SignedTransaction] = {
+    for {
+      recoveredPointSign <- SignedTransaction.getRecoveredPointSign(pointSign)
+      sender <- SignedTransaction.getSender(tx, pointSign, signatureRandom, signature, recoveredPointSign)
+    } yield SignedTransaction(tx, pointSign, signatureRandom, signature, sender)
   }
+}
+
+case class SignedTransaction protected (
```
rtkaczyk: `protected` will not work with case class the way you expect to, because there's a factory method with this signature which is always public - you can only truly hide constructors with regular classes. But I think it's fine to have this one public, for tests if nothing else.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -137,34 +138,23 @@ object Ledger extends Logger {
     accumGasLimit: BigInt,
     block: Block): Either[String, SignedTransaction] = {
     for {
-      _ <- checkSyntacticValidity(stx)
-      _ <- validateSignature(stx)
+      _ <- SignedTransactionValidator.validateTransaction(stx, fromBeforeHomestead = block.header.number < Config.Blockchain.HomesteadBlock)
+        .left.map(_.toString)
```
rtkaczyk: @LukasGasior1 does your work in #128 address such cases in the `Ledger`?

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -23,98 +23,94 @@ object SignedTransaction {
...
+      key <- recoveredPublicKey
+      addrBytes = crypto.kec256(key).slice(FirstByteOfAddress, LastByteOfAddress)
+      if addrBytes.length == Address.Length
+    } yield Address(addrBytes)
+  }
 
-    import tx._
-    import Transaction._
+  def apply(tx: Transaction, pointSign: Byte, signatureRandom: ByteString, signature: ByteString): Option[SignedTransaction] =
+    SignedTransaction(tx, pointSign, BigInt(1, signatureRandom.toArray), BigInt(1, signature.toArray))
```
rtkaczyk: Now that we have this factory method, perhaps it would make more sense if `SignedTransaction` had `signature: ECDSASignature` instead of the individual components?

src/main/scala/io/iohk/ethereum/network/p2p/validators/SignedTransactionValidator.scala
```diff
@@ -0,0 +1,80 @@
...
+
+    def byteLength(b: BigInt): Int = b.toByteArray.length
+
+    if (byteLength(nonce) > NonceLength)
+      Left(TransactionSyntaxError(s"Invalid nonce length: ${byteLength(nonce)}"))
+    else if(byteLength(gasLimit) > GasLength)
+      Left(TransactionSyntaxError(s"Invalid gasLimit length: ${byteLength(gasLimit)}"))
+    else if(byteLength(gasPrice) > GasLength)
+      Left(TransactionSyntaxError(s"Invalid gasPrice length: ${byteLength(gasPrice)}"))
+    else if(byteLength(value) > ValueLength)
```
rtkaczyk: Basically all these values are 256-bit integers. So we still, at least theoretically, have the problem of `BigInt` requiring extra bit (byte) for sign. Let's get rid off `byteLength` and do simple numeric comparison against `val maxValue = BigInt(2).pow(256) - 1`

src/main/scala/io/iohk/ethereum/network/p2p/validators/SignedTransactionValidator.scala
```diff
@@ -0,0 +1,80 @@
+package io.iohk.ethereum.network.p2p.validators
```
rtkaczyk: Is this the right place to put it?

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
ntallar: You are right, I would be more clear :+1:

src/main/scala/io/iohk/ethereum/network/p2p/validators/SignedTransactionValidator.scala
ntallar: You are right, I'll create a separate package for the `validators`, I don't know why they were all there :stuck_out_tongue_winking_eye:

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
LukasGasior1: Not at the moment, but we can use similar mechanism here

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -23,98 +23,93 @@ object SignedTransaction {
...
+
+  def apply(tx: Transaction, signature: ECDSASignature): Option[SignedTransaction] = {
+    for {
+      recoveredPointSign <- SignedTransaction.getRecoveredPointSign(signature.v)
+      sender <- SignedTransaction.getSender(tx, signature, recoveredPointSign)
+    } yield SignedTransaction(tx, signature, sender)
+  }
+}
+
+case class SignedTransaction protected (
```
rtkaczyk: `protected` will not give you the protection you expect 😉

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -23,98 +23,93 @@ object SignedTransaction {
...
-    signatureRandom.length <= ECDSASignature.RLength &&
-    signature.length <= ECDSASignature.SLength &&
-    recoveredSenderAddress.isDefined
+  def apply(tx: Transaction, pointSign: Byte, signatureRandom: ByteString, signature: ByteString): Option[SignedTransaction] = {
+    val txSignature = ECDSASignature(r = new BigInteger(1, signatureRandom.toArray), s = new BigInteger(1, signature.toArray), v = pointSign)
+    SignedTransaction(tx, txSignature)
   }
 
+  def apply(tx: Transaction, pointSign: Int, signatureRandom: ByteString, signature: ByteString): Option[SignedTransaction] = {
+    if(pointSign.toByte != pointSign)
```
adamsmo: in some cases equations `v = CHAIN_ID * 2 + 35 or v = CHAIN_ID * 2 + 36` may produce numer out of range of byte, implementation in parity just cast it to byte which result in overflow and negative value like -98, so in this case we are assuming that integer that we are getting is already truncated to byte?

rtkaczyk: Good catch, I would just skip this condition. If pointSign is invalid, you will not be able to recover public key, and this will result in `None` anyway.

PR #98, opened 2017-03-16
Rename application.conf to reference.conf

whysoserious:

PR #105, opened 2017-03-20
[EC-140] Avoid duplication in configuration files

AlanVerbner:
## Description

In order to avoid duplicating our config in universal config and src/main/resources/application.conf...

1. Rename src/main/resources/application.conf  to src/main/resources/resources.conf 
2. Remove all config from the ./src/universal/conf/application.conf
3. Put in the line include "resources.conf"  in ./src/universal/conf/application.conf
4. Put in an example override of the kind etc-client.network.server-address.port - 90

### Notes
Please test overwriting a config, performing `sbt dist` and running the app


src/main/resources/application.conf
```diff
@@ -1,104 +1 @@
...
-
-}
-
-akka {
-  loggers = ["akka.event.slf4j.Slf4jLogger"]
-  loglevel = "DEBUG"
-  logging-filter = "akka.event.slf4j.Slf4jLoggingFilter"
-  logger-startup-timeout = 30s
-}
+include "resources.conf"
```
mcsherrylabs: Presuming this was necessary to run it from sbt?

AlanVerbner: @mcsherrylabs yes...without it I cannot `sbt run`

mcsherrylabs: Actually I think the correct name for that file might be 'reference.conf'... not resources.conf.

AlanVerbner:
@rtkaczyk i've addressed your comment including #98 changes in this PR

PR #107, opened 2017-03-21
Add Rolling file appender

mcsherrylabs:
Use rolling file appender for both `sbt run` and `sbt dist` and make the `logback` in `sbt dist` the same as in `run` in order to effectively debug.

### DIFF
--- a/src/universal/conf/logback.xml
+++ b/src/universal/conf/logback.xml
@@ -1,42 +1,37 @@
 <configuration>
 
-    <property name="consoleEncoderPattern" value="%d{HH:mm:ss.SSS} - %msg%n" />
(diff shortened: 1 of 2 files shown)


src/universal/conf/logback.xml
```diff
@@ -1,42 +1,37 @@
 <configuration>
 
-    <property name="consoleEncoderPattern" value="%d{HH:mm:ss.SSS} - %msg%n" />
+    <property name="stdoutEncoderPattern" value="%d{HH:mm:ss.SSS} %logger{36} - %msg%n" />
```
rtkaczyk: As long as we're changing that, could we also add the log level?

PR #109, opened 2017-03-21
[EC-79] [EC-133] Add support for private blockchains

LukasGasior1:
Added:
- fork resolver abstraction in PeerActor instead of hard-coded dao fork check
- on startup, genesis data is loaded from json and saved to storages
- possibility to specify custom genesis file


src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -177,38 +181,36 @@ class PeerActor(
...
-        case Some(_) if daoForkValidator.validate(msg).isEmpty =>
-          log.info("Peer is running the ETC chain")
-          context become new MessageHandler(rlpxConnection, remoteStatus, daoForkBlockNumber, Chain.ETC).receive
+      val forkBlockHeaderOpt = blockHeaders.find(_.number == forkResolver.forkBlockNumber)
 
-        case Some(_) if getBestBlockHeader().difficulty < daoForkBlockTotalDifficulty =>
-          log.warning("Peer is not running the ETC fork, but we're not there yet. Keeping the connection until then.")
-          context become new MessageHandler(rlpxConnection, remoteStatus, daoForkBlockNumber, Chain.ETH).receive
+      forkBlockHeaderOpt match {
+        case Some(forkBlockHeader) if forkBlockHeader.number == forkResolver.forkBlockNumber =>
```
ntallar: Minor detail: How is the case when `forkBlockHeader.number != forkResolver.forkBlockNumber` is handled? Wouldn't it always be true?

LukasGasior1: nice catch. I wonder why I didn't get error because of unhandled match case. It should always be true, but you never know what does other peer send you, better make sure.

ntallar: But as `forkBlockHeaderOpt = blockHeaders.find(_.number == forkResolver.forkBlockNumber)`, wouldn't the `forkBlockHeader` always satisfy that condition?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -49,8 +49,6 @@ class PeerActor(
 
   val peerId: String = self.path.name
 
-  val daoForkValidator = ForkValidator(daoForkBlockNumber, daoForkBlockHash)
```
ntallar: Should the `ForkValidator` and `MessageValidator` classes be removed as they are no longer used?

src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -74,38 +73,13 @@ object Config {
   object Blockchain {
     private val blockchainConfig = config.getConfig("blockchain")
 
-    val genesisDifficulty: Long = blockchainConfig.getLong("genesis-difficulty")
-    val genesisHash = ByteString(Hex.decode(blockchainConfig.getString("genesis-hash")))
```
ntallar: Should the `genesis-difficulty` and `genesis-hash` be removed from `reference.conf` as they are no longer used?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
LukasGasior1: absolutely right, fixed

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
LukasGasior1: removing

src/main/scala/io/iohk/ethereum/utils/Config.scala
LukasGasior1: yes, removing

whysoserious:
Could you fix commit history?

ntallar:
Minor detail: While testing, I saw that by default (if the peer from the private chain has less than `target-block-offset` blocks) we never start fast sync from it as the target block number calculated is less than 0, should `targetBlock` be set to 0 if that's the case?

LukasGasior1:
@ntallar I think in such case we should disable fast sync

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
```diff
@@ -0,0 +1,142 @@
...
+    val genesisJson = customGenesisFileOpt match {
+      case Some(customGenesisFile) =>
+        log.debug(s"Trying to load custom genesis data from file: $customGenesisFile")
+        Try(Source.fromFile(customGenesisFile).getLines().mkString) match {
+          case Success(customGenesis) =>
+            log.info(s"Using custom genesis data from file: $customGenesisFile")
+            customGenesis
+          case Failure(ex) =>
+            log.error(s"Cannot load custom genesis data from file: $customGenesisFile", ex)
+            sys.exit(1)
```
rtkaczyk: I don't like `sys.exit`. That's like `throw` or `return` in the functional paradigm. Non-blocking remark though.

rtkaczyk:
How can we test this? Have you tried connecting to Morden or Ropsten? Also, can we connect to the non-classic chain with this change?

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
mcsherrylabs: Missed that altogether. Should be a rethrow rather sys.exit.

whysoserious: I missed that one too. Please remove.

LukasGasior1:
@rtkaczyk for testing locally:

_CustomGenesis.json_
```
{
    "nonce": "0x0000000000000042",     "timestamp": "0x0",
    "parentHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "extraData": "0x0",     "gasLimit": "0x8000000",     "difficulty": "0x400",
    "mixhash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "coinbase": "0x3333333333333333333333333333333333333333",     "alloc": {     }
}
```

```
geth-classic --rpc --rpcapi eth,web3,personal,admin --datadir /tmp/priv-data-dir --rpcport 8542 --port 30309 --nodiscover --identity "TestnetMainNode" --networkid 1999 init CustomGenesis.json
```

then same cmd with `console` at the end instead of `init CustomGenesis.json`

bingo, you're running geth with private blockchain. `miner.start()` in console to mine some blocks

Connect from etc-client:
conf changes:
- `custom-genesis-file = "/path/to/CustomGenesis.json"`
- `network.network-id = 1999` (match geth)
- `min-peers-to-choose-target-block = 1`
- and of course `bootstrap-nodes`
that's it.

LukasGasior1:
@rtkaczyk I did not test with any public test network yet

src/main/resources/blockchain/default-genesis.json
```diff
@@ -0,0 +1,8905 @@
...
+    "15aa530dc36958b4edb38eee6dd9e3c77d4c9145": {"balance": "2000000000000000000000"},
+    "15acb61568ec4af7ea2819386181b116a6c5ee70": {"balance": "31000000000000000000000"},
+    "15b96f30c23b8664e7490651066b00c4391fbf84": {"balance": "410650000000000000000"},
+    "15c7edb8118ee27b342285eb5926b47a855bc7a5": {"balance": "20000000000000000000"},
+    "15d99468507aa0413fb60dca2adc7f569cb36b54": {"balance": "2000000000000000000000"},
+    "15dbb48c98309764f99ced3692dcca35ee306bac": {"balance": "150000000000000000000000"},
+    "15dcafcc2bace7b55b54c01a1c514626bf61ebd8": {"balance": "9400000000000000000000"},
+    "15e3b584056b62c973cf5eb096f1733e54c15c91": {"balance": "936702000000000000000"},
+    "15ebd1c7cad2aff19275c657c4d808d010efa0f5": {"balance": "200550000000000000000"},
+    "15ee0fc63ebf1b1fc49d7bb38f8863823a2e17d2": {"balance": "191000000
```
rtkaczyk: Any idea why precompiled contracts' accounts (addresses `0x1` to `0x4`) are not included here? They are given some balance on most test networks

src/main/resources/blockchain/default-genesis.json
```diff
@@ -0,0 +1,8905 @@
+{
+  "difficulty": "0x0400000000",
+  "extraData": "0x11bbe8db4e347b4e8c937c1c8370e4b5ed33adb3db69cbdb7a38e1e50b1b82fa",
+  "gasLimit": "0x1388",
+  "nonce": "0x0000000000000042",
+  "ommersHash": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
+  "timestamp": "0x0",
+  "coinbase": "0x0000000000000000000000000000000000000000",
+  "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
```
rtkaczyk: Are all those zeroes required? Can't we simply write `"0"` ?

src/main/resources/blockchain/default-genesis.json
```diff
@@ -0,0 +1,8905 @@
...
+  "alloc": {
+    "000d836201318ec6899a67540690382780743280": {"balance": "200000000000000000000"},
```
rtkaczyk: A detail: 
geth would accept such input: 

```
"alloc": {
  "666": {"balance": "100000000000000"}
}
```

Our client will not. This might be annoying if we want to use predefined genesis JSONs

src/main/resources/blockchain/default-genesis.json
LukasGasior1: That's good point. I don't know. This genesis block is straight from geth classic

src/main/resources/blockchain/default-genesis.json
LukasGasior1: we need all the zeros. a single "0" is not the same as 64 zeros :) I could add a line that adds padding when this is too short but I'm not sure if it's worth doing so

rtkaczyk: Well, it is. I understand this is a 32-byte hash, so it is a number `[0, 2 ^ 256 - 1]`. So it makes sense to write 0, 1, 2...

Anyway I'm just cause I want to know how to configure, no changes required.

LukasGasior1: in block header mixHash is a series of bytes (ByteString) even though numeric representation of "00" and "0000" is the same, these are different ByteStrings (and thus produce different hash, etc)

src/main/resources/blockchain/default-genesis.json
LukasGasior1: I'll add padding to 40 for addresses

LukasGasior1: It should work now.

src/main/resources/blockchain/default-genesis.json
rtkaczyk: But this is not a message to be hashed, but the hash itself, right?

LukasGasior1: mixHash together with a nonce is a proof of work for this block. Both these values are used to calculate the `hash` of the block. A block hash of a block with mixHash "00" is different from the one with mixHash "0000" (even if all other fields are equal). According to doc it's a series of 256bits, not a number.

rtkaczyk: > series of 256bits

so it's a number `[0, 2 ^ 256 - 1]` :)
Logically a textual representation "0", is the same as bit representation of 256 zeroes.
Same as `(0x00000000: Int) == (0: Int)` - in both cases you get 32 zero bits.

LukasGasior1: it's a series of 256 bits, not a number :)
`00` and `0000` are two different series of bits although they represent the same number.
They can also represent in image, you can interpret them however you want, still these are just bits.

rtkaczyk: It's a hash, so it's a number. It is something you compute by applying arithmetic transformations. Yes, it is a series 256 bits, because it's a representation of this number. Same as an `Int` is a series of 32 bits, and it is a number.

> They can also represent in image, you can interpret them however you want, still these are just bits.

No, it's either a number or an image, you can _represent_ them however you want. And if you're representation requires all digits to be specified, then `00` is not different from `0000` - they are both illegal.

LukasGasior1: "It's a hash, so it's a number" - hash is an output of hash function, which is a fixed size byte array, not a number. 

"It is something you compute by applying arithmetic transformations" - are you sure `xor`, `and`, `not` are arithmetic operations? ;)

"Same as an Int is a series of 32 bits, and it is a number." - Int is not just a series of 32 bits, it's a series of 32 bits AND an information for JVM to treat these bits as number.

`"abcd"` is also a series of 32 bits, this time with associated information to treat them as characters

rtkaczyk: > are you sure xor, and, not are arithmetic operations? ;)

I should have said _algebraic_, but yes. You apply this algebra to numbers and you get a number.

> associated information to treat them as characters

And in our code we have information what to treat the mixHash as. And it's textual representation doesn't need to depend on that.

LukasGasior1: Our implementation treats is exactly as yellowpaper suggests, 256 bits, nothing else.

LukasGasior1: "I should have said algebraic" algebra is an abstraction, you can as well have an algebra that operates on birds and dogs and an algebra that operates on numbers.

rtkaczyk: > Our implementation treats is exactly as yellowpaper suggests, 256 bits, nothing else.

And that is fine, but its true nature is a number. In the implementation you simply ignore that fact because it happens to be convenient.

> you can as well have an algebra that operates on birds and dogs

Yes, but we're talking here about a well defined algebra, which happens to operate on numbers.

LukasGasior1: I have already forgotten what the initial point was :) Let's just say, in order to get correct block `hash` you need this be true `blockHeader.mixHash.length == 32` no matter when the bits underneath are and what they represent.

rtkaczyk: The original point was that I wasn't sure what values are accepted in the JSON, and I was trying to convince why simply putting `0` makes sense to me. 

This should be documented somewhere, but that's a task for the future (post-MVR).

LukasGasior1: Right, since we need exactly 32 bytes a solution for that could be to pad the input when parsing.

rtkaczyk: And I just checked: `"mixhash": "0",` works for geth.
But we will be revisiting defining the genesis block anyway. This is good enough for now.

rtkaczyk: I mean, not only does it work, it also results in the same genesis block hash: `ee8486fc3536b38e980fd51263eea65c30a1d4193d839717902e7150f28301c6`. So geth treats this value as a number (at least at user input).

LukasGasior1: my guess is it just adds the padding :)

PR #110, opened 2017-03-21
[EC-112] Block validations before execution

ntallar:
## Description

Includes doing block validations before executing the transactions of a block, which include previously implemented validations and `OmmersValidator`, which contains checks from section 11.1 from the YP.


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -16,7 +17,7 @@ object Ledger extends Logger {
     storages: BlockchainStorages,
     stateStorage: NodeStorage): Unit = {
 
-    val blockError = validateBlockBeforeExecution(block)
+    val blockError = validateBlockBeforeExecution(block, BlockchainImpl(storages))
```
AlanVerbner: Blockchain is also being created in `io.iohk.ethereum.ledger.Ledger#executeBlockTransactions` maybe it's a good idea to use this one as parameter on that method too

src/main/scala/io/iohk/ethereum/network/p2p/validators/OmmersValidator.scala
```diff
@@ -0,0 +1,138 @@
...
+      .collect{ case i if blockNumber - i >= 0 => blockNumber - i }
+      .map{ num => blockchain.getBlockHeaderByNumber(num) }
+    if(ancestors.exists(_.isEmpty)) None
+    else Some(ancestors.flatten)
+  }
+
+  private def getUsedOmmers(blockNumber: BigInt, numberOfBlocks: Int, blockchain: Blockchain): Option[Seq[BlockHeader]] = {
+    val usedOmmers: Seq[Option[Seq[BlockHeader]]] = (1 to numberOfBlocks)
+      .collect{ case i if blockNumber - i >= 0 => blockNumber - i }
+      .map{ num => blockchain.getBlockByNumber(num).map(_.body.uncleNodesList) }
```
AlanVerbner: why not `flatMap` ?

src/main/scala/io/iohk/ethereum/network/p2p/validators/OmmersValidator.scala
```diff
@@ -0,0 +1,138 @@
...
+    * @param blockchain     from where the previous blocks are obtained
+    * @return ommers if valid, an [[OmmersError]] otherwise
+    */
+  def validate(blockNumber: BigInt, ommers: Seq[BlockHeader], blockchain: Blockchain): Either[OmmersError, Seq[BlockHeader]] = {
+    for {
+      _ <- validateOmmersLength(ommers)
+      _ <- validateOmmersHeaders(ommers, blockchain)
+      _ <- validateOmmersAncestors(blockNumber, ommers, blockchain)
+      _ <- validateOmmersNotUsed(blockNumber, ommers, blockchain)
+      _ <- validateDuplicatedOmmers(ommers)
```
AlanVerbner: Maybe this validation can be put before validateOmmersHeaders as it's _les expensive_ and might make validation fail earlier rather than getting ommers from db

src/main/scala/io/iohk/ethereum/network/p2p/validators/OmmersValidator.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  private def getAncestors(blockNumber: BigInt, numberOfBlocks: Int, blockchain: Blockchain): Option[Seq[BlockHeader]] = {
+    val ancestors: Seq[Option[BlockHeader]] = (1 to numberOfBlocks)
+      .collect{ case i if blockNumber - i >= 0 => blockNumber - i }
+      .map{ num => blockchain.getBlockHeaderByNumber(num) }
+    if(ancestors.exists(_.isEmpty)) None
+    else Some(ancestors.flatten)
+  }
+
+  private def getUsedOmmers(blockNumber: BigInt, numberOfBlocks: Int, blockchain: Blockchain): Option[Seq[BlockHeader]] = {
```
AlanVerbner: minor comment. I would change the name of this method to something like `collectOmmersFromParents` or something like that

src/main/scala/io/iohk/ethereum/network/p2p/validators/OmmersValidator.scala
```diff
@@ -0,0 +1,140 @@
+package io.iohk.ethereum.network.p2p.validators
```
rtkaczyk: Why in network?

src/main/scala/io/iohk/ethereum/validators/OmmersValidator.scala
```diff
@@ -0,0 +1,140 @@
...
+    * It also includes validations mentioned in the white paper (https://github.com/ethereum/wiki/wiki/White-Paper)
+    * and implemented in the different ETC clients:
+    *   - OmmersValidator.validateOmmersNotUsed
+    *   - OmmersValidator.validateDuplicatedOmmers
+    * @param blockNumber    The number of the block to which the ommers belong
+    * @param ommers         The list of ommers to validate
+    * @param blockchain     from where the previous blocks are obtained
+    * @return ommers if valid, an [[OmmersError]] otherwise
+    */
+  def validate(blockNumber: BigInt, ommers: Seq[BlockHeader], blockchain: Blockchain): Either[OmmersError, Seq[BlockHeader]] = {
```
LukasGasior1: how about just `Either[OmmersError, Unit]` and then `yield ()`? And similar for remaining methods

src/main/scala/io/iohk/ethereum/validators/OmmersValidator.scala
```diff
@@ -0,0 +1,140 @@
+package io.iohk.ethereum.validators
+
+import io.iohk.ethereum.domain.{BlockHeader, Blockchain}
+import io.iohk.ethereum.validators.OmmersValidator.OmmersError._
+
+object OmmersValidator {
+
+  val OmmerGenerationLimit: Int = 6
```
LukasGasior1: is this defined in YP or just what other peers use?

rtkaczyk: YP, section 11.1, eq. (143)

@ntallar we could add a comment

ntallar: I will add one :+1:

src/main/scala/io/iohk/ethereum/validators/OmmersValidator.scala
ntallar: Fixed! Should we also change the the type of the functions of the other validators as well?

ntallar: I'm merging this PR, if this change is needed we can change it in a later PR

PR #111, opened 2017-03-22
Add logging of black list reason

adamsmo:


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncReceiptsRequestHandler.scala
```diff
@@ -25,6 +26,8 @@ class FastSyncReceiptsRequestHandler(
     updateBestBlockIfNeeded(receivedHashes)
 
     if (receipts.receiptsForBlocks.isEmpty) {
+      log.info(s"Blacklisting peer (${peer.path.name}), " +
+        s"got empty receipts for known hashes: ${requestedHashes.map(h => Hex.toHexString(h.toArray[Byte]))}")
```
mcsherrylabs: Can we make a lil function out of this `requestedHashes.map(h => Hex.toHexString(h.toArray[Byte]))`

whysoserious:
I don't like the design of this feature, which shouldn't stop us from merging this PR.

We'll have to add log message in each place where a peer is blacklisted. 

This is the right place to start working on that: https://github.com/input-output-hk/etc-client/blob/e8367fff1a376bb75b455bac3799f389fe8d7f5b/src/main/scala/io/iohk/ethereum/blockchain/sync/BlacklistSupport.scala#L17

`def blacklist(peer: ActorRef, duration: FiniteDuration, reason: String): Unit`

adamsmo:
@whysoserious ok will refactor this

adamsmo:
@whysoserious / @mcsherrylabs could you take a look?

whysoserious:
👀  with 👓

PR #112, opened 2017-03-22
[EC-143] Code throws 'data too short' exception regularly

adamsmo:
Issue is related to too big message containing receipts, fix for now is to limit number of receipts that can be send with one message
Final fix is adding support for multi frame messages


adamsmo:
@mcsherrylabs / @rtkaczyk could you take a look?

PR #115, opened 2017-03-23
Application configuration substitution

AlanVerbner:
## Description

Before this PR, when overriding `datadir` in dist configuration file, all dependant values (like `keys-file`)

## Tested cases:
- Overriding `datadir` in order to check we can run two different instances without overriding nodeId.keys 
- Tested akka config override puting, for example, `akka.loglevel = "ERROR"`


AlanVerbner:
@mcsherrylabs done! sorry for the confusion :sweat:

AlanVerbner:
@rtkaczyk 

1) :cry: 
2) @mcsherrylabs what are your thoughts about that? I think it's a good suggestion

rtkaczyk:
re 1. I already merged txExecution into this one, but then I was waiting for tests to pass and forgot to merge it back

PR #118, opened 2017-03-27
EC-144: Add peers limit

LukasGasior1:


src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -39,10 +42,19 @@ class PeerManagerActor(
...
+        log.info("Maximum number of connected peers reached. Peer {} will be disconnected.", peer.id)
+        peer.ref ! PeerActor.DisconnectPeer(Disconnect.Reasons.TooManyPeers)
+      }
+
     case ConnectToPeer(uri) =>
-      val peer = createPeer(new InetSocketAddress(uri.getHost, uri.getPort))
-      peer.ref ! PeerActor.ConnectTo(uri)
-      sender() ! PeerCreated(peer)
+      if (peers.size >= peerConfiguration.maxPeers) {
+        log.info("Maximum number of connected peers reached. Discarding connecting to {}", uri)
```
adamsmo: `Discarding connecting to` => `Not connecting to` ?

adamsmo:
as previously
I have issue with connecting to peers because app is considering peers from list that are not yet conneted as working connections and even if I am connected to only 1 peer and limit is set to 5 I am getting
```
19:19:19.483 i.i.e.network.PeerManagerActor - Maximum number of connected peers reached. Not connecting to enode:
```
for 4 peers out of 9 on the list from application.conf

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
```diff
@@ -65,17 +65,21 @@ class SyncController(
   }
 
   def handlePeerUpdates: Receive = {
-    case PeerManagerActor.PeersResponse(peers) =>
-      peers.foreach(_.ref ! PeerActor.GetStatus)
+    case peers: PeerManagerActor.Peers =>
+      val newHandshakedPeers = peers.handshaked.filterNot(p => handshakedPeers.contains(p._1.ref))
+      newHandshakedPeers.foreach(p => context watch p._1.ref)
```
ntallar: Minor detail: As we later start watching all handshaked peers in `if (!handshakedPeers.contains(peer.ref)) context watch peer.ref`, is watching this peers necessary here?

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -69,14 +90,50 @@ class PeerManagerActor(
...
+  def tryDiscardPeersToFreeUpLimit(): Future[Unit] = {
+    getPeers() flatMap { peers =>
+      val peersToPotentiallyDisconnect = peers.peers.filter {
+        case (_, Handshaking(numRetries)) if numRetries > 0 => true /* still handshaking and retried at least once */
+        case (_, PeerActor.Status.Disconnected) => true /* already disconnected */
+        case _ => false
+      }
+
+      if (peers.peers.size - peersToPotentiallyDisconnect.size >= peerConfiguration.maxPeers) {
+        Future.failed(new RuntimeException("Too many peers"))
```
ntallar: I'm wondering, should we try to connect to at least some of the handshaked peers here?
For example I tried having `max-peers=2` and I got the following logs:
```
17:51:15.337 i.i.e.network.PeerManagerActor - Maximum number of connected peers reached. Not connecting to enode://18a551bee469c2e02de660ab01dede06503c986f6b8520cb5a65ad122df88b17b285e3fef09a40a0d44f99e014f8616cf1ebc2e094f96c6e09e2f390f5d34857@47.90.36.129:30303
17:51:15.339 i.i.e.network.PeerManagerActor - Maximum number of connected peers reached. Not connecting to enode://98b863da48ab8bef2339a823d552f3619fd8e892425ae40c6812c6f7e4a0afb4f9591b012183e89a63bb01c5085d0e96aa5f0812652335fb0ac946d6aaf15881@118.178.57.121:30303
17:51:15.363 i.i.e.blockchain.sync.SyncController - Block: 66176/3484815. Peers: 1/1 (0 blacklisted). State: 97425/108279 known nodes.
```
Shouldn't we in this case connect to one of the two nodes as the number of connected nodes is less than `max-peers`?

ntallar: This also happened to me while `max-peers=3`, could that be the case? I think it has something to do with the fact that many of the peers have a `Connecting` status and this is not considered in the `peersToPotentiallyDisconnect` peers.

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
LukasGasior1: Right, it's not needed anymore

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
LukasGasior1: Yes. It's not really that easy. If there are, for example 2 peers that are `Connecting` and you have 2 more peers on the list that you may try connecting to (for example, as in your first case). What should we do?
If we try to connect to them anyway, and they succeed but also the `Connecting` peers succeed, we end up with more peers than the limit.
The logic here is, if we try to connect to new peer, and there's another that's `Connecting` (without any retries) they're treated "equally", i.e that `Connecting` one was first so it should not be disconnected. If at later point this peer fails to connect (or has multiple reconnect retry attempts) it will be considered "worse" and put in `peersToPotentiallyDisconnect` list to free up space. This check will happen every `bootstrap-nodes-scan-interval`

ntallar: Sounds a good strategy, thanks for the explanation :+1:
Regarding this, while testing this branch I found out that we were connected to 6 peers while `max-peers=3` with the status of 5 of them being `Connecting` and `Handshaked` for the remaining peer. Shouldn't that be avoided if I understood correctly?

LukasGasior1: @ntallar I've pushed a fix

ntallar: I will take a look at it :+1:
There seem to be a test currently failing

PR #120, opened 2017-03-29
[EC-132] ERROR Connection closed unexpectedly too frequent

adamsmo:
Add more descriptive message for logging disconnection


src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
```diff
@@ -109,23 +110,23 @@ class RLPxConnectionHandler(nodeKey: AsymmetricCipherKeyPair)
...
+        log.warning(s"[Stopping Connection] Auth handshake timeout for peer $peerId")
         context.parent ! ConnectionFailed
         context stop self
     }
 
     def processHandshakeResult(result: AuthHandshakeResult, remainingData: ByteString): Unit =
       result match {
         case AuthHandshakeSuccess(secrets) =>
-          log.debug("Auth handshake succeeded")
+          log.debug(s"Auth handshake succeeded for peer $peerId")
```
LukasGasior1: I just recalled why I haven't added peerId logging here. It's already a part of a log string - the actor name - but only in the file encoder.

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
```diff
@@ -151,14 +152,20 @@ class RLPxConnectionHandler(nodeKey: AsymmetricCipherKeyPair)
       }
 
     def handleWriteFailed: Receive = {
-      case CommandFailed(_: Write) =>
-        log.warning("Write failed")
+      case CommandFailed(cmd: Write) =>
+        log.warning(s"[Stopping Connection] Write to peer $peerId failed, trying to send ${Hex.toHexString(cmd.data.toArray[Byte])}")
```
LukasGasior1: this log can be quite big depending on the message

rtkaczyk: Won't this cause floods in the logs?

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
adamsmo: ok, I think it is more visible this way, when labeled as `peer`

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
adamsmo: yes it can, but I think we want to have as much information about failure as possible, also if it is big you will immediately find it scrolling through log, we are zipping log files after all so file size is not the issue

adamsmo: It should be rear cases, so I think no

PR #123, opened 2017-03-30
[EC-139] Process logs and receipts

ntallar:
## Description

Includes the creation of a receipt for each transaction executed which primary involves the creation of the `Bloom Filter` from the logs created by its execution.


src/main/scala/io/iohk/ethereum/ledger/BloomFilter.scala
```diff
@@ -0,0 +1,55 @@
...
+object BloomFilter {
+
+  val BloomFilterByteSize: Int = 256
+  private val BloomFilterBitSize: Int = BloomFilterByteSize * 8
+  private val EmptyBloomFilter: Array[Byte] = Array.fill(BloomFilterByteSize)(0.toByte)
+  private val IntIndexesToAccess: Set[Int] = Set(0, 2, 4)
+
+  /**
+    * Given the logs of a receipt creates the bloom filter associated with them
+    * as stated in section 4.4.1 of http://paper.gavwood.com/
```
LukasGasior1: side note: let's stay consistent and use `YP` to refer to yellow paper

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -2,13 +2,14 @@ package io.iohk.ethereum.ledger
 
 import io.iohk.ethereum.db.storage.NodeStorage
 import io.iohk.ethereum.domain._
+import io.iohk.ethereum.network.p2p.messages.PV63.Receipt
```
LukasGasior1: I don't like the fact that we're using network classes here. Can we create `Receipt` class in `domain` or `ledger` (whichever you think is better).

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -2,13 +2,14 @@ package io.iohk.ethereum.ledger
...
 import io.iohk.ethereum.utils.{Config, Logger}
 import io.iohk.ethereum.vm.{GasFee, _}
 
 object Ledger extends Logger {
 
   type PR = ProgramResult[InMemoryWorldStateProxy, InMemoryWorldStateProxyStorage]
-  type ExecResult = (InMemoryWorldStateProxy, BigInt)
+  type ExecResult = (InMemoryWorldStateProxy, BigInt, Seq[Receipt])
```
LukasGasior1: do we expect any more values here in the future? If so maybe it's good time to make it a `case class`

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
ntallar: You are right, I'm moving it to `domain` :+1:

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
ntallar: I don't think we will be needing to add any more values, however I will add a `case class` so that it is more clear

src/main/scala/io/iohk/ethereum/domain/Receipt.scala
```diff
@@ -0,0 +1,41 @@
...
+       |postTransactionStateHash: ${Hex.toHexString(postTransactionStateHash.toArray[Byte])}
+       |cumulativeGasUsed: $cumulativeGasUsed
+       |logsBloomFilter: ${Hex.toHexString(logsBloomFilter.toArray[Byte])}
+       |logs: $logs
+       |}
+       """.stripMargin
+  }
+}
+
+object Receipt {
```
adamsmo: not sure if this should be part of domain, it is related to networking and communications, maybe something like
https://github.com/input-output-hk/etc-client/blob/phase/3/txExecution/src/main/scala/io/iohk/ethereum/network/p2p/messages/PV62.scala#L82

src/main/scala/io/iohk/ethereum/ledger/BloomFilter.scala
```diff
@@ -0,0 +1,55 @@
...
+    val bloomFilters = dataForBloomFilter.map(bytes => bloomFilter(bytes.toArray))
+
+    or(bloomFilters: _*)
+  }
+
+  //Bloom filter that sets 3 bits out of 2048 based on equations 25-28 of http://paper.gavwood.com/
+  private def bloomFilter(bytes: Array[Byte]): Array[Byte] = {
+    val hashedBytes = kec256(bytes)
+    val bitsToSet = IntIndexesToAccess.map{ i =>
+      ((hashedBytes(i + 1) & 0xFF) + ((hashedBytes(i) & 0xFF) << 8)) % BloomFilterBitSize
```
adamsmo: what we are doing in this line?
it looks like we are getting 2 bytes (16 bits) from 3 first pairs of bytes from kec256
in YP it is required to get the low-order 11 bits, are we selecting this 11 bits somewhere else?

ntallar: You are right that in this line we get 11 bits from the 3 first pairs of bytes from kec256, 5 of the bits from the 16 bits are discarded while doing `% BloomFilterBitSize` (I tried to stick with how the formula is defined in the YP). I will add a comment and do some refactoring to make that clearer.

adamsmo:
LGTM

PR #125, opened 2017-04-03
[EC-113] Validations after tx execution

ntallar:
## Description

Includes the validation of the block with the results obtained when executing the transactions included in it.


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -22,7 +25,8 @@ object Ledger extends Logger {
       log.debug(s"All txs from block ${block.header} were executed")
 
       val worldToPersist = payBlockReward(Config.Blockchain.BlockReward, block, resultingWorldStateProxy)
-      val afterExecutionBlockError = validateBlockAfterExecution(block, worldToPersist)
+      val receipts = Seq() //FIXME: This should be obtained from block execution (done in EC-139)
```
adamsmo: does that mean that this PR should be merged after https://github.com/input-output-hk/etc-client/pull/123 ?

ntallar: It involves a small change so I think that any of the two PRs could be merged first. When that's the case I will update the use of the calculated receipts for the validation after the execution.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -81,7 +83,30 @@ object Ledger extends Logger {
...
+                                                  receipts: Seq[Receipt], gasUsed: BigInt): Option[String] = {
+    lazy val blockAndReceiptsValidation = BlockValidator.validateBlockAndReceipts(block, receipts)
+    if(block.header.gasUsed != gasUsed)
+      Some(s"Block has invalid gas used: ${block.header.gasUsed} != $gasUsed")
+    else if(block.header.stateRoot != stateRootHash)
+      Some(s"Block has invalid state root hash: ${block.header.stateRoot} != $stateRootHash")
+    else if(blockAndReceiptsValidation.isLeft)
+      Some(blockAndReceiptsValidation.left.get.toString)
+    else
+      None
```
adamsmo: only issue that I see here is that we are defaulting to None that in this case means valid result
I would like it to be more like
`if (conditions) valid else failure` this way it is harder to produce false valid result

rtkaczyk: I have pondered similar cases a few times before, and still haven't reached a good conclusion...

Because what type should `valid` be? `Block`? What's the point in returning the same unmodified block?

adamsmo: maybe sth like this:
```
a = block.header.gasUsed != gasUsed
b = block.header.stateRoot != stateRootHash
c = blockAndReceiptsValidation.isLeft
if(a && b && c) {
 None
} else {
 val msgs = Seq(a,b,c).zip(msgA,msgB,msgC).collect{case (check,msg) if check => msg}
 Some(msgs)
}
```

rtkaczyk: So that just gives us the possibility accumulate checks. All the validations we currently give up on first failure, so I wouldn't do anything else here. Shouldn't be a problem to change that in the future.

adamsmo: 👍

PR #126, opened 2017-04-04
[EC-99][EC-110] gas related payments in TX execution

rtkaczyk:


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -44,32 +45,41 @@ object Ledger extends Logger {
...
+    val result = runVM(stx, blockHeader, world1)
+
+    val gasUsed = if(result.error.isDefined) gasLimit else gasLimit - result.gasRemaining
+    val gasRefund = calcGasRefund(stx, result)
+
+    val refundGasFn = pay(stx.recoveredSenderAddress.get, gasRefund * gasPrice) _
+    val payMinerForGasFn = pay(Address(blockHeader.beneficiary), (gasLimit - gasRefund) * gasPrice) _
+    val deleteAccountsFn = deleteAccounts(result.addressesToDelete) _
+
+    val world2 = (refundGasFn andThen payMinerForGasFn andThen deleteAccountsFn) (result.world)
```
ntallar: Is it okay for accounts to be deleted in case the execution resulted in an error? At least `EthereumJ` seems to clear the list of accounts to be deleted when an exception ocurrs.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -44,32 +45,41 @@ object Ledger extends Logger {
...
-              if (theResult.gasRemaining > 0) (refundGas _).curried(Address(block.header.beneficiary))(gasUsed)
-              else identity[InMemoryWorldStateProxy] _
-            }
-            val payBeneficiariesFn = (payForGasUsedToBeneficiary _).curried(stx)
-            val deleteAccountsFn = (deleteAccounts _).curried(theResult.addressesToDelete)
-            (refundGasFn andThen payBeneficiariesFn andThen deleteAccountsFn) (theResult.world) -> (gasUsed + acumGas)
-          case Left(error) => throw new RuntimeException(error)
+    block.body.transactionList.foldLeft[ExecResult](initialWorld -> 0) {
+      case ((world, acumGas), stx) =>
+        //TODO: change 0 -> acumGas ?
```
ntallar: Should this be changed in this PR? I think you are right about having to do this change.

src/main/scala/io/iohk/ethereum/crypto/ECDSASignature.scala
```diff
@@ -72,6 +74,19 @@ object ECDSASignature {
     else s
   }
 
+  def sign(message: Array[Byte], keyPair: AsymmetricCipherKeyPair): ECDSASignature = {
```
ntallar: Minor detail: Should we add some tests for this function?

src/main/scala/io/iohk/ethereum/crypto/ECDSASignature.scala
```diff
@@ -72,6 +74,19 @@ object ECDSASignature {
...
+    val signer = new ECDSASigner(new HMacDSAKCalculator(new SHA256Digest))
+    signer.init(true, keyPair.getPrivate)
+    val components = signer.generateSignature(message)
+    val r = components(0)
+    val s = ECDSASignature.canonicalise(components(1))
+    val v = ECDSASignature
+      .calculateV(r, s, keyPair, message)
+      .getOrElse(throw new RuntimeException("Failed to calculate signature rec id"))
```
ntallar: Should this exception be handled or it will be handled by the user of it?
As so far this is used only for testing this is not something urgent to solve if it needs to be.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -258,7 +281,7 @@ object Ledger extends Logger {
     * @param worldStateProxy
     * @return
     */
-  private def payForGasUsedToBeneficiary(stx: SignedTransaction, worldStateProxy: InMemoryWorldStateProxy): InMemoryWorldStateProxy = worldStateProxy //TODO
+  private def payForGasUsedToBeneficiary(stx: SignedTransaction)(worldStateProxy: InMemoryWorldStateProxy): InMemoryWorldStateProxy = worldStateProxy //TODO
```
ntallar: Isn't this function implemented with the `payMinerForGasFn` function in `executeTransaction`?

src/main/scala/io/iohk/ethereum/crypto/ECDSASignature.scala
rtkaczyk: Probably, though it's kinda strange this wasn't there from the start. This PR already does several things beyond its planned scope so I'd rather not add even more to it. 

There's actually a couple of things to be refactored about the `ECDSASignature`, IMO. I will make task of that add a note about the test. Pinging @LukasGasior1

src/main/scala/io/iohk/ethereum/crypto/ECDSASignature.scala
rtkaczyk: We use that for signing and I don't think the result should be dependent on user input (correct me if I'm wrong) - if there's an error it means we messed up. An exception is fine is such situation

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
rtkaczyk: 👍 I will remove it

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -45,43 +47,52 @@ object Ledger extends Logger {
...
+
+    val gasUsed = if(result.error.isDefined) gasLimit else gasLimit - result.gasRemaining
+    val gasRefund = calcGasRefund(stx, result)
+
+    val refundGasFn = pay(stx.senderAddress, gasRefund * gasPrice) _
+    val payMinerForGasFn = pay(Address(blockHeader.beneficiary), (gasLimit - gasRefund) * gasPrice) _
+    val deleteAccountsFn = deleteAccounts(result.addressesToDelete) _
+    val persistStateFn = InMemoryWorldStateProxy.persistState _
+
+    val world2 = (refundGasFn andThen payMinerForGasFn andThen deleteAccountsFn andThen persistStateFn)(result.world)
```
rtkaczyk: @ntallar you asked: 

> Is it okay for accounts to be deleted in case the execution resulted in an error? At least EthereumJ seems to clear the list of accounts to be deleted when an exception ocurrs.

You're right, but let's make that part of EC-151

src/main/scala/io/iohk/ethereum/crypto/ECDSASignature.scala
ntallar: I agree, we can add this in the refactoring task or in another later one :+1:

LukasGasior1: Just a suggestion, maybe we can move that function to `crypto` package obj? This seems to be good place for this kind of functions

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -250,29 +259,44 @@ object Ledger extends Logger {
...
-    if (result.error.isDefined) Left(result.error.get.toString)
-    else {
-      if (stx.tx.isContractInit) payContractCreationCost(result).map(saveCreatedCost(programContext.env.ownerAddr, _))
-      else Right(result)
-    }
+  private def runVM(stx: SignedTransaction, blockHeader: BlockHeader, worldStateProxy: InMemoryWorldStateProxy): PR = {
+    val programContext: PC = ProgramContext(stx, blockHeader, worldStateProxy)
+    val result = vm.run(programContext)
+    if (stx.tx.isContractInit && result.error.isEmpty)
+      (payContractCreationCost _ andThen saveCreatedCode(programContext.env.ownerAddr))(result)
```
LukasGasior1: note to myself: EIP150 (externalities.rs, line:269)

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -94,6 +96,25 @@ object SignedTransaction {
       sender <- SignedTransaction.getSender(tx, signature, recoveredPointSign)
     } yield SignedTransaction(tx, signature, sender)
   }
+
+  def sign(tx: Transaction, keyPair: AsymmetricCipherKeyPair): SignedTransaction = {
+    val bytes = crypto.kec256(
```
ntallar: I'm wondering, should we obtain the bytes to sign as in `getSender` (https://github.com/input-output-hk/etc-client/blob/86c57a8ed9a702b67a18cbbc9b5f01a3121c2c3b/src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala#L45)? At least EthereumJ seems to do so in that way, or maybe I'm missing something

src/main/scala/io/iohk/ethereum/domain/Transaction.scala
```diff
@@ -6,9 +6,9 @@ import org.spongycastle.util.encoders.Hex
 
 object Transaction {
 
-  val NonceLength = 32
-  val GasLength = 32
-  val ValueLength = 32
+  val NonceLength = 33
+  val GasLength = 33
+  val ValueLength = 33
```
ntallar: As the way this are checked in `SignedTransactionValidator` was fixed, should we return this lengths to `32` bytes?

src/main/scala/io/iohk/ethereum/crypto/ECDSASignature.scala
rtkaczyk: I think `sign` does belong here. Also what if we add differenet signature schemes? Anyway, checkout task EC-157

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
ntallar: Minor detail: In case `payContractCreationCost` results in an `OutOfGas` error, should we call `saveCreatedCode` either way?

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
rtkaczyk: The signing method has changed slightly. When receiving transactions we need to deal with both the current and the historic one.

But we will only be signing new transactions, so only the current method will be used.

It would still be nice to avoid the code repetition, but let's close one eye for now 😉

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
rtkaczyk: 🤔

rtkaczyk: Fixing...

rtkaczyk:
Conflicts merged, ready for another round of review

ntallar:
The tests seems to be failing, apart from that LGTM!

rtkaczyk:
Tests failed because of timeouts:

```
[info] - should handle block header resolution request *** FAILED *** (5 seconds, 417 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds)

[info] - should blacklist if the response is empty *** FAILED *** (4 seconds, 90 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds)

[info] - should handle timeout *** FAILED *** (3 seconds, 857 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) 


[info] *** 3 TESTS FAILED ***
[error] Failed: Total 589, Failed 3, Errors 0, Passed 586, Ignored 1
[error] Failed tests:
[error] 	io.iohk.ethereum.blockchain.sync.FastSyncReceiptsRequestHandlerSpec
[error] 	io.iohk.ethereum.blockchain.sync.FastSyncBlockHeadersRequestHandlerSpec
[error] 	io.iohk.ethereum.blockchain.sync.FastSyncBlockBodiesRequestHandlerSpec
```

PR #128, opened 2017-04-04
Add EVM config

LukasGasior1:


LukasGasior1:
WIP as there are still tests failures, but it doesn't hold up reviews

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -0,0 +1,156 @@
...
+    feeSchedule = FeeSchedule.HomesteadFeeSchedule,
+    opCodes = OpCodes.HomesteadOpCodes)
+
+  // TODO: sub_gas_cap_divisor, no_empty, kill_empty from schedule.rs, parity
+  val PostEIP150Config = EvmConfig(
+    feeSchedule = FeeSchedule.PostEIP150FeeSchedule,
+    opCodes = OpCodes.PostEIP150OpCodes)
+
+  val frontierBlockNumber = BigInt(0)
+  val homesteadBlockNumber = BigInt(0)
```
ntallar: Should this value be obtained from `HomesteadBlock` in `Config.scala` or how should we handle having both values?

LukasGasior1: I think ultimately we should obtain these values from config files (which may be different for real/test/private networks).

rtkaczyk: So, are these hardcoded now?

src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -83,8 +87,6 @@ object Config {
 
     val chainId: Byte = Hex.decode(blockchainConfig.getString("chain-id")).head
 
-    val HomesteadBlock: BigInt = 1150000
-
     // YP eq 150
     val BlockReward = UInt256(BigInt("5000000000000000000"))
```
AlanVerbner: @LukasGasior1 even if this is defined in the YP, do you think it's worthy to put it inside a Config (not EVMConfig)?

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -0,0 +1,173 @@
...
+    subGasCapDivisor: Option[Long] = None) {
+
+  val byteToOpCode: Map[Byte, OpCode] =
+    opCodes.map(op => op.code -> op).toMap
+}
+
+object FeeSchedule {
+
+  // See YP, appendix G
+  sealed trait Key
```
AlanVerbner: Just a super small comment. I would name this `Cost` or `GasCost` instead of `Key`. Obviously it's not a blocker for a merge.

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -0,0 +1,173 @@
...
+  val PostEIP150FeeSchedule: FeeSchedule =
+    HomesteadFeeSchedule.copy(values = HomesteadFeeSchedule.values ++ Map(
+      Key.G_sload -> UInt256(200),
+      Key.G_call -> UInt256(700),
+      Key.G_balance -> UInt256(400),
+      Key.G_selfdestruct -> UInt256(5000),
+      Key.G_extcodesize -> UInt256(700),
+      Key.G_extcodecopy_base -> UInt256(700),
+      Key.G_selfdestruct_to_new_account -> UInt256(25000),
+      Key.G_expbyte -> UInt256(10)))
```
AlanVerbner: @LukasGasior1 this value seems to be the same as Frontier config, has it changed? I haven't found a new definition in EIP150 neither....am I doing something wrong?

src/main/scala/io/iohk/ethereum/utils/Config.scala
LukasGasior1: I'll move it to conf

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
LukasGasior1: Good catch! Seems it's implemented in yet another EIP (EIP160)

src/main/resources/application.conf
```diff
@@ -69,6 +70,9 @@ etc-client {
     chain-id = "3d"
 
     custom-genesis-file = null
+
+    // YP eq 150
+    block-reward = "5000000000000000000"
```
AlanVerbner: @LukasGasior1 I was talking about changing this reward based on block number actually, do you think that makes sense or not? Maybe it's just a _feature_ that will never be used

LukasGasior1: oh, didn't get that. If needed we cal always add it later

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -0,0 +1,177 @@
...
+    Config.Blockchain.eip160BlockNumber -> PostEIP160Config)
+
+}
+
+case class EvmConfig(
+    feeSchedule: FeeSchedule,
+    opCodes: List[OpCode],
+    exceptionalFailedCodeDeposit: Boolean,
+    maxMemory: UInt256 = UInt256(Long.MaxValue), /* used to artificially limit memory usage by incurring maximum gas cost */
+    maxCallDepth: Int = 1024,
```
rtkaczyk: What EIPS affect that?

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -0,0 +1,177 @@
...
+    Config.Blockchain.eip150BlockNumber -> PostEIP150Config,
+    Config.Blockchain.eip160BlockNumber -> PostEIP160Config)
+
+}
+
+case class EvmConfig(
+    feeSchedule: FeeSchedule,
+    opCodes: List[OpCode],
+    exceptionalFailedCodeDeposit: Boolean,
+    maxMemory: UInt256 = UInt256(Long.MaxValue), /* used to artificially limit memory usage by incurring maximum gas cost */
```
rtkaczyk: Should this really be part of this config? This isn't part of any specification - it's something we added as a preventive measure (not really thought out). Also, `Long.MaxValue` doesn't really make sense.

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -0,0 +1,177 @@
...
+    case object G_sha3word extends GasCost
+    case object G_copy extends GasCost
+    case object G_blockhash extends GasCost
+    case object G_extcodesize extends GasCost
+    case object G_extcodecopy_base extends GasCost
+    case object G_selfdestruct_to_new_account extends GasCost
+  }
+
+  val FrontierFeeSchedule: FeeSchedule = FeeSchedule(Map(
+      GasCost.G_zero -> UInt256(0),
```
rtkaczyk: couldn't they all be vals?

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -0,0 +1,177 @@
...
+    case object G_log extends GasCost
+    case object G_logdata extends GasCost
+    case object G_logtopic extends GasCost
+    case object G_sha3 extends GasCost
+    case object G_sha3word extends GasCost
+    case object G_copy extends GasCost
+    case object G_blockhash extends GasCost
+    case object G_extcodesize extends GasCost
+    case object G_extcodecopy_base extends GasCost
+    case object G_selfdestruct_to_new_account extends GasCost
```
rtkaczyk: What EIP is that?

src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpec.scala
```diff
@@ -1,21 +1,26 @@
...
 import io.iohk.ethereum.vm.MockWorldState.PS
-
+import scala.language.implicitConversions
 
 // scalastyle:off magic.number
 class OpCodeGasSpec extends FunSuite with OpCodeTesting with Matchers with PropertyChecks {
 
-  val stackOpsFees = (pushOps ++ dupOps ++ swapOps).map(_ -> G_verylow)
-  val constOpsFees = constOps.map(_ -> G_base)
+  override val config = EvmConfig.FrontierConfig
```
rtkaczyk: Should we test them with the latest config?

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
LukasGasior1: I don't think any EIP does. I moved it to config as it seems a good place for such constants.

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
LukasGasior1: I moved it here same way as maxCallDepth - config just seems a good place. `Long.MaxValue` was already there. Agree, it doesn't make much sense. Is this defined in YP or somewhere? If not, can we get rid of this as it's not really used?

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
LukasGasior1: they could

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
LukasGasior1: EIP150 (https://github.com/ethereum/EIPs/issues/150 version 1c)

src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpec.scala
LukasGasior1: ideally we should test will all of them

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
LukasGasior1: actually that wouldn't be that straightforward. It would require more changes in OpCode - currently we pass the key as constant gas in parent constructor. This change would require to make it config-dependent (which can be done, but I'm not sure if there are any benefits of doing so).

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
rtkaczyk: No, it was `Int.MaxValue` which is 2G of memory. The thing is I don't think that should be configurable, maybe it shouldn't even be here at all - relying on high gas cost preventing us from going OOM.

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
rtkaczyk: So maybe we should make it a constant actually? Like a val in `EmvConfig`?

src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
```diff
@@ -4,7 +4,10 @@ import io.iohk.ethereum.domain._
 
 
 object ProgramContext {
-  def apply[W <: WorldStateProxy[W, S], S <: Storage[S]](stx: SignedTransaction, blockHeader: BlockHeader, world: W): ProgramContext[W, S] = {
+  def apply[W <: WorldStateProxy[W, S], S <: Storage[S]](stx: SignedTransaction,
+                                                         blockHeader: BlockHeader,
+                                                         world: W,
+                                                         config: EvmConfig): ProgramContext[W, S] = {
```
rtkaczyk: I'd suggest formatting that is more resilient to method signature changes:

```
def apply[W <: WorldStateProxy[W, S], S <: Storage[S]](
  stx: SignedTransaction,
  blockHeader: BlockHeader,
...
```

src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpec.scala
rtkaczyk: Ideally... :)
We should first of all support the latest version, I guess 99% of our users will only need that. 
And by the way, apart from running full sync from genesis, do we have a plan for testing these changeovers?

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
mcsherrylabs: Assuming we need to keep it, `maxMemory` is something we can tune, so a config param, right?

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
```diff
@@ -1,14 +1,22 @@
...
+import io.iohk.ethereum.vm.FeeSchedule.GasCost._
 import org.scalatest.{Matchers, WordSpec}
 import Assembly._
 import GasFee._
 import io.iohk.ethereum.domain.{Account, Address}
 import io.iohk.ethereum.vm.MockWorldState._
 
+import scala.language.implicitConversions
+
+// scalastyle:off
```
rtkaczyk: What rules were broken? Should this be changed to turn off specific rules?

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -0,0 +1,177 @@
...
+      GasCost.G_transaction -> UInt256(21000),
+      GasCost.G_log -> UInt256(375),
+      GasCost.G_logdata -> UInt256(8),
+      GasCost.G_logtopic -> UInt256(375),
+      GasCost.G_sha3 -> UInt256(30),
+      GasCost.G_sha3word -> UInt256(6),
+      GasCost.G_copy -> UInt256(3),
+      GasCost.G_blockhash -> UInt256(20),
+      GasCost.G_extcodesize -> UInt256(20),
+      GasCost.G_extcodecopy_base -> UInt256(20),
```
rtkaczyk: What's this? How is it different from `G_extcode` (which doesn't appear to be used anymore)?

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
LukasGasior1: :+1:

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -0,0 +1,177 @@
...
+
+}
+
+case class EvmConfig(
+    feeSchedule: FeeSchedule,
+    opCodes: List[OpCode],
+    exceptionalFailedCodeDeposit: Boolean,
+    maxMemory: UInt256 = UInt256(Long.MaxValue), /* used to artificially limit memory usage by incurring maximum gas cost */
+    maxCallDepth: Int = 1024,
+    subGasCapDivisor: Option[Long] = None) {
```
rtkaczyk: It seems a bit too generic for my taste. This is a well defined function (YP eq (224), also see [here](https://github.com/input-output-hk/etc-client/blob/359f999a46af45ce0db43b67ad04972127cb6ba9/src/main/scala/io/iohk/ethereum/vm/GasFee.scala#L69-L73)). 

I'd rather have this as a boolean indicating whether to use that function or not.

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
rtkaczyk: I don't know, what if users of the client start adjusting this param to their liking, resulting in forks?

rtkaczyk: Probably that's not very likely... but I'm kinda uneasy about it.

LukasGasior1: I'll make it a constant in EvmCOnfig (like MaxCallDepth)

src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpec.scala
LukasGasior1: @adamsmo do you think we could use the testing mechanism you're working on to test evm config changes?

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
LukasGasior1: magic.number + object.name I'll change that

src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpec.scala
adamsmo: I think yes, I am using `Ledger.executeBlock` so if we can change configuration of eve that is called there we can check if execution resulted in correct changes in storage

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
LukasGasior1: Seems you're right - it's the same just renamed. I'll remove `G_extcode` then

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
rtkaczyk: I see why you did that. The thing is all those "keys" correspond directly the fee schedule defined in the YP appending G. And the definition of SELFDESTRUCT refers to `G_newaccount` (which is the same as for CALL, and that's in line with EIP150). 

It really makes it easier to reason about code vs specification when the names correspond, so if you could find a way for SELFDESTRUCT to utilise `G_newaccount` (a boolean flag?) that would be preferred, IMO.

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
AlanVerbner: I was thinking about the same thing @rtkaczyk

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
rtkaczyk: magic.number should definitely be disabled globally for tests

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
rtkaczyk: I'd rather suggest you would leave `G_extcode` and remove `G_extcodecopy_base`, see [this comment](https://github.com/input-output-hk/etc-client/pull/128#discussion_r111100392)

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
LukasGasior1: As we know, just because it's defined in YP doesn't mean it won't change :) Even though I also think it won't change in the nearest future, I'd still leave that configurable

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
LukasGasior1: G_newaccount is not the same. Even though they share the same value after EIP150, I'd leave them in separate vals. How about adding a comment for those that do not correspond to YP?

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
rtkaczyk: But if it's going to change, it is unlikely that simply that divisor would change. Rather we will use a whole different function, and how would you configure that? 

Rather, functions should be hardcoded and the config should indicate which functions to use.

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
rtkaczyk: What do you mean it's not the same? This is the value that is specified in the YP in the definition of the opcode.

src/main/scala/io/iohk/ethereum/vm/GasFee.scala
```diff
@@ -34,15 +34,14 @@ object GasFee {
     * Calculates transaction intrinsic gas. See YP section 6.2
     *
     */
-  def calcTransactionIntrinsicGas(txData: ByteString, isContractCreation: Boolean, blockNumber: BigInt): UInt256 = {
+  def calcTransactionIntrinsicGas(txData: ByteString, isContractCreation: Boolean, blockNumber: BigInt, config: EvmConfig): UInt256 = {
```
rtkaczyk: Should we maybe place all these methods in `EvmConfig`? Certainly the current object name `GasFee` does not sound appropriate (it never did for these methods).

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
LukasGasior1: True...it is. EIP description doesn't state that clearly.

LukasGasior1: Also parity uses separate configs for these (`suicide_to_new_account_cost` and `call_new_account_gas`). But if Yp defines so, I'll just use G_newaccount

LukasGasior1: Well, no. Actually we need those separate. In config before that EIP G_selfdestruct_to_new_account is 0 while G_newaccount is 25000.

rtkaczyk: Yes, that's one way of looking at it. Another way would be for example:

```
val FrontierConfig = EvmConfig(
    chargeSelfDesctructForNewAccount = false,
    ...)

val PostEIP150Config  = EvmConfig(
    chargeSelfDesctructForNewAccount = false,
    ...)
```

and in `SELFDESCTRUCT`: 

```
if (state.config.chargeSelfDesctructForNewAccount && !state.world.accountExists(refundAddress)) {
  baseGas + state.config.feeSchedule(G_newaccount)
} else baseGas
```

And I think this is preferred because it better corresponds to the YP.

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -0,0 +1,177 @@
...
+    subGasCapDivisor = Some(64))
+
+  val PostEIP160Config = PostEIP150Config.copy(
+    feeSchedule = FeeSchedule.PostEIP160FeeSchedule)
+
+  private val transitionBlockToConfigMapping: Map[BigInt, EvmConfig] = Map(
+    Config.Blockchain.frontierBlockNumber -> FrontierConfig,
+    Config.Blockchain.homesteadBlockNumber -> HomesteadConfig,
+    Config.Blockchain.eip150BlockNumber -> PostEIP150Config,
+    Config.Blockchain.eip160BlockNumber -> PostEIP160Config)
```
mcsherrylabs: For all new code and eventually for all code, the Config as above should be passed into EvmConfig making EvmConfig a class. This is the pattern!

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
LukasGasior1: I like both ways, if you want I can change it to a flag.

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
LukasGasior1: you mean to use it like this: `val config = EvmConfig.forBlock(blockNumber, Config.Blockchain)` ?

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -0,0 +1,175 @@
...
+    chargeSelfDestructForNewAccount: Boolean) {
+
+  val byteToOpCode: Map[Byte, OpCode] =
+    opCodes.map(op => op.code -> op).toMap
+}
+
+object FeeSchedule {
+
+  class FrontierFeeSchedule extends FeeSchedule {
+      override val G_zero = UInt256(0)
```
rtkaczyk: I would run replace on this file: `UInt256\((.+?)\)` -> `$1`
Ints, Longs and Bytes are implicitly casted to `UInt256` because they fit.

src/main/resources/application.conf
```diff
@@ -57,6 +57,11 @@ etc-client {
   }
 
   blockchain {
+    frontier-block-number = "0"
+    homestead-block-number = "1150000"
+    eip150-block-number = "2500000"
+    eip160-block-number = "3000000"
```
rtkaczyk: How would we configure our private net to start with the latest version? Would all the numbers, except for `eip160-block-number`, have to be negative?

LukasGasior1: Yes, either negative or really high (like "0x7fffffffffffffff"). For example:
```
frontier-block-number = "-1"
homestead-block-number = "-1"
eip150-block-number = "-1"
eip160-block-number = "0"
```

rtkaczyk: Ah right, they don't have to be in order

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
rtkaczyk: This also pertains to `MaxMemory`

src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpec.scala
```diff
@@ -302,7 +306,7 @@ class OpCodeGasSpec extends FunSuite with OpCodeTesting with Matchers with Prope
       verifyGas(expectedGas, stateIn, stateOut, allowOOG = false)
     }
 
-    val maxGas = G_verylow + G_copy * 8
+    val maxGas = 123// G_verylow + G_copy * 8
```
rtkaczyk: What's that?

LukasGasior1: oops

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -0,0 +1,175 @@
...
+  val G_txdatanonzero: UInt256
+  val G_transaction: UInt256
+  val G_log: UInt256
+  val G_logdata: UInt256
+  val G_logtopic: UInt256
+  val G_sha3: UInt256
+  val G_sha3word: UInt256
+  val G_copy: UInt256
+  val G_blockhash: UInt256
+  val G_extcodesize: UInt256
```
rtkaczyk: This one seems to be reduntant.

LukasGasior1: true

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
LukasGasior1: to be done in separate PR

src/main/scala/io/iohk/ethereum/vm/GasFee.scala
LukasGasior1: I'll move them to EvmConfig then

LukasGasior1:
@rtkaczyk I've addressed (hopefully all) your concerns.

src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpec.scala
```diff
@@ -313,17 +316,17 @@ class OpCodeGasSpec extends FunSuite with OpCodeTesting with Matchers with Prope
       val stateOut = op.execute(stateIn)
 
       val (Seq(_, offset, _, size), _) = stateIn.stack.pop(4)
-      val memCost = calcMemCost(stateIn.memory.size, offset, size)
+      val memCost = config.calcMemCost(stateIn.memory.size, offset, size)
       val copyCost = G_copy * wordsForBytes(size)
-      val expectedGas = G_extcode + memCost + copyCost
+      val expectedGas = G_extcode + memCost + copyCost + 123
```
rtkaczyk: So I was wondering about this `123` and realised that this tests flawed. The reason is that when the execution goes OOG then `verifyGas` only checks that less than `expectedGas` was used.

The fix for now is to change `maxGas` to something like: 

```
val maxGas = 2 * (G_extcode + G_copy * 8)
```

and remove `123` of course.

Unfortunately there may more tests flawed like that. I'll stick a task in to revise this spec later.

src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpec.scala
```diff
@@ -3,15 +3,17 @@ package io.iohk.ethereum.vm
 import org.scalatest.{FunSuite, Matchers}
 import org.scalatest.prop.PropertyChecks
 import Generators._
-import GasFee._
 import UInt256.{One, Two, Zero}
 import io.iohk.ethereum.domain.{Account, Address}
 import io.iohk.ethereum.vm.MockWorldState.PS
 
-
 // scalastyle:off magic.number
```
rtkaczyk: This should not be necessary: see https://github.com/input-output-hk/etc-client/blob/phase/3/txExecution/scalastyle-test-config.xml#L48

rtkaczyk:
@LukasGasior1 just those 2 comments about `OpCodeGasSpec` and it should be good to go

LukasGasior1:
@rtkaczyk done

PR #129, opened 2017-04-04
[EC-111] Delete accounts after tx execution

ntallar:
## Description

Completion of `deleteAccounts` function for after deleting accounts after tx execution.


src/test/scala/io/iohk/ethereum/ledger/DeleteAccountsSpec.scala
```diff
@@ -0,0 +1,44 @@
+package io.iohk.ethereum.ledger
+
+import io.iohk.ethereum.blockchain.sync.EphemBlockchainTestSetup
+import io.iohk.ethereum.domain.{Account, Address}
+import org.scalatest.{FlatSpec, Matchers}
+
+class DeleteAccountsSpec extends FlatSpec with Matchers {
+
+  it should "delete no accounts when none of them should be deleted" in new TestSetup {
+    val newWorld = InMemoryWorldStateProxy.persistState(Ledger.deleteAccounts(Seq(), worldState))
```
AlanVerbner: `Seq.empty` ?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -292,12 +292,15 @@ object Ledger extends Logger {
   private def payForGasUsedToBeneficiary(stx: SignedTransaction, worldStateProxy: InMemoryWorldStateProxy): InMemoryWorldStateProxy = worldStateProxy //TODO
 
   /**
-    * Delete all accounts (that appear in SUICIDE list). YP eq (78)
+    * Delete all accounts (that appear in SUICIDE list). YP eq (78).
+    * The contract storage should be cleared during pruning as nodes could be used in other tries.
+    * The contract code is also not deleted as it could be duplicated between different contracts. FIXME: Should we keep track of this for deletion?
```
AlanVerbner: Probably while pruning? Maybe we can update the comment to state that deleting this is risky if other account is using the same code?

PR #131, opened 2017-04-06
[EC-152] Test TX execution with predefined blocks

adamsmo:
This PR is about creating tools to test transaction execution.
It should allow to create fixtures based on executed transactions and provide mocks for Ledger
blocks / receipts / nodes are encoded with RLP and stored in text files in hex format with their hashes

* commands/configuration for running parity with ethminer to prepare private network are in  `private_chain_conf`
* contract code that was used as example contract for dumping transactions with contract interactions is in `private_chain_conf/contract.sol`
* private keys for wallet used in example are in `private_chain_conf/keys`
* application for dumping chain parts is in `src/test/scala/io/iohk/ethereum/evm/util/DumpChainApp.scala` it loads configuration from `src/test/resources/chainDump.conf`
* dumped chain is stored in files, chain for unit tests from `src/test/scala/io/iohk/ethereum/evm/ContractTest.scala` is stored in `src/test/resources/evm_test/purchaseContruct/`

`src/test/scala/io/iohk/ethereum/evm/ContractTest.scala` contains example test that can be executed using this PR


private_chain_conf/contract.sol
```diff
@@ -0,0 +1,83 @@
...
+
+    function Purchase() payable {
+        seller = msg.sender;
+        value = msg.value / 2;
+        if (2 * value != msg.value) throw;
+    }
+
+    modifier require(bool _condition) {
+        if (!_condition) throw;
+        _;
```
rtkaczyk: This is valid solidity? What does it mean?

AlanVerbner: http://solidity.readthedocs.io/en/develop/contracts.html#function-modifiers

> Modifiers can be used to easily change the behaviour of functions, for example to automatically check a condition prior to executing the function. They are inheritable properties of contracts and may be overridden by derived contracts.

[...]

> The function body is inserted where the special symbol "_;" in the definition of a modifier appears.

rtkaczyk:
@adamsmo could you please provide some summary of what has been done here and how to use it?

rtkaczyk:
A couple of words about organisation/structure:
- first of all these are supposed to be TX execution tests, not EVM tests. Having `evm` appear in names is inappropriate and may be confused with `src/evmTest`
- should they be unit tests? `it` seems more appropriate.
- `private_chain_conf` at the top level? is that meant to be used by something other than this test?
- if you're going to use a shared `resources` folder, I would suggest putting all resources relevant to these tests in a subfolder
- I think this functionality/tool really deserves a name. It was hard to even write this note without it. My mind is currently blank though.

src/it/resources/transactionTest/documentation/readme.md
```diff
@@ -0,0 +1,29 @@
+command for running parity and ethminer
+
+```
+parity -jw --jsonrpc-interface 127.0.0.1 --jsonrpc-port 8545 --author 004A44Da31Bb735cB20171AfF69E60E3e50d26De --identity parity_node --geth --chain /Users/adam/iohk/etc-client/private_chain_conf/prv.json --base-path /Users/adam/etc_prv  --pruning archive
```
ntallar: Minor detail: What if our user is not adam? :stuck_out_tongue_winking_eye:

src/it/scala/io/iohk/ethereum/transactionTest/ContractTest.scala
```diff
@@ -0,0 +1,40 @@
...
+    val fixtures: FixtureProvider.Fixture = FixtureProvider.loadFixtures("/transactionTest/purchaseContruct/")
+
+    val (storage, stateStorage) = FixtureProvider.prepareStorages(0, fixtures)
+
+    //block only with ether transfers
+    Ledger.executeBlock(fixtures.blockByNumber(1), storage, stateStorage)
+
+    //block 2 contains contract creation
+    //block 3 and 4 contains contract calls with contract storage modification
+    //block 4 contains contract paying to 2 accounts
```
ntallar: As all this blocks apart from block 4 are executed in other tests, should we remove this comments from here? Should we also add some test for block 4?

src/it/scala/io/iohk/ethereum/transactionTest/ContractTest.scala
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  it should "deploy contract" in {
+    val fixtures: FixtureProvider.Fixture = FixtureProvider.loadFixtures("/transactionTest/purchaseContruct/")
+
+    val (storage, stateStorage) = FixtureProvider.prepareStorages(1, fixtures)
+
+    //contract creation
+    Ledger.executeBlock(fixtures.blockByNumber(2), storage, stateStorage)
```
ntallar: As `Ledger.executeBlock` doesn't throw an exception now, should we check that the result of this function is `Right(())`? I just detected that this test currently fails as the execution returns a `ValidationAfterExec` error.

src/it/scala/io/iohk/ethereum/transactionTest/ContractTest.scala
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  ignore should "execute contract call" in {
```
ntallar: Why does this test has a `ignore`? At least right now (maybe it changed with the latest PRs merged) the execution of the block seems to work properly, should we not ignore it now?

src/it/scala/io/iohk/ethereum/transactionTest/util/FixtureProvider.scala
```diff
@@ -0,0 +1,140 @@
...
+        case Some(m: MptBranch) =>
+          storages.mptNodeStorage.put(m)
+          stateStorage.put(m.hash, encode(m: MptNode)(MptNode.rlpEncDec))
+          m.children.collect { case Left(MptHash(hash)) if hash.nonEmpty => hash }.foreach(traverse)
+
+        case Some(m: MptExtension) =>
+          storages.mptNodeStorage.put(m)
+          stateStorage.put(m.hash, encode(m: MptNode)(MptNode.rlpEncDec))
+          m.child match {
+            case Left(MptHash(hash)) if hash.nonEmpty => traverse(hash)
```
ntallar: Should we also consider here the case where `m.child` is a `Right`?

src/it/scala/io/iohk/ethereum/transactionTest/ContractTest.scala
ntallar: BTW, I just checked and this test's bug is the one that's is fixed in #139

src/it/scala/io/iohk/ethereum/transactionTest/util/FixtureProvider.scala
```diff
@@ -0,0 +1,140 @@
...
+      }
+
+      traverse(block.header.stateRoot)
+    }
+
+    (storages, stateStorage)
+  }
+
+  def loadFixtures(path: String): Fixture = {
+    val bodies: Map[ByteString, BlockBody] = Source.fromFile(getClass.getResource(s"$path/bodies.txt").getPath).getLines()
```
ntallar: Minor detail: Should we close the `Source` opened here and in this function?

src/it/scala/io/iohk/ethereum/transactionTest/util/FixtureProvider.scala
adamsmo: right is value so it is already there and was saved in file with the node

src/it/scala/io/iohk/ethereum/transactionTest/ContractTest.scala
adamsmo: yes I will add checking if it is `Right(())`, and mark this as ignored to be able to merge, when #139 is merged than we can unignore this test

or should I commit failing test @mcsherrylabs ?

src/it/scala/io/iohk/ethereum/transactionTest/util/FixtureProvider.scala
```diff
@@ -86,55 +88,57 @@ object FixtureProvider {
...
+    }.toMap)
 
     val blocks = headers.toList.sortBy { case (hash, _) => Hex.toHexString(hash.toArray[Byte]) }.map { case (_, header) => header }
       .zip(bodies.toList.sortBy { case (hash, _) => Hex.toHexString(hash.toArray[Byte]) }.map { case (_, body) => body })
       .map { case (h, b) => Block(h, b) }
 
     Fixture(blocks.map(b => b.header.number -> b).toMap, blocks.map(b => b.header.hash -> b).toMap, headers, bodies, receipts, stateTree, contractTrees, evmCode)
   }
+
+  private def withClose[A, B <: Closeable] (closeable: B) (f: B => A): A = try { f(closeable) } finally { closeable.close() }
```
ntallar: :+1:

src/it/scala/io/iohk/ethereum/transactionTest/util/FixtureProvider.scala
ntallar: You are right :+1: But shouldn't we consider that case here so that this pattern matching considers all the cases and then the matching is exhaustive?

adamsmo: you are right, will change this

adamsmo: done https://github.com/input-output-hk/etc-client/pull/131/files#diff-fdb17fe23efd4d165a4582fcdff22666R67

src/it/scala/io/iohk/ethereum/transactionTest/util/FixtureProvider.scala
```diff
@@ -0,0 +1,145 @@
...
+
+    val blocksToInclude = fixtures.blockByNumber.toSeq.sortBy { case (number, _) => number }.takeWhile { case (number, _) => number <= blockNumber }
+
+    blocksToInclude.foreach { case (_, block) =>
+      storages.blockBodiesStorage.put(block.header.hash, fixtures.blockBodies(block.header.hash))
+      storages.blockHeadersStorage.put(block.header.hash, fixtures.blockHeaders(block.header.hash))
+      storages.blockNumberMappingStorage.put(block.header.number, block.header.hash)
+      fixtures.receipts.get(block.header.hash).foreach(r => storages.receiptStorage.put(block.header.hash, r))
+
+      //storages.totalDifficultyStorage
```
ntallar: Minor detail: Should we remove this or maybe its needed to insert data into this storage?

adamsmo: Not sure but I think we should remove it because it is used more for conflict resolution in chain. AFAIK it is not required for tx execution

src/it/scala/io/iohk/ethereum/transactionTest/util/DumpChainActor.scala
```diff
@@ -0,0 +1,174 @@
...
+      val mptRoots: Seq[ByteString] = m.headers.map(_.stateRoot)
+
+      blockHeadersHashes = m.headers.map(_.hash)
+      m.headers.foreach { h =>
+        blockHeadersStorage = blockHeadersStorage + (h.hash -> h)
+      }
+
+      peers.headOption.foreach { case Peer(_, actor) =>
+        actor ! SendMessage(GetBlockBodies(headerHashes))
+        actor ! SendMessage(GetReceipts(headerHashes.drop(1)))
```
ntallar: Minor detail: Why is `headerHashes.drop(1)` needed? Is it for not asking for the receipts for the genesis block? As `startBlock` is configurable, should we take into account the case where there is no genesis block to drop from `headerHashes`?

src/it/scala/io/iohk/ethereum/transactionTest/util/DumpChainActor.scala
```diff
@@ -0,0 +1,174 @@
...
+      }
+
+      val cNodes = NodeData(contractNodes).values.indices.map(i => NodeData(contractNodes).getMptNode(i))
+      cChildren = cChildren ++ cNodes.flatMap {
+        case n: MptBranch => n.children.collect { case Left(h: MptHash) if h.hash.nonEmpty => h.hash }
+        case MptExtension(_, Left(h)) => Seq(h.hash)
+        case _ => Seq.empty
+      }
+      peers.headOption.foreach { case Peer(_, actor) =>
+        actor ! SendMessage(GetNodeData(children ++ cChildren ++ evmTorequest))
```
ntallar: Minor detail: Should we check whether there is any node to ask for before asking for it? I tried using our dumper with Geth (which seems to be working :smile:) and it got in an infinite request for no data with `GetNodeData`

ntallar:
Apart for my last comments, LGTM!

src/it/scala/io/iohk/ethereum/transactionTest/ContractTest.scala
```diff
@@ -0,0 +1,47 @@
+package io.iohk.ethereum.transactionTest
```
rtkaczyk: Let's stick to [naming conventions](http://stackoverflow.com/questions/12534302/why-should-java-package-name-be-lowercase)

Also could the name refer to execution? Maybe `TxExecTest` / `txexectest`? Those test are about transaction execution, not transaction signing for example.

src/it/scala/io/iohk/ethereum/transactionTest/util/DumpChainApp.scala
```diff
@@ -0,0 +1,98 @@
...
+import io.iohk.ethereum.network.{PeerManagerActor, loadAsymmetricCipherKeyPair}
+import io.iohk.ethereum.utils.{Config, NodeStatus, ServerStatus}
+import org.spongycastle.util.encoders.Hex
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration.FiniteDuration
+
+object DumpChainApp {
+  // scalastyle:off
+  def main(args: Array[String]): Unit = {
```
rtkaczyk: Is there a reason we avoid `extends App`?

src/it/scala/io/iohk/ethereum/transactionTest/util/DumpChainApp.scala
```diff
@@ -0,0 +1,98 @@
...
+import io.iohk.ethereum.network.p2p.messages.{PV62, PV63}
+import io.iohk.ethereum.network.{PeerManagerActor, loadAsymmetricCipherKeyPair}
+import io.iohk.ethereum.utils.{Config, NodeStatus, ServerStatus}
+import org.spongycastle.util.encoders.Hex
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration.FiniteDuration
+
+object DumpChainApp {
+  // scalastyle:off
```
rtkaczyk: Does this have any effect? Are we checking `it` code with scalastyle?
If it does, I'd prefer too see specific rules turned off or tweaked in the xml, rather than `// scalastyle:off`

src/main/scala/io/iohk/ethereum/domain/Account.scala
```diff
@@ -30,4 +31,9 @@ case class Account(
 
   def withStorage(storageRoot: ByteString): Account =
     copy(storageRoot = storageRoot)
+
+  override def toString: String =
+    s"Account(nonce: $nonce, balance: $balance, " +
+      s"storageRoot: ${Hex.toHexString(storageRoot.toArray[Byte])}, codeHash: ${Hex.toHexString(codeHash.toArray[Byte])})"
```
rtkaczyk: Did you actually need that or just added it as a convention?

src/it/scala/io/iohk/ethereum/transactionTest/util/DumpChainApp.scala
adamsmo: it affects intellij - code is no longer all red, will remove it

src/main/scala/io/iohk/ethereum/domain/Account.scala
adamsmo: I added this to be able to easily debug state MPT

src/it/scala/io/iohk/ethereum/transactionTest/util/DumpChainApp.scala
rtkaczyk: > code is no longer all red

Why? Because you moved it to `it`?
@mcsherrylabs should we check `it` code with scalastyle?

adamsmo: no, I mean it affects coloring in intellij editor, it ignores rules and highlights test code with rules for application code

src/main/scala/io/iohk/ethereum/domain/Account.scala
rtkaczyk: So I thought. The number of `Hex.toHexString(something.toArray[Byte])` is appalling. It would be nice to have our own type for `ByteString` and `Array[Byte]`, even if it's just a wrapper. 

But that's nothing that relates to this PR. Just complaining / throwing ideas.

adamsmo: yes it would, it would also allow as to easily swap aka ByteString to something else

mcsherrylabs: (Refactor is coming post milestone 2, Hex.toHexStting can be implicit value class, ByteString might become scorex implementation of same ...)

PR #132, opened 2017-04-07
Fixed bugs in tx execution

ntallar:
## Description

Includes fixes to several issues that where detected during tx execution, this includes:
* Before running `validateBlockAfterExecution` we persist the trie (so as to obtain the correct state root hash).
* The state trie is now a secure trie (the keys are hashed before inserting them).
* Fixing how the sender is recovered from a transaction (in https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/crypto/ECKey.java#L439 the head of the public key is dropped).
* Fixed in `BlockValidator.validateLogBloom` how we handle the case where there are no receipts.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
@@ -75,6 +75,9 @@ object SignedTransaction {
 
     for {
       key <- recoveredPublicKey
+
+      //TODO: Parity doesn't do this (https://github.com/paritytech/parity/blob/master/ethkey/src/keypair.rs#L23-L28)
+      //      more research is needed on why `key.tail` is needed
       addrBytes = crypto.kec256(key.tail).slice(FirstByteOfAddress, LastByteOfAddress)
       if addrBytes.length == Address.Length
     } yield Address(addrBytes)
--- a/src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
@@ -75,7 +75,7 @@ object SignedTransaction {
 
     for {
       key <- recoveredPublicKey
-      addrBytes = crypto.kec256(key).slice(FirstByteOfAddress, LastByteOfAddress)
+      addrBytes = crypto.kec256(key.tail).slice(FirstByteOfAddress, LastByteOfAddress)
       if addrBytes.length == Address.Length
     } yield Address(addrBytes)
   }
--- a/src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala
+++ b/src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala
@@ -33,12 +33,6 @@ object InMemoryWorldStateProxy {
     new InMemoryWorldStateProxy(stateStorage, accountsStateTrieProxy, Map.empty, storages.evmCodeStorage, Map.empty, getBlockHashByNumber)
   }
 
-  def persistIfHashMatches(stateRootHash: ByteString, worldState: InMemoryWorldStateProxy): InMemoryWorldStateProxy = {
-    val commited = persistState(worldState)
-    if (commited.stateRootHash == stateRootHash) commited
-    else worldState
-  }
-
   /**
     * Updates state trie with current changes but does not persist them into the storages. To do so it:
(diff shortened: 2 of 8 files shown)


rtkaczyk:
> We now start executing the block from its parent's state trie.

This was already fixed in #126 

> The state trie is now a secure trie (the keys are hashed before inserting them).

Could you explain that a bit more?

> Fixing how the sender is recovered from a transaction

Huh, that's strange... could it be related to byte representation of a BigInteger?

ntallar:
> This was already fixed in #126

I had to add it to run regular sync with execution, but you are right, I'll remove it from here.
> Could you explain that a bit more?

Both the state trie and the storage tries use a `SecureTrie` (as called in EthereumJ and other clients) which uses a trie but before, operating with it, the keys are hashed with kec256 (there is a brief explanation on why this was done this way in https://github.com/ethereum/wiki/wiki/Design-Rationale#merkle-patricia-trees). So instead of inserting `key -> value` we insert `kec256(key) -> value` to the trie.
> Huh, that's strange... could it be related to byte representation of a BigInteger?

That would explain this but I actually don't know

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -75,7 +75,7 @@ object SignedTransaction {
 
     for {
       key <- recoveredPublicKey
-      addrBytes = crypto.kec256(key).slice(FirstByteOfAddress, LastByteOfAddress)
+      addrBytes = crypto.kec256(key.tail).slice(FirstByteOfAddress, LastByteOfAddress)
```
rtkaczyk: As much as I can read rust, parity doesn't seem to be doing this: https://github.com/paritytech/parity/blob/master/ethkey/src/keypair.rs#L23-L28

I'd rather suspect the `key` is not in fact the public key having an extra byte. For now, could you add TODO comment reminding to investigate it further?

src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala
```diff
@@ -82,7 +76,7 @@ object InMemoryWorldStateProxy {
 
   /**
     * Returns an [[InMemorySimpleMapProxy]] of the accounts state trie "The world state (state), is a mapping
-    * between addresses (160-bit identifiers) and account states (a data structure serialised as RLP [...]).
+    * between Keccak 256-bit hashes of the addresses (160-bit identifiers) and account states (a data structure serialised as RLP [...]).
```
AlanVerbner: :cry:

src/test/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxySpec.scala
```diff
@@ -9,6 +9,12 @@ import org.spongycastle.util.encoders.Hex
 
 class InMemoryWorldStateProxySpec extends FlatSpec with Matchers {
 
+  def persistIfHashMatches(stateRootHash: ByteString, worldState: InMemoryWorldStateProxy): InMemoryWorldStateProxy = {
```
AlanVerbner: @ntallar I would remove this and just call persist in tests. Actually, if you just persist the test works too.

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -77,7 +77,10 @@ object SignedTransaction {
 
     for {
       key <- recoveredPublicKey
-      addrBytes = crypto.kec256(key).slice(FirstByteOfAddress, LastByteOfAddress)
+
+      //TODO: Parity doesn't do this (https://github.com/paritytech/parity/blob/master/ethkey/src/keypair.rs#L23-L28)
+      //      more research is needed on why `key.tail` is needed
+      addrBytes = crypto.kec256(key.tail).slice(FirstByteOfAddress, LastByteOfAddress)
```
adamsmo: 👍 this is needed, I checked with data dumped from parity and only with this change recovered addresses matches original ones

src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala
adamsmo: 👌

adamsmo:
LGTM

PR #133, opened 2017-04-07
[EC-150] Link tx execution with regular sync

ntallar:
## Description

Includes executing the blocks received during regular sync and handling the case when execution fails.
We are currently not able to always resume the regular sync after a failure due to the removal of nodes in `MPT` (i.e. if validation of a block after execution fails, we already removed the root of the previous `MPT` so we can't resume regular sync).
This PR is dependant on all the fixes from ~~#132~~ and on also on the fix to get the `initialWorldState` from ~~#126~~.


ntallar:
Using this implementation we are currently able to execute all the blocks till 46147 which is the first one that has a tx. This PR is WIP as the current failure is (at least partially) related to functionality not implemented but with open PR.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -156,13 +156,45 @@ trait RegularSync {
...
+    * Inserts and executes the transactions of the blocks, up to the point to which one of them fails (or we run out of blocks).
+    * If the execution of any block were to fail, newBlocks only contains the NewBlock msgs for all the blocks executed before it.
+    *
+    * @param blocks to execute
+    * @param blockParentTd, td of the parent of the blocks.head block
+    * @param newBlocks which, after adding the corresponding NewBlock msg for blocks, will be broadcasted
+    * @return list of NewBlocks to broadcast (one per block successfully executed) and an error if one happened during execution
+    */
+  @tailrec
+  private def insertBlocks(blocks: Seq[Block], blockParentTd: BigInt,
```
LukasGasior1: How much time does it take to process a block? Just a suggestion: can we make that asynchronous and still keep asking other peers for next blocks in parallel? Maybe have something like a queue of blocks to process (maybe even separate actor)?

LukasGasior1: btw can we rename this method to `processBlocks` or something similar, it does not only insert (and it does not always insert)

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -42,27 +40,27 @@ class Ledger(vm: VM) extends Logger {
...
     * @param storages
-    * @param stateStorage
+    * @param validators
     */
   private def executeBlockTransactions(
     block: Block,
     blockchain: Blockchain,
     storages: BlockchainStorages,
-    stateStorage: NodeStorage):
+    validators: Validators):
```
LukasGasior1: This looks like something that we could put inside the config I am working on. If we do that, we won't have to check for homestead block in signedTransactionValidator anymore (we could just use different instances in configs for different blocks). Do you think it makes sense?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
ntallar: I haven't formally checked that but it does seem to take longer to run regular sync with tx execution enabled than not so it should be worth taking a look at it. Shall we create a task for checking that? As this PR already contains changes to many files I would prefer to take a look at it in another PR

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
ntallar: That makes a lot of sense :+1: I did this as it was previously done with the `blockValidator` function (in `SyncController`) only for simpler testing, but it seems it would also work for the config you are working in :smile:

src/main/scala/io/iohk/ethereum/validators/Validators.scala
```diff
@@ -0,0 +1,19 @@
...
+}
+
+object ValidatorsImpl extends Validators {
+
+  override val blockValidator: BlockValidator = BlockValidator
+  override val blockHeaderValidator: BlockHeaderValidator = BlockHeaderValidator
+  override val ommersValidator: OmmersValidator = OmmersValidator
+  override val signedTransactionValidator: SignedTransactionValidator = SignedTransactionValidator
+
+}
```
mcsherrylabs: I would think this can be moved to 'NodeBuilder'. 
```ValidatorsBuilder {
 val validators = new Validators { val blockValidator = BlockValidator ... } ```
Don't hold up the merge, but if you've got to do one more commit, please consider moving this...

**Comment by ntallar** [comment]
The current errors got during execution of the etc blockchain are:
* Block 46000~: State not being rollbacked if a OutOfGas exception happens when calling vm.run (fixed in https://github.com/input-output-hk/etc-client/pull/134)
* Block 47205: https://iohk.myjetbrains.com/youtrack/issue/EC-159
* Block 48462: https://github.com/input-output-hk/etc-client/pull/137
* Block 49018: We currently are creating contracts according to the rules after homestead (as mentioned in https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.mediawiki note 3.).
* Block 49157: ...

File: src/main/scala/io/iohk/ethereum/ledger/Ledger.scala (hunk 2 above)
Comment (ntallar) [inline src/main/scala/io/iohk/ethereum/ledger/Ledger.scala]: @LukasGasior1 should we create a separate task to work on that after this PR gets merged?

Comment (LukasGasior1) [inline src/main/scala/io/iohk/ethereum/ledger/Ledger.scala]: Yes, let's create a task for that.

File: src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala [hunk 4]
```diff
@@ -112,31 +115,28 @@ trait RegularSync {
 
   private def handleBlockBodies(peer: ActorRef, m: Seq[BlockBody]) = {
     if (m.nonEmpty) {
-      val result = headersQueue.zip(m).map { case (h, b) => blockValidator(h, b) }
+      val result = headersQueue.zip(m).map { case (h, b) => validators.blockValidator.validateHeaderAndBody(h, b) }
```
AlanVerbner: I think we are running the same checks within the Ledger as one of the first steps before executing the transactions, so...if most of the blocks are ok we are running this twice most of the times. I won't tag this as a blocker for merging tho.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -112,31 +115,28 @@ trait RegularSync {
...
-            context.self ! BroadcastBlocks(newBlocks)
-            log.info(s"got new blocks up till block: ${newBlocks.last.block.header.number} " +
-              s"with hash ${Hex.toHexString(newBlocks.last.block.header.hash.toArray[Byte])}")
+            if(error.isDefined){
+              val numberBlockFailed = blocks.head.header.number + newBlocks.length
+              resumeWithDifferentPeer(peer, reason = s"a block execution error: ${error.get.toString}, in block $numberBlockFailed")
+            }
           case None =>
-            log.error("no total difficulty for latest block")
+            log.error("no total difficulty for latest block") //FIXME: How do we handle this error on our blockchain?
```
AlanVerbner: Should we address this in the current PR? Maybe we need to create a task in order to properly handle this error (which might indicate data corruption, right?)

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -45,25 +44,26 @@ class Ledger(vm: VM) extends Logger {
...
   Either[BlockExecutionError, BlockResult] = {
     val parentStateRoot = blockchain.getBlockHeaderByHash(block.header.parentHash).map(_.stateRoot)
-    val initialWorld = InMemoryWorldStateProxy(storages, stateStorage, parentStateRoot)
+    val initialWorld = InMemoryWorldStateProxy(storages, storages.nodeStorage, parentStateRoot)
 
     val config = EvmConfig.forBlock(block.header.number)
 
-    log.debug(s"About to execute txs from block ${block.header}")
-    val blockTxsExecResult = executeTransactions(block.body.transactionList, initialWorld, block.header, config)
+    log.debug(s"About to execute ${block.body.transactionList.size} txs from block ${Hex.toHexString(block.header.hash.toArray)}")
```
AlanVerbner: As it seems we will print this in several places, maybe it's a good idea to create `toHexString` method within the block, what do you think?

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
```diff
@@ -332,6 +342,123 @@ class SyncControllerSpec extends FlatSpec with Matchers {
     peer4.expectMsg(PeerActor.Unsubscribe)
   }
 
+  it should "broadcast all blocks if they were all valid" in new TestSetup {
```
AlanVerbner: This test is calling the Ledger in order to validate the block so if by any reason there is a bug within the Ledger, this test fail even if SyncController logic is ok. imo we have two options:

1. Move this test to it
2. Stub ledger execution in this test which might involve in being able to configure Ledger for SyncController

Maybe this is work for another task ( @mcsherrylabs, any thoughts on this? )

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
AlanVerbner: Super minor detail. I think it's helpful to also print block number here.

ntallar: I agree, I'll add a `hashAsHexString` method in `BlockHeader` :+1:

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
ntallar: I agree :+1: I think it should only be caused by data corruption (or by a bug), I'll create a task to handle this error

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
mcsherrylabs: @AlanVerbner Looks like the test is in the right place. We should be able to stub ledger execution.

PR #134, opened 2017-04-10
[EC-151] Block execution correct error handling

ntallar:
## Description

Includes the correct handling of the errors that are raised during block execution and the removal of throwing exceptions when that were to happen.
This also fixes:
 * In case an error during execution happens, no account is deleted and the world state is rollbacked to its previous state


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -50,18 +53,38 @@ class Ledger(vm: VM) extends Logger {
...
     val initialWorld = InMemoryWorldStateProxy(storages, stateStorage, parentStateRoot)
 
-    block.body.transactionList.foldLeft[BlockResult](BlockResult(worldState = initialWorld)) {
-      case (BlockResult(world, acumGas, receipts), stx)=>
-        validateTransaction(stx, world, acumGas, block.header) match {
-          case Left(err) =>
-            throw new RuntimeException(err)
+    log.debug(s"About to execute txs from block ${block.header}")
+    val blockTxsExecResult = executeTransactions(block.body.transactionList, initialWorld, block.header)
+    log.debug(s"All txs from block ${block.header} were executed")
```
LukasGasior1: actually we are not sure __all__ txs have been executed, are we?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -271,7 +294,9 @@ class Ledger(vm: VM) extends Logger {
   private def runVM(stx: SignedTransaction, blockHeader: BlockHeader, worldStateProxy: InMemoryWorldStateProxy): PR = {
     val context: PC = ProgramContext(stx, blockHeader, worldStateProxy)
     val result = vm.run(context)
-    if (stx.tx.isContractInit && result.error.isEmpty)
+    if(result.error.isDefined)
+      result.copy(world = worldStateProxy, addressesToDelete = Nil) //Rollback to the previous state if an error happened
+    else if (stx.tx.isContractInit && result.error.isEmpty)
```
LukasGasior1: `result.error.isEmpty` isn't it always true here?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
ntallar: You are right, I'll update the logging to clarify that :+1:

ntallar:
@LukasGasior1 I fixed your comments and updated the logging, please take a look at it and let me know if there's something else to fix.

ntallar:
@rtkaczyk @LukasGasior1 I just fixed a small bug I detected in how the rollback was done, in case the `OutOfGasError` happened in `saveNewContract` the world state wasn't rollbacked

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -70,38 +96,46 @@ class Ledger(vm: VM) extends Logger {
...
   private[ledger] def executeTransaction(stx: SignedTransaction, blockHeader: BlockHeader, world: InMemoryWorldStateProxy): TxResult = {
     val gasPrice = UInt256(stx.tx.gasPrice)
     val gasLimit = UInt256(stx.tx.gasLimit)
 
-    val world1 = updateSenderAccountBeforeExecution(stx, world)
-    val result = runVM(stx, blockHeader, world1)
+    val worldBeforeTransfer = updateSenderAccountBeforeExecution(stx, world)
+    val result = runVM(stx, blockHeader, worldBeforeTransfer)
+
+    //FIXME: This only implements error handling as done in Homestead
```
rtkaczyk: What is needed for pre-homestead?

ntallar: We should create an empty contract in case an `OutOfGas` exception is produced in `saveNewContract` (as mentioned in https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.mediawiki note 3.). This currently causes tx execution to fail in block 49018.

rtkaczyk: Oh man... yes, that's so... ugh...

PR #135, opened 2017-04-11
Fixed more bugs in tx execution

ntallar:
Includes fixes to the bugs:
* Both `Ledger.updateSenderAccountBeforeExecution` and `ProgramContext.callOrCreate` decreased tx.value from sender of a tx in the state world so the sender's balance was decreased by `tx.value` twice
* In case a tx was for contract creation (so `receivingAddress` was empty) the `recevingAdress` was padded with zeros (resulting in the address `0x00000000000000000000` instead of an empty one). This changed the result of encoding the `SignedTransaction`for `BlockValidator.validateTransactionRoot`. The `tx.recevingAddress` was changed to an `Option[Address]` with `None` signifying that the tx is for contract creation.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/domain/Transaction.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/Transaction.scala
@@ -10,13 +10,16 @@ object Transaction {
   val GasLength = 32
   val ValueLength = 32
 
+  def apply(nonce: BigInt, gasPrice: BigInt, gasLimit: BigInt, receivingAddress: Address, value: BigInt, payload: ByteString): Transaction =
+    Transaction(nonce, gasPrice, gasLimit, Some(receivingAddress), value, payload)
+
 }
 
 case class Transaction(
   nonce: BigInt,
   gasPrice: BigInt,
   gasLimit: BigInt,
-  receivingAddress: Address,
+  receivingAddress: Option[Address],
   value: BigInt,
   payload: ByteString) {
 
@@ -27,7 +30,7 @@ case class Transaction(
          |nonce: $nonce
          |gasPrice: $gasPrice
          |gasLimit: $gasLimit
-         |receivingAddress: ${Hex.toHexString(receivingAddress.toArray)}
+         |receivingAddress: ${if(receivingAddress.isDefined) Hex.toHexString(receivingAddress.get.toArray) else "[Contract creation]"}
          |value: $value wei
          |payload: ${if (isContractInit) "ContractInit: " else "TransactionData: "}${Hex.toHexString(payload.toArray[Byte])}
          |}""".stripMargin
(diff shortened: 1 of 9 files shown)


src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
```diff
@@ -30,10 +30,10 @@ object ProgramContext {
 
     } else {
       // message call
-      val world1 = world.transfer(senderAddress, tx.receivingAddress, UInt256(tx.value))
-      val code = world1.getCode(tx.receivingAddress)
+      val world1 = world.transfer(senderAddress, tx.receivingAddress.get, UInt256(tx.value))
```
LukasGasior1: how about
```
tx.receivingAddress match {
  case Some(address) => // message call
  case None => // contract create
}
```

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -249,7 +249,7 @@ class Ledger(vm: VM) extends Logger {
   private[ledger] def updateSenderAccountBeforeExecution(stx: SignedTransaction, worldStateProxy: InMemoryWorldStateProxy): InMemoryWorldStateProxy = {
     val senderAddress = stx.senderAddress
     val account = worldStateProxy.getGuaranteedAccount(senderAddress)
-    worldStateProxy.saveAccount(senderAddress, account.increaseBalance(-calculateUpfrontCost(stx.tx)).increaseNonce)
+    worldStateProxy.saveAccount(senderAddress, account.increaseBalance(-calculateUpfrontGas(stx.tx)).increaseNonce)
```
rtkaczyk: Is calculate `calculateUpfrontCost` still used?

I still think that `ProgramContext` should not be tasked with such updates (accounts balance). But we can leave, I don't want to focus prematurely on refactoring either.

ntallar: Yes, it is currently used in the `validateAccountHasEnoughGasToPayUpfrontCost` function to validate that the account has enough ether to pay for the tx execution. However I agree that it would be better to have the transfer not done in `ProgramContext` it also confused me that it was done that way.

src/main/scala/io/iohk/ethereum/domain/Transaction.scala
```diff
@@ -10,13 +10,16 @@ object Transaction {
...
+    Transaction(nonce, gasPrice, gasLimit, Some(receivingAddress), value, payload)
+
 }
 
 case class Transaction(
   nonce: BigInt,
   gasPrice: BigInt,
   gasLimit: BigInt,
-  receivingAddress: Address,
+  receivingAddress: Option[Address],
```
rtkaczyk: If we are going this way then we should remove `Address.empty` - it was supposed to handle the case of contract creation. Maybe an `Option` actually fits the domain better, but it would have been much easier to fix `Address.apply` to correctly produce `Address.empty`.

ntallar: Yes, both `Address.empty` and the `ìsEmpty` function where removed from `Address.scala` in this PR as they were no longer used :+1:

rtkaczyk: Ahh, 👓 :+1:

PR #136, opened 2017-04-11
[EC-92][EC-161] CREATE opcode

rtkaczyk:


src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -656,6 +656,70 @@ case object LOG3 extends LogOp(0xa3)
...
+      } else {
+        val stack2 = stack1.push(newAddress.toUInt256)
+        val world3 = result.world.saveCode(newAddress, result.returnData)
+        state
+          .withStack(stack2)
+          .withWorld(world3)
+          .withAddressesToDelete(result.addressesToDelete)
+          .withMemory(memory1)
+          .spendGas(gasUsed)
+          .step()
```
ntallar: Should the new state also have the logs that resulted from the execution of `initCode`? According to the YP it seems to be that way.

rtkaczyk: Nice catch! This was based on CALL, which also doesn't propagate the logs (it was created before logging was fully implemented). I will update both.

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -656,6 +656,70 @@ case object LOG3 extends LogOp(0xa3)
...
+      state.withStack(stack2).step()
+    } else {
+
+      val (initCode, memory1) = state.memory.load(inOffset, inSize)
+      val (newAddress, world1) = state.world.newAddress(state.env.ownerAddr)
+      val world2 = world1.transfer(state.env.ownerAddr, newAddress, endowment)
+
+      val newEnv = ExecEnv(
+        newAddress,
+        state.env.callerAddr,
```
ntallar: I'm wondering, could this have to be `state.env.ownerAddr`?
BTW, we should probably at least a comment in `ExecEnv` with the relation between `ownerAddr`, `callerAddr` and `originAddr` and the different `I_x` in the YP, that would make it easier to follow with the paper :stuck_out_tongue_winking_eye:

rtkaczyk: You're on 🔥 
Yes, it should be `ownerAddr` and it's good idea to add those symbols to the comments.

rtkaczyk:
So, as EC-102 was already merged, it turned that implementing the EIP changes was just too easy, so now this PR is for both EC-92 and EC-161

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -697,15 +697,22 @@ case object CREATE extends OpCode(0xf0, 3, 1, _.G_create) {
...
       val codeDepositGas = state.config.calcCodeDepositCost(result.returnData)
       val gasUsed = startGas - result.gasRemaining + codeDepositGas
+      val enoughGasForDeposit = gasUsed <= availableGas
 
       if (result.error.isDefined) {
         val stack2 = stack1.push(UInt256.Zero)
         state.withStack(stack2).spendGas(startGas)
-      } else if (gasUsed > availableGas) {
+
+      } else if (!enoughGasForDeposit && state.config.exceptionalFailedCodeDeposit) {
```
ntallar: I'm wondering, in this case is it okay that nothing is pushed to the stack? If I understood correctly, on https://github.com/ethereumproject/go-ethereum/blob/master/core/vm/instructions.go#L430 they push a `UInt256.Zero` to the stack

rtkaczyk: That doesn't make sense. In case of exceptional halting the contents of stack and all of the current machine state become irrelevant. 

In particular, whatever is pushed to the stack could only be used by a subsequent instruction, but there will be no subsequent instruction because we're halting with OOG.

Btw, I was looking at: https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L224 - ethereum non-classic. It's interesting how much these projects diverged.

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -697,15 +697,22 @@ case object CREATE extends OpCode(0xf0, 3, 1, _.G_create) {
...
         state.withError(OutOfGas)
+
       } else {
         val stack2 = stack1.push(newAddress.toUInt256)
-        val world3 = result.world.saveCode(newAddress, result.returnData)
+
+        val world3 =
+          if (!enoughGasForDeposit) result.world
```
ntallar: In this case, should we call `spendGas(gasUsed)` spending also the `codeDepositGas` if nothing is saved? At least in Geth (https://github.com/ethereumproject/go-ethereum/blob/master/core/execution.go#L125) it seems that the `codeDepositGas` is not subtracted from the gas in that case.

rtkaczyk: Dammit

ntallar: Yes :confused: It's quite difficult to find this details

rtkaczyk: Well, you're doing great 😄

rtkaczyk: Should be fixed now

src/main/scala/io/iohk/ethereum/vm/ExecEnv.scala
```diff
@@ -5,7 +5,15 @@ import io.iohk.ethereum.domain.{ Address, BlockHeader }
...
-  *  @param ownerAddr address of accounts that owns the code
+  *  @param ownerAddr   I_a: address of the account that owns the code
+  *  @param callerAddr  I_s: address of the account which caused the code to be executing
+  *  @param originAddr  I_o: sender address of the transaction that originated this execution
+  *  @param gasPrice    I_p
+  *  @param inputData   I_d
+  *  @param value       I_v
+  *  @param program     I_b
+  *  @param blockHeader I_H
+  *  @param callDepth   I_e
```
AlanVerbner: <3

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -663,8 +663,79 @@ case object LOG3 extends LogOp(0xa3)
...
+      val result = VM.run(context)
+
+      val codeDepositGas = state.config.calcCodeDepositCost(result.returnData)
+      val gasUsedInVm = startGas - result.gasRemaining
+      val totalGasRequired = gasUsedInVm + codeDepositGas
+      val enoughGasForDeposit = totalGasRequired <= availableGas
+
+      if (result.error.isDefined) {
+        val stack2 = stack1.push(UInt256.Zero)
+        state.withStack(stack2).spendGas(startGas)
```
AlanVerbner: I'm not sure why `startGas` and not `availableGas` (I'm not thinking about OOG exception). Can you point me out in YP where it's defined?

rtkaczyk: `availableGas` is the current machine gas minus the part that is spent [here](https://github.com/input-output-hk/etc-client/blob/master/src/main/scala/io/iohk/ethereum/vm/OpCode.scala#L178). This the gas available to execute this instruction and all the remaining program.

`startGas` is a fraction of `availableGas` and it is the maximum value that can be used by the initialisation code, which runs in the recursively spawned child VM. If that code runs correctly most likely less gas will be used, but if it results in an error we need to consume all of `startGas`.

To be exact `startGas == availableGas * 63/64`. In the YP it is defined in the `CREATE` opcode definition. See the 4th argument to the _Λ_ function, i.e. _L(μ<sub>g</sub>)_, and the _L_ function is defined in the eq (224).

Note that before EIP-150 it was `startGas == availableGas`.

AlanVerbner: ok. maybe what confused mem is that _L(μg)_ for me is like a _reserve_ to store the contract within this opcode so if that process fails, would make sense to spend it all.

PR #137, opened 2017-04-12
Fixed transfer in WorldState with same origin and destination

ntallar:
## Description

Includes a fix to the `transfer` function in `WorldState` for the case that the origin and the destination of the transfer is the same account. This for example happens in tx `0x6fd112af869acb860895a29c1ee55b8edcf1c3081ce13ab2e5aa171b931746e1` of block `48462`


src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
```diff
@@ -14,7 +14,7 @@ import io.iohk.ethereum.rlp.RLPList
   * should be kept in memory and applied only after a transaction completes without errors. This does not forbid mutable
   * caches for DB retrieval operations.
   */
-trait WorldStateProxy[WS <: WorldStateProxy[WS, S], S <: Storage[S]] {
+trait WorldStateProxy[WS <: WorldStateProxy[WS, S], S <: Storage[S]] { self: WS =>
```
rtkaczyk: Why this?

ntallar: It was needed so that `this: WS` in the `transfer` function. If compilation failed due to:

PR #139, opened 2017-04-12
[EC-159] Fixed nonce being increased twice in contract creation

ntallar:
## Description

Includes a fix to nonce being increased twice in contract creation (before it was increased both in `WorldState.newAddress` and `Ledger.updateSenderAccountBeforeExecution`).

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
@@ -675,7 +675,7 @@ case object CREATE extends OpCode(0xf0, 3, 1, _.G_create) {
     } else {
 
       val (initCode, memory1) = state.memory.load(inOffset, inSize)
-      val (newAddress, world1) = state.world.increaseCreatorNonceAndCreateAddress(state.env.ownerAddr)
+      val (newAddress, world1) = state.world.createAddressWithOpCode(state.env.ownerAddr)
       val world2 = world1.transfer(state.env.ownerAddr, newAddress, endowment)
 
       val newEnv = state.env.copy(
--- a/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
@@ -675,7 +675,7 @@ case object CREATE extends OpCode(0xf0, 3, 1, _.G_create) {
     } else {
 
       val (initCode, memory1) = state.memory.load(inOffset, inSize)
-      val (newAddress, world1) = state.world.newAddress(state.env.ownerAddr)
+      val (newAddress, world1) = state.world.increaseCreatorNonceAndCreateAddress(state.env.ownerAddr)
       val world2 = world1.transfer(state.env.ownerAddr, newAddress, endowment)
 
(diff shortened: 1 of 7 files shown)


src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -675,7 +675,7 @@ case object CREATE extends OpCode(0xf0, 3, 1, _.G_create) {
     } else {
 
       val (initCode, memory1) = state.memory.load(inOffset, inSize)
-      val (newAddress, world1) = state.world.newAddress(state.env.ownerAddr)
+      val (newAddress, world1) = state.world.increaseCreatorNonceAndCreateAddress(state.env.ownerAddr)
```
AlanVerbner: I'm not completely sure about this in terms of having a single function that does this. I would create an `increaseNonce` fn and then creating the address. That would look more similar to YP `CREATE` rule that states `σ ∗ ≡ σ except σ ∗ [I a ] n = σ[I a ] n + 1`.

But, I would like to stick to @rtkaczyk preference on this. @rtkaczyk are you ok with this?

AlanVerbner:
I've merged this into https://github.com/input-output-hk/etc-client/pull/133 and got until 49018! ... so, if @rtkaczyk approves, It's ok for me too

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
adamsmo: @ntallar it is necessary to perform this operations always together (I mean increase creator nonce and create address) if yes then maybe only name is misleading?

src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
```diff
@@ -54,11 +54,27 @@ trait WorldStateProxy[WS <: WorldStateProxy[WS, S], S <: Storage[S]] { self: WS
...
+    Address(hash)
+  }
+
+  /**
+    * Increases the creator's nonce and creates a new address based on the address and the new nonce of the creator
+    *
+    * @param creatorAddr, the address of the creator of the new address
+    * @return the new address and the state world after the creator's nonce was increased
+    */
+  def increaseCreatorNonceAndCreateAddress(creatorAddr: Address): (Address, WS) = {
```
adamsmo: so we are using this method when contract is created by op code and not by transaction resulting in contract creation, am I right?
so maybe `increaseCreatorNonceAndCreateAddress` => `createAddressWithOpCode` ?

ntallar: I agree, I'll change it to that :+1: An in any case we can rename it or divide it into 2 methods later

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
ntallar: Not always, this is not done when the tx is for contract creation (in which case the nonce is increased and the address created but in separate places). I agree that the name is misleading so I'll follow your name recommendation :+1:

adamsmo:
LGTM 👍

PR #140, opened 2017-04-13
[EC-156] Remove erasing of nodes in MPT

AlanVerbner:
## Description

This commit removes old mpt nodes deletion from underlying db. From now on, a pruning system needs to be performed in order to free space.

## Implementation note

Mpt now doesn't _know_ that some nodes are useless now. If we think our pruning implementation requires to record "deletes" - for example if it's made using reference counting method - this commit can be rolled back.

### DIFF
--- a/src/it/scala/io/iohk/ethereum/mpt/MerklePatriciaTreeIntegrationSuite.scala
+++ b/src/it/scala/io/iohk/ethereum/mpt/MerklePatriciaTreeIntegrationSuite.scala
@@ -1,11 +1,13 @@
 package io.iohk.ethereum.mpt
 
+import java.io.File
 import java.nio.ByteBuffer
+import java.nio.file.Files
 import java.security.MessageDigest
 
 import io.iohk.ethereum.ObjectGenerators
 import io.iohk.ethereum.crypto.kec256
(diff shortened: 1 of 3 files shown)


src/it/scala/io/iohk/ethereum/mpt/MerklePatriciaTreeIntegrationSuite.scala
```diff
@@ -1,11 +1,13 @@
 package io.iohk.ethereum.mpt
```
AlanVerbner: Sorry for the formatting changes on this file

adamsmo: maybe use diff to revert the changes, I mean formatting in places that you did not change code?

adamsmo:
besides formatting LGTM

src/it/scala/io/iohk/ethereum/mpt/MerklePatriciaTreeIntegrationSuite.scala
rtkaczyk: These formatting changes are the result of functional changes in the test code and they are required. He just feels sorry it makes a bit harder to review.

PR #141, opened 2017-04-13
Add BlockchainConfig

LukasGasior1:


src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
```diff
@@ -8,12 +8,16 @@ import io.iohk.ethereum.db.components.{SharedLevelDBDataSources, Storages}
...
 import io.iohk.ethereum.network.{PeerManagerActor, ServerActor}
 import io.iohk.ethereum.rpc.{JsonRpcServer, RpcServerConfig}
+import io.iohk.ethereum.utils.BlockchainConfig
 import io.iohk.ethereum.utils.{Config, NodeStatus, ServerStatus}
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import io.iohk.ethereum.network._
 import io.iohk.ethereum.validators.BlockValidator
 
+trait BlockchainConfigBuilder {
```
LukasGasior1: ideally, we should move all configs from objects to classes and here have a `ConfigBuilder` that would provide all needed config instances. (we can create a separate task for that)

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -267,15 +266,6 @@ class PeerActor(
       messageSubscribers = messageSubscribers.filterNot(_.ref == sender())
   }
 
-  def handlePeerChainCheck(rlpxConnection: RLPxConnection): Receive = {
```
LukasGasior1: I think, it's not needed anymore since we have `FastUploadHost`, is that right ( @ntallar ? )

ntallar: I think you are right, `FastSyncHost` could also be used for this request :+1:. However, as in our client we first ask for the fork block before fully connecting to a peer, shouldn't we able to respond to the same request for the fork block during this setup phase? For example, before the removal of this function (`handlePeerChainCheck`) we also used it in `waitingForForkHeader`, but we currently only handle the block requests (with `FastSyncHost`) in the `MessageHandler`.
Maybe we should also rename `FastSyncHost` if it's used outside of FastSync

adamsmo:
maybe we can add some tests to check if configuration changes affects gas calculation / other functionalities properly?

ntallar:
Apart from @adamsmo's comment and mine, LGTM!

src/it/scala/io/iohk/ethereum/rlp/RLPSpeedSuite.scala
```diff
@@ -80,7 +80,8 @@ class RLPSpeedSuite extends FunSuite
       payload = ByteString.empty),
     pointSign = 28,
     signatureRandom = ByteString(Hex.decode("cfe3ad31d6612f8d787c45f115cc5b43fb22bcc210b62ae71dc7cbf0a6bea8df")),
-    signature = ByteString(Hex.decode("57db8998114fae3c337e99dbd8573d4085691880f4576c6c1f6c5bbfe67d6cf0"))
+    signature = ByteString(Hex.decode("57db8998114fae3c337e99dbd8573d4085691880f4576c6c1f6c5bbfe67d6cf0")),
+    chainId = 0x3d.toByte
```
adamsmo: <3

rtkaczyk:
@LukasGasior1 could we have either a task or a description for this PR?

LukasGasior1:
@rtkaczyk this PR doesn't really change much. The only change is: `Config.Blockchain` is no longer a singleton object, instead we now pass it around as `BlockchainConfig`

LukasGasior1:
@adamsmo I'll add a tests for this functionality as integration tests

rtkaczyk:
@LukasGasior1 yes, I see that now. But a description/task is always useful, especially when you see 23 files changed after coming back from vacation 😉

PR #142, opened 2017-04-13
fix G_txcreate fees

rtkaczyk:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.mediawiki was misread. (21000 + 32000 = 53000)
Consult YP appendix G

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
@@ -160,7 +160,7 @@ object FeeSchedule {
       override val G_exp = 10
       override val G_expbyte = 10
       override val G_memory = 3
-      override val G_txcreate = 21000
+      override val G_txcreate = 0
       override val G_txdatazero = 4
       override val G_txdatanonzero = 68
       override val G_transaction = 21000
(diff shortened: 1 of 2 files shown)


src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -175,7 +175,7 @@ object FeeSchedule {
   }
 
   class HomesteadFeeSchedule extends FrontierFeeSchedule {
-    override val G_txcreate = 53000
+    override val G_txcreate = 32000
```
LukasGasior1: so what happens with 53000? parity seems to keep them in separate config entries: create_gas = 32000 and tx_create_gas which is 53000.

LukasGasior1:
Note, just to clarify:

The confusion arose from parity treating these values differently.
Here's how they do it:
```
if (is_create) tx_create_gas else tx_gas

// which is:
                frnt     hmst
if (is_create) (21000 or 53000) else 21000
```

what we do is:
```
(if (isContractCreation) G_txcreate else 0) + G_transaction
```

That's why this config entries need to be lower by 21000 (the G_transaction aka tx_gas) in comparison to parity.

I think the way we do it corresponds to YP more directly.

rtkaczyk:
@LukasGasior1 

> I think the way we do it corresponds to YP more directly.

Exactly, that's why we're better 😄

PR #143, opened 2017-04-17
Static analysis tools fixes

ntallar:
## Description

Includes several fixes to issues detected while running different static analysis tools.
The main problems fixed were:
* Removed unused method parameters.
* Parantheless call to methods with effect and returning unit.


LukasGasior1:
LGTM. It's surprising how often we leave no longer used method params :P

PR #144, opened 2017-04-18
Handle pre-Homestead contract creation when not enough gas

LukasGasior1:
We need this logic in two places: CREATE op code (already there) and Ledger (this PR).

### DIFF
--- a/src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
+++ b/src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
@@ -317,13 +317,16 @@ class Ledger(vm: VM) extends Logger {
 
   private def saveNewContract(address: Address, result: PR, config: EvmConfig): PR = {
     val codeDepositCost = config.calcCodeDepositCost(result.returnData)
-    if (result.gasRemaining < codeDepositCost)
-      result.copy(error = Some(OutOfGas))
-    else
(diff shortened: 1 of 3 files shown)


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -317,13 +317,16 @@ class Ledger(vm: VM) extends Logger {
 
   private def saveNewContract(address: Address, result: PR, config: EvmConfig): PR = {
```
AlanVerbner: The code looks good and i'm still running to see what block do I get. Do you think it's too much trouble to create a test to verify this if branches are ok?

AlanVerbner:
@ntallar there is a FIXME message in Ledger line 116, is it related to this fix?

LukasGasior1:
It seems related. I don't think there are any other changes in Homestead with regard to error handling

ntallar:
It is related, it was just a remainder that the fix from this PR was needed, so it can be removed :+1:

ntallar:
Apart from @AlanVerbner's comment on having tests for this, LGTM!

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
LukasGasior1: @AlanVerbner I've added a test

PR #147, opened 2017-04-19
Cumulative gas calculation fix

AlanVerbner:
## Description

At block #50111 there is some gas being refunded. It seems our calculation is wrong and because of that, block header validation after execution is returning an error.

## Fix

The cumulative gas was being calculated as `gasLimit - gasRemaining` but it the working version needs to be `gasLimit - gasRefund`. 

This leads to execution up to block #53145

_Note: That's not strictly stated in YP (at least I wasn't able to find that yet) and some reasearch is still being done with the help of @rtkaczyk_


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -136,7 +135,7 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
 
     val world2 = (refundGasFn andThen payMinerForGasFn andThen deleteAccountsFn andThen persistStateFn)(resultWithErrorHandling.world)
 
-    TxResult(world2, gasUsed, resultWithErrorHandling.logs)
+    TxResult(world2, gasLimit - gasRefund, resultWithErrorHandling.logs)
```
rtkaczyk: Did you mean `gasUsed - gasRefund`?

src/test/scala/io/iohk/ethereum/ledger/LedgerSpec.scala
```diff
@@ -43,15 +43,15 @@ class LedgerSpec extends FlatSpec with PropertyChecks with Matchers {
...
     val initialMinerBalance: UInt256 = 2000000
 
-    val table = Table[UInt256, UInt256, Option[ProgramError], UInt256](
-      ("gasUsed", "refundsFromVM", "maybeError", "balanceDelta"),
-      (25000, 20000, None, (25000 / 2) * defaultGasPrice),
-      (25000, 10000, None, (25000 - 10000) * 10),
-      (125000, 10000, Some(OutOfGas), defaultGasLimit * defaultGasPrice),
-      (125000, 100000, Some(OutOfGas), defaultGasLimit * defaultGasPrice)
+    val table = Table[UInt256, UInt256, Option[ProgramError], UInt256, BigInt](
+      ("execGasUsed", "refundsFromVM", "maybeError", "balanceDelta", "gasRefund"),
```
rtkaczyk: 4th and 5th column seems to be highly correlated.

src/test/scala/io/iohk/ethereum/ledger/LedgerSpec.scala
```diff
@@ -62,11 +62,13 @@ class LedgerSpec extends FlatSpec with PropertyChecks with Matchers {
...
 
-      val mockVM = new MockVM(createResult(_, gasUsed, defaultGasLimit, gasRefund, error))
+      val mockVM = new MockVM(createResult(_, execGasUsed, defaultGasLimit, gasRefundFromVM, error))
       val ledger = new LedgerImpl(mockVM, blockchainConfig)
 
-      val postTxWorld = ledger.executeTransaction(stx, header, initialWorld).worldState
+      val execResult = ledger.executeTransaction(stx, header, initialWorld)
+      val postTxWorld = execResult.worldState
 
+      execResult.gasUsed shouldEqual gasUsed
```
rtkaczyk: I would say this is a separate test

AlanVerbner: well... it seemed to me that too but as it's related to gas thought I might not be so different (a separate test would have almost the same code).

I will split it.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
AlanVerbner: I'm afraid not. It's late and maybe i'm not following it but check out this pls: https://github.com/paritytech/parity/blob/master/ethcore/src/executive.rs#L462

```rust
let gas_left_prerefund = match result { Ok(x) => x, _ => 0.into() };
let refunded = cmp::min(refunds_bound, (t.gas - gas_left_prerefund) >> 1);
let gas_left = gas_left_prerefund + refunded;

let gas_used = t.gas - gas_left;
let refund_value = gas_left * t.gas_price;
```

rtkaczyk: I see why I was confused: [this comment](https://github.com/input-output-hk/etc-client/blob/phase/3/txExecution/src/main/scala/io/iohk/ethereum/ledger/Ledger.scala#L346-L349) is inaccurate - it's the whole sum. Could you update it please?

Also the `gasRefund` name may be confusing - compare with `ProgramResult#gasRefund`. Maybe rename to `totalGasRefunded`. Also adding `val gasUsed = gasLimit - totalGasRefunded` could help readability, I think.

src/test/scala/io/iohk/ethereum/ledger/LedgerSpec.scala
rtkaczyk: It could stay if the spec description was updated...

src/test/scala/io/iohk/ethereum/ledger/LedgerSpec.scala
rtkaczyk: Also the 5th was named `gasRefund` but is matched as `gasUsed`.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
AlanVerbner: @rtkaczyk I think i've addressed your comments. I think the only one left is the one that states:

> this comment is inaccurate  - it's the whole sum. 

Can you rephrase pls?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -360,7 +359,7 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
     * Calculate gas refund
     * See YP, eq (72) - only the right addend
     */
-  private def calcGasRefund(stx: SignedTransaction, result: PR): UInt256 = {
+  private def calcTotalGasToRefund(stx: SignedTransaction, result: PR): UInt256 = {
```
rtkaczyk: please update the scaladoc comment, remove `only the right addend`

AlanVerbner: done

PR #148, opened 2017-04-19
Fixed regular sync headers request error fails

ntallar:
## Description

When execution of a block failed, due to a bug in `RegularSync`, two `ResumeRegularSync` were sent to the `RegularSync` actor:
* Due to [resuming download from another peer](https://github.com/input-output-hk/etc-client/blob/phase/3/txExecution/src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala#L132) 
* Due to [trying to resume download with the same peer](https://github.com/input-output-hk/etc-client/blob/phase/3/txExecution/src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala#L143)

This caused an exception when execution of a block failed (as is the case with bugs found), by sending the request for a block header twice to the same peer which later caused doing `head` of an empty list in https://github.com/input-output-hk/etc-client/blob/phase/3/txExecution/src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala#L119

## Fix

We should only try to resume download from the same peer in case of correct the execution of the blocks received.


src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -130,18 +130,19 @@ trait RegularSync {
...
+            headersQueue = headersQueue.drop(blocks.length)
+            if (headersQueue.nonEmpty) {
+              val hashes = headersQueue.take(blockBodiesPerRequest).map(_.hash)
+              waitingForActor = Some(context.actorOf(SyncBlockBodiesRequestHandler.props(peer, hashes)))
+            } else {
+              context.self ! ResumeRegularSync
+            }
           }
         case None =>
           log.error("no total difficulty for latest block") //FIXME: How do we handle this error on our blockchain?
```
mcsherrylabs: Can we add some detail to the error, ideally the block height and hash? 
If this is an "impossible" state and we can't recover we should throw new IllegalStateException

ntallar: At least right now we don't recover from this state, so I'll change it to throw an IllegalStateException. There's a task to research into this problem https://iohk.myjetbrains.com/youtrack/issue/EC-165

adamsmo:
@AlanVerbner actually I did not implemented logic for executing blocks in regular sync but can take a look tomorrow

AlanVerbner:
@adamsmo I thought you did, my bad, no worries then

PR #150, opened 2017-04-20
Fix a bug after using self destruct

LukasGasior1:
It fixes a bug when a contract account was removed (using SelfDestruct), but later in `persistState.persistContractStorage` we still tried to use it (and call `getGuaranteedAccount`).


src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala
```diff
@@ -155,7 +155,8 @@ class InMemoryWorldStateProxy private(
     copyWith(accountsStateTrie = accountsStateTrie.put(address.bytes, account))
 
   override def deleteAccount(address: Address): InMemoryWorldStateProxy =
-    copyWith(accountsStateTrie = accountsStateTrie.remove(address.bytes))
+    copyWith(accountsStateTrie = accountsStateTrie.remove(address.bytes),
+      contractStorages = contractStorages - address)
```
LukasGasior1: Removing from `contractStorages` was all that was needed to fix this particular bug, I wonder if we should also remove `evmCodeStorage` and `accountCodes` here?

rtkaczyk: Probably. But how does this fix prevent the `getGuaranteedAccount` call?

LukasGasior1: It tried to persist contract storage (of contract that no longer exists) in `persist`. Here's the exact line: https://github.com/input-output-hk/etc-client/blob/86bf3ad6fd1dcbe7f81d41a5025e22f2d63b2a5a/src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala#L67

LukasGasior1: `accountCodes` must be updated elsewhere, as my tests work without changing it. Or maybe there is another bug, that we should solve so that we have a bug with `accountCodes` that we can solve.

rtkaczyk: So should we improve this so that it's consistent?

AlanVerbner: I think it's a bug. When account delete account is called, it's contract storage and the related code should be removed, at least from the proxy and not to be used within any persist* method

adamsmo: what in case when contract MPT was shared with other contracts as well as EVM code?

AlanVerbner: I think nothing, nodes won't be deleted from mpt db

LukasGasior1: I think it's just a mapping entry address -> storage we are removing, not the actual storage. Is that right @AlanVerbner ?

AlanVerbner: @LukasGasior1 correct. As the api uses addresses, that's the mapping we use and store when persisting (if necessary)

ntallar: I evaluated deleting the code from `evmCodeStorage` of contracts when deleting accounts in `Ledger.deleteAccouts` but I don't think it would work properly if the code is duplicated among different contracts (for example as mentioned in https://www.reddit.com/r/ethereum/comments/571j84/lets_say_i_want_to_make_millions_of_contracts/). I left a comment regarding this in https://github.com/input-output-hk/etc-client/blob/phase/3/txExecution/src/main/scala/io/iohk/ethereum/ledger/Ledger.scala#L380

adamsmo: back to accountCodes
so we should add `accountCodes = accountCodes - address` ?
and have
```
    copyWith(accountsStateTrie = accountsStateTrie.remove(address.bytes),
      contractStorages = contractStorages - address, accountCodes = accountCodes - address)
```

AlanVerbner: @adamsmo :+1:

LukasGasior1:
@adamsmo @AlanVerbner added removal from accountCodes and updated with phase/3/txExec

PR #151, opened 2017-04-20
[EC-166] Add blockchain forks integration tests

LukasGasior1:
Test data contains few deployments/method calls/self destructs of ForksTestContract at different fork stages of the blockchain (frontier, homestead, eip150, eip160 - blocks 0, 3, 5, 7). Dumped from Parity/v1.5.4-beta-74b850e-20170223/x86_64-linux-gnu/rustc1.15.1

Depends on: #147 and #150


src/it/scala/io/iohk/ethereum/txExecTest/ForksTest.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  val validators = new Validators {
+    val blockValidator: BlockValidator = BlockValidator
+    val blockHeaderValidator: BlockHeaderValidator = new BlockHeaderValidatorImpl(blockchainConfig)
+    val ommersValidator: OmmersValidator = new OmmersValidatorImpl(blockchainConfig)
+    val signedTransactionValidator: SignedTransactionValidator = SignedTransactionValidator
+  }
+
+  "Ledger" should "execute blocks with respect to forks" in {
+    val fixtures: FixtureProvider.Fixture = FixtureProvider.loadFixtures("/transactionTest/forksTest/")
```
ntallar: Minor detail: as we renamed the folder containing the code for the tests from `transactionTest` -> `txExecTest`, should we do the same for the folder containing the resources for the tests as well?

adamsmo: maybe this is the reason why test is failing on CI
```
[info] - should execute blocks with respect to forks *** FAILED *** (6 seconds, 145 milliseconds)
```

LukasGasior1: the test is falling because:
 > Depends on: #147 and #150

LukasGasior1: @ntallar I've updated the package name

PR #154, opened 2017-04-21
CallOps: extend memory according to outSize arg, rather actual return…

rtkaczyk:
… data size

See explanation in: https://github.com/input-output-hk/etc-client/wiki/TX-Execution---problematic-blocks#68460

Additionally this PR:

- fixes how memory expansion (as spotted by @ntallar, see comments)
- changes MSIZE so the perceived memory size is rounded up to the next 32 byte word (block 179332)


ntallar:
There's an edge case that I'm not sure we are considering (at least according to geth), if the size of the data stored in memory is 0, it doesn't seem to be expanded (https://github.com/ethereumproject/go-ethereum/blob/master/core/vm/memory.go#L40). For example in this case if `outOffset` is too big but `outSize` is 0 we would be extending the Memory.

rtkaczyk:
@ntallar 

> the size of the data stored in memory 

Not sure what you mean by that. The memory contents prior to storing the return data? Or the return data from the called contract?

In either case, if I'm not mistake the `size` in geth code, refers to the stack argument to CALL, not the actual data size.

In our code we only extend memory when we store/load data of size > 0.

> For example in this case if outOffset is too big but outSize is 0 we would be extending the Memory.

I don't think we do. Are you saying we should?

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
```diff
@@ -138,6 +138,18 @@ class CallOpcodesSpec extends WordSpec with Matchers {
         call.stateOut.stack.pop._1 shouldEqual UInt256.One
       }
 
+      "should store contract's return data in memory" in {
+        //here the passed data size is equal to the contract's return data size
+
+        val expectedData = fxt.inputData.take(fxt.inputData.size / 2)
```
AlanVerbner: minor detail. Can we explain with a short comment where this `2` comes from? I think it's related with the code being called but...

rtkaczyk:
@ntallar correction: the [memory](https://github.com/input-output-hk/etc-client/blob/phase/3/txExecution/src/main/scala/io/iohk/ethereum/vm/Memory.scala#L47) actual does get expanded, but the [gas](https://github.com/input-output-hk/etc-client/blob/phase/3/txExecution/src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala#L83) is calculated correctly. So we are correct, but we have a small performance bug.

ntallar:
> @ntallar correction: the memory actual does get expanded, but the gas is calculated correctly. So we are correct, but we have a small performance bug

@rtkaczyk :1: I was just looking into a test to show that.
Apart for it being a performance bug, as the memory is changed, wouldn't that affect the gas cost of the next attempts to `read` or `store` data to memory?

rtkaczyk:
Good point @ntallar

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
rtkaczyk: Here's the [explanation](https://github.com/input-output-hk/etc-client/blob/bugfix/callOpsStoreResultInMemoryFix/src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala#L44). Do you think it's not enough?

AlanVerbner: For me it wasn't matched immediately (on a quick PR review for example) with the comment you point out. If you think it's self explanatory, no need to do anything.

rtkaczyk:
I think `EvmConfig#calcMemCost` (or usages of it) might be wrong. It should the word count, **_μ<sub>i</sub>_,** rather than memory size in bytes.

adamsmo:
About mentioned OOM errors, I think gas limit may prevent it, am I right?

adamsmo:
`calcMemCost` calculates how many words will be needed to store given amount of bytes, do you mean that dataSize is in words not in bytes?

rtkaczyk:
No, `calcMemSize` takes memory size and data size in bytes, but it converts them to size in words. In other words, forget what I said 😛

rtkaczyk:
> About mentioned OOM errors, I think gas limit may prevent it, am I right?

Theoretically you could have enough gas (or Ether to pay for it) to cause OOM.

adamsmo:
what about gas per block limit?

rtkaczyk:
> what about gas per block limit?

I don't know. Feel free to do the calculations :)
However as per YP:

> it is highly unlikely addresses will ever go above 32-bit bounds. That said, implementations must be able to manage this eventuality.

rtkaczyk:
Thanks, but I'm adding one more test here 😄

PR #155, opened 2017-04-21
Ledger refactoring: moved SignedTransaction validations to SignedTransactionValidator

ntallar:
## Description

Includes moving all the validations done in the `Ledger` of `SignedTransaction` to the `SignedTransactionValidator`. This refactoring is needed for simplifying the `Ledger` testing.


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -89,14 +83,18 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
...
                                   acumGas: BigInt = 0, acumReceipts: Seq[Receipt] = Nil): Either[TxsExecutionError, BlockResult] =
     signedTransactions match {
       case Nil =>
         Right(BlockResult(worldState = world, gasUsed = acumGas, receipts = acumReceipts))
 
       case Seq(stx, otherStxs@_*) =>
-        validateTransaction(stx, world, acumGas, blockHeader, signedTransactionValidator, config) match {
+        val senderAccount = world.getAccount(stx.senderAddress)
+        val validatedStx = senderAccount
+          .toRight(Left(TxsExecutionError(s"Account of tx sender ${Hex.toHexString(stx.senderAddress.toArray)} not found")))
```
AlanVerbner: maybe we can use `io.iohk.ethereum.domain.Address#toString` to print the address

src/main/scala/io/iohk/ethereum/validators/SignedTransactionValidator.scala
```diff
@@ -83,11 +90,72 @@ object SignedTransactionValidator extends SignedTransactionValidator {
...
 }
 
 sealed trait SignedTransactionError
 
 object SignedTransactionError {
   case object TransactionSignatureError extends SignedTransactionError
-  case class TransactionSyntaxError(msg: String) extends SignedTransactionError
+  case class TransactionSyntaxError(reason: String) extends SignedTransactionError
+  case class TransactionNonceError(reason: String) extends SignedTransactionError
+  case class TransactionNotEnoughGasForIntrinsicError(reason: String) extends SignedTransactionError
```
AlanVerbner: mmm...maybe this error message can be constructed based on `stx.tx.gasLimit` and `txIntrinsicGas` and not received as parameter

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -20,9 +20,6 @@ trait Ledger {
 
 class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with Logger {
 
-  case class BlockResult(worldState: InMemoryWorldStateProxy, gasUsed: BigInt = 0, receipts: Seq[Receipt] = Nil)
```
AlanVerbner: why this was moved?

src/main/scala/io/iohk/ethereum/validators/SignedTransactionValidator.scala
ntallar: You are right, it's much better to do it that way :+1:

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
ntallar: I made this change as it's needed for the `LedgerSpec` PR but you are right that this isn't a change needed for this PR, I'll remove it :+1:

AlanVerbner: Nvm, for me it's ok to leave it as it is because the other PR is going to happen soon

src/main/scala/io/iohk/ethereum/validators/SignedTransactionValidator.scala
```diff
@@ -38,7 +45,7 @@ object SignedTransactionValidator extends SignedTransactionValidator {
     * @param stx Transaction to validate
     * @return Either the validated transaction or TransactionSyntaxError if an error was detected
     */
-  private def checkSyntacticValidity(stx: SignedTransaction): Either[SignedTransactionError, SignedTransaction] = {
+  private def checkSyntacticValidity(stx: SignedTransaction): Either[SignedTransactionError, Unit] = {
```
adamsmo: maybe just `Option[SignedTransactionError]` ?

adamsmo:
LGTM

src/main/scala/io/iohk/ethereum/validators/SignedTransactionValidator.scala
ntallar: I see your point. I'll add a `TransactionValid` case object :+1:
As we were using very similar return types for the other validators, should we also change their return type? Maybe this should be done in another PR but it would be quite easy to incorporate here.

adamsmo: yes it can be done is separate PR, it should be quick change

PR #156, opened 2017-04-21
[EC-155] Ledger tests

ntallar:
## Description

Includes tests for the `executeBlockTransactions` and `executeBlock` functions for which different scenarios are considered (having no txs, having a tx that fails to execute, having more than one tx).

For being able to make testing of the `Ledger` easier, some refactoring of the `Ledger` and `SignedTransactionValidator` was needed, so this PR is dependent on #155.


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -33,7 +30,7 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
     val blockExecResult = for {
       _ <- validateBlockBeforeExecution(block, blockchain, validators)
 
-      execResult <- executeBlockTransactions(block, blockchain, storages, validators)
+      execResult <- executeBlockTransactions(block, blockchain, storages, validators.signedTransactionValidator)
```
LukasGasior1: Just thinking loud, if we could add `val validators = new Validators(blockchainConfig)` or something like that in `Ledger` we could avoid passing these vals all around

src/main/scala/io/iohk/ethereum/validators/SignedTransactionValidator.scala
```diff
@@ -1,34 +1,41 @@
...
+    * @param stx                        Transaction to validate
+    * @param senderAccount              Account of the sender of the tx
+    * @param blockHeader                Container block
+    * @param calculateUpfrontGasCost    Function used to calculate the upfront gas cost
+    * @param accumGasUsed               Total amount of gas spent prior this transaction within the container block
     * @return Transaction if valid, error otherwise
     */
-  def validateTransaction(stx: SignedTransaction, fromBeforeHomestead: Boolean): Either[SignedTransactionError, SignedTransaction] = {
+  def validate(stx: SignedTransaction, senderAccount: Account, blockHeader: BlockHeader,
+               calculateUpfrontGasCost: Transaction => UInt256, accumGasUsed: BigInt): Either[SignedTransactionError, SignedTransaction] = {
```
LukasGasior1: why not just `upfrontGasCost: UInt256`? and than pass `calculateUpfrontCost(stx.tx)` in Ledger?

LukasGasior1: I'd also change it to return `Either[SignedTransactionError, Unit]`. What's the point of returning back SignedTransaction?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
mcsherrylabs: You are right @LukasGasior1. The only question is when we do it. I'm happy to leave the shape as is until we get through the txs. I *think* that's the most efficient.

src/main/scala/io/iohk/ethereum/validators/SignedTransactionValidator.scala
ntallar: You are right about this changes :+1:
As this changes are related to code introduced in the dependant PR #155, what about if I fix them there? I could then pull from that PR or maybe you could take a look at that PR that only needs one reviewer more, so as to merge it before this one

adamsmo:
test failure on CI was fixed in `phase/3/txExecution`

src/main/scala/io/iohk/ethereum/validators/SignedTransactionValidator.scala
ntallar: This comments were fixed on PR #155 on https://github.com/input-output-hk/etc-client/pull/155/commits/0019b44b7b4467a62b609ec60d5c1abc1064abcf

rtkaczyk:
Why did the coverage decrease here, if we're only adding tests?

adamsmo:
I think reporting is misleading, I assume it decreased by 0.08 comparing to max coverage score on this branch

PR #158, opened 2017-04-21
Fix gas refund from contract called not passed to caller/creator

ntallar:
## Description

At block 81383 there is a CALL opcode to a contract where several SSTORE opcodes are called, some of which result in gas being added to be refunded (that is `ProgramState.gasRefund` is increased). In the code of the `CallOp`s this `gasRefund` from the `ProgramResult` of the call to the contract wasn't being added to the gas to be refunded.

This for example is done in EthereumJ in https://github.com/ethereum/ethereumj/blob/master/ethereumj-core/src/main/java/org/ethereum/vm/program/Program.java#L571.

## Fix

Update the `CallOp`s so that the gas refunded in the contract being called are passed to the caller of them.
With this fix we are able to execute all blocks till block 153259.

_Note_: It also includes the same fix applied to the CREATE opcode, which means that the `ProgramResult.gasRefund` of the execution of the initialization code of the contract is now being passed to the creator.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
@@ -719,6 +719,7 @@ case object CREATE extends OpCode(0xf0, 3, 1, _.G_create) {
           }
 
         state1
+          .refundGas(result.gasRefund)
           .withStack(stack2)
           .withAddressesToDelete(result.addressesToDelete)
           .withLogs(result.logs)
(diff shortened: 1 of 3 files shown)


ntallar:
That's a good point @rtkaczyk :smile:
I hadn't thought about that, I will add it to this PR as it doesn't look too difficult to add that fix :+1:

ntallar:
@rtkaczyk @AlanVerbner I just updated this PR with fixes to the problem for the gas refund for the CREATE opcode, please take a look at it

PR #159, opened 2017-04-24
Fix evm inputData when executing contract init transaction

AlanVerbner:
## Description

At block #153259 a contract init is being called and the resulting gas does not match.

## Fix

When it's a contract init transaction, payload shouldn't be used as ExecEnv inputData. YP eq 91

This leads to block #179332

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
@@ -16,13 +16,13 @@ object ProgramContext {
     import stx.tx
 
     // YP eq (91)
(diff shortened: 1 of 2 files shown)


src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
```diff
@@ -14,9 +15,14 @@ object ProgramContext {
 
     import stx.tx
 
+    // YP eq (91)
+    val programInput =
```
rtkaczyk: Minor: for constistency I would rename it to `inputData`. That's how it's referred to everywhere else.

PR #160, opened 2017-04-24
fix for CALL gas prior to EIP-150

rtkaczyk:
Fixes block 196647


src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -834,9 +834,9 @@ sealed abstract class CallOp(code: Int, delta: Int, alpha: Int) extends OpCode(c
   }
 
   private def gasCap[W <: WorldStateProxy[W, S], S <: Storage[S]](state: ProgramState[W, S], g: UInt256, gExtra: UInt256): UInt256 = {
-   if (state.gas >= gExtra)
+    if (state.config.subGasCapDivisor.isDefined && state.gas >= gExtra)
```
LukasGasior1: I don't like that we're checking `subGasCapDivisor` in two places (here and inside `state.config.gasCap`). Do you think we can make it just once? Not a blocker of course, but would be nice to have.

rtkaczyk: The problem is that `subGasCapDivisor` carries 2 pieces of information: whether to perform a calculation and how to perform it. If we did both the check and the calculation here, then we would have to repeat it for CREATE. So I don't see a simple solution for that :(

ntallar:
Looking at the Frontier YP, I think there might be an edge case we are not considering from it. In case we are executing a `CALL` with `callValue (u_2) ==0`, our current `startGas` for the contract being called is `gCap` (as `endowment==0`). As we would be executing a block from Frontier then `gCap` would be the `gas` parameter (u_0 in the YP) but (at least according to the Frontier YP) it should be 0. Could there be something that I'm missing or maybe the clients implement something different?

rtkaczyk:
@ntallar I'm not sure if I follow. The addend that is dependent on `callValue` is called _C<sub>CALLXFER</sub>_ in frontier YP and _C<sub>XFER</sub>_ in the new one. In both cases it is part of `gExtra` - frontier YP doesn't explicitly specify _C<sub>EXTRA</sub>_ but its constituents are part of _C<sub>CALL</sub>_ definition (subtract _μ<sub>s</sub>[0]_ and you get _C<sub>EXTRA</sub>_)

Does that clarify things or am I missing something?

ntallar:
Sorry, I think I didn't explain myself properly. What I'm comparing is the C_callgas function for the case that `u_2 (in our code `callValue`)==0`, before EIP150 it is always 0 (at least according to the YP) while after it, it is C_gascap. In our code it seems that we are only following the case after EIP150, being the starting gas C_gascap but maybe I'm missing something or the implementation in other clients differ from this.

rtkaczyk:
Ah, __C<sub>CALLGAS</sub>_ yes, good catch! But does it even make sense? That would essentially disallow calling a contract when `callValue` is zero.

rtkaczyk:
I am not seeing anything like that in [geth](https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go#L588) or [ethereumJ](https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/vm/VM.java#L1150). Looks like they simply add the _G<sub>callstipend</sub>_ when `callValue != 0`.

ntallar:
That's weird. But I agree that it makes more sense to do it as done in geth and ethereumJ  (which is probably why they did it that way) At least this YP difference simplifies our code

rtkaczyk:
@ntallar This concludes our discussion: https://github.com/ethereum/yellowpaper/pull/181 😄

adamsmo:
scala style fails on CI

adamsmo:
other then that LGTM

PR #161, opened 2017-04-24
[EC-158] Implement ECIP1010

LukasGasior1:
This PR implements difficulty bomb delay (see: https://github.com/ethereumproject/ECIPs/blob/master/ECIPs/ECIP-1010.md)


src/it/scala/io/iohk/ethereum/txExecTest/ForksTest.scala
```diff
@@ -24,6 +24,8 @@ class ForksTest extends FlatSpec with Matchers {
     override val daoForkBlockNumber: BigInt = 10000
     override val daoForkBlockHash: ByteString = ByteString("unused")
     override val daoForkBlockTotalDifficulty: BigInt = 1000000
+    override val difficultyBombPauseBlockNumber: BigInt = Long.MaxValue
+    override val difficultyBombContinueBlockNumber: BigInt = Long.MaxValue
```
rtkaczyk: No tests for these?

LukasGasior1: Not here. Although this may actually be good idea to export some post ecip1010 blocks from parity and try to execute them as integration tests. But for that I'd probably need to sync from scratch with `--pruning archive` is that right @adamsmo ?

adamsmo: yes `--pruning archive`, but MPT tree may be large

LukasGasior1: I could probably try running with different config and make this fork much sooner

adamsmo: 👍

src/main/scala/io/iohk/ethereum/validators/BlockHeaderValidator.scala
```diff
@@ -142,20 +142,33 @@ class BlockHeaderValidatorImpl(blockchainConfig: BlockchainConfig) extends Block
...
-    val difficultyBombExponent: Int = (blockHeader.number / ExpDifficultyPeriod - 2).toInt
+
+    val difficultyBombExponent: Int =
+      if (blockHeader.number < difficultyBombPauseBlockNumber)
+        (blockHeader.number / ExpDifficultyPeriod - 2).toInt
+      else if (blockHeader.number < difficultyBombContinueBlockNumber)
+        ((difficultyBombPauseBlockNumber / 100000) - 2).toInt
+      else {
+        val delay = (difficultyBombContinueBlockNumber - difficultyBombPauseBlockNumber) / 100000
+        ((blockHeader.number / 100000) - delay - 2).toInt
```
ntallar: Very minor detail: Should we use the defined `ExpDifficultyPeriod` instead of `100000`?

src/it/scala/io/iohk/ethereum/txExecTest/ForksTest.scala
LukasGasior1: @rtkaczyk Since difficulty bomb starts growing only after block 200 000 we would have to dump at least that many blocks for integration test data. I think it doesn't make sense to do that. I could adjust the algorithm for tests but what's the point of doing integrations test where the function to be tested is not exactly the same? Let's stick to unit tests with this one

PR #162, opened 2017-04-24
Fix for contract init that might be using an account that already exists

AlanVerbner:
## Description

At Block #243826 When executing a contract init, tx fails because the account already exists

## Fix

As stated in YP EIP-150 REVISION eq 86 the destination address might have already received funds .

This leads up to #244793 execution.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
+++ b/src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
@@ -320,12 +320,8 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
     stx.tx.receivingAddress match {
       case None =>
         val address = worldStateProxy.createAddress(stx.senderAddress)
-        val world1 =
-          // YP eq (86) implies that the address might have already received funds before contract init. We need to check that.
-          if(worldStateProxy.getAccount(address).isDefined) worldStateProxy
-          else worldStateProxy.newEmptyAccount(address)
-        val world2 = world1.transfer(stx.senderAddress, address, UInt256(stx.tx.value))
-        ProgramContext(stx, address,  Program(stx.tx.payload), blockHeader, world2, config)
+        val world1 = worldStateProxy.transfer(stx.senderAddress, address, UInt256(stx.tx.value))
+        ProgramContext(stx, address,  Program(stx.tx.payload), blockHeader, world1, config)
 
       case Some(txReceivingAddress) =>
         val world1 = worldStateProxy.transfer(stx.senderAddress, txReceivingAddress, UInt256(stx.tx.value))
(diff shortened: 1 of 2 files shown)


adamsmo:
so in this case somebody precalculated contract address and sent founds to that address before contract creation, this can be done in separate transaction? if so then we are modifying EVM code of existing account?

AlanVerbner:
@adamsmo afaik you can send a tx to any address whenever you want to. 

Code will be stored when a contract init tx is sent. The address where the code will be stored is calculated based on sender account nonce that is always increased before executing anything. That being said, there is no chance to overwrite a stored contract code (as nonces won't match for any 2 init TXS for a given sender)...does it make sense?

adamsmo:
yes it does, but when you are sending ether to non existing address then account at that address is created with codeHash pointing to empty evm code
so after that if you sent contract init tx that creates account with same address than you are changing codeHash pointing to empty code to codeHash pointing to contract code

AlanVerbner:
I see your point. I'm not sure the immutability statement is about "after storing some code". Can you imagine another case where that's possible?

BTW, checkout YP EIP-150 REVISION eq 86  where it's implied that the contract account might already exist when contract init tx is executed

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -320,7 +320,10 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
     stx.tx.receivingAddress match {
       case None =>
         val address = worldStateProxy.createAddress(stx.senderAddress)
-        val world1 = worldStateProxy.newEmptyAccount(address)
+        val world1 =
+          // YP eq (86) implies that the address might have already received funds before contract init. We need to check that.
+          if(worldStateProxy.getAccount(address).isDefined) worldStateProxy
+          else worldStateProxy.newEmptyAccount(address)
```
rtkaczyk: Why do we even bother with this step? `WorldStateProxy#transfer` is already an upsert. Look at the implementation in `CREATE` opcode.

mcsherrylabs:
It's an interesting point. One we can check in the block explorer?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
AlanVerbner: I think you are right @rtkaczyk ..will try that. Good catch!!

rtkaczyk:
To me it is evident from YP that this is allowed. Contract are immutable, but accounts aren't. An account that has no code is not a contract, but may become one.

This situation already occurs, at least for a brief moment. Upon contract creation an empty account _A_ is created first (if not exists) and endowed with Ether, and this account is accessible when running contract initialisation code. But during that run, any message calls to account _A_ will result in no code being executed, because account _A_ will only have the code (codeHash changed) after the initialisation code has finished.

AlanVerbner:
@adamsmo @mcsherrylabs Just pushed an improvement based on @rtkaczyk comment (it was a better solution). Can you please take a quick look again? Sorry about this.

adamsmo:
I did, LGTM

PR #164, opened 2017-04-25
fix memory expansion

rtkaczyk:
See the comment on `Memory#expand` for details. Related blocks are: 68460 and 244793


AlanVerbner:
I'm working on the tests atm

src/main/scala/io/iohk/ethereum/vm/Memory.scala
```diff
@@ -70,6 +70,21 @@ class Memory private(private val underlying: ByteString) {
...
     }
 
+  /** This function will expand the Memory size as if storing data given the `offset` and `size`.
+    * If the memory is already initialised at that region it will not be modified, otherwise it will be filled with
+    * zeroes.
+    * This is required to satisfy memory expansion semantics for *CALL* opcodes.
+    */
+  def expand(offset: UInt256, size: UInt256): Memory = {
+    val totalSize = (offset + size).toInt
+    if (this.size >= totalSize || size.isZero)
```
AlanVerbner: @rtkaczyk found this `size.isZero` was missing. Based on YP EIP-150 REVISION above eq 222

> Referencing a zero length range (e.g. by attempting to pass it as the input range to a CALL) does not require memory to be extended to the beginning of the range.

rtkaczyk: Yep, this is also a direct consequence of eq 216 (_M(s,f,l) = ..._). Good catch!

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
```diff
@@ -656,4 +670,42 @@ class CallOpcodesSpec extends WordSpec with Matchers {
...
+    Seq(CALL, CALLCODE, DELEGATECALL).foreach { opCode =>
+
+      s"$opCode processes returned data" should {
+
+        "handle memory expansion properly" in {
+
+          val inputData = ByteString(Array[Byte](1).padTo(32, 1.toByte))
+          val context: PC = fxt.context.copy(world = fxt.worldWithReturnSingleByteCode)
+
+          val outOffsets = Seq(0, inputData.size / 2, inputData.size * 2)
```
rtkaczyk: I would recommend using a `Table` here, even if it's just a single column. A `Table` makes it easier to see which input caused a test failure.

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
```diff
@@ -656,4 +670,42 @@ class CallOpcodesSpec extends WordSpec with Matchers {
...
+
+          outOffsets.foreach { outOffset =>
+
+            val call = CallResult(
+              op = CALL,
```
rtkaczyk: `op = opCode` ?

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
rtkaczyk: And what about `outSize` as the second column to this table?

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
```diff
@@ -7,6 +7,7 @@ import io.iohk.ethereum.crypto._
 import io.iohk.ethereum.domain.{Account, Address}
 import io.iohk.ethereum.utils.ByteUtils
 import io.iohk.ethereum.vm.MockWorldState._
+import org.scalatest.prop.PropertyChecks
```
rtkaczyk: I don't think this was used in the end, but see the comment about tables.

src/main/scala/io/iohk/ethereum/vm/Memory.scala
rtkaczyk: BTW, we were talking about the same equation 😄  
Referencing the YP version like `YP EIP-150 REVISION` is not enough cause I also have this revision, and my eq numbers are different. My version is `EIP-150 REVISION (1e18248 - 2017-04-12)`

AlanVerbner: Lol..no, it's eq 223, my bad.

I feel stupid but, where do you get that revision numbers? Mine only says 
screenshot from 2017-04-26 07-47-15

rtkaczyk: The latest one can be downloaded here: https://ethereum.github.io/yellowpaper/paper.pdf
It has all that info in the heading.

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
AlanVerbner: `outSize` it's the same calculation so I wouldn't add it as `Table`

src/main/scala/io/iohk/ethereum/vm/Memory.scala
AlanVerbner: ok, mine is from Gavin Wood page :confused:

rtkaczyk:
```
Pull request authors can't their own pull request
```

:+1: LGTM :+1:

PR #167, opened 2017-04-25
Fixed varGas calculation having an overflow in CALL opcode

ntallar:
## Description

At block 299804 there is a CALL opcode that has it's `gas` parameter (from which the starting gas for the contract execution is calculated) too big that causes an overflow in the calculation on `varGas` enabling its execution although it should cause an OutOfGas exception.

## Fix

Replacing the use of `UInt256`to `BigInt` for manipulating gas.
This fix allows us to execute at least till block `505404`.


src/main/scala/io/iohk/ethereum/vm/UInt256.scala
```diff
@@ -38,7 +38,7 @@ object UInt256 {
   def apply(n: Long): UInt256 =
     apply(BigInt(n))
 
-
+  implicit def bigIntToUInt256(bigInt: BigInt): UInt256 = UInt256(bigInt)
```
rtkaczyk: Perhaps I wasn't clear enough, but this is what I wanted to **avoid**. `BigInt` is "bigger" than `UInt256`, so such lossy conversion should not be implicit. Same as with `Int` vs `Long`:

```
scala> val l: Long = 1: Int
l: Long = 1

scala> val i: Int = 1L
<console>:7: error: type mismatch;
 found   : Long(1L)
 required: Int
       val i: Int = 1L
```

What I proposed was to add an implicit value class:

```
implicit class BigIntAsUInt256(big: BigInt) extends AnyVal {
  def toUInt256: UInt256 = UInt256(big)
}
```

Then we can write `bigInt.toUInt256`, instead of `UInt256(bigInt)`. Somehow I find the former more readable, but it's really optional.

PR #168, opened 2017-04-26
precompiled contract should use sha256, not kec256 (sha3)

rtkaczyk:
Fixes block 505404


rtkaczyk:
>  kec256 [...] used in most places

That's why I made this mistake. But there already is an opcode SHA3 (Keccak), and SHA-256 is still widely used accross the world.

PR #169, opened 2017-04-26
VM handle large values when accessing byte arrays - remove optimistic…

rtkaczyk:
… toInt conversions

Fixes block 549413

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
@@ -164,12 +164,9 @@ object OpCodes {
 
 object OpCode {
   def sliceBytes(bytes: ByteString, offset: UInt256, size: UInt256): ByteString = {
-    if (size > Int.MaxValue)
-      throw new Exception(s"We're not supporting arrays of length > ${Int.MaxValue} ($size requested)")
-    else if (offset >= bytes.size)
-      ByteString(Array.fill[Byte](size.toInt)(0))
-    else
-      bytes.slice(offset.toInt, offset.toInt + size.toInt).padTo(size.toInt, 0.toByte)
+    val start = offset.min(bytes.size).toInt
+    val end = (offset + size).min(bytes.size).toInt
+    bytes.slice(start, end).padTo(size.toInt, 0.toByte)
   }
 }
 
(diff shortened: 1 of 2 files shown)


rtkaczyk:
Tests added

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -163,8 +163,11 @@ object OpCodes {
 }
 
 object OpCode {
-  def sliceBytes(bytes: ByteString, offset: Int, size: Int): ByteString =
-    bytes.slice(offset, offset + size).padTo(size, 0.toByte)
+  def sliceBytes(bytes: ByteString, offset: UInt256, size: UInt256): ByteString = {
+    val start = offset.min(bytes.size).toInt
+    val end = (offset + size).min(bytes.size).toInt
+    bytes.slice(start, end).padTo(size.toInt, 0.toByte)
```
rtkaczyk: I think `.padTo(end - start, 0.toByte)` would make more sense. But this is what [geth does](https://github.com/ethereum/go-ethereum/blob/master/core/vm/common.go#L42). See the consequence in the tests.

adamsmo: maybe but for me it is more readable this way

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -367,7 +370,7 @@ case object CODESIZE extends ConstOp(0x38)(_.env.program.length)
 case object CODECOPY extends OpCode(0x39, 3, 0, _.G_verylow) {
   protected def exec[W <: WorldStateProxy[W, S], S <: Storage[S]](state: ProgramState[W, S]): ProgramState[W, S] = {
     val (Seq(memOffset, codeOffset, size), stack1) = state.stack.pop(3)
-    val bytes = state.program.getBytes(codeOffset.toInt, size.toInt)
+    val bytes = OpCode.sliceBytes(state.program.code, codeOffset, size)
```
LukasGasior1: Do we still need `state.program.getBytes`?

rtkaczyk: Yes, it's used for PUSHx (the semantics would be the same, but no unnecessary UInt256 -> Int conversions)

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
rtkaczyk: It's not about readability...

rtkaczyk: Compare the 5th and 6th table rows (counting from 1)

adamsmo: 👍

PR #170, opened 2017-04-27
Fix for ECDSA recovery not failing

adamsmo:


rtkaczyk:
Even better with the refactoring 👍

PR #173, opened 2017-05-02
[EC-171] Error in block 3003036

adamsmo:
The issue was that we did not create accounts for sender address when it does not exists which is an issue for transactions with gas price set to 0 which are valid transactions.

### DIFF
--- a/src/test/scala/io/iohk/ethereum/ledger/LedgerSpec.scala
+++ b/src/test/scala/io/iohk/ethereum/ledger/LedgerSpec.scala
@@ -663,6 +663,9 @@ class LedgerSpec extends FlatSpec with PropertyChecks with Matchers {
     }
   }
 
+  it should "create sender account if it does not exists" in new TestSetup {
+    //todo add test
+  }
 
   trait TestSetup {
     val originKeyPair = generateKeyPair()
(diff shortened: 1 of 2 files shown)


src/test/scala/io/iohk/ethereum/ledger/LedgerSpec.scala
```diff
@@ -663,10 +663,35 @@ class LedgerSpec extends FlatSpec with PropertyChecks with Matchers {
...
+    val newAccountKeyPair: AsymmetricCipherKeyPair = generateKeyPair()
+    val newAccountAddress = Address(kec256(newAccountKeyPair.getPublic.asInstanceOf[ECPublicKeyParameters].getQ.getEncoded(false).tail))
+
+    val mockVM = new MockVM((pc: Ledger.PC) => {
+      pc.env.inputData shouldEqual ByteString.empty
+      createResult(pc, defaultGasLimit, defaultGasLimit, 0, None, returnData = ByteString("contract code"))
+    })
+    val ledger = new LedgerImpl(mockVM, blockchainConfig)
+
+    val tx: Transaction = defaultTx.copy(gasPrice = 0, receivingAddress = None, payload = inputData)
```
AlanVerbner: Do you think it's worthy to write a test where the execution fails because `gasPrice` is not zero?

adamsmo: it may be, but it is the same case as if you do not have ether to pay for gas

PR #175, opened 2017-05-04
Fast sync download fix on restart	

LukasGasior1:
This fixes the issue when, after restarting during fast sync, some mpt nodes were "lost". Solution: we need to also save requested items in sync state (the ones that are not yet in db, but not in queues anymore).

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
@@ -162,10 +162,10 @@ trait FastSync {
       scheduler.schedule(persistStateSnapshotInterval, persistStateSnapshotInterval) {
         syncStateStorageActor ! SyncState(
           initialSyncState.targetBlock,
-          mptNodesQueue ++ requestedMptNodes.values.flatten.toSeq.distinct,
-          nonMptNodesQueue ++ requestedNonMptNodes.values.flatten.toSeq.distinct,
-          blockBodiesQueue ++ requestedBlockBodies.values.flatten.toSeq.distinct,
-          receiptsQueue ++ requestedReceipts.values.flatten.toSeq.distinct,
+          requestedMptNodes.values.flatten.toSeq.distinct ++ mptNodesQueue,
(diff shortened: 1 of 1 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -195,11 +200,19 @@ trait FastSync {
...
+        requestedReceipts = requestedReceipts - sender()
         processSyncing()
 
       case Terminated(ref) if assignedHandlers.contains(ref) =>
         context unwatch ref
         assignedHandlers -= ref
+        requestedMptNodes = requestedMptNodes - ref
+        requestedNonMptNodes = requestedNonMptNodes - ref
+        requestedBlockBodies = requestedBlockBodies - ref
+        requestedReceipts = requestedReceipts - ref
```
ntallar: I'm wondering, should we add any request being handled by the terminated handler to our queues of pending requests? For example if a `FastSyncNodesRequestHandler` terminated we may have to add the nodes it was in charge of to `mptNodesQueue` and `nonMptNodesQueue` lists

LukasGasior1: it's done in the `FastSyncNodesRequestHandler` itself:
```
  override def handleTerminated(): Unit = {
    syncController ! FastSync.EnqueueNodes(requestedHashes)
    cleanupAndStop()
  }
```

adamsmo: Terminated message will happen after requested entities will be added back to queue?

LukasGasior1: hmm actually that's what happens when underlying `peer` actor is terminated, not the handler

LukasGasior1: let me double check

LukasGasior1: seems you're right @adamsmo we should enqueue these items here

PR #176, opened 2017-05-08
Add initial JSON RPC

LukasGasior1:
- add JsonRpcController for handling RPC requests
- add JsonRpcHttpServer to handle RPC requests over HTTP
- remove spray json in favour of json4s

wip:
needs tests


src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import scala.util.{Success, Try}
+
+object JsonMethodsImplicits {
+
+  implicit val formats: Formats = DefaultFormats + ByteStringSerializer
+
+  implicit val web3_sha3 = new JsonDecoder[Sha3Request] with JsonEncoder[Sha3Response] {
+    override def decodeJson(params: JsonAST.JArray): Try[Sha3Request] =
+      Try(Sha3Request(params.arr.head.extract[ByteString]))
```
ntallar: Should we check here or somewhere else that the `ByteString` parameter has a `0x` prefix? It seems to be needed by https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding, and geth fails when asked for request such as `{"jsonrpc":"2.0","method":"web3_sha3","params":["68656c6c6f20776f726c64"],"id":64}`

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala
```diff
@@ -0,0 +1,46 @@
...
+import scala.util.Try
+import scala.concurrent.ExecutionContext.Implicits.global
+
+object JsonRpcController {
+  trait JsonDecoder[T] {
+    def decodeJson(params: JArray): Try[T]
+  }
+
+  trait JsonEncoder[T] {
+    def encodeJson(t: T): JArray
```
adamsmo: it can also be JValue like true / false
for example in eth_submitHashrate

src/main/scala/io/iohk/ethereum/jsonrpc/Web3Service.scala
```diff
@@ -0,0 +1,27 @@
...
+
+class Web3Service {
+  import Web3Service._
+
+  def sha3(req: Sha3Request): Future[Sha3Response] = {
+    Future.successful(Sha3Response(crypto.kec256(req.data)))
+  }
+
+  def clientVersion(req: ClientVersionRequest): Future[ClientVersionResponse] = {
+    Future.successful(ClientVersionResponse("etc-client/v0.1"))
```
AlanVerbner: Maybe it' better to move this to `Config`

src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  implicit val web3_clientVersion = new JsonDecoder[ClientVersionRequest] with JsonEncoder[ClientVersionResponse] {
+    override def decodeJson(params: JsonAST.JArray): Try[ClientVersionRequest] = Success(ClientVersionRequest())
+
+    override def encodeJson(t: ClientVersionResponse): JsonAST.JArray = JArray(List(t.value))
+  }
+
+  object ByteStringSerializer extends CustomSerializer[ByteString](formats =>
+    (
+      { case JString(str) => ByteString(Hex.decode(str.drop(2)))},
```
AlanVerbner: what about if Hex.decode fails? I think other clients return an error saying it's an invalid param

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala
LukasGasior1: True, I just noticed `sha3` and others are also just a single values

src/main/scala/io/iohk/ethereum/jsonrpc/Web3Service.scala
LukasGasior1: :+1: sure it is

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcError.scala
```diff
@@ -0,0 +1,5 @@
+package io.iohk.ethereum.jsonrpc
+
+import org.json4s.JsonAST.JValue
+
+case class JsonRpcError(code: Int, message: String, data: Option[JValue])
```
ntallar: Should we incorporate in this PR handling the different errors with their respective codes? (as specified in http://www.jsonrpc.org/specification). Maybe we could create a separate task to handle this.
For example we currently fail to send a response when the request is: `{"jsonrpc":"2.0","method":"web3_clientVersion",params":[],"id":67}`

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala
```diff
@@ -0,0 +1,46 @@
...
+}
+
+class JsonRpcController(web3Service: Web3Service) {
+
+  import JsonMethodsImplicits._
+
+  def handleRequest(request: JsonRpcRequest): Future[JsonRpcResponse] = {
+    request.method match {
+      case "web3_sha3" => handle[Sha3Request, Sha3Response](web3Service.sha3, request)
+      case "web3_clientVersion" => handle[ClientVersionRequest, ClientVersionResponse](web3Service.clientVersion, request)
```
AlanVerbner: It seems that this api call is failing if we don't send an empty array as parameters. In parity it works with or without it

ntallar: It seems that they could be ommited according to the specification:

LukasGasior1:
@AlanVerbner @ntallar @adamsmo I've addresses your concerns

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala
```diff
@@ -0,0 +1,52 @@
...
+  }
+}
+
+class JsonRpcController(web3Service: Web3Service) {
+
+  import JsonMethodsImplicits._
+  import JsonRpcErrors._
+
+  def handleRequest(request: JsonRpcRequest): Future[JsonRpcResponse] = {
+    request.method match {
```
AlanVerbner: I think this should be mixed from different handlers. I'm thinking about this in order to disable different apis. This is not a blocker tho, it can be refactored later.

src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
```diff
@@ -0,0 +1,41 @@
...
+object JsonMethodsImplicits {
+
+  import JsonRpcErrors._
+
+  implicit val formats: Formats = DefaultFormats
+
+  implicit val web3_sha3 = new JsonDecoder[Sha3Request] with JsonEncoder[Sha3Response] {
+    override def decodeJson(params: Option[JArray]): Either[JsonRpcError, Sha3Request] =
+      params match {
+        case Some(JArray((input: JString) :: Nil)) if input.s.startsWith("0x") =>
```
AlanVerbner: As we we are dealing with `DATA` encoded as ByteString, this will be used in other parts too, can't we extract some common function here?

AlanVerbner: I sketched something here (haven't tested it tho): https://github.com/input-output-hk/etc-client/blob/feature/jsonrpcBlockNumberGetTransactionByBlockHashAndIndex/src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala#L53

LukasGasior1: Agree, we should have a util for that. We can use the one from your branch

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala
LukasGasior1: since you've already started working off this branch, let's merge it and to it in separate PR

LukasGasior1:
@AlanVerbner I've added util methods for extracting/encoding `unformatted data` and `quantity`. I don't think they need to extend JsonEncoder and Decoder

src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
```diff
@@ -0,0 +1,60 @@
...
+    override def encodeJson(t: ClientVersionResponse): JValue = t.value
+  }
+
+  private def encodeAsHex(input: ByteString): JString =
+    JString(s"0x${Hex.toHexString(input.toArray[Byte])}")
+
+  private def encodeAsHex(input: BigInt): JString =
+    JString(s"0x${Hex.toHexString(input.toByteArray)}")
+
+  private def tryExtractUnformattedData(input: JString): Either[JsonRpcError, ByteString] = {
```
adamsmo: it odes not work for `0x55109` because there are 5 (odd number) digits after 0x
it does not work for `0x0` for the same reason

src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
```diff
@@ -0,0 +1,60 @@
...
+    if (input.s.startsWith("0x")) {
+      Try(ByteString(Hex.decode(input.s.drop(2)))) match {
+        case Success(bs) => Right(bs)
+        case Failure(_) => Left(InvalidParams.copy(message = s"Unable to parse data from '${input.s}'"))
+      }
+    } else Left(InvalidParams.copy(message = s"Data '${input.s}' should have 0x prefix"))
+  }
+
+  private def tryExtractQuantity(input: JString): Either[JsonRpcError, BigInt] = {
```
adamsmo: same here

src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
LukasGasior1: these are not valid values for unformatted data see: https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding they are, however, valid when interpreted as `QUANTITY` type (see `tryExtractQuantity`)

adamsmo: 👍 yes, and they are interpreted as quantity for `eth_submitHashrate`

adamsmo: maybe
```
  private def formatForByteString(input: String): String = {
    val trimmed = input.drop(2)
    if (trimmed.length % 2 == 0) {
      trimmed
    } else {
      "0" + trimmed
    }
  }
```

adamsmo: or should I use BigInt in this case?

LukasGasior1: you should use BigInt :) or if you need to accept QUANTITY but as ByteString instead of BigInt you can write separate method for that.
For `unformatted data` `0x123` is not a valid input.

build.sbt
```diff
@@ -8,7 +8,7 @@ val commonSettings = Seq(
 
 val dep = {
   val akkaVersion = "2.4.17"
-  val akkaHttpVersion = "10.0.3"
+  val akkaHttpVersion = "10.0.6"
```
AlanVerbner: <3

src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
```diff
@@ -0,0 +1,60 @@
...
+    if (input.s.startsWith("0x")) {
+      val noPrefix = input.s.replace("0x", "")
+      Try(BigInt(noPrefix, 16)) match {
```
ntallar: Minor detail: Should we check here that `noPrefix` doesn't have an odd length (apart from `0x0`)?

LukasGasior1: @ntallar Odd length is allowed for `QUANTITY`

ntallar: You are right, my mistake :+1:

PR #178, opened 2017-05-09
Add net_ rpc methods

LukasGasior1:
wip: needs tests

PR #179, opened 2017-05-09
Added some eth_ methods

ntallar:
## Description

Includes the implementation of `eth_getBlockTransactionCountByHash`, `eth_getBlockByHash` and `eth_getUncleByBlockHashAndIndex` JSON RPC methods. Both the transaction and block objects (as described in https://github.com/ethereum/wiki/wiki/JSON-RPC) were implemented, as `TransactionResponse` and `BlockResponse` respectively.


src/test/scala/io/iohk/ethereum/jsonrpc/EthJsonRpcControllerSpec.scala
```diff
@@ -0,0 +1,228 @@
...
+import org.scalatest.{Matchers, WordSpec}
+import org.scalatest.prop.PropertyChecks
+import org.spongycastle.util.encoders.Hex
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
+class EthJsonRpcControllerSpec extends WordSpec with Matchers with PropertyChecks {
```
rtkaczyk: I had a different concept about these tests, see PR #180. Basically:
- `JsonRpcControllerSpec` should test all the methods, just in terms of their availability
- `EthServiceSpec` should tests the methods' functionality in depth - cover different cases like done here, but without awareness of http/json level.

What do you think?

src/main/scala/io/iohk/ethereum/jsonrpc/BlockResponse.scala
```diff
@@ -0,0 +1,69 @@
...
+                          difficulty: BigInt,
+                          totalDifficulty: Option[BigInt],
+                          extraData: ByteString,
+                          size: BigInt,
+                          gasLimit: BigInt,
+                          gasUsed: BigInt,
+                          timestamp: BigInt,
+                          transactions: Either[Seq[ByteString], Seq[TransactionResponse]],
+                          uncles: Seq[ByteString]
+                        )
```
rtkaczyk: Is this indentation intentional?

src/test/scala/io/iohk/ethereum/jsonrpc/EthJsonRpcControllerSpec.scala
AlanVerbner: imo we need to test 3 things:

1. XXXService returns expected data and handle corner cases and errors if needed
2. We decode/encode requests & responses properly
3. The service is handled by the routes

What do you think? Also, I'm  not sure if the cases you mention, @rtkaczyk , cover this.

ntallar: It agree it sounds better to do it that way, so as to separate the testing. Also, my previous version of the tests were done as you recommend for `EthServiceSpec` so I'll be able to have `JsonRpcControllerSpec` and `EthServiceSpec` tests quite fast :+1:

rtkaczyk: > The service is handled by the routes

I would say 2. and 3. are the same thing because a route is just a method name in the JSON request. And this is covered by `JsonRpcControllerSpec`. Unless you think we also need handled some corner cases *for each method* when encoding/decoding.

Anyway, it seems the testing pattern I proposed was also assumed by @LukasGasior1 in his #178. Lukasz, what's your opinion on this?

AlanVerbner: @rtkaczyk well... there are some optional fields in some responses I would like to test.

That being said, as @LukasGasior1 already followed that pattern, let's stick to it and tests corner cases if needed.

src/main/scala/io/iohk/ethereum/jsonrpc/JsonSerializers.scala
```diff
@@ -0,0 +1,66 @@
...
+  object SignedTransactionResponseSerializer extends CustomSerializer[TransactionResponse](_ =>
+    (
+      {PartialFunction.empty},
+      { case stxResponse: TransactionResponse =>
+          implicit val formats = DefaultFormats.preservingEmptyValues +
+            UnformattedDataJsonSerializer +
+            QuantitiesSerializer +
+            OptionNoneToJNullSerializer
+
+          Extraction.decompose(stxResponse)
```
LukasGasior1: maybe we can just have `Extraction.decompose(stxResponse)` in JsonEncoder?

ntallar: I mainly did it this way to have how we treat `None` values separated for `TransactionResponse` and `BlockResponse` (for using `OptionNoneToJNullSerializer`) but AFAIK, there shouldn't be any problem of including it in `JsonMethodsImplicits`. So, I could add that and remove both `SignedTransactionResponseSerializer` and `BlockResponseSerializer`, do you also think we should do so?

LukasGasior1: I'd do that if it won't cause problems

ntallar: It doesn't cause any problems and it even simplifies our code so I'll add it right now :+1:

PR #180, opened 2017-05-10
json rpc: eth_protocolVersion

rtkaczyk:


src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -0,0 +1,19 @@
+package io.iohk.ethereum.jsonrpc
+
+
+import scala.concurrent.Future
+
+
+object EthService {
+  val CurrentProtocolVersion = 63
```
ntallar: Minor detail: Maybe it's better to move this to `Config`?

rtkaczyk: I thought about that, but I decided hardcoding is better. After all we even have it in the object name: `io.iohk.ethereum.network.p2p.messages.PV63`, and protocol version shouldn't be changed without recompilation. It's also true for test networks.

Maybe it could be useful if we wanted to switch to an older protocol version, but that would require additional functionality and I don't see the point at this moment.

PR #182, opened 2017-05-12
Add eth_syncing and possibility to enable/disable apis

LukasGasior1:
There's one potential issue here: when not doing fast-sync at all (i.e regular sync from block 0), we don't have any estimate for highest block. It would require to ask nodes about their highest blocks to get it. The questions is whether we should do it or not.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
@@ -57,10 +57,6 @@ trait FastSync {
     case PeerActor.MessageReceived(BlockHeaders(Seq(blockHeader))) =>
       sender() ! PeerActor.Unsubscribe
 
-      if (appStateStorage.getEstimatedHighestBlock() < blockHeader.number) {
-        appStateStorage.putEstimatedHighestBlock(blockHeader.number)
-      }
-
       val newWaitingFor = waitingFor - sender()
       val newReceived = received + (sender() -> blockHeader)
 
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
@@ -57,6 +57,10 @@ trait FastSync {
     case PeerActor.MessageReceived(BlockHeaders(Seq(blockHeader))) =>
       sender() ! PeerActor.Unsubscribe
 
+      if (appStateStorage.getEstimatedHighestBlock() < blockHeader.number) {
+        appStateStorage.putEstimatedHighestBlock(blockHeader.number)
+      }
+
       val newWaitingFor = waitingFor - sender()
       val newReceived = received + (sender() -> blockHeader)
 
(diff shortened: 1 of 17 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -57,6 +57,10 @@ trait FastSync {
     case PeerActor.MessageReceived(BlockHeaders(Seq(blockHeader))) =>
       sender() ! PeerActor.Unsubscribe
 
+      if (appStateStorage.getEstimatedHighestBlock() < blockHeader.number) {
+        appStateStorage.putEstimatedHighestBlock(blockHeader.number)
+      }
```
rtkaczyk: Don't we also need this for regular sync?

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala
```diff
@@ -11,30 +11,73 @@ import scala.concurrent.Future
...
 
+  val apisHandleFns: Map[String, PartialFunction[JsonRpcRequest, Future[JsonRpcResponse]]] = Map(
+    Apis.Eth -> handleEthRequest,
+    Apis.Web3 -> handleWeb3Request,
+    Apis.Net -> handleNetRequest,
+    Apis.Db -> PartialFunction.empty,
+    Apis.Personal -> PartialFunction.empty,
+    Apis.Admin -> PartialFunction.empty,
+    Apis.Debug -> PartialFunction.empty
+  )
```
rtkaczyk: Looks nice

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
LukasGasior1: That's the question I've asked in the description. It would require us to additionally ask peers for their best blocks

rtkaczyk: > That's the question I've asked in the description
Yup, sorry 😛. But isn't that just the highest block we downloaded?

LukasGasior1: It can be, but it wouldn't be particularly accurate estimate of highest block on the blockchain :)

adamsmo:
In my opinion we should not ask other nodes
highest block should be highest that we know of at given time, in case of fast sync it is target block and in case of regular sync latest block

LukasGasior1:
@adamsmo in case of fast sync we already ask peers for their highest blocks

adamsmo:
I mean issue additional request just for this RPC

rtkaczyk:
How about estimating the highest block based on time/speed?
Consider this:

```
h = getCurrentHeight()
dt = B[h].timestamp - B[h - N].timestamp
avg = N / dt
x = avg * (now - B[h].timestamp)
heightestEstimatedBlock = h + x
```

LukasGasior1:
@rtkaczyk I think that makes sense!

LukasGasior1:
another possibility is to always ask peers for their latest block (no matter is it's fast or regular sync)

rtkaczyk:
Do you mean to ask the peers via RPC or via the Ethereum protocol?
In either could we simply ask/estimate this value when executing this RPC method, rather than continuously update the app state storage while syncing?

LukasGasior1:
Ethereum protocol, we could do that once when starting syncing, it would provide pretty accurate estimate

rtkaczyk:
So once per runtime? Depends on what this method is really used for. Say I already had my node up to date and kept it running. But I notice I'm lagging a little (maybe because of network problems), and I want to see how far behind I am. In that case it's no good - I would have to restart the client.

Also, we could use `eth_blockNumber` to ask the peers, just saying.

LukasGasior1:
Nodes don't talk to each other using rpc @rtkaczyk

rtkaczyk:
If they knew their RPC port numbers they could. But they don't, right?

LukasGasior1:
@rtkaczyk right. I guess there's quite a number of nodes that do not expose rpc

LukasGasior1:
@rtkaczyk how about asking a peer for their best block when we connect? that way we could update our estimates every time we connect to new peer

LukasGasior1:
Actually it was even simpler. I've added two changes:
1. ask for highest block when peer connects (and update estimate on response)
2. update estimate when new blocks are broadcasted

PR #183, opened 2017-05-17
Include mined blocks

adamsmo:
~This PR depends on https://github.com/input-output-hk/etc-client/pull/177~
This PR contains logic for adding mined blocks to blockchain


src/main/resources/blockchain/default-genesis.json
```diff
@@ -1,5 +1,5 @@
 {
-  "difficulty": "0x0400000000",
+  "difficulty": "0x040000",
```
rtkaczyk: Was this intentional? I'm somewhat confused because according to YP neither value is correct. It should be 2^17 which is `0x20000`

adamsmo: no it was not intentional there should be `0x0400000000` this is the value from main chain

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -62,6 +73,25 @@ trait RegularSync {
       }
   }
 
+  private def insertMinedBlock(block: Block, parentTd: BigInt) = {
+    val result: Either[BlockExecutionError, Seq[Receipt]] = ledger.executeBlock(block, blockchainStorages, validators)
```
rtkaczyk: If I'm not mistaken, we are executing the block twice. First when preparing the block for mining, then here. Can we do that only when preparing the block?

adamsmo: when executing block in prepare we do not perform full validation because there is no nonce/mixHash so I decided to execute it again to check consistency and to not be force to cache and pass around mpt hash
so: we can, but it would make code more complicated
I am not convinced about performance gain

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -98,10 +128,24 @@ trait RegularSync {
...
+            }.sum
+
+          val newBranchTotalDifficulty = headersQueue.map(_.difficulty).sum
+
+          if (currentBranchTotalDifficulty < newBranchTotalDifficulty) {
+            val hashes = headersQueue.take(blockBodiesPerRequest).map(_.hash)
+            waitingForActor = Some(context.actorOf(SyncBlockBodiesRequestHandler.props(peer, hashes)))
+          } else {
+            scheduleResume()
+          }
```
rtkaczyk: Could you say a few words about this change? I don't understand what it's for.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -122,7 +166,7 @@ trait RegularSync {
           val (newBlocks, errorOpt) = processBlocks(blocks, blockParentTd)
 
           if(newBlocks.nonEmpty){
-            context.self ! BroadcastBlocks(newBlocks)
+            handshakedPeers.keys.foreach(actor => actor ! BroadcastBlocks(newBlocks))
```
rtkaczyk: Does that mean we'll send the same block to the same peer we received it from? Should it be "smarter"?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -141,17 +144,49 @@ class EthService(blockchain: Blockchain, blockGenerator: BlockGenerator, appStat
...
     import io.iohk.ethereum.mining.pow.PowCache._
-    val block = blockGenerator.generateBlockForMining()
-    Future.successful(Right(GetWorkResponse(
-      powHeaderHash = ByteString(kec256(BlockHeader.getEncodedWithoutNonce(block.header))),
-      dagSeed = seedForBlock(block.header.number),
-      target = ByteString((BigInt(2).pow(256) / block.header.difficulty).toByteArray)
-    )))
+
+    val blockNumber = appStateStorage.getBestBlockNumber() + 1
+    //todo delete stub
```
rtkaczyk: I would elaborate on this comment. If I hadn't talked to you I would have no idea what it's about.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
adamsmo: this is for calculation of total difficulty of branch that we currently have vs branch that we are getting from peer, if ours is "heavier" then we are are ignoring branch from peer and sheduling resume if peers branch is "heavier" then we are asking for block bodies and start replacing ours branch with branch from peer

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
adamsmo: We will send same block to the same peerActor we received it from
smart part is in peerActor and it will just ignore this message

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
```diff
@@ -75,6 +73,22 @@ class BlockGenerator(blockchainStorages: BlockchainStorages, blockchainConfig: B
         }
       }.getOrElse(Left(NoParent))
     }
+
+    result.right.foreach(b => cache.updateAndGet(new UnaryOperator[List[Block]] {
+      override def apply(t: List[Block]): List[Block] =
+        (b :: t).take(cacheLimit)
+    }))
```
rtkaczyk: Do we need this cache? We cannot produce a new block for mining until the recently mined block has been included, can we?

adamsmo: yes we can produce new block before previous was mined
for example we will get new transaction from network

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerSpec.scala
```diff
@@ -1,5 +1,9 @@
 package io.iohk.ethereum.jsonrpc
 
+import akka.actor.ActorSystem
+import akka.testkit.TestProbe
+import io.circe.Json.JString
```
rtkaczyk: We're using json4s, aren't we? I believe circe shouldn't even be in the dependencies anymore (sadly!)

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerSpec.scala
```diff
@@ -304,10 +308,12 @@ class JsonRpcControllerSpec extends FlatSpec with Matchers with ScalaFutures wit
     val storagesInstance = new SharedEphemDataSources with Storages.DefaultStorages
     val blockchain = BlockchainImpl(storagesInstance.storages)
     val blockGenerator: BlockGenerator = mock[BlockGenerator]
+    implicit val system = ActorSystem("JsonRpcControllerSpec_System")
+    val syncingController = TestProbe()
 
     val appStateStorage = mock[AppStateStorage]
     val web3Service = new Web3Service
-    val ethService = new EthService(blockchain, blockGenerator, appStateStorage)
+    val ethService = new EthService(blockchain, blockGenerator, appStateStorage, syncingController.ref)
```
rtkaczyk: We need tests here for `getWork`, `submitWork` and `submitHashrate`.

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerSpec.scala
adamsmo: 👍  good catch

rtkaczyk: Sorry, `circe` is used in `evmTest`. Only get rid off the import here

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
rtkaczyk: Maybe the question shouldn't be "can we?`, but "should we?". It's up to us what we mine. We can simply refuse to give new work to `ethminer` until we received the mined block.

rtkaczyk: Say we continuously receive new transaction and submit new blocks for mining, without including the mining results. Yes we can mine them, but without keeping the order (the parent - child relation) those block cannot form a chain, so we waste resources.

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerSpec.scala
adamsmo: added

PR #184, opened 2017-05-17
Pending transaction management & send raw transaction rpc method

LukasGasior1:
Changed:
- changed the way to subscribe to messages from peer (one PeerMessageBus instead of subscribing to PeerActor directly)
- added PendingTransactionsManager (to be used by miner)
- added sendRawTransaction rpc method

~WIP:~
~- need to update remaining tests~


src/main/scala/io/iohk/ethereum/network/PeerMessageBusActor.scala
```diff
@@ -0,0 +1,86 @@
...
+        false
+      }
+    }
+
+    override def unsubscribe(subscriber: ActorRef): Unit = {
+      subscriptions = subscriptions.filterNot(_.subscriber == subscriber)
+    }
+
+    override def publish(event: MessageFromPeer): Unit = {
+      subscriptions.filter(_.classifier.peerSelector.contains(event.peerId)).foreach(_.subscriber ! event)
```
ntallar: I'm wondering, should we also check the `messageCodes` of each subscription to see whether a subscriber is subscribe to that event in particular?

src/main/scala/io/iohk/ethereum/network/PeerMessageBusActor.scala
```diff
@@ -0,0 +1,86 @@
...
+      val subscription = Subscription(subscriber, from)
+      if (subscriptions.contains(subscription)) {
+        subscriptions = subscriptions.filterNot(_ == subscription)
+        true
+      } else {
+        false
+      }
+    }
+
+    override def unsubscribe(subscriber: ActorRef): Unit = {
```
ntallar: I'm wondering, should we have a separate `Unsubscribe` message for this function, so that we are able to call it as with `override def unsubscribe(subscriber: ActorRef, from: Classifier): Boolean`? We could also consider adding it later if necessary.

src/main/scala/io/iohk/ethereum/network/PeerMessageBusActor.scala
LukasGasior1: of course we should! nice catch

src/main/scala/io/iohk/ethereum/network/PeerMessageBusActor.scala
LukasGasior1: done

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -126,4 +132,11 @@ class EthService(blockchain: Blockchain, appStateStorage: AppStateStorage) {
       highestBlock = appStateStorage.getEstimatedHighestBlock()))
   }
 
+  def sendRawTransaction(req: SendRawTransactionRequest): Future[SendRawTransactionResponse] = {
+    import SignedTransactions._
+    val signedTransaction = rlp.decode[SignedTransaction](req.data.toArray[Byte])
```
ntallar: Should we consider the case when the decoding fails due to a wrongly encoded tx? It seems Geth returns an empty transaction hash if that were to happen.

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
```diff
@@ -0,0 +1,49 @@
...
+    case BroadcastTransaction(signedTransaction) =>
+      if (!pendingTransactions.contains(signedTransaction)) {
+        peerManager ! PeerManagerActor.BroadcastMessage(SignedTransactions(Seq(signedTransaction)))
+        pendingTransactions :+= signedTransaction
+      }
+
+    case GetPendingTransactions =>
+      sender() ! PendingTransactions(pendingTransactions)
+
+    case RemoveTransactions(signedTransactions) =>
```
ntallar: Should this be used to remove the transactions from processed blocks on regular sync? This may become complicated as we should handle when there is some blockchain reorganization and we should add txs from dropped block to the list of `pendingTransactions`, so we could create a separate task for this if needed.

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
LukasGasior1: I think more appropriate response would be JsonRpcError, won't it?

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
LukasGasior1: That definately sounds lika a separate task

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
```diff
@@ -0,0 +1,49 @@
...
+      pendingTransactions = pendingTransactions.filterNot(signedTransactions.contains)
+
+    case MessageFromPeer(signedTransactions: SignedTransactions, peerId) =>
```
ntallar: I'm wondering, AFAIK the purpose of this is to receive the broadcasted txs from peers, should we also broadcast them? Also regarding this, geth seems to be doing a more intelligent tx broadcasting considering if the peer knows a tx (https://github.com/ethereumproject/go-ethereum/blob/master/eth/handler.go#L708), maybe we should consider incorporating something similar?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
ntallar: I think are right, my mistake while reading geth code :+1:
I just tested this with geth and it does return various different error messages when the input doesn't represent an encoded tx.

LukasGasior1: This change however depends on @rtkaczyk PR as he added possibility to return error from services

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
LukasGasior1: Let's also do it in separate task (I can take it right next)

AlanVerbner: I think this is related to [this](https://github.com/ethereum/wiki/wiki/Ethereum-Wire-Protocol)

> Nodes must not resend the same transaction to a peer in the same session.

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
```diff
@@ -497,23 +553,24 @@ class SyncControllerSpec extends FlatSpec with Matchers {
...
+    peer1TestProbe.expectMsg(PeerActor.SendMessage(GetBlockBodies(Seq(newBlockHeader.hash))))
+    peerMessageBus.expectMsg(Subscribe(MessageClassifier(Set(BlockBodies.code), PeerSelector.WithId(peer1.id))))
+    peerMessageBus.reply(MessageFromPeer(BlockBodies(Seq(BlockBody(Nil, Nil))), peer1.id))
 
-    peer1.expectMsg(PeerActor.SendMessage(GetBlockBodies(Seq(newBlockHeader.hash))))
-    peer1.expectMsg(PeerActor.Subscribe(Set(BlockBodies.code)))
-    peer1.reply(PeerActor.MessageReceived(BlockBodies(Seq(BlockBody(Nil, Nil)))))
+    peerMessageBus.expectMsg(Unsubscribe(MessageClassifier(Set(BlockHeaders.code), PeerSelector.WithId(peer1.id))))
+    peerMessageBus.expectMsg(Subscribe(MessageClassifier(Set(BlockHeaders.code), PeerSelector.WithId(peer2.id))))
+    peerMessageBus.expectMsg(Unsubscribe(MessageClassifier(Set(BlockBodies.code), PeerSelector.WithId(peer1.id))))
```
ntallar: Maybe we should do `expectMsgAllOf` between this last messages sent to `peerMessageBus`? This test seems to be randomly failing: 
```
[info] - should only ask a peer once for BlockHeaders in case execution of a block were to fail *** FAILED *** (12 milliseconds)
[info]   java.lang.AssertionError: assertion failed: expected Subscribe(MessageClassifier(Set(22),WithId(PeerId(testProbe-157)))), found Unsubscribe(Some(MessageClassifier(Set(20),WithId(PeerId(testProbe-157)))))
[info]   at scala.Predef$.assert(Predef.scala:219)
[info]   at akka.testkit.TestKitBase.expectMsg_internal(TestKit.scala:388)
[info]   at akka.testkit.TestKitBase.expectMsg(TestKit.scala:364)
[info]   at akka.testkit.TestKitBase.expectMsg$(TestKit.scala:364)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:814)
[info]   at io.iohk.ethereum.blockchain.sync.SyncControllerSpec$$anon$9.<init>(SyncControllerSpec.scala:561)
[info]   at io.iohk.ethereum.blockchain.sync.SyncControllerSpec.$anonfun$new$9(SyncControllerSpec.scala:522)
```

LukasGasior1: Yes, I must have missed this place

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
LukasGasior1: @ntallar I've improved PendingTransactionsManager to handle this

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
LukasGasior1: Doing this also depends on changes from https://github.com/input-output-hk/etc-client/pull/183

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
```diff
@@ -21,29 +27,63 @@ object PendingTransactionsManager {
...
 
+  /**
+    * stores all pending transactions
+   */
   var pendingTransactions: Seq[SignedTransaction] = Nil
 
+  /**
+    * stores information which tx hashes are "known" by which peers
+    */
+  var knownTransactions: Map[ByteString, Set[PeerId]] = Map.empty
```
ntallar: Geth seems to keep at most a certain number of txs per peer https://github.com/ethereumproject/go-ethereum/blob/master/eth/peer.go#L42, should we also incorporate this? This isn't a blocker for this PR getting merged and could be a future optimization.

ntallar:
Apart from the problem on catching exceptions on decoding wrongly encoded tx, which should be easily fixed, and my last optional comment, LGTM!

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
```diff
@@ -0,0 +1,89 @@
...
+
+import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext.Implicits.global
+
+object PendingTransactionsManager {
+  def props(peerManager: ActorRef, peerMessageBus: ActorRef): Props =
+    Props(new PendingTransactionsManager(peerManager, peerMessageBus))
+
+  case class AddTransaction(signedTransaction: SignedTransaction)
+  private case class NotifyPeer(signedTransactions: Seq[SignedTransaction], peer: Peer)
```
rtkaczyk: Couldn't it just be a private method (instead of sending this message to self)?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -161,4 +165,11 @@ class EthService(blockchain: Blockchain, blockGenerator: BlockGenerator, appStat
       highestBlock = appStateStorage.getEstimatedHighestBlock())))
   }
 
+  def sendRawTransaction(req: SendRawTransactionRequest): ServiceResponse[SendRawTransactionResponse] = {
+    import io.iohk.ethereum.network.p2p.messages.CommonMessages.SignedTransactions._
+    val signedTransaction = rlp.decode[SignedTransaction](req.data.toArray[Byte])
```
rtkaczyk: A kind reminder of the now [hidden comment](https://github.com/input-output-hk/etc-client/pull/184#discussion_r117232357)

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
LukasGasior1: Not really. It's send from inside the future (where I can't update the var directly)

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
adamsmo: maybe this should use blockchain to only drop transactions if they are at certain depth in the chain, for example depth of transaction confirmation + x

adamsmo:
LGTM

PR #185, opened 2017-05-17
Decoupling handshaker from PeerActor

ntallar:
## Description

Includes decoupling the `Handshaker` from the `PeerActor` and creating a `handshaker` package for containing the handshaking. The new package contains:
* Handshaking interface: `Handshaker` and `HandshakerState`
* Ethereum handshaker implementation: the remaining of the files.

Decoupling the handshaker enables us to easily exchange the handshaking mechanism used for our future network layer for an alternative one. The handshaker is designed as a state machine, where each message and timeout processed moves it to a new state.


src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -73,42 +74,72 @@ class PeerActor(
...
-    case GetStatus => sender() ! StatusResponse(Connecting)
-  }
+        case GetStatus => sender() ! StatusResponse(Connecting)
+    }
+
+  def processingHandshaking(handshaker: Handshaker[PeerInfo], rlpxConnection: RLPxConnection,
+                            timeout: Cancellable, numRetries: Int): Receive =
+    handleSubscriptions orElse handleTerminated(rlpxConnection) orElse
+      handleDisconnectMsg orElse handlePingMsg(rlpxConnection) orElse
+      handleBlockFastDownload(rlpxConnection) orElse stashMessages orElse {
```
LukasGasior1: should we handle `handleBlockFastDownload` even though handshake isn't finished yet?

LukasGasior1: nvm, I guess it was because other peers ask us for the fork block

src/main/scala/io/iohk/ethereum/network/handshaker/HelloExchangeState.scala
```diff
@@ -0,0 +1,58 @@
+package io.iohk.ethereum.network.handshaker
+
+import akka.util.ByteString
+import io.iohk.ethereum.network.PeerActor.PeerInfo
+import io.iohk.ethereum.network.handshaker.Handshaker.NextMessage
+import io.iohk.ethereum.network.p2p.Message
+import io.iohk.ethereum.network.p2p.messages.WireProtocol.{Capability, Disconnect, Hello}
+import io.iohk.ethereum.utils.{Config, Logger, ServerStatus}
+
+case class HelloExchangeState(handshakerConfiguration: EtcHandshakerConfiguration) extends InProgressState[PeerInfo] with Logger {
```
LukasGasior1: should we also prefix this class with `Etc`?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
ntallar: Yes, you are right, we need to handle `handleBlockFastDownload` to respond for the request for the fork block

src/main/scala/io/iohk/ethereum/network/handshaker/HelloExchangeState.scala
ntallar: I agree, I will also add it for the other handshake states :+1:

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
ntallar: I was wondering, what do you think @LukasGasior1 about incorporating the handling of the response of the fork block on the handshaker? It seems to make more sense as it is part of the handshaking (and also with how the `PeerActor` will be after the network refactoring AFAIK), but we would need something similar to our previously removed `handlePeerChainCheck` https://github.com/input-output-hk/etc-client/commit/4979c7524105391c84de20f1f9296424b535b800#diff-d88dc0e9076bb9b1c58875187ebcd4acL270.

ntallar:
@rtkaczyk @LukasGasior1 I just updated this PR and did some minor changes (regarding the dependent `HandshakerResult` and `PeerInfo`) to properly merge it with `phase/4/jsonrpc`

PR #186, opened 2017-05-17
Improve network RLP implicits handling

AlanVerbner:
## Description

This refactor is the first of a series of changes in order to create different modules that would be, eventually, be used as a building blocks for future tests and changes. 

The goal is to create a network module where all the network related componentes will be located

### Changes

- All implicit conversions passed through all layers up to `FrameCodec` where replaced by the interface  `MessageSerializable`. Doing so, our code looks cleaner and no implicits need to be sent up to the bottom classes to do RLPEncoding
- As a consecuence of previous item, the serialization for each message can be easily changed (RLPx handshaking cannot be changed though)
- Decoding is also configurable, checkout `EthereumMessageDecoder`

### Next steps (in another PR)
- [ ] An sbt multi-project build will be configured
- [ ] Split files between different projects
- [ ] Create a clean interface to interact with network layer


src/evmTest/scala/io/iohk/ethereum/vm/utils/EvmTestEnv.scala
```diff
@@ -15,7 +15,7 @@ import io.iohk.ethereum.domain.{Account, Address}
 import io.iohk.ethereum.vm._
 
 object EvmTestEnv {
-  val ContractsDir = new File("target/contracts")
+  val ContractsDir = new File("node/target/contracts")
```
AlanVerbner: Fix this

src/it/scala/io/iohk/ethereum/rlp/RLPSpeedSuite.scala
```diff
@@ -1,117 +1,115 @@
...
-      case RLPList(header, txs: RLPList, uncles: RLPList) =>
-        TestBlock(headerRlpEncDec.decode(header),
-                  txs.items.map(SignedTransactions.txRlpEncDec.decode),
-                  uncles.items.map(headerRlpEncDec.decode))
-      case _ => throw new RuntimeException("Invalid Block encodeable")
-    }
-
-  }
-}
+//package io.iohk.ethereum.rlp
```
AlanVerbner: Fix this test too

src/main/scala/io/iohk/ethereum/network/p2p/Message.scala
```diff
@@ -1,60 +1,20 @@
...
-    case _ => throw new RuntimeException(s"Unknown message type: ${`type`}")
-  }
 }
 
 trait Message {
   def code: Int
 }
+
+trait MessageSerializable extends Message {
+  def toBytes(implicit di: DummyImplicit): ByteString
```
LukasGasior1: nice, I didn't know this trick

src/main/scala/io/iohk/ethereum/network/p2p/MessageSerializableImplicit.scala
```diff
@@ -0,0 +1,15 @@
+package io.iohk.ethereum.network.p2p
+
+/**
+  * Helper class
+  */
+abstract class MessageSerializableImplicit[T <: Message](val msg: T) extends MessageSerializable {
```
LukasGasior1: what actually is this class for?

AlanVerbner: It defines `equals` and `hashCode` using unlying message

LukasGasior1: oh I see, and since we no longer use type-class like mechanism, we need those, right?

AlanVerbner: yes :cry:

src/it/scala/io/iohk/ethereum/rlp/RLPSpeedSuite.scala
```diff
@@ -2,49 +2,49 @@ package io.iohk.ethereum.rlp
...
+  */
 class RLPSpeedSuite extends FunSuite
-    with PropertyChecks
-    with GeneratorDrivenPropertyChecks
-    with ObjectGenerators
-    with Logger {
+  with PropertyChecks
+  with GeneratorDrivenPropertyChecks
+  with ObjectGenerators
+  with Logger {
```
adamsmo: revert of formatting?

src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainActor.scala
```diff
@@ -145,13 +143,13 @@ class DumpChainActor(peerManager: ActorRef, peerMessageBus: ActorRef, startBlock
...
-        def dumpToFile[T](fw: FileWriter, element: (ByteString, T))(implicit enc: RLPEncoder[T]): Unit = element match {
-          case (h, v) => fw.write(s"${Hex.toHexString(h.toArray[Byte])} ${Hex.toHexString(encode(v))}\n")
+        def dumpToFile(fw: FileWriter, element: (ByteString, ByteString)): Unit = element match {
+          case (h, v) => fw.write(s"${Hex.toHexString(h.toArray[Byte])} ${Hex.toHexString(v.toArray)}\n")
         }
 
-        blockHeadersStorage.foreach(dumpToFile(headersFile, _)(BlockHeaderImplicits.headerRlpEncDec))
-        stateStorage.foreach(dumpToFile(stateTreeFile, _))
-        contractStorage.foreach(dumpToFile(contractTreesFile, _))
+        blockHeadersStorage.foreach(h => dumpToFile(headersFile, h._1 -> h._2.toBytes))
```
adamsmo: maybe instead of `h` `case (hash, value)` ?

src/main/scala/io/iohk/ethereum/network/p2p/Message.scala
adamsmo: it looks nice, I worried about maintenance, if somebody will have to change this will he know that this is just to add 2 methods with no arguments with same name, maybe some comment would help?

src/main/scala/io/iohk/ethereum/rlp/package.scala
```diff
@@ -43,4 +44,11 @@ package object rlp {
     * @throws RLPException if there is any error
     */
   def nextElementIndex(data: Array[Byte], pos: Int): Int = RLP.getItemBounds(data, pos).end + 1
+
+  trait RLPSerializable {
+    def toRLPEncodable: RLPEncodeable
+    def toBytes(implicit di: DummyImplicit): ByteString = ByteString(toBytes: Array[Byte])
+    def toBytes: Array[Byte] = encode(this.toRLPEncodable)
```
adamsmo: should this trait extend/share toBytes method with `MessageSerializable` in
https://github.com/input-output-hk/etc-client/pull/186/files#diff-07c07211a1c9ffb837c7b3895bd5321bR14

src/main/scala/io/iohk/ethereum/network/p2p/Message.scala
ntallar: You are right, I'll add a comment on this :+1:

src/main/scala/io/iohk/ethereum/rlp/package.scala
ntallar: I'm not sure about that, the idea of `MessageSerializable` not extending from `RLPSerializable` was to have `RLPSerializable` and `MessageSerializable` as decoupled as possible, so that in the future we could easily exchange `RLPSerializable` for another serializing algorithm. We also have some classes (such as `Block`) which are `RLPSerializable`s but not messages.

adamsmo: 👍

PR #187, opened 2017-05-18
keystore & personal api

rtkaczyk:


src/main/scala/io/iohk/ethereum/keystore/KeyStore.scala
```diff
@@ -0,0 +1,90 @@
...
+  def importPrivateKey(prvKey: Array[Byte], passphrase: String): Either[KeyStoreError, Address] = {
+    val pubKey = pubKeyFromPrvKey(prvKey)
+    val address = Address(kec256(pubKey))
+    save(address, prvKey, passphrase).map(_ => address)
+  }
+
+  def listAccounts(): Either[KeyStoreError, List[Address]] = {
+    val dir = new File(keyStoreDir)
+    Try {
+      if (!dir.exists() || !dir.isDirectory())
```
ntallar: I'm wondering, should we return an empty list if `dir` wasn't created yet? It seems currently return `"Could not read /tmp/etc-client/keystore"` if no account was created, and geth  returns an empty list in that scenario.

src/main/scala/io/iohk/ethereum/crypto/package.scala
```diff
@@ -45,6 +46,24 @@ package object crypto {
     generator.generateKeyPair()
   }
 
+  /** @return (privateKey, publicKey) pair */
+  def keyPairToByteArrays(keyPair: AsymmetricCipherKeyPair): (Array[Byte], Array[Byte]) = {
```
ntallar: Minor detail: We currently have a very similar method `AsymmetricCipherKeyPairSerializable.toHexStrings`, should we keep only one of them?

src/test/resources/application.conf
```diff
@@ -1,4 +1,6 @@
 etc-client {
+  keystore-dir = "/tmp/etc-client/keystore"
```
ntallar: Minor detail: Should we use a different `keystore-dir` for our tests? If not we currently clear the `KeyStore` of our node while running the `KeyStoreImpl` tests.

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcError.scala
```diff
@@ -5,10 +5,12 @@ import org.json4s.JsonAST.JValue
 case class JsonRpcError(code: Int, message: String, data: Option[JValue])
 
 // scalastyle:off magic.number
+// scalastyle:off public.methods.have.type
 object JsonRpcErrors {
   val ParseError = JsonRpcError(-32700, "An error occurred on the server while parsing the JSON text", None)
   val InvalidRequest = JsonRpcError(-32600, "The JSON sent is not a valid Request object", None)
   val MethodNotFound = JsonRpcError(-32601, "The method does not exist / is not available", None)
-  val InvalidParams = JsonRpcError(-32602, "Invalid method parameters", None)
+  def InvalidParams(msg: String = "Invalid method parameters") = JsonRpcError(-32602, msg, None)
```
LukasGasior1: minor detail, but shouldn't method names be lowercase?

src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
```diff
@@ -0,0 +1,59 @@
...
+    keyStore.listAccounts
+      .map(addresses => ListAccountsResponse(addresses.map(_.toString)))
+      .left.map(handleError)
+  }
+
+
+  private val handleError: PartialFunction[KeyStore.KeyStoreError, JsonRpcError] = {
+    case KeyStore.WrongPassphrase => InvalidPassphrase
+    case KeyStore.KeyNotFound => KeyNotFound
+    case KeyStore.IOError(msg) => LogicError(msg)
```
LukasGasior1: IOError sounds like InternalError, doesn't it?

src/main/scala/io/iohk/ethereum/keystore/Wallet.scala
```diff
@@ -0,0 +1,9 @@
+package io.iohk.ethereum.keystore
+
+
+import akka.util.ByteString
+import io.iohk.ethereum.domain.{Address, SignedTransaction, Transaction}
+
+case class Wallet(address: Address, privKey: ByteString) {
+  def signTx(tx: Transaction): SignedTransaction = ???
```
LukasGasior1: We already have `SignedTransaction.sign` maybe we can overload it with the one that accepts `Wallet`?

src/main/scala/io/iohk/ethereum/keystore/KeyStore.scala
rtkaczyk: Or maybe the dir should be created when `KeyStore` is first initialized.

src/test/resources/application.conf
rtkaczyk: I disagree. It's not very minor. It's a good point and I will change that 😀

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcError.scala
rtkaczyk: Normally, yes. But here I preferred the capitalised version for consistency with the vals. I don't have strong feeling one way or the other, would you like me to change this?

src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
rtkaczyk: Yes, it does! However I checked what geth returns. I changed permissions on its keystore directory and the error code was -32000 (different than InternalError). The name `LogicError` came from [this comment](https://github.com/ethereumproject/go-ethereum/blob/master/rpc/errors.go#L74) - can be changed.

src/main/scala/io/iohk/ethereum/keystore/Wallet.scala
rtkaczyk: And this method will be used here. I don't like making `SignedTransaction` dependent on `Wallet` because `Wallet` is a higher-level concept and that should determine the dependency direction.

src/main/scala/io/iohk/ethereum/crypto/package.scala
rtkaczyk: Well, one returns strings and the other byte arrays. I'd rather say that `AsymmetricCipherKeyPairSerializable.toHexStrings` should use `keyPairToByteArrays` underneath, instead of doing the same things manually.

OTOH, the whole `AsymmetricCipherKeyPairSerializable` seems like a bit of an overkill, especially, that we only need to store private key (we can derive the public key). So, maybe it's best to remove it altogether?

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcError.scala
LukasGasior1: Let's leave it uppercase

src/main/scala/io/iohk/ethereum/keystore/Wallet.scala
LukasGasior1: Makes sense :+1:

src/main/scala/io/iohk/ethereum/keystore/KeyStore.scala
ntallar: That seems to be a better idea :+1:

src/main/scala/io/iohk/ethereum/crypto/package.scala
ntallar: I agree, it seems better to remove `AsymmetricCipherKeyPairSerializable` and save all the keys in the same format with only the private key :+1:

PR #188, opened 2017-05-18
[EC-177] Inesrt receipts in regulat sync

adamsmo:


ntallar:
LGTM! Maybe we should add a test for the receipts returned by the Ledger if it's not too complicated?

adamsmo:
I think it is tested - receipt root is part of block header and we have tests for that what was missing is returning this list from executeBlock

PR #189, opened 2017-05-24
Decoupled handling of messages from PeerActor

ntallar:
## Description

Includes decoupling the processing of `Ethereum` messages from the `PeerActor`. This further removes the non-network aspects from the `PeerActor` so as to then move it to a separate `Network` module.


src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -48,9 +49,15 @@ trait RegularSync {
...
-    case block: BroadcastBlocks if broadcasting =>
+    case BroadcastBlocks(newBlocks) if broadcasting =>
       //FIXME: Decide block propagation algorithm (for now we send block to every peer) [EC-87]
-      peersToDownloadFrom.keys.foreach(_ ! block)
+      val blocksToSendToEachPeer = for {
+        peer <- peersToDownloadFrom.keys
+        block <- newBlocks
+      } yield (peer, block)
+      blocksToSendToEachPeer.foreach{ case (peer, block) =>
+        peer ! PeerActor.SendMessage(block)
```
adamsmo: 👍 I also fixed it in https://github.com/input-output-hk/etc-client/pull/183

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -29,16 +32,19 @@ import org.spongycastle.crypto.AsymmetricCipherKeyPair
   * and `Status` exchange).
   * Once that's done it can send/receive messages with peer (HandshakedHandler.receive).
   */
+//FIXME: MessageHandler type should be configurable, this is dependant on PR 185
```
adamsmo: does it mean that we have to merge #185 first?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -258,126 +268,78 @@ class PeerActor(
       context stop self
   }
 
+  //FIXME: As the peer chain check is part of the handshaker, it should be incorporated into the handshake
```
adamsmo: should this be addressed now?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
ntallar: Not necessarily, this could be fixed in a later PR, I left this comment to remind us of changing this for the last version of the `Network` module.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
ntallar: Maybe this was a wrongly written comment but the idea was to justify why this function was brought back (it was removed in https://github.com/input-output-hk/etc-client/commit/4979c7524105391c84de20f1f9296424b535b800#diff-d88dc0e9076bb9b1c58875187ebcd4acL270). I left a comment on PR #185 about this.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
adamsmo: 👍 maybe we can also add this comment to task to not forget about it

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
adamsmo: maybe it can just stay here, because we have other dependency on `RLPxConnectionHandler.MessageReceived` for example here:
https://github.com/input-output-hk/etc-client/pull/189/files/d4fcce558c4cb0272c967c714e1992aa6558b905#diff-d88dc0e9076bb9b1c58875187ebcd4acR297

src/main/scala/io/iohk/ethereum/network/EtcMessageHandler.scala
```diff
@@ -0,0 +1,152 @@
...
+  /**
+    * Processes the message and updates if the fork block was accepted from the peer
+    *
+    * @param message to be processed
+    * @param initialPeerInfo from before the message was processed
+    * @return new peer info with the fork block accepted value updated
+    */
+  private def handleSendingNewBlock(message: Message, initialPeerInfo: EtcPeerInfo): MessageAction = message match {
+    case b: NewBlock =>
+      if (b.block.header.number > initialPeerInfo.maxBlockNumber)
```
adamsmo: what about case when block number is equal but has greater total difficulty?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
ntallar: So we would leave this `handlePeerChainCheck` in the `PeerActor` instead of it going in the `Handshaker`? I mainly thought about moving it there as the idea for our final design would be completely decoupling the `Ethereum` messages from the `PeerActor`, and maybe add a configurable message decoder if needed (as mentioned in https://github.com/input-output-hk/etc-client/pull/186/files#diff-d88dc0e9076bb9b1c58875187ebcd4acR407).

adamsmo: 👍 makes sens

src/main/scala/io/iohk/ethereum/network/EtcMessageHandler.scala
ntallar: Nice catch! Will add a fix to send new blocks anyway if that's the case :+1:
In relation to that, although unlikely, what if the block number of `b` (the block about to send) is less that the best block number known by a peer but it's total difficulty is larger? Maybe we should we also send the block `b` to the peer anyway?

adamsmo: I think we should send block if its total difficulty is greater becaus eblockchain should be reorganised

src/main/scala/io/iohk/ethereum/network/EtcMessageHandler.scala
```diff
@@ -0,0 +1,153 @@
...
+    * Processes the message and updates if the fork block was accepted from the peer
+    *
+    * @param message to be processed
+    * @param initialPeerInfo from before the message was processed
+    * @return new peer info with the fork block accepted value updated
+    */
+  private def handleSendingNewBlock(message: Message, initialPeerInfo: EtcPeerInfo): MessageAction = message match {
+    case newBlock: NewBlock =>
+      if (newBlock.block.header.number > initialPeerInfo.maxBlockNumber ||
+        newBlock.totalDifficulty > initialPeerInfo.totalDifficulty)
```
adamsmo: I am wondering if it makes sense to send block with higher number but lower total difficulty as they will not be part of main, one reason that I can think of is to inform peer about ommers
What do you think?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -49,9 +48,15 @@ trait RegularSync {
       waitingForActor = None
       handleBlockBodies(peer, blockBodies)
 
-    case block: BroadcastBlocks if broadcasting =>
+    case BroadcastBlocks(newBlocks) if broadcasting =>
       //FIXME: Decide block propagation algorithm (for now we send block to every peer) [EC-87]
-      peersToDownloadFrom.keys.foreach(_.ref ! block)
+      val blocksToSendToEachPeer = for {
+        peer <- peersToDownloadFrom.keys
```
ntallar: That makes sense :+1:
I'm wondering, what do you think about solving this by adding a `PeerConnection` event to the `PeerEventBus` so that the `peersToDownloadFrom` is up-to-date? As you suggested this would be a nice optimization instead of periodically asking for peers in the `SyncController` and it seems to solve this issue.

LukasGasior1: That's one possible solution. Another one (and I think I like it slightly more) is to add possibility to send messages to peers in peerEventBus, so that we can do, for example: `peerEventBus ! MessageToPeers(block, PeerSelector.AllPeers)` or `MessageToPeers(block, PeerSelector.Handshaked)` etc

src/main/scala/io/iohk/ethereum/network/EtcMessageHandler.scala
ntallar: You are right, it makes sense to send this blocks so that they can be included as ommers :+1:
Regarding that, I'm wondering, should we also send new blocks that have a low block number so that they are considered for being ommers? For example geth checks the block number to see if it is a possible ommer (https://github.com/ethereumproject/go-ethereum/blob/master/eth/fetcher/fetcher.go#L312), however that would require changing how we store the known blocks by a peer (instead of only using the max block number) to something similar to how we handle known txs. What do you think about this?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
ntallar: I like your idea :+1:
Maybe it's better to create a separate task for this or add it in PR #191 as various changes were already done to `PeerEventBus` in that PR?

ntallar: Although I think having a `MessageToPeers` it's a nice solution, as adding the `PeerConnection` event (and the `PeerStatusUpdated` event for when the peer status is updated) are needed for not asking periodically for the peers in the `SyncController`, maybe we could take advantage of that to solve the problem you found as well?

src/main/scala/io/iohk/ethereum/network/EtcMessageHandler.scala
adamsmo: I think that this change may be too big, for now, I would leave it as is and allow sending block with lower total difficulty
I remember that it is considered bad practice to send blocks with lower number to peers

PR #193, opened 2017-05-25
[EC-180][EC-181] Add ommers and transactions from network to new blocks

adamsmo:
for convenience this PR is for
[EC-180] Use transaction from pool to construct new block for mining
[EC-181] Create ommers pool and use them in block creation

which are the last pieces of functionality for mining new blocks

~this PR depends on https://github.com/input-output-hk/etc-client/pull/183~


src/main/resources/application.conf
```diff
@@ -84,7 +84,7 @@ etc-client {
   }
 
   fast-sync {
-    do-fast-sync = true
+    do-fast-sync = false
```
LukasGasior1: forgot to change it back or is it intended change?

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
```diff
@@ -103,23 +103,26 @@ class SyncController(
...
             ledger: Ledger,
             validators: Validators,
-            peerMessageBus: ActorRef):
-  Props = Props(new SyncController(peerManager, appStateStorage, blockchain, blockchainStorages,
-    syncStateStorage, ledger, validators, peerMessageBus))
+            actors: DependencyActors):
+  Props = Props(new SyncController(appStateStorage, blockchain, blockchainStorages,
+    syncStateStorage, ledger, validators, actors))
+
+  case class DependencyActors(peerManager: ActorRef, peerMessageBus: ActorRef, pendingTransactionsManager: ActorRef)
```
LukasGasior1: what is the value of such grouping?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -134,7 +134,7 @@ trait FastSync {
 
     case TargetBlockTimeout =>
       blacklist(peer.id, blacklistDuration, s"did not respond with target block header (timeout), blacklisting and scheduling retry in $startRetryInterval")
-      peerMessageBus ! Unsubscribe(MessageClassifier(Set(BlockHeaders.code), PeerSelector.WithId(peer.id)))
+      actors.peerMessageBus ! Unsubscribe(MessageClassifier(Set(BlockHeaders.code), PeerSelector.WithId(peer.id)))
```
LukasGasior1: maybe we could just add `import actors._` at the top?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -53,6 +54,19 @@ trait RegularSync {
       //FIXME: Decide block propagation algorithm (for now we send block to every peer) [EC-87]
       peersToDownloadFrom.keys.foreach(_.ref ! block)
 
+    //todo improve mined block handling - add info that block was not included because of syncing
+    //we allow inclusion of mined block only if we are not syncing / reorganising chain
```
LukasGasior1: is there a separate task created for it (just so that we remember to fix this later)?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -63,12 +77,31 @@ trait RegularSync {
       }
   }
 
+  private def insertMinedBlock(block: Block, parentTd: BigInt) = {
+    val result: Either[BlockExecutionError, Seq[Receipt]] = ledger.executeBlock(block, blockchainStorages, validators)
+
+    result match {
+      case Right(receipts) =>
```
LukasGasior1: is this logic different than executing blocks we receive from others? maybe we could reuse one method to handle both cases

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -146,17 +156,44 @@ class EthService(blockchain: Blockchain, blockGenerator: BlockGenerator, appStat
...
-      target = ByteString((BigInt(2).pow(256) / block.header.difficulty).toByteArray)
-    )))
+
+    val blockNumber = appStateStorage.getBestBlockNumber() + 1
+
+    import scala.concurrent.ExecutionContext.Implicits.global
+    implicit val timeout = Timeout(5 seconds)
+    (pendingTransactionsManager ? PendingTransactionsManager.GetPendingTransactions).map{
+      case PendingTransactions(txList) =>
+        val fakeAddress = 42
```
LukasGasior1: let's not forget to change that later ;)

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -146,17 +156,44 @@ class EthService(blockchain: Blockchain, blockGenerator: BlockGenerator, appStat
...
+            Right(GetWorkResponse(
+              powHeaderHash = ByteString(kec256(BlockHeader.getEncodedWithoutNonce(b.header))),
+              dagSeed = seedForBlock(b.header.number),
+              target = ByteString((BigInt(2).pow(256) / b.header.difficulty).toByteArray)
+            ))
+          case Left(err) =>
+            log.error(s"unable to prepare block because of $err")
+            Left(JsonRpcErrors.InternalError)
+        }
+      case a =>
```
LukasGasior1: you can just do `(pendingTransactionsManager ? PendingTransactionsManager.GetPendingTransactions).mapTo[PendingTransactoins].map { pendingTransactinons => ... }`
you don't have to handle this error then (future will fail automatically)

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -146,17 +156,44 @@ class EthService(blockchain: Blockchain, blockGenerator: BlockGenerator, appStat
...
+    }
   }
 
   def submitWork(req: SubmitWorkRequest): ServiceResponse[SubmitWorkResponse] = {
-    //todo add logic for including mined block into blockchain
-    Future.successful(Right(SubmitWorkResponse(true)))
+    blockGenerator.getPrepared(req.powHeaderHash) match {
+      case Some(block) if appStateStorage.getBestBlockNumber() <= block.header.number =>
+        syncingController ! MinedBlock(block.copy(header = block.header.copy(nonce = req.nonce, mixHash = req.mixHash)))
+        pendingTransactionsManager ! PendingTransactionsManager.RemoveTransactions(block.body.transactionList)
```
LukasGasior1: there may be a case that we send MinedBlock to syncingController and it will not execute it (because it is resolving fork, for example). In such case we just remove transactions and they're lost, aren't they?

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
```diff
@@ -15,7 +15,7 @@ object PendingTransactionsManager {
   def props(peerManager: ActorRef, peerMessageBus: ActorRef): Props =
     Props(new PendingTransactionsManager(peerManager, peerMessageBus))
 
-  case class AddTransaction(signedTransaction: SignedTransaction)
+  case class AddTransaction(signedTransactions: SignedTransaction*)
```
LukasGasior1: rename to AddTransaction __s__ ?

LukasGasior1: don't you also need to update a method that handles this message?

src/main/resources/application.conf
adamsmo: forgot

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
adamsmo: no scalastyle error, and you have actors in one place which is handy because they do not have distinct type

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
adamsmo: I will add task

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
adamsmo: will add todo

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
adamsmo: yes they are, will check if we can prevent that, but it is not trivial because then we would add transactions when doing reorganization of the chain which may led to inconsistent transaction pool
for example:
while reorganizing chain we removed some transactions because they were included in blocks that we got from peer and then we include them because we received mined block while reorg in that case we should somehow filter transactions from ours newly mined block

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
adamsmo: return type is different and in case of mining we are broadcasting new block to other peers

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -126,24 +128,27 @@ trait RegularSync {
         //we have same chain prefix
         if (parent.hash == headers.head.parentHash) {
 
-          val currentBranchTotalDifficulty = headersQueue.map(_.number)
+          val oldBlocks: Seq[Option[(BlockBody, BigInt)]] = headersQueue.map(_.number)
```
rtkaczyk: `Seq[(BlockBody, BigInt)]`? It's seems we don't have any interest in those Nones.
Also, if I'm reading this correctly, blocks in `headersQueue` should be in order. So shouldn't we give up after first `getBlockByNumber` that returns `None`?

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
rtkaczyk: What scalastyle error exactly? 
This is used only for sync controller, what if we want to pass a different set of actors elsewhere? Will we create a new class for each such case?
Also the name `DependencyActors` is kind of like calling a variable `aVariable`. But it's hard to find a good name since they are unrelated actors from different packages - and this is probably the cue that they shouldn't be grouped.

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -12,11 +16,14 @@ import io.iohk.ethereum.crypto._
...
 import io.iohk.ethereum.ledger.Ledger
 import io.iohk.ethereum.mining.BlockGenerator
+import io.iohk.ethereum.utils.Logger
 import io.iohk.ethereum.utils.{BlockchainConfig, Logger}
 import org.spongycastle.util.encoders.Hex
 import io.iohk.ethereum.transactions.PendingTransactionsManager
+import io.iohk.ethereum.transactions.PendingTransactionsManager.PendingTransactions
 
 import scala.concurrent.Future
+import scala.language.postfixOps
```
rtkaczyk: Why?

rtkaczyk: Ahh, I see. It's not worth it - just write `5.seconds`

adamsmo: ok

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -182,43 +189,36 @@ class EthService(
...
-          dagSeed = seedForBlock(b.header.number),
-          target = ByteString((BigInt(2).pow(256) / b.header.difficulty).toByteArray)
-        )))
-      case Left(err) =>
-        log.error(s"unable to prepare block because of $err")
-        Future.successful(Left(JsonRpcErrors.InternalError))
-    }
+
+    import scala.concurrent.ExecutionContext.Implicits.global
+    implicit val timeout = Timeout(5 seconds)
```
rtkaczyk: Should we use a single configurable timeout for all the services?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -126,24 +128,27 @@ trait RegularSync {
...
+
+          val currentBranchTotalDifficulty: BigInt = oldBlocks.collect {
+            case Some((_, difficulty)) => difficulty
+          }.sum
 
           val newBranchTotalDifficulty = headersQueue.map(_.difficulty).sum
 
           if (currentBranchTotalDifficulty < newBranchTotalDifficulty) {
+            val transactionsToAdd = oldBlocks.collect{case Some((blockBody,_)) => blockBody.transactionList}.flatten
+            actors.pendingTransactionsManager ! PendingTransactionsManager.AddTransactions(transactionsToAdd.toList)
```
rtkaczyk: Do we test this?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -182,43 +189,36 @@ class EthService(
...
+
+    (pendingTransactionsManager ? PendingTransactionsManager.GetPendingTransactions)
+      .mapTo[PendingTransactions]
+      .map { pendingTxs =>
+        //todo get address from conf
+        val fakeAddress = 42
```
rtkaczyk: Maybe it can be initialised from conf, but we should also be able to set this using `miner_setEtherbase` RPC.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
adamsmo: 👍  for `Seq[(BlockBody, BigInt)]`
but I would go with collect

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
adamsmo: yes we do i added tests in https://github.com/input-output-hk/etc-client/pull/196 I can merge them back here
also I was thinking about merging #193 and #196 together to have one PR

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
adamsmo: it is in https://github.com/input-output-hk/etc-client/pull/196 similar as above

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
adamsmo: will add it to configuration

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: Why?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: I think that would be convenient

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
adamsmo: too many arguments for method, and I agree that it was to many
maybe we should provide actors to sync controller in different way

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
adamsmo: for me it is more convenient

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
```diff
@@ -44,11 +50,12 @@ class PendingTransactionsManager(peerManager: ActorRef, peerMessageBus: ActorRef
   peerMessageBus ! Subscribe(MessageClassifier(Set(SignedTransactions.code), PeerSelector.AllPeers))
 
   override def receive: Receive = {
-    case AddTransaction(signedTransaction) =>
-      if (!pendingTransactions.contains(signedTransaction)) {
-        pendingTransactions :+= signedTransaction
+    case AddTransactions(signedTransactions) =>
+      val transactionsToAdd = signedTransactions.filterNot(t => pendingTransactions.contains(t))
+      if (transactionsToAdd.nonEmpty) {
+        pendingTransactions = (pendingTransactions ++ transactionsToAdd).takeRight(miningConfig.txPoolSize)
```
adamsmo: @LukasGasior1 I added limit for transactions pool, I think we need this for long running node

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -126,19 +135,29 @@ trait RegularSync {
...
+          val oldBranch = headersQueue.map(_.number)
             .map(blockNumber => blockchain.getBlockByNumber(blockNumber))
-            .collect{
-              case Some(b) => b.header.difficulty
-              case None => BigInt(0)
-            }.sum
+
+          val oldBlocks: Seq[(BlockBody, BigInt)] = oldBranch
+            .collect { case Some(b) => b }
+            .map{b => (b.body, b.header.difficulty)}
```
rtkaczyk: > adamsmo 4 hours ago Member
> for me it is more convenient

It's not a matter of convenience. It's about performance - `getBlockByNumber` is a DB call, is it not?

adamsmo: 👍 yes it is, will change that

rtkaczyk: BTW, if we wanted to retain this functionality then we could simply do `val oldBranch = headersQueue.map(_.number).flatMap(blockchain.getBlockByNumbe)`. But that's not the point.

src/main/resources/application.conf
```diff
@@ -62,6 +62,14 @@ etc-client {
     }
   }
 
+  mining {
+    tx-pool-size = 1000
+    ommers-pool-size = 30
+    block-cashe-size = 30
+    coin-base = "0011223344556677889900112233445566778899"
```
LukasGasior1: minor detail: it's usually written as one word: `coinbase`
of course it's not a blocker for this PR

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -194,43 +201,42 @@ class EthService(
...
-          dagSeed = seedForBlock(b.header.number),
-          target = ByteString((BigInt(2).pow(256) / b.header.difficulty).toByteArray)
-        )))
-      case Left(err) =>
-        log.error(s"unable to prepare block because of $err")
-        Future.successful(Left(JsonRpcErrors.InternalError))
-    }
+
+    import scala.concurrent.ExecutionContext.Implicits.global
+    implicit val timeout = Timeout(miningConfig.poolingServicesTimeout)
```
LukasGasior1: minor detail: do we need that as a config param? sounds more like it should be a constant in code

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -194,43 +201,42 @@ class EthService(
...
+
+    Future.sequence(Seq(
```
LukasGasior1: if we split these into methods, like: `getOmmersFromPool`, `getTransactionsFromPool` we could make it more readable:
`(getOmmersFromPool() zip getTransactionsFromPool()) map { (ommers, transactions) => ... }`

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -194,43 +201,42 @@ class EthService(
...
+    )).map {
+      case (ommers: OmmersPool.Ommers) :: (pendingTxs: PendingTransactions) :: Nil =>
+        blockGenerator.generateBlockForMining(blockNumber, pendingTxs.signedTransactions, ommers.headers, miningConfig.coinBase) match {
+          case Right(b) =>
+            Right(GetWorkResponse(
+              powHeaderHash = ByteString(kec256(BlockHeader.getEncodedWithoutNonce(b.header))),
+              dagSeed = seedForBlock(b.header.number),
+              target = ByteString((BigInt(2).pow(256) / b.header.difficulty).toByteArray)
+            ))
+          case Left(err) =>
```
LukasGasior1: do we handle situation when block generation fails because, for example, block gas limit is exceeded? should we retry with less transactions?

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
```diff
@@ -14,19 +14,18 @@ import io.iohk.ethereum.mining.BlockGenerator.{InvalidOmmers, NoParent}
...
+class BlockGenerator(blockchainStorages: BlockchainStorages, blockchainConfig: BlockchainConfig, miningConfig: MiningConfig,
+  ledger: Ledger, validators: Validators) {
 
   val difficulty = new DifficultyCalculator(blockchainConfig)
 
-  //todo add logic
-  private val cacheLimit = 30
   private val cache: AtomicReference[List[Block]] = new AtomicReference(Nil)
 
   def generateBlockForMining(blockNumber: BigInt, transactions: Seq[SignedTransaction], ommers: Seq[BlockHeader], beneficiary: Address):
```
LukasGasior1: do we store a block we're currently mining somewhere? is that the `cache`? can we use that to implement "pending" block param (see: https://github.com/ethereum/wiki/wiki/JSON-RPC#the-default-block-parameter )

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
```diff
@@ -72,7 +71,7 @@ class BlockGenerator(blockchainStorages: BlockchainStorages, blockchainConfig: B
 
     result.right.foreach(b => cache.updateAndGet(new UnaryOperator[List[Block]] {
       override def apply(t: List[Block]): List[Block] =
-        (b :: t).take(cacheLimit)
+        (b :: t).take(miningConfig.blockCasheSize)
```
LukasGasior1: typo: blockCa**c**heSize

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
LukasGasior1: should we implement a strategy for which transactions do we drop? like: dropHead or dropTail ?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
rtkaczyk: I would rather see one global timeout for all the services (and all the asks) - should be easier that way. Then having that single timeout configured should not be polluting.

src/main/resources/application.conf
rtkaczyk: Also, as mentioned in the previous, comment we need to be able to modify the coinbase via RPC. Will having this as a config setting not interfere with its mutability?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
adamsmo: @rtkaczyk disagree ;) https://github.com/input-output-hk/etc-client/pull/193#discussion-diff-119289594R194

adamsmo:
we can do this but I would create separate task for transactions dropping strategy

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
adamsmo: currently we are returning error form service, but we should add that functionality to filter out failing transactions.
Will create task for that

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
adamsmo: we can do this but I would create separate task for transactions dropping strategy

src/main/resources/application.conf
adamsmo: other clients allow to specify coinbase in configuration so I would leave this and add override mechanism when we will implement that RPC method

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
adamsmo: changed so it will not ask again if it gets empty response for block

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
adamsmo: added EC-185

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
adamsmo: added EC-186

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -126,19 +135,35 @@ trait RegularSync {
         //we have same chain prefix
         if (parent.hash == headers.head.parentHash) {
 
-          val currentBranchTotalDifficulty = headersQueue.map(_.number)
-            .map(blockNumber => blockchain.getBlockByNumber(blockNumber))
-            .collect{
-              case Some(b) => b.header.difficulty
-              case None => BigInt(0)
-            }.sum
+          var oldBranch: Seq[Option[Block]] = Nil
```
rtkaczyk: We can do it in a more functional way and get rid off those unnecessary Options:

```scala
          def getOldBlocks(headers: List[BlockHeader]): List[Block] = headers match {
            case h :: tail =>
              blockchain.getBlockByNumber(h.number).map(_ :: getOldBlocks(tail)).getOrElse(Nil)
            case Nil =>
              Nil
          }

          val oldBranch = getOldBlocks(headersQueue.toList)


          val currentBranchTotalDifficulty: BigInt = oldBranch.map(_.header.difficulty).sum

          val newBranchTotalDifficulty = headersQueue.map(_.difficulty).sum

          if (currentBranchTotalDifficulty < newBranchTotalDifficulty) {
            val transactionsToAdd = oldBranch.flatMap(_.body.transactionList)
            pendingTransactionsManager ! PendingTransactionsManager.AddTransactions(transactionsToAdd)
            val hashes = headersQueue.take(blockBodiesPerRequest).map(_.hash)
            waitingForActor = Some(context.actorOf(SyncBlockBodiesRequestHandler.props(peer, peerMessageBus, hashes)))
            //add first block from branch as ommer
            oldBranch.headOption.foreach { h => ommersPool ! AddOmmers(h.header) }
          } else {
            //add first block from branch as ommer
            headersQueue.headOption.foreach { h => ommersPool ! AddOmmers(h) }
            scheduleResume()
          }
```

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
```diff
@@ -103,16 +103,21 @@ class SyncController(
...
-    syncStateStorage, ledger, validators, peerMessageBus))
+            actors: DependencyActors):
+  Props = Props(new SyncController(appStateStorage, blockchain, blockchainStorages,
+    syncStateStorage, ledger, validators, actors))
+
+  case class DependencyActors(
+    peerManager: ActorRef,
+    peerMessageBus: ActorRef,
+    pendingTransactionsManager: ActorRef,
+    ommersPool: ActorRef)
```
rtkaczyk: We are we doing about this?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -126,19 +134,27 @@ trait RegularSync {
...
-          val currentBranchTotalDifficulty = headersQueue.map(_.number)
-            .map(blockNumber => blockchain.getBlockByNumber(blockNumber))
-            .collect{
-              case Some(b) => b.header.difficulty
-              case None => BigInt(0)
-            }.sum
+          val oldBranch: Seq[Block] = getOldBlocks(headersQueue)
+
+          val oldBlocks: Seq[(BlockBody, BigInt)] = oldBranch
+            .map{b => (b.body, b.header.difficulty)}
```
rtkaczyk: As shown in the code snippet I pasted, this is not needed. In fact it's easier to use `oldBranch` directly.

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -194,43 +201,52 @@ class EthService(
...
+            log.error(s"unable to prepare block because of $err")
+            Left(JsonRpcErrors.InternalError)
+        }
+      }
+  }
+
+  private def getOmmersFromPool = {
+    import scala.concurrent.ExecutionContext.Implicits.global
```
rtkaczyk: We have 3 of those imports now? Should we just import it once for the whole class?

adamsmo: Importing it once for the whole class does not work

rtkaczyk: So the import doesn't work because we're passing the execution context to each method now? Why? Can't we pass it to the whole class? @LukasGasior1 do you know something about it?

adamsmo: i think it is for testing

LukasGasior1: why doesn't it work when you import it just once at the top? compiler error?

adamsmo: yes compilation error, you can check it locally

rtkaczyk: ok, leave it, I will prepare a separate PR which will fix it.

PR #194, opened 2017-05-25
Add eth_call rpc method

LukasGasior1:
This pr adds the eth_call method to simulate tx execution


src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -215,4 +237,44 @@ class EthService(blockchain: Blockchain, blockGenerator: BlockGenerator, appStat
...
+  private def resolveBlock(blockParam: Either[BigInt, String]): Either[JsonRpcError, Block] = {
+    def getBlock(number: BigInt): Either[JsonRpcError, Block] = {
+      blockchain.getBlockByNumber(number)
+        .map(Right.apply)
+        .getOrElse(Left(JsonRpcErrors.InvalidParams(s"Block $number not found")))
+    }
+
+    blockParam match {
+      case Left(blockNumber) => getBlock(blockNumber)
+      case Right("earliest") => getBlock(0)
```
adamsmo: maybe case object instead of string?

src/it/scala/io/iohk/ethereum/txExecTest/ContractTest.scala
```diff
@@ -56,4 +56,5 @@ class ContractTest extends FlatSpec with Matchers {
     //block contains contract paying 2 accounts
     new LedgerImpl(VM, blockchainConfig).executeBlock(fixtures.blockByNumber(3), storage, validators) shouldBe noErrors
   }
+
```
adamsmo: revert?

adamsmo:
other then mentioned LGTM

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -140,4 +141,39 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
...
+      }
+
+    def encodeJson(t: CallResponse): JValue = encodeAsHex(t.returnData)
+
+    def extractTx(obj: JObject): Either[JsonRpcError, CallTx] = {
+      def toEitherOpt[A, B](opt: Option[Either[A, B]]): Either[A, Option[B]] =
+        opt.map(_.right.map(Some.apply)).getOrElse(Right(None))
+
+      for {
+        from <- toEitherOpt((obj \ "from").extractOpt[String].map(tryExtractUnformattedData))
```
rtkaczyk: What does `(obj \ "from").extractOpt[String]` return if the `from` not a string (eg. `{"from": 42}`) ?

LukasGasior1: so for int, like 42, it will actually work and extract a String out of it, but it would return `None` in case of nested object, for example

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -136,6 +138,26 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
...
 
+  override def simulateTransaction(stx: SignedTransaction, blockHeader: BlockHeader, storages: BlockchainStorages, validators: Validators): TxResult = {
+    val blockchain = BlockchainImpl(storages)
+    val stateRoot = blockchain.getBlockHeaderByHash(blockHeader.hash).map(_.stateRoot)
+    val world = InMemoryWorldStateProxy(storages, stateRoot)
+
+    val gasLimit = stx.tx.gasLimit
+    val config = EvmConfig.forBlock(blockHeader.number, blockchainConfig)
+
+    val context: PC = stx.tx.receivingAddress match {
```
ntallar: I'm wondering, could we use the `prepareProgramContext` function to create this `ProgramContext`? The only difference that I see is that we currently don't transfer the tx value to the receiver of the tx, should we transfer it?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -136,6 +138,26 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
         }
   }
 
+  override def simulateTransaction(stx: SignedTransaction, blockHeader: BlockHeader, storages: BlockchainStorages, validators: Validators): TxResult = {
```
ntallar: I'm wondering, should `updateSenderAccountBeforeExecution` be called here as in `Ledger.executeTransaction`?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
LukasGasior1: The problem was that sender account may not exist (and it's required for the transfer inside `prepareProgramContext`). But I can just make sure there's at least an empty account for sender before calling this fn.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
LukasGasior1: You're right. I'll add it

LukasGasior1:
@ntallar I've addressed your concerns

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -215,4 +247,39 @@ class EthService(blockchain: Blockchain, blockGenerator: BlockGenerator, appStat
...
+      blockchain.getBlockByNumber(number)
+        .map(Right.apply)
+        .getOrElse(Left(JsonRpcErrors.InvalidParams(s"Block $number not found")))
+    }
+
+    blockParam match {
+      case BlockParam.WithNumber(blockNumber) => getBlock(blockNumber)
+      case BlockParam.Earliest => getBlock(0)
+      case BlockParam.Latest => getBlock(appStateStorage.getBestBlockNumber())
+      case BlockParam.Pending => getBlock(appStateStorage.getBestBlockNumber())
```
ntallar: The pending block should be obtained from the pending txs, righ? Should we add a comment and create a task for obtaining it?

LukasGasior1: Yes definately, I'll create a task for it.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -136,6 +138,30 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
...
```
ntallar: The `validators` parameter seems to currently not being used, should we start using it or maybe remove it?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -136,6 +138,30 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
...
+    val blockchain = BlockchainImpl(storages)
+    val stateRoot = blockchain.getBlockHeaderByHash(blockHeader.hash).map(_.stateRoot)
```
ntallar: Minor detail: I'm wondering, do we need to access the `blockchain` to get the `stateRoot`? Could we just access the `blockHeader` parameter?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
LukasGasior1: It's not used indeed, removing

PR #195, opened 2017-05-29
personal API: unlock wallet and sign TX

rtkaczyk:


src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
```diff
@@ -43,6 +46,71 @@ trait JsonMethodsImplicits {
...
+  protected def extractAddress(input: JString): Either[JsonRpcError, Address] =
+    extractAddress(input.s)
+
+  protected def extractBytes(input: String): Either[JsonRpcError, ByteString] =
+    Try(ByteString(decode(input))).toEither.left.map(_ => InvalidParams())
+
+  protected def extractBytes(input: JString): Either[JsonRpcError, ByteString] =
+    extractBytes(input.s)
+
+  protected def extractQuantity(input: JValue): Either[JsonRpcError, BigInt] =
```
ntallar: We currently have a `tryExtractQuantity` function for extracting the quantity out of a `JString`, as it is not being use in this function, maybe we should remove it and replace it with this one?

rtkaczyk: Yeah, the other extractors didn't suit me (though the differences are small), so I created my own versions. I agree it's better we stick with one version, but I wanted to get this reviewed first.

rtkaczyk:
scalastyle problems will be fixed along with review updates

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -80,6 +80,13 @@ trait Blockchain {
   def getMptNodeByHash(hash: ByteString): Option[MptNode]
 
 
+  def getAccount(address: Address): Option[Account] = {
+    val hash = crypto.kec256(address.bytes)
+    getMptNodeByHash(hash).collect {
```
ntallar: I'm wondering, for getting the account given a certain address shouldn't we use a MPT and use state root hash of the best block?

src/main/scala/io/iohk/ethereum/keystore/Wallet.scala
```diff
@@ -2,8 +2,13 @@ package io.iohk.ethereum.keystore
...
 import io.iohk.ethereum.domain.{Address, SignedTransaction, Transaction}
+import org.spongycastle.crypto.AsymmetricCipherKeyPair
 
-case class Wallet(address: Address, privKey: ByteString) {
-  def signTx(tx: Transaction): SignedTransaction = ???
+case class Wallet(address: Address, prvKey: ByteString) {
+  lazy val keyPair: AsymmetricCipherKeyPair = keyPairFromPrvKey(prvKey.toArray)
+
+  def signTx(tx: Transaction): SignedTransaction =
+    SignedTransaction.sign(tx, keyPair, None)
```
ntallar: I'm wondering should we use our node's `chainId` for signing the tx?

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
rtkaczyk: I assumed that's exactly what `getMptNodeByHash` does. Does it not?

src/main/scala/io/iohk/ethereum/keystore/Wallet.scala
rtkaczyk: I believe @adamsmo has discovered that Parity is not using the new signing scheme for ETC (no chain ID). Adam, can you confirm?

adamsmo: yes, exactly that, parity accepts new signing schema only when chainId = networkId
but in case of ETC chainId = 3d and networkId = 1

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
ntallar: Not exactly, `getMptNodeByHash` fetches a MPT node by its hash from the storage, in this case we would need to construct a new MPT from the state root hash to get the corresponding leaf.

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -151,6 +160,16 @@ class BlockchainImpl(
...
 
+  override def getAccount(address: Address, blockNumber: BigInt): Option[Account] =
+    getBlockHeaderByNumber(blockNumber).flatMap { bh =>
+      val mpt = new MerklePatriciaTrie[Address, Account](
+        Some(bh.stateRoot.toArray),
+        nodeStorage,
+        crypto.kec256(_: Array[Byte])
+      )
+      mpt.get(address)
+    }
```
rtkaczyk: @ntallar the correct (hopefully) impl of `getAccount`. I'm still not sure if `Blockchain` is the best place to put this method.

adamsmo:
there was compilation failure

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -92,10 +94,10 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
   implicit val eth_getTransactionByBlockHashAndIndex =
     new JsonDecoder[GetTransactionByBlockHashAndIndexRequest] with JsonEncoder[GetTransactionByBlockHashAndIndexResponse] {
       override def decodeJson(params: Option[JArray]): Either[JsonRpcError, GetTransactionByBlockHashAndIndexRequest] = params match {
-        case Some(JArray((blockHash: JString) :: (transactionIndex: JString) :: Nil)) =>
+        case Some(JArray(JString(blockHash) :: transactionIndex :: Nil)) =>
```
adamsmo: `case Some(JArray(JString(blockHash) :: JString(transactionIndex) :: Nil))` ?
params are specified as
```
params: [
   '0xe670ec64341771606e55d6b4ca35a1a6b75ee3d5145a99d05921026d1527331',
   '0x0' // 0
]
```

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -107,10 +109,10 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
   implicit val eth_getUncleByBlockHashAndIndex = new JsonDecoder[UncleByBlockHashAndIndexRequest] with JsonEncoder[UncleByBlockHashAndIndexResponse] {
     override def decodeJson(params: Option[JArray]): Either[JsonRpcError, UncleByBlockHashAndIndexRequest] =
       params match {
-        case Some(JArray((blockHash: JString) :: (uncleIndex: JString) :: Nil)) =>
+        case Some(JArray(JString(blockHash) :: uncleIndex :: Nil)) =>
```
adamsmo: `uncleIndex` => `JString(uncleIndex)` ?

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
rtkaczyk: Geth allows you to specify numbers both a hex strings, and as regular JSON numbers (decimal). I don't see why we would deprive ourselves of such convenience.

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
rtkaczyk: same as above

rtkaczyk:
> there was compilation failure

yeah, I did not verify the merge, should be good now

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -151,6 +160,16 @@ class BlockchainImpl(
 
   override def getTotalDifficultyByHash(blockhash: ByteString): Option[BigInt] = totalDifficultyStorage.get(blockhash)
 
+  override def getAccount(address: Address, blockNumber: BigInt): Option[Account] =
+    getBlockHeaderByNumber(blockNumber).flatMap { bh =>
+      val mpt = new MerklePatriciaTrie[Address, Account](
+        Some(bh.stateRoot.toArray),
```
ntallar: Minor detail: Maybe we should use the `def apply[K, V](rootHash: Array[Byte], source: NodeStorage, hashFn: HashFn): MerklePatriciaTrie[K, V]` constructor? I think this constructor should be private right?

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -92,10 +94,10 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
   implicit val eth_getTransactionByBlockHashAndIndex =
     new JsonDecoder[GetTransactionByBlockHashAndIndexRequest] with JsonEncoder[GetTransactionByBlockHashAndIndexResponse] {
       override def decodeJson(params: Option[JArray]): Either[JsonRpcError, GetTransactionByBlockHashAndIndexRequest] = params match {
-        case Some(JArray((blockHash: JString) :: (transactionIndex: JString) :: Nil)) =>
+        case Some(JArray(JString(blockHash) :: transactionIndex :: Nil)) =>
           for {
-            parsedBlockHash <- tryExtractBlockHash(blockHash)
-            parsedTransactionIndex <- tryExtractQuantity(transactionIndex)
+            parsedBlockHash <- extractBytes(blockHash)
```
ntallar: I'm wondering, maybe we should we use `extractHash` here?

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
rtkaczyk: Agree, I haven't event noticed the factory methods.

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -151,6 +160,16 @@ class BlockchainImpl(
 
   override def getTotalDifficultyByHash(blockhash: ByteString): Option[BigInt] = totalDifficultyStorage.get(blockhash)
 
+  override def getAccount(address: Address, blockNumber: BigInt): Option[Account] =
+    getBlockHeaderByNumber(blockNumber).flatMap { bh =>
+      val mpt = new MerklePatriciaTrie[Address, Account](
```
ntallar: Should we use something similar to https://github.com/input-output-hk/etc-client/blob/master/src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala#L96 and use `HashByteArraySerializable` so that the addresses are hashed before using them to access the MPT? It should be just adding the use of `(HashByteArraySerializable(addressSerializer), accountSerializer)` or maybe extracting this in some way (as it is shared with `InMemoryWorldStateProxy`).

rtkaczyk: I thought that the `hashFn` hashes the address (what is it for then?). Good catch. The serializer in `Address` comp obj has been updated.

ntallar: That `hashFn` is used to configure the function used to hash the encoded nodes for storing them in the `nodeStorage` (so it's a used as a key-value storage as `hashFn(encode(node))->encode(node)`, which is different from the key-value map `hashFn2(address)->account` that is the MPT created). Although in Ethereum the hash function used for the storage is always `kec256`, it is configurable on the MPT.

ntallar:
Apart from CircleCI currently failing, LGTM!

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
rtkaczyk: Thanks, I think I get it, but don't be surprised if ask again ;)

src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
```diff
@@ -29,58 +29,84 @@ trait JsonMethodsImplicits {
...
+      case JInt(n) =>
+        Right(n)
 
-  protected def tryExtractUnformattedData(input: JString, requiredLength: Int, errorMsg: String): Either[JsonRpcError, ByteString] =
-    tryExtractUnformattedData(input.values) match {
-      case Right(bytes) if bytes.length == requiredLength => Right(bytes)
-      case Right(_) => Left(JsonRpcErrors.InvalidParams(errorMsg))
-      case Left(err) => Left(err)
+      case JString(s) =>
+        Try(BigInt(decode(s))).toEither.left.map(_ => InvalidParams())
```
ntallar: I'm wondering should we check here (or in `decode`) that `s` has no leading zeroes as mentioned in https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding?

rtkaczyk: I deliberately relaxed the requirements listed in the wiki to match Geth behaviour (which also happens to be more convenient when manually playing with the RPC). However, I admit I haven't checked this case with Geth. I can do that on Monday. Either way, I can't see this causing any trouble.

ntallar: I agree, this shouldn't be a blocker for merging this PR :+1:

rtkaczyk: I checked and Geth accepts leading zeroes

PR #196, opened 2017-05-29
[EC-181] Create ommers pool and use them in block creation

adamsmo:
this PR depends on https://github.com/input-output-hk/etc-client/pull/193


adamsmo:
this is now part of #193

PR #197, opened 2017-05-30
Add some eth_ RPC methods

LukasGasior1:


src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -19,7 +19,7 @@ import io.iohk.ethereum.transactions.PendingTransactionsManager
 import scala.concurrent.Future
 import scala.util.{Failure, Success, Try}
 
-
+// scalastyle:off number.of.methods number.of.types
```
adamsmo: maybe we should split EthService into logical parts?
for example: mining and blockchain

LukasGasior1: We could, but as long as methods implementations are concise I don't see much value in doing so (other than make scalastyle happy).

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -264,6 +276,42 @@ class EthService(
     }
   }
 
+  def getCode(req: GetCodeRequest): ServiceResponse[GetCodeResponse] = {
+    Future.successful {
+      resolveBlock(req.block).map { block =>
+        val world = InMemoryWorldStateProxy(blockchainStorages, Some(block.header.stateRoot))
```
ntallar: I'm wondering, although it is quite easy to use an `InMemoryWorldStateProxy`, as it has a structure and design around it for saving changes before persisting them (for the `Ledger`), should we only use a simple MPT for this? Maybe we could have also a `WorldState` without this extra cache? This shouldn't be a blocker for merging this PR though

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -264,6 +276,42 @@ class EthService(
...
+    }
+  }
+
+  def getUncleCountByBlockHash(req: GetUncleCountByBlockHashRequest): ServiceResponse[GetUncleCountByBlockHashResponse] = {
+    Future.successful {
+      blockchain.getBlockBodyByHash(req.blockHash) match {
+        case Some(blockBody) =>
+          Right(GetUncleCountByBlockHashResponse(blockBody.uncleNodesList.size))
+        case None =>
+          Left(JsonRpcErrors.InvalidParams(s"Block with hash ${req.blockHash} not found"))
```
ntallar: Minor detail: should we use `Hex.toHexString` for printing the block hash?

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -174,6 +175,60 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
...
         data = data.getOrElse(ByteString("")))
     }
+  }
+
+  implicit val eth_getCode = new JsonDecoder[GetCodeRequest] with JsonEncoder[GetCodeResponse] {
+    def decodeJson(params: Option[JArray]): Either[JsonRpcError, GetCodeRequest] =
+      params match {
+        case Some(JArray((address: JString) :: (blockStr: JString) :: Nil)) =>
+          for {
+            addr <- tryExtractUnformattedData(address)
```
ntallar: Minor detail: Should we check whether the length of `addr` is correct? At least `geth` returns a `"invalid argument 0: hex string has length 42, want 40 for Address" ` in that case

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -174,6 +175,60 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
...
+
+    def encodeJson(t: GetUncleCountByBlockNumberResponse): JValue = encodeAsHex(t.result)
+  }
+
+  implicit val eth_getUncleCountByBlockHash = new JsonDecoder[GetUncleCountByBlockHashRequest] with JsonEncoder[GetUncleCountByBlockHashResponse] {
+    def decodeJson(params: Option[JArray]): Either[JsonRpcError, GetUncleCountByBlockHashRequest] =
+      params match {
+        case Some(JArray((blockHashStr: JString) :: Nil)) =>
+          for {
+            blockHash <- tryExtractUnformattedData(blockHashStr)
```
ntallar: Minor detail: As with `eth_getCode` should we also check here that the `blockHashStr` has the correct length and return an error message instead of `null`?

ntallar:
Apart from my comments (and that there was a timeout on CircleCI), LGTM!

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
LukasGasior1: I think these additional features don't cause any overhead when just calling `getCode`, do they?

ntallar: You are right, it doesn't cause any additional overhead. So I think we can leave as it is :+1:

src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
```diff
@@ -29,6 +32,19 @@ trait JsonMethodsImplicits {
   protected def tryExtractUnformattedData(input: JString): Either[JsonRpcError, ByteString] =
     tryExtractUnformattedData(input.values)
 
+  protected def tryExtractAddress(input: JString): Either[JsonRpcError, ByteString] =
+    tryExtractUnformattedData(input, AddressLength, s"Invalid length for address ${input.values}, expected 40 bytes")
```
adamsmo: `expected 20 bytes` ?

LukasGasior1: Fixed

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
adamsmo: are we testing those methods? I mean `getCode`, `getUncleCountByBlockNumber`, `getUncleCountByBlockHash` and `getBlockTransactionCountByNumber`

LukasGasior1: @adamsmo I've added the tests for `getUncleCountByBlockNumber`, `getUncleCountByBlockHash` and `getBlockTransactionCountByNumber`. I'm still struggling to provide tests data for `getCode`.

LukasGasior1: Added test for `getCode`

adamsmo: 👍

rtkaczyk: > I think these additional features don't cause any overhead when just calling getCode, do they?

They induce additional level of coupling which may increase the later effort of separating the Ledger or the VM.

AlanVerbner: I'm not sure if it's too late but it seems a little bit strange for me too

LukasGasior1: I'm going to change this in my next PR @rtkaczyk @AlanVerbner @ntallar

PR #200, opened 2017-06-01
Add eth_balance, getStorageAt and getTransactionCount rpc methods

LukasGasior1:


src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -278,9 +290,8 @@ class EthService(
 
   def getCode(req: GetCodeRequest): ServiceResponse[GetCodeResponse] = {
     Future.successful {
-      resolveBlock(req.block).map { block =>
-        val world = InMemoryWorldStateProxy(blockchainStorages, Some(block.header.stateRoot))
```
LukasGasior1: no more `InMemoryWorldStateProxy` in EthService :)

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -312,6 +323,60 @@ class EthService(
     }
   }
 
+  def getBalance(req: GetBalanceRequest): ServiceResponse[GetBalanceResponse] = {
+    Future.successful {
```
rtkaczyk: `Future.successful`? It's a DB call, isn't it?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -312,6 +323,60 @@ class EthService(
...
+      withAccount(Address(req.address), req.block) { account =>
+        GetTransactionCountResponse(account.nonce)
+      }
+    }
+  }
+
+  private def withAccount[T](address: Address, blockParam: BlockParam)(f: Account => T): Either[JsonRpcError, T] = {
+    withStateMpt(blockParam) { stateMpt =>
+      val account = stateMpt.get(kec256(address.toArray)).getOrElse(Account.Empty)
```
rtkaczyk: We will need to access accounts in several places, so I'd rather have address serialization defined in a single place (it has caused us problems before), like [here](https://github.com/input-output-hk/etc-client/pull/195/files#diff-112cd375a8fe272c1310b5f66c605c9dR13)

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -312,6 +323,60 @@ class EthService(
...
+      f(account)
+    }
+  }
+
+  private def withStateMpt[T](blockParam: BlockParam)(f: MerklePatriciaTrie[Array[Byte], Account] => T): Either[JsonRpcError, T] = {
+    import MerklePatriciaTrie.defaultByteArraySerializable
+    resolveBlock(blockParam).map { block =>
+      val stateMpt = MerklePatriciaTrie[Array[Byte], Account](block.header.stateRoot.toArray[Byte],
+        blockchainStorages.nodeStorage, (input: Array[Byte]) => crypto.kec256(input))
```
rtkaczyk: Since `PersonalService` also builds, the MPT in a similar way, I'd rather have this as in a trait. I will try adding on my PR.

rtkaczyk: Wait... I forgot I already [done that differently](https://github.com/input-output-hk/etc-client/pull/195/files#diff-abb3f698c5418831dcb5f183fbf61937R163) 😉 (though I'm still hoping for some confirmation whether putting it in `Blockchain` is OK)

This is only for account. We could add a similar for storage. What else do we need to access via MPT?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -312,6 +323,60 @@ class EthService(
...
+  }
+
+  def getStorageAt(req: GetStorageAtRequest): ServiceResponse[GetStorageAtResponse] = {
+    import io.iohk.ethereum.rlp.UInt256RLPImplicits._
+
+    val byteArrayUInt256Serializer = new ByteArrayEncoder[UInt256] {
+      override def toBytes(input: UInt256): Array[Byte] = input.bytes.toArray[Byte]
+    }
+
+    val rlpUInt256Serializer = new ByteArraySerializable[UInt256] {
```
rtkaczyk: We already have those serializers in `InMemoryWorldStateProxy`. We should probably move them to a common place.

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
LukasGasior1: Will change (also some other methods should use `Future {` as well)

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
LukasGasior1: How about we just define it in MPT companion obj? So that we can obtain it using just:
```
val mpt = MerklePatriciaTrie.stateMpt(block.stateRootHash, nodeStorage)
mpt.get(address)
...
```

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
LukasGasior1: how about `mpt` package obj? (where ByteArraySerializable is defined)

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
rtkaczyk: That would be OK. We'd also need `MerklePatriciaTrie.accountStorageMpt(...)`.

But what do you think about `blockchain.getAccount(address, stateRootHash)` and `blockain.getAccountStorage(address, stateRootHash)`?  I think that's somewhat nicer to look at, I'm just not sure if `Blockchain` was meant to be used that way.

Also having those in `Blockchain` separates MPT implementation from its application.

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
rtkaczyk: Ideally MPT doesn't need to know about the objects it stores, but it should be easy enough to move those serializers elsewhere if we want to make the `mpt` package more uncoupled in the future.

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
rtkaczyk: So I added the serializers to `Address` comp object because of how it was done for `Account`. But now that I think of it, those serializers should also reside in `mpt` package object. It makes more sense to me for the storage to know about the objects it stores, rather than the objects knowing how they are stored.

[This remark](https://github.com/input-output-hk/etc-client/pull/200#discussion_r119828406) still holds, though.

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
AlanVerbner: Imo mpt should only have rlp (and any other common stuff) as dependency but not any model or storage class. It's just a data structure, we could potentially replace it by a hashmap for example.

LukasGasior1: sounds good

LukasGasior1: I'll wait for your PR and upade the code

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
LukasGasior1: how about a trait MptSerializers in `domain` package?

rtkaczyk: Then we make the `domain` package dependent on `mpt` which, imo, is even worse. I think that should be a third layer. But for now I'm fine with either approach as long those serializers are easy to move (single file).

@mcsherrylabs @AlanVerbner @ntallar as this possibly relates to the ongoing refactoring effort (beyond `network`) maybe you'd like to voice your opinion?

AlanVerbner: Mpt will become a module or a part of another one (I'm actually towards creating a a separate one as Mpt is a data structure itself and can be used in other contexts) 

That being said, I think that your solution, @rtkaczyk, it's the best one.

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
rtkaczyk: #195 has been merged

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -148,6 +158,16 @@ class BlockchainImpl(
                       protected val totalDifficultyStorage: TotalDifficultyStorage
                     ) extends Blockchain {
 
+
+  private val byteArrayUInt256Serializer = new ByteArrayEncoder[UInt256] {
```
AlanVerbner: The following two are also implemented here https://github.com/input-output-hk/etc-client/blob/phase/4/jsonrpc/src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala#L17

Maybe it should be extracted to a common place, what do you think?

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -170,6 +190,14 @@ class BlockchainImpl(
       mpt.get(address)
     }
 
+  override def getAccountStorageAt(rootHash: ByteString, position: BigInt): ByteString = {
+    val storageMpt =
+      MerklePatriciaTrie[UInt256, UInt256](rootHash.toArray[Byte], nodeStorage,
```
AlanVerbner: This is done here as well https://github.com/input-output-hk/etc-client/blob/phase/4/jsonrpc/src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala#L118

LukasGasior1:
@AlanVerbner I've moved those to `domain` package obj

PR #201, opened 2017-06-01
[EC-183] Add hashrate reporting

adamsmo:
This PR adds `eth_hashrate` and implements `eth_submitHashrate`
it depends on #193

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
@@ -24,11 +24,16 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
   }
 
   implicit val eth_submitHashrate = new JsonDecoder[SubmitHashRateRequest] with JsonEncoder[SubmitHashRateResponse] {
+    val lengthWithPrefix = 66
+
     override def decodeJson(params: Option[JsonAST.JArray]): Either[JsonRpcError, SubmitHashRateRequest] = params match {
-      case Some(JArray((hashRate: JString) :: (id: JString) :: Nil)) =>
-        tryExtractQuantity(hashRate)
-          .flatMap(h => tryExtractUnformattedData(id).map(i => (h, i)))
-          .map { case (h, i) => SubmitHashRateRequest(h, i) }
+      case Some(JArray((hashRate: JString) :: (id: JString) :: Nil))
+        if hashRate.s.length == lengthWithPrefix && id.s.length == lengthWithPrefix =>
+        val result: Either[JsonRpcError, SubmitHashRateRequest] = for {
+          rate <- tryExtractQuantity(hashRate)
+          miner <- tryExtractUnformattedData(id)
+        } yield SubmitHashRateRequest(rate, miner)
+        result
       case _ =>
         Left(InvalidParams())
     }
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
@@ -25,14 +25,11 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
   }
 
   implicit val eth_submitHashrate = new JsonDecoder[SubmitHashRateRequest] with JsonEncoder[SubmitHashRateResponse] {
-    val lengthWithPrefix = 66
-
     override def decodeJson(params: Option[JsonAST.JArray]): Either[JsonRpcError, SubmitHashRateRequest] = params match {
-      case Some(JArray((hashRate: JString) :: (id: JString) :: Nil))
-        if hashRate.s.length == lengthWithPrefix && id.s.length == lengthWithPrefix =>
+      case Some(JArray(hashRate :: JString(id) :: Nil)) =>
         val result: Either[JsonRpcError, SubmitHashRateRequest] = for {
           rate <- extractQuantity(hashRate)
-          miner <- extractBytes(id)
+          miner <- extractHash(id)
         } yield SubmitHashRateRequest(rate, miner)
         result
       case _ =>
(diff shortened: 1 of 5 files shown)


src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -190,10 +200,34 @@ class EthService(
   }
 
   def submitHashRate(req: SubmitHashRateRequest): ServiceResponse[SubmitHashRateResponse] = {
-    //todo do we care about hash rate for now?
+    hashRate.updateAndGet(new UnaryOperator[Map[ByteString, (BigInt, Date)]] {
+      override def apply(t: Map[ByteString, (BigInt, Date)]): Map[ByteString, (BigInt, Date)] = {
+        val now = new Date
+        clearTimeouts(now, t + (req.id -> (req.hashRate, now)))
```
ntallar: Should we check here or in `EthJsonMethodImplicits` that `req.id` and `req.hashrate` have a length of 32 bytes as mentioned in https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_submithashrate

ntallar:
Apart from my comment, LGTM!

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -190,10 +200,34 @@ class EthService(
...
+      override def apply(t: Map[ByteString, (BigInt, Date)]): Map[ByteString, (BigInt, Date)] = {
+        clearTimeouts(new Date, t)
+      }
+    })
+
+    //sum all reported hashRates
+    Future.successful(Right(GetHashRateResponse(hashRates.mapValues { case (hr, _) => hr }.values.sum)))
+  }
+
+  private def clearTimeouts(now: Date, rates: Map[ByteString, (BigInt, Date)]):Map[ByteString, (BigInt, Date)]={
```
LukasGasior1: maybe let's rename it to `removeObsoleteHashrates` or something similar?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
rtkaczyk: `EthJsonMethodImplicits` is definitely the place (`extractHash`), but it seems `eth_submitHashrate` method has not been added.

adamsmo: @rtkaczyk https://github.com/input-output-hk/etc-client/pull/201/files#diff-324f2e7f056cd8ea85ee130bf26f18f0L26 ?

rtkaczyk: Ah, I was expecting it to be added in this PR.
Why 66?
BTW, this won't compile after merge with phase branch.

adamsmo: because 64 for bytes and 2 for `0x`
ok will update

rtkaczyk: Wait, hashrate is a number, so it should parsed as such. For example geth is fine with input like `123` (JSON int). This is what `extractQuantity` method should be used for.

Regarding `0x`, see comments on #195

adamsmo: ok but in documentation it says
`Hashrate, a hexadecimal string representation (32 bytes) of the hash rate`

rtkaczyk: Yes, documentation says... ;)
Actually there is no conflict here, because the doc specifies strict rules on the input, and we're simply relaxing them (just like Geth, for convenience of manual interaction).  The input accepted by us is a superset of what's accepted in the doc, without any ambiguity.

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -25,18 +25,32 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
...
 
   implicit val eth_submitHashrate = new JsonDecoder[SubmitHashRateRequest] with JsonEncoder[SubmitHashRateResponse] {
+    val lengthWithPrefix = 66
+
     override def decodeJson(params: Option[JsonAST.JArray]): Either[JsonRpcError, SubmitHashRateRequest] = params match {
-      case Some(JArray(hashRate :: JString(id) :: Nil)) =>
-        extractQuantity(hashRate)
-          .flatMap(h => extractBytes(id).map(i => (h, i)))
-          .map { case (h, i) => SubmitHashRateRequest(h, i) }
+      case Some(JArray((hashRate: JString) :: (id: JString) :: Nil))
```
rtkaczyk: `case Some(JArray(hashRate :: JString(id) :: Nil))`

`hashRate` - because we want to accept JSON ints as input.
`JString(id)` - because it's just nicer :)

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -25,18 +25,32 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
...
+        if hashRate.s.length == lengthWithPrefix && id.s.length == lengthWithPrefix =>
+        val result: Either[JsonRpcError, SubmitHashRateRequest] = for {
+          rate <- extractQuantity(hashRate)
+          miner <- extractBytes(id)
```
rtkaczyk: `extractHash` for `id` will check the length of `id`. `lengthWithPrefix` check is unnecessary.

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
adamsmo: ok, so you saying that we should just ignore length check?

rtkaczyk: yes

PR #204, opened 2017-06-02
PeerMessageBus refactor

AlanVerbner:
## Description

Peer message bus implementation changed in order to:

- Don't send the notification twice if a subscription is made with two `MessageClassifier` with same message code, eg: `MessageClassifier(1,2,3)` and `MessageClassifier(2)` should only trigger once
- Allow users to Unsubscribe by code message code instead of the entire `MessageClassifier`

### DIFF
--- a/src/main/scala/io/iohk/ethereum/network/PeerMessageBusActor.scala
+++ b/src/main/scala/io/iohk/ethereum/network/PeerMessageBusActor.scala
@@ -27,57 +27,64 @@ object PeerMessageBusActor {
 
   case class MessageFromPeer(message: Message, peerId: PeerId)
 
-  case class Subscription(subscriber: ActorRef, classifier: MessageClassifier)
-
   class PeerMessageBus extends ActorEventBus {
 
     override type Event = MessageFromPeer
     override type Classifier = MessageClassifier
 
-    private var subscriptions: Seq[Subscription] = Nil
+    //FIXME Remove this var
+    private var subscriptions: Map[(Subscriber, PeerSelector), Set[Int]] = Map.empty
 
     override def subscribe(subscriber: ActorRef, to: Classifier): Boolean = {
-      val subscription = Subscription(subscriber, to)
-      if (subscriptions.contains(subscription)) {
-        false
-      } else {
-        subscriptions = subscriptions :+ subscription
+      val newSubscriptions = subscriptions.get((subscriber, to.peerSelector)) match {
+        case Some(messageCodes) =>
+          if (messageCodes == to.messageCodes) subscriptions
+          else subscriptions + ((subscriber, to.peerSelector) -> (messageCodes ++ to.messageCodes))
+        case None =>  subscriptions + ((subscriber, to.peerSelector) -> to.messageCodes)
+      }
+      if(newSubscriptions == subscriptions) false
+      else {
+        subscriptions = newSubscriptions
         true
       }
     }
 
     override def unsubscribe(subscriber: ActorRef, from: Classifier): Boolean = {
-      val subscription = Subscription(subscriber, from)
-      if (subscriptions.contains(subscription)) {
-        subscriptions = subscriptions.filterNot(_ == subscription)
-        true
-      } else {
-        false
+      subscriptions.get((subscriber, from.peerSelector)).exists { messageCodes =>
+        val newMessageCodes = messageCodes -- from.messageCodes
+        if (messageCodes == newMessageCodes) false
+        else {
+          if(newMessageCodes.isEmpty) subscriptions = subscriptions - (subscriber -> from.peerSelector)
+          else subscriptions = subscriptions + ((subscriber, from.peerSelector) -> newMessageCodes)
+          true
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/io/iohk/ethereum/network/PeerMessageBusActor.scala
```diff
@@ -27,57 +27,64 @@ object PeerMessageBusActor {
...
-  case class Subscription(subscriber: ActorRef, classifier: MessageClassifier)
-
   class PeerMessageBus extends ActorEventBus {
 
     override type Event = MessageFromPeer
     override type Classifier = MessageClassifier
 
-    private var subscriptions: Seq[Subscription] = Nil
+    //FIXME Remove this var
+    private var subscriptions: Map[(Subscriber, PeerSelector), Set[Int]] = Map.empty
```
adamsmo: should we remove this var now?

AlanVerbner: We can but I rather to do it on another PR as it was not introduced in this fix. 

@adamsmo do you think is worthy to do it now? :innocent:

adamsmo: ok, I think it is worth creating task to not forgot about this

AlanVerbner: done `EC-197`

adamsmo:
LGTM

src/main/scala/io/iohk/ethereum/network/PeerMessageBusActor.scala
```diff
@@ -27,57 +27,64 @@ object PeerMessageBusActor {
...
-      if (subscriptions.contains(subscription)) {
-        subscriptions = subscriptions.filterNot(_ == subscription)
-        true
-      } else {
-        false
+      subscriptions.get((subscriber, from.peerSelector)).exists { messageCodes =>
+        val newMessageCodes = messageCodes -- from.messageCodes
+        if (messageCodes == newMessageCodes) false
+        else {
+          if(newMessageCodes.isEmpty) subscriptions = subscriptions - (subscriber -> from.peerSelector)
```
adamsmo: I think changing `from.peerSelector` to `messageCodes` would make this easier to read

adamsmo: maybe `(subscriber, from.peerSelector)` to be consistent?

AlanVerbner: Fixed!

src/main/scala/io/iohk/ethereum/network/PeerMessageBusActor.scala
```diff
@@ -27,57 +27,64 @@ object PeerMessageBusActor {
...
-      if (subscriptions.contains(subscription)) {
-        subscriptions = subscriptions.filterNot(_ == subscription)
-        true
-      } else {
-        false
+      subscriptions.get((subscriber, from.peerSelector)).exists { messageCodes =>
+        val newMessageCodes = messageCodes -- from.messageCodes
+        if (messageCodes == newMessageCodes) false
+        else {
+          if(newMessageCodes.isEmpty) subscriptions = subscriptions - ((subscriber, from.peerSelector))
```
adamsmo: `((subscriber, from.peerSelector))` <- are double parentheses required?

src/main/scala/io/iohk/ethereum/network/PeerMessageBusActor.scala
LukasGasior1: I think this var doesn't hurt much as long as it's encapsulated in an actor

src/main/scala/io/iohk/ethereum/network/PeerMessageBusActor.scala
```diff
@@ -27,57 +27,64 @@ object PeerMessageBusActor {
...
 
     override def subscribe(subscriber: ActorRef, to: Classifier): Boolean = {
-      val subscription = Subscription(subscriber, to)
-      if (subscriptions.contains(subscription)) {
-        false
-      } else {
-        subscriptions = subscriptions :+ subscription
+      val newSubscriptions = subscriptions.get((subscriber, to.peerSelector)) match {
+        case Some(messageCodes) =>
+          if (messageCodes == to.messageCodes) subscriptions
```
LukasGasior1: this `if` isn't really needed, since it's a `Set` we can just return `subscriptions + ((subscriber, to.peerSelector) -> (messageCodes ++ to.messageCodes))`, can't we?

AlanVerbner: you are right! fixed.

src/main/scala/io/iohk/ethereum/network/PeerMessageBusActor.scala
AlanVerbner: At least for me, it doesn't compile otherwise

PR #205, opened 2017-06-05
EC-194 Fix regular sync not working while importing blocks

adamsmo:


src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -162,9 +162,9 @@ trait RegularSync {
   }
 
   def getOldBlocks(headers: Seq[BlockHeader]): List[Block] = headers match {
```
ntallar: I'm wondering, if it's not too difficult, maybe we can adapt this function to be tail recursive and use the `@tailrec` annotation?

PR #206, opened 2017-06-05
EC-192 Implement eth_mining

adamsmo:


src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -213,25 +221,43 @@ class EthService(
...
+  }
+
   def getHashRate(req: GetHashRateRequest): ServiceResponse[GetHashRateResponse] = {
     val hashRates: Map[ByteString, (BigInt, Date)] = hashRate.updateAndGet(new UnaryOperator[Map[ByteString, (BigInt, Date)]] {
       override def apply(t: Map[ByteString, (BigInt, Date)]): Map[ByteString, (BigInt, Date)] = {
         removeObsoleteHashrates(new Date, t)
       }
     })
-
+    reportActive()
```
ntallar: Is `getHashRate` used by `ethminer`? I wasn't able to find if it's used, but If it isn't we might not need to `reportActive` when this RPC method is called, right? I might be missing something.

ntallar:
Apart from my comment, LGTM!

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
adamsmo: it is not, it is a mistake, will fix

PR #207, opened 2017-06-05
Added eth_getBlockByNumber and eth_getUncleByBlockNumberAndIndex

ntallar:
## Description

Includes `eth_getBlockByNumber` and `eth_getUncleByBlockNumberAndIndex` JSON RPC methods.

**Note:** Although according to https://github.com/ethereum/wiki/wiki/JSON-RPC#returns-26 `number`, `hash`, `nonce` and `logsBloom` should be `null` when the block tag is `Pending`, to keep the compatibility with geth the parameters selected as `null` were chosen as mentioned in https://github.com/ethereumproject/go-ethereum/blob/master/eth/api.go#L578.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
@@ -233,10 +233,9 @@ class EthService(
   def getUncleByBlockNumberAndIndex(request: UncleByBlockNumberAndIndexRequest): ServiceResponse[UncleByBlockNumberAndIndexResponse] = Future {
     val UncleByBlockNumberAndIndexRequest(blockParam, uncleIndex) = request
     val uncleHeaderOpt = resolveBlock(blockParam).toOption
-      .flatMap { uncleBlock =>
-        val uncleBody = uncleBlock.body
-        if (uncleIndex >= 0 && uncleIndex < uncleBody.uncleNodesList.size)
-          Some(uncleBody.uncleNodesList.apply(uncleIndex.toInt))
+      .flatMap { block =>
+        if (uncleIndex >= 0 && uncleIndex < block.body.uncleNodesList.size)
+          Some(block.body.uncleNodesList.apply(uncleIndex.toInt))
         else
           None
       }
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
@@ -41,11 +41,17 @@ object EthService {
   case class BlockByBlockHashRequest(blockHash: ByteString, fullTxs: Boolean)
   case class BlockByBlockHashResponse(blockResponse: Option[BlockResponse])
 
+  case class BlockByNumberRequest(block: BlockParam, fullTxs: Boolean)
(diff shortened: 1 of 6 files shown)


src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -198,10 +219,34 @@ class EthService(
...
+        if (uncleIndex >= 0 && uncleIndex < block.body.uncleNodesList.size)
+          Some(block.body.uncleNodesList.apply(uncleIndex.toInt))
+        else
+          None
+      }
+    val totalDifficulty = uncleHeaderOpt.flatMap(uncleHeader => blockchain.getTotalDifficultyByHash(uncleHeader.hash))
+
+    //The block in the response will not have any txs or uncles
+    val uncleBlockResponseOpt = uncleHeaderOpt.map { uncleHeader =>
+      BlockResponse(blockHeader = uncleHeader, totalDifficulty = totalDifficulty, pendingBlock = blockParam == BlockParam.Pending) }
```
adamsmo: maybe `resolveBlock` should return information if block is pending
maybe something like
`val (uncleHeaderOpt, pending) = resolveBlock(blockParam)` ?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -157,11 +163,26 @@ class EthService(
...
+    * Implements the eth_getBlockByNumber method that fetches a requested block.
+    *
+    * @param request with the block requested (by it's number or by tag)
+    * @return the block requested or None if the client doesn't have the block
+    */
+  def getBlockByNumber(request: BlockByNumberRequest): ServiceResponse[BlockByNumberResponse] = Future {
+    val BlockByNumberRequest(blockParam, fullTxs) = request
+    val blockResponseOpt = resolveBlock(blockParam).toOption.map { block =>
+      val totalDifficulty = blockchain.getTotalDifficultyByHash(block.header.hash)
+      BlockResponse(block, totalDifficulty, fullTxs = fullTxs, pendingBlock = blockParam == BlockParam.Pending)
```
adamsmo: same as in line 246

src/test/scala/io/iohk/ethereum/jsonrpc/EthServiceSpec.scala
```diff
@@ -105,6 +105,70 @@ class EthServiceSpec extends FlatSpec with Matchers with ScalaFutures with MockF
     response.transactionResponse shouldBe Some(expectedTxResponse)
   }
 
+  //FIXME: This test should be changed once the pending block doesn't equal to the latest one
```
adamsmo: should we address it now?

ntallar: We should probably address it soon (there's currently an open task for that https://iohk.myjetbrains.com/youtrack/issue/EC-184) but I think it will be better to handle this in a separate PR.

adamsmo: 👍 added comment to that task

PR #208, opened 2017-06-05
[EC-198] eth_estimateGas

AlanVerbner:
## Description

`eth_estimateGas` implementation based on the same code as `eth_call`. As this is a the first approach, `EC-199` was created in order to do further investigation to check if we also need to do something like https://github.com/ethereum/go-ethereum/pull/3587 (which is implemented in parity too  https://github.com/paritytech/parity/blob/e6a31e7543e33b925908c8abc424060a28127290/ethcore/src/client/client.rs#L1045)

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
@@ -213,11 +213,15 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
       } yield CallTx(
         from = from,
         to = to,
-        gas = gas.getOrElse(0),
+        gas = gas,
         gasPrice = gasPrice.getOrElse(0),
         value = value.getOrElse(0),
         data = data.getOrElse(ByteString("")))
     }
+  }
+
+  implicit val eth_estimateGas = new JsonEncoder[EstimateGasResponse] {
+    override def encodeJson(t: EstimateGasResponse): JValue = encodeAsHex(t.gas)
   }
 
   implicit val eth_getCode = new JsonDecoder[GetCodeRequest] with JsonEncoder[GetCodeResponse] {
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
@@ -386,7 +386,8 @@ class EthService(
       }
     }
     val tx = Transaction(0, req.tx.gasPrice, gasLimit, toAddress, req.tx.value, req.tx.data)
-    val stx = SignedTransaction(tx, ECDSASignature(0, 0, 0.toByte), fromAddress)
+    val fakeSignature = ECDSASignature(0, 0, 0.toByte)
+    val stx = SignedTransaction(tx, fakeSignature, fromAddress)
 
     resolveBlock(req.block).map { block =>
       ledger.simulateTransaction(stx, block.header, blockchainStorages)
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
@@ -444,18 +444,21 @@ class EthService(
 
     val toAddress = req.tx.to.map(Address.apply)
 
-    val gasLimit = req.tx.gas.getOrElse {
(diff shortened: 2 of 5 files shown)


src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -213,13 +213,17 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
       } yield CallTx(
         from = from,
         to = to,
-        gas = gas.getOrElse(0),
```
AlanVerbner: gas is an optional value but settining it to 0 leads to a 0 as result

AlanVerbner: > See eth_call parameters, expect that all properties are optional. If no gas limit is specified geth uses the block gas limit from the pending block as an upper bound.

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -376,4 +369,27 @@ class EthService(
...
+      keyStore
+        .listAccounts().getOrElse(Nil).headOption // first account, if exists and `from` param not specified
+        .getOrElse(Address(0))) // 0x0 default
+
+    val toAddress = req.tx.to.map(Address.apply)
+
+    val gasLimit = req.tx.gas.getOrElse {
+      resolveBlock(BlockParam.Latest) match {
+        case Right(b) => b.header.gasLimit
+        case Left(err) => BigInt(0)
```
adamsmo: if latest is not avaliable maybe we can provide some constant value that allows to execute transaction, with 0 tx will always fail, it requires gas even to do ether transfer
we can also treat case when we are not able to get latest block as error and return JsonRpcError for internal error

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -376,4 +369,27 @@ class EthService(
...
+      }
+    }
+    val tx = Transaction(0, req.tx.gasPrice, gasLimit, toAddress, req.tx.value, req.tx.data)
+    val stx = SignedTransaction(tx, ECDSASignature(0, 0, 0.toByte), fromAddress)
```
adamsmo: `ECDSASignature(0, 0, 0.toByte)` <- this is not obviouse and at first glance looks strange, maybe we can extract this to val and add name
something like `fakeSignature` ?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
AlanVerbner: @adamsmo please checkout `EC-199` where some investigation in order to improve this method is proposed. Let's chat If you think your suggestion should be done before that.

adamsmo: we can handle this in separate PR, but I would add todo here or add link to this line in EC-199 with some comments to not forget about this

AlanVerbner: @adamsmo just addressed this comment based on your suggestion :wink:

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala
```diff
@@ -120,6 +120,8 @@ class JsonRpcController(
       handle[SendTransactionRequest, SendTransactionResponse](personalService.sendTransaction, req)
     case req @ JsonRpcRequest(_, "eth_call", _, _) =>
       handle[CallRequest, CallResponse](ethService.call, req)
+    case req @ JsonRpcRequest(_, "eth_estimateGas", _, _) =>
+      handle[CallRequest, EstimateGasResponse](ethService.estimateGas, req)
```
ntallar: I've been testing `eth_estimateGas` and could it be that in geth the block parameter is not configurable (as in `eth_call`)? It seems to be always `Pending` https://github.com/ethereumproject/go-ethereum/blob/master/eth/api.go#L792.

AlanVerbner: Well.....[the reference](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_estimategas) doesn't state that :disappointed:

PR #209, opened 2017-06-06
[EC-195] Implement eth_gasPrice

adamsmo:


rtkaczyk:
The code looks good, I'm not sure about the logic though. We are supposed to recommend a gas price. I'm not sure if the average is a good recommendation (I don't think it's terrible). 

Here's what [geth does](https://github.com/ethereum/go-ethereum/blob/c7a4d9cf8ae307a94d16dd8197824eecd320cdf7/eth/gasprice/gasprice.go#L76), adn here's [geth classic](https://github.com/ethereumproject/go-ethereum/blob/5ad3edba36368a8fb61b34d1f9ab33a151e33890/eth/gasprice.go#L196). I didn't figure out what exactly they do, but neither looks like it's simply taking the average.

adamsmo:
for me it looks like geth is returning lowest price from a block with `getBlockPrices` than if the price is greater than some max value it clips it and if not it sorts results and pics one with from the last position after discarding top `gpo.percentile`
and geth classic is estimating minimal price with which transaction would be included into block base on historical blocks and there is more logic there, but I am not sure if it helps to estimate better price because this is similar (if not the same) as predicting stock market prices

We can not predict the future so even with more complicated logic we can predict price which will be wrong (transaction will not be included), I will argue that when picking lowest possible price probability of not including transaction is higher

So I decided to stick with simples solution - ruling average, I can change that to picking lowest price from X latest blocks and  then getting max price from those after discarding Y percentiles (geth implementation)

rtkaczyk:
Makes sense, but the problem I see with average is that a single TX could greatly offset the value. If we're after a simple solution maybe a median would fit better?

> change that to picking lowest price from X latest blocks and then getting max price from those after discarding Y percentiles (geth implementation)

This is also interesting and perhaps not too laborious?

The other issue I see with this, is that in a fixed X number of blocks there could be very few or no TXs at all (at least theoretically).

With all that said, I'm fine with this solution. If you're going to stick with it I would just appreciate a comment like "this is simply taking the average, may require improvement in the future".

adamsmo:
👍 will switch to max from latest min
about
> The other issue I see with this, is that in a fixed X number of blocks there could be very few or no TXs at all (at least theoretically).

theoretically yes but if you set X high enough probability is negligible

adamsmo:
I was checking few latest blocks and found that price is fixed at 20000 000 000 wei

rtkaczyk:
Yes, it's the [default gas price](https://ethereum.stackexchange.com/a/12588)

adamsmo:
as discussed at VSU we will stick with this naive estimate and add new task to improve it in the next PR

PR #212, opened 2017-06-08
[EC-200] Implement eth_getTransactionByBlockNumberAndIndex

adamsmo:


src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -137,6 +137,21 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
         t.transactionResponse.map(Extraction.decompose).getOrElse(JNull)
     }
 
+  implicit val eth_d  =
+    new JsonDecoder[GetTransactionByBlockNumberAndIndexRequest] with JsonEncoder[GetTransactionByBlockNumberAndIndexResponse] {
+      override def decodeJson(params: Option[JArray]): Either[JsonRpcError, GetTransactionByBlockNumberAndIndexRequest] = params match {
+        case Some(JArray(JString(blockStr) :: transactionIndex :: Nil)) =>
```
ntallar: Should we use `blockStr: JString` instead of `JString(blockStr)` to allow `blockStr` to contain a `JInt` (as done for example in `eth_getBalance`)?

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -137,6 +137,21 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
         t.transactionResponse.map(Extraction.decompose).getOrElse(JNull)
     }
 
+  implicit val eth_d  =
```
ntallar: Minor detail: Maybe we could name this `val` as `eth_getTransactionByBlockNumberAndIndex`?

AlanVerbner:
@adamsmo I'm having an issue when testing this. I don't think is related to this PR so feel free to create a new task if you think it's out of scope. Other tests I ran were ok so It might not be a blocker to merge.

The following request works with other clients but not in ours:

`jsonrpc="2.0" "method"="eth_getTransactionByBlockNumberAndIndex" params:='["0xC005", "0x0"]' id=1`

If I change block number by `0x0C005` it does

adamsmo:
@AlanVerbner this is because:
```
scala> BigInt(0xC005)
res0: scala.math.BigInt = 49157
```
but
```
scala> BigInt(Hex.decode("C005"))
res2: scala.math.BigInt = -16379
```
fix is:
```
scala> BigInt(1,Hex.decode("C005"))
res4: scala.math.BigInt = 49157
```
so yes this is not related to this PR, but will fix it, good catch

PR #213, opened 2017-06-08
[EC-202] Implement personal_ecRecover

adamsmo:


src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
```diff
@@ -87,6 +91,19 @@ class PersonalService(
     Right(LockAccountResponse(true))
   }
 
+  def ecRecover(req: EcRecoverRequest): ServiceResponse[EcRecoverResponse] = Future {
+    import req._
+    val prefixed: Array[Byte] =
+      Array[Byte](19.toByte) ++
+        s"Ethereum Signed Message:\n${message.length}".map(c => c.toByte).toArray[Byte] ++
+        message.toArray[Byte]
```
rtkaczyk: [In geth](https://github.com/ethereum/go-ethereum/blob/1a0eb903f1e84772b14f399e519b4a3408137fdc/internal/ethapi/api.go#L361) the character is in hexadecimal, here you have a decimal. Could that be the problem?

src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
```diff
@@ -192,22 +195,19 @@ object JsonMethodsImplicits extends JsonMethodsImplicits {
   }
 
   implicit val personal_ecRecover = new Codec[EcRecoverRequest, EcRecoverResponse] {
-    val signatureLength = 65
-    val rLength = 32
-    val vLength = 32
 
     def decodeJson(params: Option[JArray]): Either[JsonRpcError, EcRecoverRequest] =
       params match {
-        case Some(JArray(JString(message) :: JString(signature) :: _)) if signature.length == (signatureLength * 2 + 2) =>
```
AlanVerbner: Isn't this if statement needed?

src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
```diff
@@ -192,22 +195,19 @@ object JsonMethodsImplicits extends JsonMethodsImplicits {
...
             msg <- extractBytes(message)
             sig <- extractBytes(signature)
           } yield (msg, sig)
 
           decoded.flatMap { case (msg, sig) =>
-            val r = sig.take(rLength)
-            val s = sig.drop(rLength).take(vLength)
+            val r = sig.take(ECDSASignature.RLength)
+            val s = sig.drop(ECDSASignature.RLength).take(ECDSASignature.SLength)
             val v = sig.takeRight(1)
```
AlanVerbner: What about `ECDSASignature.VLength` ?

src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
rtkaczyk: I added a method but forgot to use it, see now

src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
rtkaczyk: 👍

PR #214, opened 2017-06-08
standardise keystore

rtkaczyk:
This PR is about making our keystore meet the standards set out in: https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition

Implemented:
- 2 AES cipher modes: `CBC` and `CTR`
- 2 key derivation functions: `scrypt` and `pbkdf2-hmac-sha256`

Both parity and geth follow this standard. Both use `AES-CTR`. Geth uses `scrypt`, and parity uses `pbkdf2`.

The names of key JSON files are not defined in the standard. This implementation name the files along the pattern `UTC--[ISO8601 date]--[account address]`. Geth and parity use similar but different (also among themselves) names. Our implementation relies on the address being part of the name (it can be changed).

~~At the moment the PR lacks good test vectors for AES-CBC and AES-CTR ciphers.~~


src/test/scala/io/iohk/ethereum/crypto/AesCbcSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+        "This is a 48-byte message (exactly 3 AES blocks)",
+        "d0a02b3836451753d493665d33f0e8862dea54cdb293abc7506939276772f8d5021c19216bad525c8579695d83ba2684"),
+
+      ("56e47a38c5598974bc46903dba290349",
+        "8ce82eefbea0da3c44699ed7db51b7d9",
+        new String(
+          Hex.decode("a0a1a2a3a4a5a6a7a8a9aaabacadaeafb0b1b2b3b4b5b6b7b8b9babbbcbdbebfc0c1c2c3c4c5c6c7c8c9cacbcccdcecfd0d1d2d3d4d5d6d7d8d9dadbdcdddedf"),
+          StandardCharsets.US_ASCII),
+        "c30e32ffedc0774e6aff6af0869f71aa0f3af07a9a31a9c684db207eb0ef8e4e35907aa632c3ffdf868bb7b29d3d46ad83ce9f9a102ee99d49a53e87f4c3da55")
+    )
```
rtkaczyk: These test vectors do not work because they are not meant to be tested with padding. The plaintext messages are divisible by the block size. The output produced by our cipher is 1 block greater in size. I believe this is correct for PKCS#5/PKCS#7 padding (should verify), but it would still be good to find test vectors with messages of irregular sizes.

Most importantly however, the last test vector doesn't work at all (the output is totally different irrespective of padding).

There are several options here:
- remove CBC - this was hard-fixed as default encryption mode in the previous versions of the keystore standard, but it isn't used by the latest versions of geth and parity.
- only rely on the second test (random) - we assume our encryption is correct and that the last test vector is broken
- find better tests vectors

src/main/scala/io/iohk/ethereum/keystore/EncryptedKey.scala
```diff
@@ -0,0 +1,87 @@
...
+  */
+case class EncryptedKey(
+  id: UUID,
+  address: Address,
+  cryptoSpec: CryptoSpec,
+  version: String
+) {
+
+  def decrypt(passphrase: String): Either[String, ByteString] = {
+    val cipher = getCipher(cryptoSpec.cipher)
```
ntallar: Minor detail: As the `cipher` is obtained when calculating the decrypted key, this variable is currently not being used.

src/main/scala/io/iohk/ethereum/keystore/EncryptedKeyJsonCodec.scala
```diff
@@ -0,0 +1,88 @@
...
+    Hex.toHexString(bs.toArray)
+
+  def toJson(encKey: EncryptedKey): String = {
+    import encKey._
+    import cryptoSpec._
+
+    val json =
+      ("id" -> id.toString) ~
+      ("address" -> asHex(address.bytes)) ~
+      ("version" -> version) ~
```
ntallar: Minor detail: Should we write the `version` as `JInt`? I tried using the keys generated by our client in geth and it seems it expects a `JInt` as the version. (I should double check but I think the same thing happens in parity)

src/main/scala/io/iohk/ethereum/crypto/SymmetricCipher.scala
```diff
@@ -0,0 +1,44 @@
...
+    val offset = cipher.processBytes(data.toArray, 0, data.size, output, 0)
+    val len = cipher.doFinal(output, offset)
+
+    ByteString(output).take(offset + len)
+  }
+}
+
+object AES_CBC extends SymmetricCipher {
+  protected def getCipher =
+    new PaddedBufferedBlockCipher(new CBCBlockCipher(new AESEngine), new PKCS7Padding)
```
ntallar: I'm wondering, why is `PaddedBufferedBlockCipher` needed here instead of `BufferedBlockCipher` as in `AES_CTR`? I tried doing so and it seems without the padding there seems to be no need to drop bytes on `AesCbcSpec`.

adamsmo: CBC mode requires padding
I see in `CBCBlockCipher` that it throws exception `DataLengthException` if there isn't enough data

build.sbt
```diff
@@ -22,7 +22,8 @@ val dep = {
     "com.typesafe.akka" %% "akka-http-testkit" % akkaHttpVersion % "it,test",
     "io.suzaku" %% "boopickle" % "1.2.6",
     "org.consensusresearch" %% "scrypto" % "1.2.0-RC3",
-    "com.madgag.spongycastle" % "core" % "1.54.0.0",
+    "com.madgag.spongycastle" % "core" % "1.56.0.0",
+    "com.madgag.spongycastle" % "prov" % "1.56.0.0",
```
adamsmo: do we need provider?

src/test/scala/io/iohk/ethereum/crypto/AesCbcSpec.scala
adamsmo: I think that plain text encoding is wrong i last test case:
I see in debug that all characters are: '�' 65533

ntallar:
I'm having an issue when testing, maybe I'm doing something wrong but it seems that the `personal_newAccount` method is taking too long to run and I'm getting the error:
```
The server was not able to produce a timely response to your request.
Please try again in a short while!
```

src/main/scala/io/iohk/ethereum/keystore/EncryptedKeyJsonCodec.scala
rtkaczyk: From the wiki:

> Version is intrinsically numeric yet is a string (structured versioning would be possible with a string, but can be considered out of scope for a rarely changing configuration file format).

I sometimes don't know why I even bother with reading the docs... Yes, it should be numeric.

rtkaczyk:
> I'm having an issue when testing, maybe I'm doing something wrong but it seems that the personal_newAccount method is taking too long to run and I'm getting the error

@ntallar I'd say it must be your setup. I used this exact branch in the demo last week where I showed `personal_newAccount` working. Is that the only method that fails for you?

build.sbt
rtkaczyk: not anymore 👍

src/test/scala/io/iohk/ethereum/crypto/AesCbcSpec.scala
rtkaczyk: Yep, the second vector worked so I thought the decoding method was OK

src/main/scala/io/iohk/ethereum/crypto/SymmetricCipher.scala
rtkaczyk: Now I'm thinking maybe we shouldn't use padding because the encrypted private keys are already aligned with the block size (only 16-byte cipher keys are used).

This all guessing though, cause I don't have any test vectors for it :/

adamsmo: but you can have case where private key is shorter, private key is just integer, for example value `42` is valid private key value

rtkaczyk: I have exchanged CBC encrypted keys with [this tool](https://github.com/ethereumjs/keythereum) (both ways) and it worked fine so I'm leaving the padding as is

rtkaczyk:
@adamsmo @ntallar I've addressed all your comments

src/test/scala/io/iohk/ethereum/crypto/ScryptSpec.scala
```diff
@@ -0,0 +1,35 @@
...
+    // https://datatracker.ietf.org/doc/rfc7914/?include_text=1
+    val testVectors = Table[String, String, Int, Int, Int, Int, String](
+      ("passphrase", "salt", "n", "r", "p", "dklen", "derivedKey"),
+      ("", "", 16, 1, 1, 64,
+        "77d6576238657b203b19ca42c18a0497f16b4844e3074ae8dfdffa3fede21442fcd0069ded0948f8326a753a0fc81f17e8d3e0fb2e0d3628cf35e20c38d18906"),
+      ("password", "NaCl", 1024, 8, 16, 64,
+        "fdbabe1c9d3472007856e7190d01e9fe7c6ad7cbc8237830e77376634b3731622eaf30d92e22a3886ff109279d9830dac727afb94a83ee6d8360cbdfa2cc0640"),
+      // takes a bit too long:
+      // ("pleaseletmein", "SodiumChloride", 1048576, 8, 1, 64,
+      //   "2101cb9b6a511aaeaddbbe09cf70f881ec568d574a2ffd4dabe5ee9820adaa478e56fd8f4ba5d09ffa1c6d927c40f4c337304049e8a952fbcbf45c6fa77a41a4"),
```
adamsmo: if it is commented maybe it should be removed?

src/test/scala/io/iohk/ethereum/crypto/Pbkdf2HMacSha256Spec.scala
```diff
@@ -0,0 +1,39 @@
...
+      ("passphrase", "salt", "c", "dklen", "derivedKey"),
+      ("password", "salt", 1, 32,
+        "120fb6cffcf8b32c43e7225256c4f837a86548c92ccc35480805987cb70be17b"),
+      ("password", "salt", 2, 32,
+        "ae4d0c95af6b46d32d0adff928f06dd02a303f8ef3c251dfd6e2d85a95474c43"),
+      ("password", "salt", 4096, 32,
+        "c5e478d59288c841aa530db6845c4c8d962893a001ce4e11a4963873aa98134a"),
+      // takes to long:
+      // ("password", "salt", 16777216, 32,
+      //  "cf81c66fe8cfc04d1f31ecb65dab4089f7f179e89b3b0bcb17ad10e3ac6eba46"),
```
adamsmo: if it is commented maybe it should be removed?

src/test/scala/io/iohk/ethereum/crypto/ScryptSpec.scala
rtkaczyk: Well, it's a valid test, it's just inconvenient to run it on the CI. I'd rather we leave it in case we need to examine this function in the future. The reason for commenting out is provided within the comment itself.

src/test/scala/io/iohk/ethereum/crypto/Pbkdf2HMacSha256Spec.scala
rtkaczyk: same as above

src/test/scala/io/iohk/ethereum/crypto/ScryptSpec.scala
adamsmo: yes but it will be never executed and it is dead code

rtkaczyk: Yes, and do you see any harm in that?
I don't like commented out code because I don't know why it was left around and it confuses me, *unless* the reason is right there in the comment...

adamsmo: 👍

rtkaczyk: @adamsmo I have made those comments more verbose to make it clear

PR #215, opened 2017-06-09
[EC-201] Removed MessageHandler from Network layer

ntallar:
## Description

Includes removing the `MessageHandler` from the `Network` layer, being the functionality provided by it replaced by:
* PeersInfoHolderActor: in charge of storing and updating the information about each peer (`PeerInfo` class)
* BlockBroadcast: used by the `RegularSync` to broadcast blocks apropiately, by taking into account the information about each peer
* BlockchainHostActor: in charge of replying to requests for syncing to our peer (for example `GetBlockHeaders`)

The `PeerMessageBus` was also replaced by a `PeerEventBus` which now handles the events `MessageFromPeer`, `MessageToPeer`, `PeerDisconnected` and `PeerHandshakeSuccessful`.


src/main/scala/io/iohk/ethereum/network/PeersInfoHolderActor.scala
```diff
@@ -0,0 +1,189 @@
...
+import io.iohk.ethereum.network.handshaker.Handshaker.HandshakeResult
+import io.iohk.ethereum.network.p2p.Message
+import io.iohk.ethereum.network.p2p.messages.CommonMessages.{NewBlock, Status}
+import io.iohk.ethereum.network.p2p.messages.PV62.{BlockHeaders, GetBlockHeaders, NewBlockHashes}
+import io.iohk.ethereum.network.p2p.messages.WireProtocol.Disconnect
+
+class PeersInfoHolderActor(peerEventBusActor: ActorRef, appStateStorage: AppStateStorage,
+                           forkResolverOpt: Option[ForkResolver]) extends Actor with ActorLogging {
+
+  val msgCodesWithInfo: Set[Int] = Set(BlockHeaders.code, NewBlock.code, NewBlockHashes.code)
```
AlanVerbner: Maybe define this ones in the companion?

src/main/scala/io/iohk/ethereum/network/PeersInfoHolderActor.scala
```diff
@@ -0,0 +1,189 @@
...
+
+    def withTotalDifficulty(totalDifficulty: BigInt): PeerInfo = copy(totalDifficulty = totalDifficulty)
+
+    def withForkAccepted(forkAccepted: Boolean): PeerInfo = copy(forkAccepted = forkAccepted)
+
+    def withMaxBlockNumber(maxBlockNumber: BigInt): PeerInfo = copy(maxBlockNumber = maxBlockNumber)
+
+  }
+
+  case class PeerWithInfo(peer: Peer, peerInfo: PeerInfo)
```
AlanVerbner: As this isn't used outside the actor, what about if we make it private?

src/main/scala/io/iohk/ethereum/network/PeersInfoHolderActor.scala
```diff
@@ -0,0 +1,189 @@
...
+      peerEventBusActor ! Subscribe(MessageSentClassifier(msgCodesWithInfo, PeerSelector.WithId(peer.id)))
+
+      //Ask for the highest block from the peer
+      peer.ref ! SendMessage(GetBlockHeaders(Right(peerInfo.remoteStatus.bestHash), 1, 0, false))
+      context become handlePeersInfoEvents(peersWithInfo + (peer.id -> PeerWithInfo(peer, peerInfo)))
+
+    case PeerDisconnected(peerId) if peersWithInfo.contains(peerId) =>
+      peerEventBusActor ! Unsubscribe(PeerDisconnectedClassifier(PeerSelector.WithId(peerId)))
+      peerEventBusActor ! Unsubscribe(MessageReceivedClassifier(msgCodesWithInfo, PeerSelector.WithId(peerId)))
+      peerEventBusActor ! Unsubscribe(MessageSentClassifier(msgCodesWithInfo, PeerSelector.WithId(peerId)))
```
AlanVerbner: If `MessageReceivedClassifier` and `MessageSentClassifier` have the same info and you are subscribing to both of them each time, what's the purpose of having two classes? Is there any case where we subscribe to one and not to the other and that's why we require to have them separated?

src/main/scala/io/iohk/ethereum/network/PeersInfoHolderActor.scala
```diff
@@ -0,0 +1,189 @@
...
+  override def receive: Receive = handlePeersInfoEvents(Map.empty)
+
+  def handlePeersInfoEvents(peersWithInfo: Map[PeerId, PeerWithInfo]): Receive = {
+
+    case MessageToPeer(message, peerId) if peersWithInfo.contains(peerId) =>
+      val PeerWithInfo(peer, oldPeerInfo) = peersWithInfo(peerId)
+      val newPeerInfo = handleSentMessage(message, oldPeerInfo)
+      context become handlePeersInfoEvents(peersWithInfo + (peerId -> PeerWithInfo(peer, newPeerInfo)))
+
+    case MessageFromPeer(message, peerId) if peersWithInfo.contains(peerId) =>
```
AlanVerbner: This case and the other one before are quite the same. Do you think we can, in some way, to share it to reduce some code?

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
```diff
@@ -71,18 +71,15 @@ class SyncController(
...
+          context watch peer.ref
 
         case _ => // nothing
       }
 
-      handshakedPeers = peers.handshaked
+      handshakedPeers = peers
 
     case Terminated(ref) if handshakedPeers.exists(_._1.ref == ref) =>
       removePeer(ref)
```
AlanVerbner: This is a little bit confusing to me. We are adding peers based on `PeersInfoHolderActor` but removing them because we are watching them. Maybe we need another opinion.

src/main/scala/io/iohk/ethereum/network/PeersInfoHolderActor.scala
ntallar: They are separate as the `MessageReceivedClassifier` is for messages received from peers while `MessageSentClassifier` is for messages sent to them.
Apart from the `PeersInfoHolder`, currently no other actor is interested in both kinds of events as for example the `SyncBlockHeadersRequestHandler` is only interested in receiving the `BlockHeaders` that it requested. This is also the case for the `PendingTransactionsManager` and all the `SyncRequestHandler`s.

src/main/scala/io/iohk/ethereum/network/PeersInfoHolderActor.scala
```diff
@@ -0,0 +1,204 @@
...
+      val newPeersWithInfo = updatePeersWithInfo(peersWithInfo, peerId, message, handleReceivedMessage)
+      context become handlePeersInfoEvents(newPeersWithInfo)
+
+    case PeerHandshakeSuccessful(peer, peerInfo: PeerInfo) =>
+      peerEventBusActor ! Subscribe(PeerDisconnectedClassifier(PeerSelector.WithId(peer.id)))
+      peerEventBusActor ! Subscribe(MessageReceivedClassifier(msgCodesWithInfo, PeerSelector.WithId(peer.id)))
+      peerEventBusActor ! Subscribe(MessageSentClassifier(msgCodesWithInfo, PeerSelector.WithId(peer.id)))
+
+      //Ask for the highest block from the peer
+      peer.ref ! SendMessage(GetBlockHeaders(Right(peerInfo.remoteStatus.bestHash), 1, 0, false))
```
AlanVerbner: Should this actor ask for this? Isn't that a logic that should be within the one that does the handshaking?

src/main/scala/io/iohk/ethereum/network/p2p/Message.scala
```diff
@@ -17,6 +17,8 @@ trait MessageSerializable extends Message {
 
   def toBytes: Array[Byte]
 
+  def underlyingMsg: Message
```
mcsherrylabs: Will we need this if the message is only serialize-able at the last minute?

src/test/scala/io/iohk/ethereum/network/PeerEventBusActorSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+import io.iohk.ethereum.network.PeerEventBusActor.PeerSelector
+import io.iohk.ethereum.network.PeerEventBusActor.SubscriptionClassifier._
+import io.iohk.ethereum.network.PeersInfoHolderActor.PeerInfo
+import io.iohk.ethereum.network.p2p.messages.CommonMessages.Status
+import io.iohk.ethereum.network.p2p.messages.Versions
+import io.iohk.ethereum.network.p2p.messages.WireProtocol.{Ping, Pong}
+import org.scalatest.{FlatSpec, Matchers}
+
+class PeerEventBusActorSpec extends FlatSpec with Matchers {
+
```
mcsherrylabs: Do we have a test for multiple subscriptions to the same code resulting in a single event raised?

src/main/scala/io/iohk/ethereum/network/PeerEventBusActor.scala
```diff
@@ -0,0 +1,131 @@
...
+        case PeerDisconnected(peerId) =>
+          subscriptions.collect {
+            case Subscription(subscriber, classifier: PeerDisconnectedClassifier)
+              if classifier.peerSelector.contains(peerId) => subscriber
+          }
+        case _: PeerHandshakeSuccessful[_] =>
+          subscriptions.collect {
+            case Subscription(subscriber, PeerHandshaked) => subscriber
+          }
+      }
```
ntallar: Not yet, @AlanVerbner's changes should get merged soon (#204) and I'll incorporate them into this PR then

src/test/scala/io/iohk/ethereum/network/PeerEventBusActorSpec.scala
ntallar: We currently don't. I agree one is needed but one was added in #204 (https://github.com/input-output-hk/etc-client/pull/204/files#diff-0a6d52064f30e33c6245bb97281c9aedR56). As it should get merged soon (it already received 2 approvals but were dismissed due to changes) it should be incorporated soon into this PR

src/main/scala/io/iohk/ethereum/blockchain/sync/BlockchainHostActor.scala
```diff
@@ -0,0 +1,94 @@
...
+    case MessageFromPeer(message, peerId) =>
+      val responseOpt = handleBlockFastDownload(message) orElse handleEvmMptFastDownload(message)
+      responseOpt.foreach{ response =>
+        peerManagerActor ! PeerManagerActor.SendMessage(response, peerId)
+      }
+  }
+
+  private def handleEvmMptFastDownload(message: Message): Option[MessageSerializable] = message match {
+    case request: GetNodeData =>
+
```
ntallar: That's right, the `GetNodeData` is used both for requesting MPT nodes and EVM code (which I agree is kind of weird), I'll refactor this code to make it clearer :+1:

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
AlanVerbner: Well...here we are _watching_ for `Terminated` and in an almost similar mapping in `PeersInfoHolderActor` we are removing based on `PeerDisconnected` message

src/main/scala/io/iohk/ethereum/network/PeersInfoHolderActor.scala
AlanVerbner: Lol...I wasn't meant to use this within handshaker, my bad. My question is if it should be done in app layer. My doubt comes from the idea that the `PeersInfoHolderActor` is in charge of _sniffing_ messages that are sent (passive way I might say) so I'm wondering if is there any other responsible for triggering this first message after handshake.

src/main/scala/io/iohk/ethereum/network/p2p/Message.scala
ntallar: In that case it won't be needed, this was added as the interaction with the `Network` layer is done through messages that are serializable (https://github.com/input-output-hk/etc-client/blob/10e5a10226e657d32cf681fbac02d9afb77590a3/src/main/scala/io/iohk/ethereum/network/PeerActor.scala#L253). What do you think the interaction being that way? This was done to take advantage of the `MessageSerializable` so that no implicits need to be passed to the Network layer for serialization.

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
ntallar: I agree that it's confusing that this is handled through the `Terminated` message and not the `PeerDisconnected` event. IMO this should be changed (possibly later) as the application shouldn't know of the use of actors for representing peers in the `Network` layer, but it could also be changed here.

src/main/scala/io/iohk/ethereum/network/PeerEventBusActor.scala
```diff
@@ -0,0 +1,131 @@
...
+      override def contains(p: PeerId): Boolean = p == peerId
+    }
+
+  }
+
+  sealed trait SubscriptionClassifier
+
+  object SubscriptionClassifier {
+    case class MessageReceivedClassifier(messageCodes: Set[Int], peerSelector: PeerSelector) extends SubscriptionClassifier
+    case class MessageSentClassifier(messageCodes: Set[Int], peerSelector: PeerSelector) extends SubscriptionClassifier
```
ntallar: It could be done that way, but what I prefer about this implementation is that there's no need for each time a message is sent (that is required by `PeersInfoHolderActor` or any future actor needing it) also sending the update to it (which we might forget).
Also regarding `PeersInfoHolderActor` being the one that forwards the message, at least the intent of the `PeersInfoHolder` was being a passive entity as much as possible, but we could rename the actor so as to also handle doing that.

src/main/scala/io/iohk/ethereum/network/p2p/Message.scala
mcsherrylabs: SendMessage looks good, but only at the point where SendMessage is called, the implicit must be available. 
`import messages._`
`peer.ref ! SendMessage(GetBlockHeaders(...).)`

In this way the implicit happens outside network?

src/main/scala/io/iohk/ethereum/network/PeerEventBusActor.scala
mcsherrylabs: :+1:

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
mcsherrylabs: @AlanVerbner goor point :thinking:

src/main/scala/io/iohk/ethereum/network/PeerEventBusActor.scala
mcsherrylabs: Right, but we use explicit coupling of actor refs for events we generate everywhere else. We have to treat the incoming messages as events. Now if there's strong case against it, I'm all ears. but PeersInfoHolderActor can be the application level PeerManager it doesn't need to be passive although that would have been nice.

ntallar: :thinking: I see your point, looking at it that way (as a `EtcPeerManager`) it would make sense for it to send the messages (as done currently with the `PeerManager.SendMessage`).
I'll take a look at it, and see if there's any problem with implementing it that way it, but it seems to make sense. I think that sending messages to the `EtcPeerManager` would also enable us to remove the use of `underlyingMsg`.

src/main/scala/io/iohk/ethereum/network/p2p/Message.scala
ntallar: That's right, the messages transmitted to be sent to the Network layer should always be `MessageSerializable`, and the use of the abstract classes for obtaining them always happens outside the network.

ntallar:
@mcsherrylabs You are right :+1:
Fixed!

PR #217, opened 2017-06-13
[EC-205] personal_sign & eth_sign methods

AlanVerbner:
## Description

As `personal_sign` and `eth_sign` methods are almost the same, this PR includes both. The only difference is:

- `eth_sign` requires the account to be unlocked when calling the method
- `personal_sign` unlocks the account but just for that call


src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -401,4 +401,17 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
     def encodeJson(t: GetTransactionCountResponse): JValue = encodeAsHex(t.value)
   }
 
+  implicit val eth_sign = new JsonDecoder[SignRequest] {
+    override def decodeJson(params: Option[JArray]): Either[JsonRpcError, SignRequest] =
+      params match {
+        case Some(JArray(JString(message) :: JString(addr) :: _)) =>
```
ntallar: Maybe the order of the parameters should be `[addr, message]`? It seems to be that way in https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sign.

AlanVerbner: good catch...let me check :cry:

ntallar:
I'm having an issue when testing this that might not have been caused by the changes on this PR (so could be changed in a separate task if needed). 
The following request works on our client but doesn't work in other clients (apart from the order of the parameters):
`"jsonrpc":"2.0","method":"eth_sign","params": ["0x2565023b2980f668ef86b11a27e628e5ff696235", "1234"],"id":1`
This seems to be caused by other clients checking whether `DATA` parameters has a `0x` prefix or not

ntallar:
Apart from those comments, LGTM!

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
AlanVerbner: Done!

AlanVerbner:
@ntallar fixed!

PR #218, opened 2017-06-13
Add rpc filter methods

LukasGasior1:
- added FilterManager for managing filters states


src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -376,15 +376,120 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
...
+      val topicsEither: Either[JsonRpcError, Seq[Seq[ByteString]]] =
+        Try((obj \ "topics").extractOpt[JArray].map(_.arr).getOrElse(Nil).map {
+          case JNull => Nil
+          case js: JString =>
+            Seq(extractBytes(js).toOption.getOrElse(throw new RuntimeException(s"Unable to parse topics, expected byte data but got  ${js.values}")))
+          case jarr: JArray => jarr.arr.map {
+            case e: JString =>
+              extractBytes(e).toOption.getOrElse(throw new RuntimeException(s"Unable to parse topics, expected byte data but got ${e.values}"))
+            case other =>
+              throw new RuntimeException(s"Unable to parse topics, expected byte data but got: $other")
```
rtkaczyk: 😨

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala
```diff
@@ -182,7 +195,10 @@ class JsonRpcController(
             case Right(success) => successResponse(rpcReq, success)
             case Left(error) => errorResponse(rpcReq, error)
           }
-          .recover { case ex => errorResponse(rpcReq, InternalError) }
+          .recover { case ex =>
+            log.error("Failed to handle RPC request", ex)
```
rtkaczyk: 👍 
We generally miss logging in this area. Would be really helpful to know which methods are being called from Ethereum Wallet.

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
```diff
@@ -0,0 +1,253 @@
...
+    keyStore: KeyStore,
+    pendingTransactionsManager: ActorRef)
+  extends Actor {
+
+  import FilterManager._
+  import akka.pattern.{ask, pipe}
+
+  val maxBlockHashesChanges = 256
+
+  var filters: Seq[Filter] = Nil
```
AlanVerbner: Why not a `Map`?

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
```diff
@@ -0,0 +1,253 @@
...
+    case GetFilterChanges(id) => getFilterChanges(id)
+  }
+
+  private def addFilterAndSendResponse(filter: Filter): Unit = {
+    filters :+= filter
+    lastCheckBlocks += (filter.id -> appStateStorage.getBestBlockNumber())
+    sender() ! NewFilterResponse(filter.id)
+  }
+
+  private def uninstallFilter(id: BigInt): Unit = {
```
AlanVerbner: Within the definition it states: _Additonally Filters timeout when they aren't requested with eth_getFilterChanges for a period of time._  ... is that done? If so please point me out where is it and discard this comment :smile:

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    recur(blockNumber + 1, Nil)
+  }
+
+  private def getPendingTransactions(): Future[Seq[SignedTransaction]] = {
+    (pendingTransactionsManager ? PendingTransactionsManager.GetPendingTransactions)
+      .mapTo[PendingTransactionsManager.PendingTransactions]
+      .flatMap { case PendingTransactionsManager.PendingTransactions(pendingTransactions) =>
+        keyStore.listAccounts() match {
```
AlanVerbner: Are you sure this filter is required?

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
```diff
@@ -0,0 +1,253 @@
...
+  def props(blockchain: Blockchain,
+            appStateStorage: AppStateStorage,
+            keyStore: KeyStore,
+            pendingTransactionsManager: ActorRef): Props =
+    Props(new FilterManager(blockchain, appStateStorage, keyStore, pendingTransactionsManager))
+
+  sealed trait Filter {
+    def id: BigInt
+  }
+  case class LogFilter(
```
AlanVerbner: Is it possible the following field is missing? 

`removed: TAG - true when the log was removed, due to a chain reorganization. false if its a valid log.`

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
LukasGasior1: According to documentation no, but geth implementation only returns pending transactions of accounts managed by this node so I think it's needed to stay consistent

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
LukasGasior1: It is, according to documentation, but...:
a) geth does not return such field
b) I don't think there's a case when this should be set to false (we never return logs from dropped blocks)

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
LukasGasior1: oops, missed it

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
LukasGasior1: oh...so you noticed that :P fixed

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
LukasGasior1: done

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
LukasGasior1: done

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
```diff
@@ -0,0 +1,271 @@
...
+      } else logsSoFar
+    }
+  }
+
+  private def topicsMatch(logTopics: Seq[ByteString], filterTopics: Seq[Seq[ByteString]]): Boolean = {
+    (filterTopics zip logTopics).forall { case (filter, log) => filter.isEmpty || filter.contains(log) }
+  }
+
+  private def getBlockHashesAfter(blockNumber: BigInt): Seq[ByteString] = {
+    val bestBlock = appStateStorage.getBestBlockNumber()
```
AlanVerbner: As this is being queried here https://github.com/input-output-hk/etc-client/pull/218/files#diff-b5ee4e4c8bcd462269de630eb5466a06R127 maybe we can send it as paramenter instead of getting that value again

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
AlanVerbner: Discard the comment then

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
```diff
@@ -0,0 +1,271 @@
...
+      } else blockchain.getBlockHeaderByNumber(currentBlockNumber) match {
+        case Some(header) => recur(currentBlockNumber + 1, hashesSoFar :+ header.hash)
+        case None => hashesSoFar
+      }
+    }
+
+    recur(blockNumber + 1, Nil)
+  }
+
+  private def getPendingTransactions(): Future[Seq[SignedTransaction]] = {
```
AlanVerbner: It seems this one doesn't work as parity does. Our implementation is returning all the pending transactions every time I query them. In parity, it only returns the ones It hasn't returned before.

This doc statement is confusing, but might be the reason _Creates a filter in the node, to notify when new pending transactions arrive_

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
AlanVerbner: :+1: 

And do you think it's worthy to join `filters` and `lastCheckBlocks` within the same `Map`?

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
AlanVerbner: Works perfectly! :metal:

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
```diff
@@ -0,0 +1,276 @@
...
+
+  case class GetLogs(fromBlock: Option[BlockParam], toBlock: Option[BlockParam], address: Option[Address], topics: Seq[Seq[ByteString]])
+
+  case class Log(
+      logIndex: BigInt,
+      transactionIndex: Option[BigInt],
+      transactionHash: Option[ByteString],
+      blockHash: ByteString,
+      blockNumber: BigInt,
+      address: Address,
```
adamsmo: since we are using `Extraction.decompose` should we use ByteString for address?

LukasGasior1: I've added custom serializer for `Address`

adamsmo: 👍  `AddressJsonSerializer`

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
```diff
@@ -0,0 +1,290 @@
...
+
+  var filters: Map[BigInt, Filter] = Map.empty
+
+  var lastCheckBlocks: Map[BigInt, BigInt] = Map.empty
+
+  var lastCheckTimestamps: Map[BigInt, Long] = Map.empty
+
+  var filterTimeouts: Map[BigInt, Cancellable] = Map.empty
+
+  implicit val timeout = Timeout(5.seconds)
```
rtkaczyk: What is this used for? Should it be configurable?

LukasGasior1: of course it should

PR #221, opened 2017-06-17
Basic Json Rpc batch requests support

AlanVerbner:
## Description

While testing using Mist I've found that there are a lot batch requests being performed even if this is not stated it must be supported by clients in documentation. 

After implementing it, Mist stopped throwing errors and noticed some fixes. For example: accounts and balances are updated when added or mining.

### Notes
- It's a basic implementation but seems to be working as parity does. 
- It doesn't check if two requests with same Id are sent in the same batch request (parity doesn't neither). Same case if the items within the batch array are the same
- Does not implement any limit / restriction on the amount of items a batch request might contain. Maybe we need to create a new task for this

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/http/JsonRpcHttpServer.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/http/JsonRpcHttpServer.scala
@@ -15,6 +15,7 @@ import org.json4s.{DefaultFormats, native}
 
 import scala.util.{Failure, Success}
 import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
 
 class JsonRpcHttpServer(jsonRpcController: JsonRpcController, config: JsonRpcHttpServerConfig)
                        (implicit val actorSystem: ActorSystem)
@@ -34,6 +35,8 @@ class JsonRpcHttpServer(jsonRpcController: JsonRpcController, config: JsonRpcHtt
   val route: Route = {
     (pathEndOrSingleSlash & post & entity(as[JsonRpcRequest])) { request =>
       handleRequest(request)
+    } ~ (pathEndOrSingleSlash & post & entity(as[Seq[JsonRpcRequest]])) { request =>
+      handleBatchRequest(request)
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/io/iohk/ethereum/jsonrpc/http/JsonRpcHttpServer.scala
```diff
@@ -52,6 +55,10 @@ class JsonRpcHttpServer(jsonRpcController: JsonRpcController, config: JsonRpcHtt
     complete(jsonRpcController.handleRequest(request))
   }
 
+  private def handleBatchRequest(requests: Seq[JsonRpcRequest]) = {
+    complete(Future.sequence(requests.map(request => jsonRpcController.handleRequest(request))))
```
rtkaczyk: I wonder if should synchronise those calls in case the caller expects them to be performed in the given order? IMO, finding this out is outside the scope of this PR.

src/main/scala/io/iohk/ethereum/jsonrpc/http/JsonRpcHttpServer.scala
```diff
@@ -34,6 +35,8 @@ class JsonRpcHttpServer(jsonRpcController: JsonRpcController, config: JsonRpcHtt
   val route: Route = {
     (pathEndOrSingleSlash & post & entity(as[JsonRpcRequest])) { request =>
       handleRequest(request)
+    } ~ (pathEndOrSingleSlash & post & entity(as[Seq[JsonRpcRequest]])) { request =>
+      handleBatchRequest(request)
```
rtkaczyk: We could avoid repetition here:

```
    (pathEndOrSingleSlash & post) {
      entity(as[JsonRpcRequest]) { request =>
        handleRequest(request)
      } ~
      entity(as[Seq[JsonRpcRequest]]) { request =>
        handleBatchRequest(request)
      }
    }
```

src/main/scala/io/iohk/ethereum/jsonrpc/http/JsonRpcHttpServer.scala
LukasGasior1: I think the order is maintained here, isn't it?

PR #222, opened 2017-06-17
Fixes in eth_estimateGas

AlanVerbner:
## Description

When testing using Mist, the following requests are performed to `eth_estimateGas` endpoint and weren't supported by our implementation:

- Optional block parameter: `{"jsonrpc":"2.0","id":"6aed168f-91b6-485f-8ac1-86969de52e49","method":"eth_estimateGas","params":[{"from":"0xa797e6b221246a61141c7d04dfe2047e5b3ba620","to":"0x480d874B16C952efC6584F7A61bf3E8c38e39EDd","value":"0x578c6b4cf1583a8000","data":"0xaaaa","gas":"0x4c4b40"}]}`
- Allow data to be sent as "": `{"jsonrpc":"2.0","id":"6aed168f-91b6-485f-8ac1-86969de52e49","method":"eth_estimateGas","params":[{"from":"0xa797e6b221246a61141c7d04dfe2047e5b3ba620","to":"0x480d874B16C952efC6584F7A61bf3E8c38e39EDd","value":"0x578c6b4cf1583a8000","data":"","gas":"0x4c4b40"}, "latest"]}`

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
@@ -31,15 +31,10 @@ trait JsonMethodsImplicits {
   protected def encodeAsHex(input: BigInt): JString =
     JString(s"0x${input.toString(16)}")
 
-  protected def decodeWithoutHexPrefix(s: String): Either[JsonRpcError, Array[Byte]] =
-    Try {
-      val normalized = if (s.length % 2 == 1) "0" + s else s
-      Hex.decode(normalized)
-    }.toEither.left.map(_ => InvalidParams())
-
-  private def decode(s: String): Either[JsonRpcError, Array[Byte]] = {
-    if(!s.isEmpty && !s.startsWith("0x")) Left(InvalidParams())
-    else decodeWithoutHexPrefix(s.drop("0x".length))
+  private def decode(s: String): Array[Byte] = {
+    val stripped = s.replaceFirst("^0x", "")
+    val normalized = if (stripped.length % 2 == 1) "0" + stripped else stripped
+    Hex.decode(normalized)
   }
 
   protected def extractAddress(input: String): Either[JsonRpcError, Address] =
@@ -49,7 +44,7 @@ trait JsonMethodsImplicits {
     extractAddress(input.s)
 
   protected def extractBytes(input: String): Either[JsonRpcError, ByteString] =
-    decode(input).map(ByteString(_))
+    Try(ByteString(decode(input))).toEither.left.map(_ => InvalidParams())
 
   protected def extractBytes(input: JString): Either[JsonRpcError, ByteString] =
     extractBytes(input.s)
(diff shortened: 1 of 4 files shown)


rtkaczyk:
We require the block parameter for all the methods that define it. I wonder if it should be optional for all of them?

src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
```diff
@@ -38,7 +38,7 @@ trait JsonMethodsImplicits {
     }.toEither.left.map(_ => InvalidParams())
 
   private def decode(s: String): Either[JsonRpcError, Array[Byte]] = {
-    if(!s.startsWith("0x")) Left(InvalidParams())
+    if(!s.isEmpty && !s.startsWith("0x")) Left(InvalidParams())
```
rtkaczyk: So I noticed you made this change:

[Data should be prefixed with 0x](https://github.com/input-output-hk/etc-client/commit/8b76fd2a6f7e854b00bb1137d446f45db74ef19c)

No, it should not. We're getting hit all the time by following the documentation to the word, and this PR is a perfect example ;)
Geth does not require bytes to be prefixed with `0x`, which is why I changed that in #195. Could you change it back, please?

AlanVerbner: Shit :disappointed:  ... I will check it myself because I've been asked to change to the current implementation. 

BTW, if other clients didn't follow that spec, does it mean we shouldn't. I understand some apps won't work. IMO we either need to decide to create a PR to update the docs or be strict following the docs, what do you think?

AlanVerbner:
Regarding the block parameter, I've seen in other implementations that it's not optional for eth_call for example. I need to run some queries again to be sure tho.

src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
mcsherrylabs: To recap, we should try to be compliant with the existing working clients over the spec unless there is a security concern. Our experience shows there are many deviations. A PR to update the docs is a good idea though!

ntallar: The strange thing is that geth doesn't seem to be consistent with asking or not for the prefix, as for example it doesn't allow the request `{"jsonrpc":"2.0","method":"eth_getBalance","params": ["c371cfecae774a728681c164360df9cec2a0a412", "0x1234"],"id":1}` (address without "0x" prefix).

rtkaczyk: @ntallar that request works for me. Using geth-classic `Geth/v3.3.0-1-gdd95f05/darwin/go1.7`.

Either way being more permissive (allowing unprefixed address for consistency) doesn't hurt in this case.

AlanVerbner: @rtkaczyk done!

PR #224, opened 2017-06-22
[EC-186] Filter out transactions that are invalid and are causing block preparation failure

adamsmo:


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -57,15 +57,23 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
...
-    blockExecResult
+    } yield BlockPreparationResult(block, execResult, worldPersisted.stateRootHash)
+
+    blockExecResult match {
+      case Right(result) =>
+        log.debug(s"Prepared for mining block with number ${block.header.number}")
+        Right(result)
+      case Left(TxsExecutionError(tx,reason)) =>
+        log.debug(s"failed to execute transaction $tx because: $reason")
+        val newBlock = block.copy(body = block.body.copy(transactionList = block.body.transactionList.filter(_.hash != tx.hash)))
```
AlanVerbner: If transaction failed (If i'm not mistaken the only reason it's because validation failed) shouldn't it be removed from pending transaction pool?

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
```diff
@@ -27,7 +27,7 @@ class BlockGenerator(blockchainStorages: BlockchainStorages, blockchainConfig: B
   val difficulty = new DifficultyCalculator(blockchainConfig)
 
   private val cache: AtomicReference[List[Block]] = new AtomicReference(Nil)
-
+  // scalastyle:off
```
AlanVerbner: :thinking:

adamsmo: forgot to remove

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
adamsmo: this is good question, it may be that after future block reorganisation it will not fail
and we have limited transaction pool so failing transactions will be push out by new one

AlanVerbner: ok...that makes sense but:

1. If a tx execution is failing while preparing a block, everytime we are asked for work (`eth_getWork`) will we run once with error and re run without that tx?
2. Related to (1) If it's a tx that has a wrong gas limit or something like that, it will stay there bothering forever?

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
```diff
@@ -40,7 +40,8 @@ class BlockGenerator(blockchainStorages: BlockchainStorages, blockchainConfig: B
           ommersHash = ByteString(kec256(ommers.toBytes: Array[Byte])),
           beneficiary = beneficiary.bytes,
           stateRoot = ByteString.empty,
-          transactionsRoot = buildMpt(transactions, SignedTransaction.byteArraySerializable),
+          //not here
```
AlanVerbner: what does this means?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -57,15 +57,23 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
...
+        prepareBlock(newBlock, storages,validators)
```
AlanVerbner: this might now be `@tailrec`, right?

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
adamsmo: it means that we cannot calculate here transaction root because we do not know which transactions will be valid, will update

AlanVerbner: Ok, maybe we can improve this comment because it seems a message you forgot to remove. What about something like _unknown at this point_ ?

src/test/scala/io/iohk/ethereum/Mocks.scala
```diff
@@ -58,18 +58,12 @@ object Mocks {
       override def validateHeaderAndBody(blockHeader: BlockHeader, blockBody: BlockBody) = Right(Block(blockHeader, blockBody))
     }
 
-    override val blockHeaderValidator: BlockHeaderValidator = new BlockHeaderValidator {
-      override def validate(blockHeader: BlockHeader, blockchain: Blockchain) = Right(blockHeader)
-    }
+    override val blockHeaderValidator: BlockHeaderValidator = (blockHeader: BlockHeader, blockchain: Blockchain) => Right(blockHeader)
```
rtkaczyk: Cool, didn't know about this. (SAM)

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
rtkaczyk: Also wouldn't hurt to say *what* is unknown :)

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -57,15 +57,23 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
...
+      case r =>
```
rtkaczyk: I think this is confusing. You matched a `Right` and a `Left` and what's this supposed to be? Either make a precise match on a `Left` with the inner type, or leave the `left.map` as it was.

src/test/scala/io/iohk/ethereum/mining/BlockGeneratorSpec.scala
```diff
@@ -52,6 +52,22 @@ class BlockGeneratorSpec extends FlatSpec with Matchers with PropertyChecks with
...
+    val result: Either[BlockPreparationError, Block] =
+      blockGenerator.generateBlockForMining(1, Seq(signedTransaction, duplicatedSignedTransaction), Nil, Address(testAddress))
+    result shouldBe a[Right[_, Block]]
+
+    //mined with etc-client + ethminer
+    val minedNonce = ByteString(Hex.decode("5e8d5c12cea7e0c7"))
+    val minedMixHash = ByteString(Hex.decode("9247b81258f97159f987a5f4f9e94df1d95e10eeabff2836020eafb27a8228b0"))
+    val miningTimestamp = 1494604913
+
+    val fulBlock: Either[BlockPreparationError, Block] = result.right
```
rtkaczyk: typo? `fullBlock`?

src/test/scala/io/iohk/ethereum/mining/BlockGeneratorSpec.scala
```diff
@@ -52,6 +52,22 @@ class BlockGeneratorSpec extends FlatSpec with Matchers with PropertyChecks with
     fulBlock.right.foreach(b => ledger.executeBlock(b, blockchainStorages.storages, validators) shouldBe a[Right[_, Seq[Receipt]]])
   }
 
+  it should "filter out failing transactions" in new Envirnoment {
```
rtkaczyk: Could you fix the type in Environment? Or better yet, for consistency, change it to `TestSetup`, as that's the name we use everywhere else.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
adamsmo: 1. yes we will re run `prepareBlock` recursively until we have no failing transactions (that may mean we end up with empty block)
2. No, it will be eventually pushed out of transaction pool by new incoming transactions, because we prepend them to the pool and we have finite pool size, so in some cases (not to many transactions) it may happen that we will keep this transaction forever

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
```diff
@@ -37,7 +37,10 @@ class BlockGenerator(blockchainStorages: BlockchainStorages, blockchainConfig: B
         val blockTimestamp = Instant.now.getEpochSecond
         val header: BlockHeader = prepareHeader(blockNumber, ommers, beneficiary, parent, blockTimestamp)
 
-        val transactionsForBlock = transactions.toSet[SignedTransaction].toList.sortBy(_.tx.nonce)
+        val transactionsForBlock = transactions
+          .filter(_.tx.gasLimit < header.gasLimit)
```
AlanVerbner: Is there any test that covers this scenario?

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
```diff
@@ -37,7 +37,10 @@ class BlockGenerator(blockchainStorages: BlockchainStorages, blockchainConfig: B
...
+          //if we have 2 transactions from same address we want first one with lower nonce
+          .sortBy(_.tx.nonce)
```
AlanVerbner: I'm not saying this is wrong at all but I would imagine that also using the fee (indirectly by gasPrice) should be used to pick tx to mine first.

This is just a commet, this shouldn't be a stopper for merging.

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
adamsmo: will add test for this

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
adamsmo: ordering by nonce is to prevent tx failure when we have 2 transactions from same account
this is not optimisation for mining, we will do that in separate PR

PR #227, opened 2017-06-23
[EC-209] Merge Log case class

adamsmo:


src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -293,10 +293,10 @@ class EthService(
         gasUsed = if (txIndex == 0) receipt.cumulativeGasUsed else receipt.cumulativeGasUsed - receipts(txIndex - 1).cumulativeGasUsed,
         contractAddress = contractAddress,
         logs = receipt.logs.zipWithIndex.map { case (txLog, index) =>
-          TxLog(
+          Log(
```
rtkaczyk: Isn't the `TxLog` name better? `Log` is already quite overloaded.

adamsmo: I think it is I just picked Log because I did not want to change @LukasGasior1 code
but you are right, will change this

PR #229, opened 2017-06-23
[EC-215] Allow mining after peer response timeout

ntallar:
## Description
The node currently doesn't allow mining after a peer response timeout was received (i.e. when asking for block headers). The scenario that causes this problem is:
* Start with a network with an ETC node and another one
* The other node is disconnected (this should happen while a request is pending which seems to be the general case)
* Miner is attached to our node.
In this case, the blocks mined were not being included into the blockchain.

## Fix
Set `waitingForActor` to `None` when a timeout happens.


src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
```diff
@@ -706,6 +707,63 @@ class SyncControllerSpec extends FlatSpec with Matchers {
     pendingTransactionsManager.expectNoMsg()
   }
 
+  it should "accept mined blocks after request timed-out" in new TestSetup() {
+    val peerTestProbe: TestProbe = TestProbe()(system)
+
+    val peer = Peer(new InetSocketAddress("127.0.0.1", 0), peerTestProbe.ref)
+
+    time.advance(1.seconds)
```
rtkaczyk: What is it for? (This particular occurence)

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
```diff
@@ -706,6 +707,63 @@ class SyncControllerSpec extends FlatSpec with Matchers {
...
+    peerMessageBus.expectMsg(Subscribe(MessageClassifier(Set(BlockHeaders.code), PeerSelector.WithId(peer.id))))
+
+    //wait for timeout
+    time.advance(2 * Config.FastSync.peerResponseTimeout)
+    peerMessageBus.expectMsg(Unsubscribe(PeerDisconnectedClassifier(PeerSelector.WithId(peer.id))))
+
+    syncController ! MinedBlock(Block(minedBlockHeader,BlockBody(Nil,Nil)))
+
+    //wait for actor to insert data
+    time.advance(3.seconds.toMillis)
```
rtkaczyk: Does this change anything? AFAIK this just a mocked time for akka scheduler, but we want to wait for the actor to complete its work.

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
ntallar: You are right, it's not needed :+1:

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
ntallar: I agree with you, it seems that even without this or `Thread.sleep` it usually works fine, so I didn't realize that. As it could be the case that it doesn't, I'll replace this with `Thread.sleep` (as I saw you left on `EC-213` for the similar test `"accept mined blocks"`).

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
```diff
@@ -706,6 +707,61 @@ class SyncControllerSpec extends FlatSpec with Matchers {
...
+    peerMessageBus.expectMsg(Subscribe(MessageClassifier(Set(BlockHeaders.code), PeerSelector.WithId(peer.id))))
+
+    //wait for timeout
+    time.advance(2 * Config.FastSync.peerResponseTimeout)
+    peerMessageBus.expectMsg(Unsubscribe(PeerDisconnectedClassifier(PeerSelector.WithId(peer.id))))
+
+    syncController ! MinedBlock(Block(minedBlockHeader,BlockBody(Nil,Nil)))
+
+    //wait for actor to insert data
+    Thread.sleep(3.seconds.toMillis)
```
rtkaczyk: It's funny but now this test failed...

PR #230, opened 2017-06-26
[EC-212] configurable secure random

rtkaczyk:


src/test/scala/io/iohk/ethereum/SecureRandomProvider.scala
```diff
@@ -0,0 +1,11 @@
...
+
+import java.security.SecureRandom
+
+import io.iohk.ethereum.utils.Config
+
+trait SecureRandomProvider {
+  val secureRandom: SecureRandom = SecureRandom.getInstance(Config.secureRandomAlgo)
+}
+
+object SecureRandomProvider extends SecureRandomProvider
```
ntallar: Minor detail: It seems this object is not being used, is it needed?

ntallar:
It seems we are also currently using `new SecureRandom` in `AuthInitiateMessageSpec`, should we also use `SecureRandomProvider` in it?

ntallar:
Apart from minor comments (and CircleCI failing due to the usage of `generateKeyPair` in `PrecompiledContractsSpecEvm`), LGTM!

rtkaczyk:
@mcsherrylabs 
> So the change ... can be abandoned

you mean close the PR?

src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
```diff
@@ -26,6 +28,7 @@ object DumpChainApp extends App{
     val privateNetworkId = conf.getInt("networkId")
     val startBlock = conf.getInt("startBlock")
     val maxBlocks = conf.getInt("maxBlocks")
+    val secureRandom = new SecureRandom // TODO: should it be configurable here as well?
```
mcsherrylabs: Add `extends App with NodeKeyBuilder
  with SecureRandomBuilder
  with AuthHandshakerBuilder` and remove `val`s `secureRandom` and `nodeKey` and fix the compile error -
 `secureRandom` in `PeerManager` should be `authHandShaker`.

src/test/scala/io/iohk/ethereum/SecureRandomProvider.scala
mcsherrylabs: `SecureRandomProvider` in test shadows `SecureRandomProvider` without linking the types. What we want is to use the type in `main` in the test area as we would building another component for test support. Then if it's required, override it using `override val secureRandom` in tests, or if it's useful across many tests use a trait (eg) `TestRandomProvider extends SecureRandomProvider`. Even if we want to shadow component builders with more test oriented ones, they have to link to the original builders, otherwise it's going to get very confusing.

PR #231, opened 2017-06-26
[EC-218] circle build steps defined in circle.yml

rtkaczyk:
TODO: remove build steps from WebUI prior to merging this


rtkaczyk:
The build failed on post-test build steps - moving files. As you can see all the relevant steps where executed twice, and this is also the reason for failure (moving already moved files). This is expected and should work fine once we remove the relevant WebUI config.

circle.yml
```diff
@@ -1,3 +1,18 @@
...
+    - sbt scalastyle
+    - sbt test:scalastyle
+    - sbt coverage test
+    - sbt coverage evm:test
+    - sbt coverageOff it:test
+    - sbt coverageReport coverageAggregate coveralls
+    - sbt dist
+  post:
+    - mkdir -p $CIRCLE_ARTIFACTS/scala-2.12
+    - mv target/scala-2.12/coverage-report  $CIRCLE_ARTIFACTS/scala-2.12/coverage-report
```
AlanVerbner: I know this is not part of your task but as we have the chance....Are this files being used? Does coveralls need them? If not, maybe we should remove them, what do you think?

circle.yml
```diff
@@ -1,3 +1,18 @@
 dependencies:
   pre:
     - sudo add-apt-repository -y ppa:ethereum/ethereum; sudo apt-get update; sudo apt-get install -y solc
+  override:
+    - echo "OK, let's go!"
```
AlanVerbner: @rtkaczyk Can you leave a comment, at least here, why did you need to put this? I recall you mentioned something but imo it's better if it's written somewhere :wink:

circle.yml
rtkaczyk: AFAIR, they are required for coveralls reports

AlanVerbner: ok...it doesn't take too much time either so, it's fine

PR #228, opened 2017-06-23
[EC-213] unify timeouts in tests

rtkaczyk:
Replaces the timeouts in tests with values defined in a single place.

Some remarks:
- ~`Thread.sleep`: arguments replaced - not sure if this is desired~ - some of those were not needed at all, others replaced with predefined values
- ~`time.advance`: mock scheduler uses values defined at call site, instead configured/predefined values - not changed~ - this will be addressed in another PR
- ~test times may have increased (?)~ - fixed

~TODO: try overriding Circle CI inferred build steps (reduntant `sbt test:compile`)~ - done

PR #232, opened 2017-06-26
[EC-184] Pending block parameter

AlanVerbner:
## Description

This PR implements the `pending` block filter which is used in several RPC methods as defined [here](https://github.com/ethereum/wiki/wiki/JSON-RPC#the-default-block-parameter)

## Notes

- When block is queried and there is no pending, latest is returned as done in other clients
- Receipts are now saved in `BlockGenerator` as they are needed when having `pending` filter when calculating logs are needed
- Maybe `FilterManagerSpec` tests needs some more :heart: but I think this should be done in a separate task

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
@@ -486,13 +486,15 @@ class EthService(
 
   def submitWork(req: SubmitWorkRequest): ServiceResponse[SubmitWorkResponse] = {
     reportActive()
-    blockGenerator.getPrepared(req.powHeaderHash) match {
-      case Some(pendingBlock) if appStateStorage.getBestBlockNumber() <= pendingBlock.block.header.number =>
-        import pendingBlock._
-        syncingController ! MinedBlock(block.copy(header = block.header.copy(nonce = req.nonce, mixHash = req.mixHash)))
-        Future.successful(Right(SubmitWorkResponse(true)))
-      case _ =>
-        Future.successful(Right(SubmitWorkResponse(false)))
+    Future {
+      blockGenerator.getPrepared(req.powHeaderHash) match {
+        case Some(pendingBlock) if appStateStorage.getBestBlockNumber() <= pendingBlock.block.header.number =>
+          import pendingBlock._
+          syncingController ! MinedBlock(block.copy(header = block.header.copy(nonce = req.nonce, mixHash = req.mixHash)))
+          Right(SubmitWorkResponse(true))
+        case _ =>
+          Right(SubmitWorkResponse(false))
+      }
     }
   }
 
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
@@ -692,7 +692,10 @@ class EthService(
       case BlockParam.WithNumber(blockNumber) => getBlock(blockNumber).map(ResolvedBlock(_, pending = false))
       case BlockParam.Earliest => getBlock(0).map(ResolvedBlock(_, pending = false))
       case BlockParam.Latest => getBlock(appStateStorage.getBestBlockNumber()).map(ResolvedBlock(_, pending = false))
-      case BlockParam.Pending => getBlock(appStateStorage.getBestBlockNumber()).map(ResolvedBlock(_, pending = true))
+      case BlockParam.Pending =>
+        blockGenerator.getPending.map(ResolvedBlock(_, pending = true))
+          .map(Right.apply)
+          .getOrElse(resolveBlock(BlockParam.Latest))
     }
   }
 
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
@@ -448,11 +448,11 @@ class EthService(
     getOmmersFromPool(blockNumber).zip(getTransactionsFromPool).map {
       case (ommers, pendingTxs) =>
         blockGenerator.generateBlockForMining(blockNumber, pendingTxs.pendingTransactions.map(_.stx), ommers.headers, miningConfig.coinbase) match {
-          case Right(b) =>
+          case Right(pb) =>
             Right(GetWorkResponse(
-              powHeaderHash = ByteString(kec256(BlockHeader.getEncodedWithoutNonce(b.header))),
-              dagSeed = seedForBlock(b.header.number),
-              target = ByteString((BigInt(2).pow(256) / b.header.difficulty).toByteArray)
+              powHeaderHash = ByteString(kec256(BlockHeader.getEncodedWithoutNonce(pb.block.header))),
+              dagSeed = seedForBlock(pb.block.header.number),
+              target = ByteString((BigInt(2).pow(256) / pb.block.header.difficulty).toByteArray)
             ))
           case Left(err) =>
             log.error(s"unable to prepare block because of $err")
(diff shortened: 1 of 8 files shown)


src/test/scala/io/iohk/ethereum/jsonrpc/EthServiceSpec.scala
```diff
@@ -107,19 +107,33 @@ class EthServiceSpec extends FlatSpec with Matchers with ScalaFutures with MockF
...
-  //FIXME: This test should be changed once the pending block doesn't equal to the latest one
-  it should "answer eth_getBlockByNumber with the correct parameters in None when the pending block is requested" in new TestSetup {
-    val bestBlockNumber = 10
+  it should "answer eth_getBlockByNumber with the correct block when the pending block is requested" in new TestSetup {
+
+    (blockGenerator.getPending _).expects().returns(Some(PendingBlock(blockToRequest, Nil)))
 
-    blockchain.save(blockToRequest.copy(header = blockToRequest.header.copy(number = bestBlockNumber)))
+    val request = BlockByNumberRequest(BlockParam.Pending, fullTxs = true)
+    val response = Await.result(ethService.getBlockByNumber(request), Duration.Inf).right.get
```
rtkaczyk: Can we use `.futureValue` instead of `Await`? At any rate, I'm not fond of `Duration.Inf`

src/test/scala/io/iohk/ethereum/jsonrpc/FilterManagerSpec.scala
```diff
@@ -29,7 +30,15 @@ class FilterManagerSpec extends FlatSpec with Matchers with ScalaFutures with De
...
-    val filterManager = TestActorRef[FilterManager](Props(new FilterManager(blockchain, appStateStorage, keyStore, pendingTransactionsManager.ref, config, Some(time.scheduler))))
+    val filterManager = TestActorRef[FilterManager](Props(new FilterManager(
+      blockchain,
+      blockGenerator,
+      appStateStorage,
+      keyStore,
+      pendingTransactionsManager.ref,
+      config,
+      Some(time.scheduler)))
+    )
```
rtkaczyk: Does this differ in any of the tests? Could we move it to `TestSetup`?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -487,7 +487,8 @@ class EthService(
   def submitWork(req: SubmitWorkRequest): ServiceResponse[SubmitWorkResponse] = {
     reportActive()
     blockGenerator.getPrepared(req.powHeaderHash) match {
-      case Some(block) if appStateStorage.getBestBlockNumber() <= block.header.number =>
+      case Some(pendingBlock) if appStateStorage.getBestBlockNumber() <= pendingBlock.block.header.number =>
+        import pendingBlock._
         syncingController ! MinedBlock(block.copy(header = block.header.copy(nonce = req.nonce, mixHash = req.mixHash)))
         Future.successful(Right(SubmitWorkResponse(true)))
```
rtkaczyk: Not related to your work, but I noticed this function accesses the DB so it should wrap the body in a `Future` rather than return a `Future.successful(...)`.

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
```diff
@@ -222,23 +235,24 @@ class FilterManager(
...
 
-  private def resolveBlockNumber(blockParam: BlockParam, appStateStorage: AppStateStorage): BigInt = {
+  private def resolveBlockNumber(blockParam: BlockParam, bestBlockNumber: BigInt): BigInt = {
     blockParam match {
       case BlockParam.WithNumber(blockNumber) => blockNumber
       case BlockParam.Earliest => 0
-      case BlockParam.Latest => appStateStorage.getBestBlockNumber()
-      case BlockParam.Pending => appStateStorage.getBestBlockNumber() + 1
+      case BlockParam.Latest => bestBlockNumber
+      case BlockParam.Pending => bestBlockNumber
```
rtkaczyk: Why was the `+ 1` dropped?

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
```diff
@@ -124,13 +132,19 @@ class FilterManager(
       }
     }
 
+    val bestBlockNumber = appStateStorage.getBestBlockNumber()
+
     val fromBlockNumber =
-      startingBlockNumber.getOrElse(resolveBlockNumber(filter.fromBlock.getOrElse(BlockParam.Latest), appStateStorage))
+      startingBlockNumber.getOrElse(resolveBlockNumber(filter.fromBlock.getOrElse(BlockParam.Latest), bestBlockNumber))
```
rtkaczyk: Perhaps this was address before, but should `Latest` be the default? I would expect that not providing `fromBlock` param would result in a block number in the past (`Earliest` ?). Otherwise, when `fromBlock` is not provided`, it only makes sense to have `toBlock == Latest`.

Is this how other clients behave?

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
AlanVerbner: For me bestBlockNumner + 1 is not exactly pending block but a special case. Let's suppose that:

- `bestBlockNumber` is 10
- you are filtering from `BlockParam.WithNumber(0)`  to `BlockParam.WithNumber(20)`
- When the time to fetch block 11 [here](https://github.com/input-output-hk/etc-client/pull/232/files#diff-b5ee4e4c8bcd462269de630eb5466a06R119), will you return `None` or `Pending` one?

I thought about tweaking the code but for me looks more confusing from a developer point of view approaching to our code. Do you think it's no?

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
AlanVerbner: @rtkaczyk let me take a look as this change wasn't introduced in this PR

src/test/scala/io/iohk/ethereum/jsonrpc/EthServiceSpec.scala
AlanVerbner: Several tests on this file are done like this...maybe it's a good idea to create a separate task and change them all (for consistency reasons)...what do you think?

src/test/scala/io/iohk/ethereum/jsonrpc/FilterManagerSpec.scala
AlanVerbner: No, you are right....I will refactor it

src/test/scala/io/iohk/ethereum/jsonrpc/EthServiceSpec.scala
rtkaczyk: Yes, exactly: #228 :)
But this is a new addition so it can be forgotten when that PR gets merged

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
AlanVerbner: I can change it ...

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
AlanVerbner: At least [geth uses the latest.](https://github.com/ethereumproject/go-ethereum/blob/master/eth/filters/api.go#L317)

I think the rationale there is the filter, if not specified, should start from the current position, ie, latest

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
rtkaczyk: Ok, I get it, makes sense :+1:

PR #234, opened 2017-06-28
[EC-222] Ignore known transactions in transactions pool

adamsmo:


src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
```diff
@@ -55,9 +55,8 @@ class PendingTransactionsManager(miningConfig: MiningConfig, peerManager: ActorR
   peerMessageBus ! Subscribe(MessageClassifier(Set(SignedTransactions.code), PeerSelector.AllPeers))
 
   override def receive: Receive = {
+    // TODO: we should check whether a transaction with the same address and nonce exists, and if so replace it with the new version
```
AlanVerbner: Is there any open task for this?

adamsmo: added EC-225

PR #235, opened 2017-06-30
Fix PeerManager not to accept connections from already connected peers

LukasGasior1:


src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -104,10 +107,15 @@ class PeerManagerActor(
...
+        context.unbecome()
+        val peer = createPeer(remoteAddress)
+        peer.ref ! PeerActor.HandleConnection(connection, remoteAddress)
+        unstashAll()
+      } else {
+        log.info("Another connection with {} is already opened. Disconnecting.", remoteAddress)
+        connection ! PoisonPill
+      }
 
     case (HandlePeerConnection(connection, remoteAddress), Failure(_)) =>
```
ntallar: As in this case we are also establishing a connection with a new peer, should we also check whether  `isConnectionHandled` is true or not?

ntallar:
As when receving `HandlePeerConnection` and `ConnectToPeer` messages during `listening` we disconnect from peers if there are too many (when calling `tryDiscardPeersToFreeUpLimit()`), maybe it would be better to check there for `isConnectionHandled` is true or not? That way we wouldn't be disconnecting from other peers to make room to a duplicated peer we won't end connecting to.

rtkaczyk:
@LukasGasior1 apart from Nico's comments, LGTM

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
LukasGasior1: In this case the peer limit has been reached (note, the `Failure` in the tuple). The peer will be disconnected and closed anyway (see `       peer.ref ! PeerActor.DisconnectPeer(Disconnect.Reasons.TooManyPeers)` few lines below)

PR #236, opened 2017-06-30
[EC-223] Fix how default tx nonce is obtained when sending x_sendTransaction

ntallar:
## Description

Currently, if 2 transactions (from the same sender) are sent without mining started (or if they arrive between the mining of 2 blocks), and the last one to arrive doesn't have a nonce field set, it will not included to the blockchain.
This happens when using both `eth_sendTransaction` and `personal_sendTransaction`, and it can be seen by using Mist as it doesn't set the nonce field when sending a tx. This can be reproduced by either:
* Starting the miner after the txs were sent.
* Increasing the mining difficulty to allow to send the txs so that they are attempted to be included in the same block.

## Fix
The pending txs of the same sender are considered when setting the nonce of tx sent.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
@@ -1,19 +1,23 @@
 package io.iohk.ethereum.jsonrpc
 
+import akka.pattern.ask
 import akka.actor.ActorRef
-import akka.util.ByteString
+import akka.util.{ByteString, Timeout}
 import io.iohk.ethereum.crypto
 import io.iohk.ethereum.crypto.ECDSASignature
 import io.iohk.ethereum.db.storage.AppStateStorage
-import io.iohk.ethereum.domain.{Account, Address, Blockchain, BlockchainStorages}
+import io.iohk.ethereum.domain.{Account, Address, Blockchain}
 import io.iohk.ethereum.jsonrpc.PersonalService._
 import io.iohk.ethereum.keystore.{KeyStore, Wallet}
 import io.iohk.ethereum.jsonrpc.JsonRpcErrors._
-import io.iohk.ethereum.transactions.PendingTransactionsManager.AddTransactions
+import io.iohk.ethereum.transactions.PendingTransactionsManager
+import io.iohk.ethereum.transactions.PendingTransactionsManager.{AddTransactions, PendingTransactionsResponse}
+import io.iohk.ethereum.utils.MiningConfig
 
 import scala.collection.mutable
 import scala.concurrent.Future
 import scala.concurrent.ExecutionContext.Implicits.global
+import scala.util.Try
 
 object PersonalService {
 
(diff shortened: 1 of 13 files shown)


src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
```diff
@@ -114,32 +119,47 @@ class PersonalService(
...
+    val pendingTxsFuture = (txPool ? PendingTransactionsManager.GetPendingTransactions).mapTo[PendingTransactionsResponse]
+    val latestPendingTxNonceFuture: Future[Option[BigInt]] = pendingTxsFuture.map { pendingTxs =>
+      val senderTxsNonces = pendingTxs.pendingTransactions
+        .collect { case ptx if ptx.stx.senderAddress == wallet.address => ptx.stx.tx.nonce }
+      Try(senderTxsNonces.max).toOption
+    }
+    latestPendingTxNonceFuture.map{ maybeLatestPendingTxNonce =>
+      val maybeCurrentNonce = getCurrentAccount(request.from).map(_.nonce.toBigInt)
+      val maybeNextTxNonce = maybeLatestPendingTxNonce.map(_ + 1) orElse maybeCurrentNonce
+      val tx = request.toTransaction(maybeNextTxNonce.getOrElse(Account.Empty.nonce))
```
rtkaczyk: Does it make sense to send a TX if account is empty (and therefore has zero balance)?

src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
```diff
@@ -114,32 +119,47 @@ class PersonalService(
...
+        Future.successful(Left(AccountLocked))
     }
   }
 
-  private def sendTransaction(request: TransactionRequest, wallet: Wallet): ByteString = {
-    val defaultNonce = getCurrentAccount(request.from).getOrElse(Account.Empty).nonce
-    val tx = request.toTransaction(defaultNonce)
-    val stx = wallet.signTx(tx)
+  private def sendTransaction(request: TransactionRequest, wallet: Wallet): Future[ByteString] = {
+    implicit val timeout = Timeout(miningConfig.poolingServicesTimeout)
```
rtkaczyk: I'm wondering whether `poolingServicesTimeout` should be moved somewhere else. Having a TX pool doesn't have to mean we're doing any mining. `PersonalService` is also not dependent on mining, so passing the mining config to feels kinda strange.

src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
ntallar: It could make sense for the case where this TX results in the creation of the sender's account (as happened in https://github.com/input-output-hk/etc-client/pull/173) if the gasPrice of the TX is 0.

rtkaczyk: Huh, forgot about this.. Still doesn't make much sense to me, but what can you do? :)
👍

PR #239, opened 2017-07-04
[EC-229] Setup for test networks

adamsmo:
this pr contains configuration for private networks and changes to allow configuring initial account nonce


src/main/resources/morden-application.conf
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+
+  txPool {
+    tx-pool-size = 1000
+    pending-tx-manager-query-timeout = 5.seconds
+  }
+
+  mining {
+    tx-pool-size = 1000
```
ntallar: Minor detail: This configuration was moved to the above `txPool` so it's no longer needed in `mining`.

src/main/resources/morden-application.conf
```diff
@@ -0,0 +1,147 @@
...
+    ommers-pool-size = 30
+    block-cashe-size = 30
+    coinbase = "0011223344556677889900112233445566778899"
+    pooling-services-timeout = 5.seconds
+  }
+
+  blockchain {
+    frontier-block-number = "0"
+    homestead-block-number = "494000"
+    eip150-block-number = "1783000" //todo add EIP155 ??
```
ntallar: Should we maybe add a task to take a look into this? There also some more parameters such as `gasLimitBoundDivisor` and `difficultyBoundDivisor` that are accepted in parity

src/main/resources/morden-application.conf
```diff
@@ -0,0 +1,147 @@
...
+    eip160-block-number = "1915000"
+    difficulty-bomb-pause-block-number = "1915000"
+    difficulty-bomb-continue-block-number = "3415000"
+
+    // Doc: https://blog.ethereum.org/2016/07/20/hard-fork-completed/
+    dao-fork-block-number = "1783000"
+    dao-fork-block-total-difficulty = "131072"//todo ?????? check that
+    dao-fork-block-hash = "f376243aeff1f256d970714c3de9fd78fa4e63cf63e32a51fe1169e375d98145"
+
+    account-start-nonce = "1048576"
```
rtkaczyk: Any idea why is it required for morden?

src/main/scala/io/iohk/ethereum/domain/Account.scala
```diff
@@ -13,7 +13,7 @@ object Account {
   val EmptyStorageRootHash = ByteString(kec256(rlp.encode(Array.empty[Byte])))
   val EmptyCodeHash: ByteString = kec256(ByteString())
 
-  val Empty = Account(0, 0, EmptyStorageRootHash, EmptyCodeHash)
+  def empty(startNonce: UInt256): Account = Account(nonce = startNonce, storageRoot = EmptyStorageRootHash, codeHash = EmptyCodeHash)
```
rtkaczyk: Should it maybe default to zero? Could save some keystrokes in the tests.

src/main/resources/morden-application.conf
adamsmo: because of this:
http://ethereum-classic-guide.readthedocs.io/en/latest/network/test-networks.html
and this
https://github.com/paritytech/parity/blob/b50fb71dd1d29dfde2a6c7e1830447cf30896c31/ethcore/res/ethereum/morden.json#L26 (

src/main/resources/morden-application.conf
adamsmo: I am running this on EC2 to get this number, but will search for online morden browser

adamsmo: 155 is signing with new schema, sow this should work without doing anything, because we are accepting txs signed with new schema
I will add task to add configuration value for this, to make us issue transactions with chain specific signature

src/main/resources/morden-application.conf
rtkaczyk: Sure, I realise it's there. But it's a theoretical question: why on Earth would we want to tinker with that?

adamsmo: to be able to connect to test networks and test our client further

rtkaczyk: Why would want to change the initial nonce on any network?

adamsmo: because this is the specification of morden test network, the test network that we can connect to without implementing EIP-158
I think it is changed to differentiate from main net

rtkaczyk: Adam, I'm asking you theoretical question and you keep saying "because it's in the spec". I'm not interested in what the doc says, I can read. I'm interested in what it does not say, which is the motivation for adjusting this parameter.

> I think it is changed to differentiate from main net

I don't think that's necessary. Ropsten has the initial nonce as zero.

I'm not questioning your code, which is fine in this regard. But while touching on certain new functionality we might as well discuss it, to get a better understanding. I certainly would like to.

adamsmo: ok, so my answers was maybe not clear, only reason to add this is because it is in spec and we cannot connect to network if we set nonce according to YP
this change is needed because someone decided to set initial account nonce to different value when setting up morden

PR #240, opened 2017-07-05
[EC-230] Support TX override

rtkaczyk:
Also taking care of EC-225, which boils down to removing a TODO comment


src/test/scala/io/iohk/ethereum/transactions/PendingTransactionsManagerSpec.scala
```diff
@@ -107,15 +108,46 @@ class PendingTransactionsManagerSpec extends FlatSpec with Matchers with ScalaFu
...
+    pendingTransactionsManager ! AddOrOverrideTransaction(overrideTx)
+    peerManager.expectMsg(PeerManagerActor.GetPeers)
+    peerManager.reply(Peers(Map(peer1 -> Handshaked)))
+
+    val pendingTxs = (pendingTransactionsManager ? GetPendingTransactions).mapTo[PendingTransactionsResponse]
+      .futureValue.pendingTransactions
+
+    pendingTxs.map(_.stx) shouldEqual List(overrideTx, otherTx)
+
+    // overriden TX will still be broadcast to peers
```
mcsherrylabs: Why do we broadcast the overriden tx? :thinking:

rtkaczyk: Because we do that immediately upon submission of a TX. We could for example introduce a timed delay, to give change for an override to discard the old TX before it gets broadcast, but do we want to go down this road?

mcsherrylabs: Nope.

PR #242, opened 2017-07-05
[EC-193] Fixed TCP connection failing (due to Write failing)

ntallar:
## Description

As stated in http://doc.akka.io/docs/akka/current/scala/io-tcp.html#throttling-reads-and-writes, Akka TCP doesn't have any internal buffering, which causes the TCP connection to fail randomly when too much stress is put into it. The problem that was raised in our client was `[Stopping Connection] Write to peer ... failed, trying to send ...`, which was caused by a `CommandFailed` message received by our client.
In order to almost always reproduce this bug, the problem can be increased by having all `NewBlock`s sent to every peer, by changing `BlockBroadcast.shouldSendNewBlock` to return always `true`.

## Fix

Akka TCP's internal `Ack`s and message buffering were used to prevent saturating the TCP Akka connection.


src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
```diff
@@ -136,21 +138,53 @@ class RLPxConnectionHandler(
...
       handleWriteFailed orElse handleConnectionClosed orElse {
         case sm: SendMessage =>
-          val out = messageCodec.encodeMessage(sm.serializable)
-          connection ! Write(out)
-          log.debug("Sent message: {}", sm.serializable)
+          if(messageAwaitingAck.isEmpty) {
+            sendMessage(messageCodec, sm.serializable)
+            context become handshaked(messageCodec, messagesNotSent, Some(sm.serializable))
+          } else
+            context become handshaked(messageCodec, messagesNotSent :+ sm.serializable, messageAwaitingAck)
```
mcsherrylabs: If we miss an Ack or the Ack is never sent because the other client dies will we queue messages forever? If so, we'll need a timeout on send, probably no need to resend?

ntallar: I agree, we should add a timeout for the `Ack`s :+1: Taking into account this losing of the `Ack` doesn't happen frequently (it doesn't seem to have happened while testing the current implementation, but i'll take a look at that while testing handling timeouts) I also think that it's not needed to resend the message.
BTW, this `Ack` is only an internal Akka message, it let's us know that the message was properly enqueued for sending by the OS (as explained in https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/io/Tcp.scala#L325), so it seems to be independent on how the other client acts.

ntallar: After further testing, I realized that sending new messages upon `Ack` timeouts so didn't work and caused `Write to ... peer failed` errors. The reason for this was the same as for the problem that is being solved in this PR, which is that a msgs was written to the TCP connection while another one's queueing was in progress (the one we are awaiting for the `Ack`).
We should then close the TCP connection when encountering that an expected `Ack` was not received.

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
```diff
@@ -49,7 +51,7 @@ class RLPxConnectionHandler(
 
     case HandleConnection(connection) =>
       connection ! Register(self)
-      val timeout = system.scheduler.scheduleOnce(3.seconds, self, AuthHandshakeTimeout)
+      val timeout = system.scheduler.scheduleOnce(rlpxConfiguration.waitForHandshakeTimeout, self, AuthHandshakeTimeout)
```
mcsherrylabs: :1st_place_medal:

src/test/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandlerSpec.scala
```diff
@@ -70,6 +73,23 @@ class RLPxConnectionHandlerSpec extends FlatSpec with Matchers with MockFactory
...
+
+    setupRLPxConnection()
+
+    rlpxConnectionParent watch rlpxConnection
+
+    rlpxConnection ! RLPxConnectionHandler.SendMessage(Ping())
+    connection.expectMsg(Tcp.Write(ByteString("ping encoded"), RLPxConnectionHandler.Ack))
+
+    //Wait for Ack timeout
+    Thread.sleep(rlpxConfiguration.waitForTcpAckTimeout.toMillis)
```
mcsherrylabs: @rtkaczyk Did you find a better way of doing this? Nico, It's not possible to use `eventually` here?

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
```diff
@@ -136,21 +140,64 @@ class RLPxConnectionHandler(
...
+          } else
+            context become handshaked(messageCodec, messagesNotSent :+ sm.serializable, tcpAckTimeout)
 
         case Received(data) =>
           val messages = messageCodec.readMessages(data)
           messages foreach processMessage
+
+        case Ack if tcpAckTimeout.nonEmpty =>
+          //Cancel pending message timeout
+          tcpAckTimeout.foreach(_.cancel())
```
adamsmo: is it possible that we get race condition where timeout will send `AckTimeout` message to us before we cancel it and as a result this timeout will be treated as timeout for message we are sending in this case
maybe we can distinguish `AckTimeout` somehow to be able to determine which `AckTimeout` is for which message?

adamsmo: is it possible that we get race condition where timeout will send AckTimeout message to us before we cancel it and as a result this timeout will be treated as timeout for message we are sending in this case
maybe we can distinguish AckTimeout somehow to be able to determine which AckTimeout is for which message?

src/test/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandlerSpec.scala
rtkaczyk: @mcsherrylabs @ntallar actually in this case there should be a better way:

```
rlpxConnectionParent.expectTerminated(rlpxConnection, max = rlpxConfiguration.waitForTcpAckTimeout)
```

An expectation with a timeout is always better than a `Thread.sleep`.
But in general, in many places we do require those sleeps, because we lack feedback from actors under test. We would have make certain design changes (new messages) - to make those actors test-oriented.

Those sleeps are the most critical factor increasing our build times. At some point (probably not now) this issue might be worth the effort.

ntallar: Added using `max` instead of `Thread.sleep` :+1: I used: `rlpxConnectionParent.expectTerminated(rlpxConnection, max = 2 * rlpxConfiguration.waitForTcpAckTimeout)` as the test was sometimes failing as it the RLPxConnecitonHandler didn't have enough time to process the timeout.

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
ntallar: You are right, just added a fix for this with the test scenario that you described.
I finally decided to distinguish AckTimeout with a sequence number (my first idea was distinguishing by message but for example the `Pong` message could be repeated).

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
```diff
@@ -136,21 +140,84 @@ class RLPxConnectionHandler(
...
+      *
+      * @param messageCodec, for encoding the messages sent
+      * @param messagesNotSent, messages not yet sent
+      * @param cancellableAckTimeout, timeout for the message sent for which we are awaiting an acknowledgement (if there is one)
+      * @param nextSeqNumber, sequence number for the next message to be sent
+      */
+    def handshaked(messageCodec: MessageCodec,
+                   messagesNotSent: Queue[MessageSerializable] = Queue.empty,
+                   cancellableAckTimeout: Option[CancellableAckTimeout] = None,
+                   nextSeqNumber: Int = 0): Receive =
```
adamsmo: should it be called `nextSeqNumber` instead of `seqNumber`?

ntallar: Do you mean that the `seqNumber` parameter from the `sendMessage` function should be renamed `nextSeqNumber`? Or that this should be called `seqNumber` instead of `nextSeqNumber`?

adamsmo: for me it looks like they are both current seq number because you increase it in sendMessage

ntallar: You are right, it does seem confusing, renamed to `seqNumber` :+1:

PR #243, opened 2017-07-06
[EC-225] Update not mined trasanction?

adamsmo:


src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
```diff
@@ -37,11 +37,13 @@ class BlockGenerator(blockchainStorages: BlockchainStorages, blockchainConfig: B
...
         val header: BlockHeader = prepareHeader(blockNumber, ommers, beneficiary, parent, blockTimestamp)
 
-        val transactionsForBlock = transactions
-          .filter(_.tx.gasLimit < header.gasLimit)
-          //if we have 2 transactions from same address we want first one with lower nonce
-          .sortBy(_.tx.gasPrice)
-          .sortBy(_.tx.nonce)
+        val sortedTransactions = transactions.groupBy(_.senderAddress).values.toList.flatMap { txsFromSender =>
+          val ordered = txsFromSender.sortBy(_.tx.gasPrice).reverse.sortBy(_.tx.nonce)
+            .takeWhile(_.tx.gasLimit <= header.gasLimit)
```
rtkaczyk: Do we need to check for gasLimit? Isn't this already taken care of by `SignedTransactionValidator`?

rtkaczyk:
I would also add a test for consecutive TXs from a single sender.

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
adamsmo: if we will not check it here all transactions after this will fail (from that account) and we will include less valid transactions in block
but this is mainly because here
https://github.com/input-output-hk/etc-client/pull/243/files#diff-6d3634062464c02321dc5a991254d727R49
we would prepare empty block if tx with gas limit set to > blockGasLimit would be first

src/test/scala/io/iohk/ethereum/mining/BlockGeneratorSpec.scala
```diff
@@ -92,6 +92,27 @@ class BlockGeneratorSpec extends FlatSpec with Matchers with PropertyChecks with
...
+    val result: Either[BlockPreparationError, PendingBlock] =
+      blockGenerator.generateBlockForMining(1, Seq(txWitSameNonceButLowerGasPrice, signedTransaction), Nil, Address(testAddress))
+    result shouldBe a[Right[_, Block]]
+
+    //mined with etc-client + ethminer
+    val minedNonce = ByteString(Hex.decode("5e8d5c12cea7e0c7"))
+    val minedMixHash = ByteString(Hex.decode("9247b81258f97159f987a5f4f9e94df1d95e10eeabff2836020eafb27a8228b0"))
+    val miningTimestamp = 1494604913
+
+    val fullBlock: Either[BlockPreparationError, Block] = result.right
```
ntallar: Minor detail: Should we maybe include a check to see that `signedTransaction` is included in the `fullBlock` and not `txWitSameNonceButLowerGasPrice`?

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
```diff
@@ -37,11 +37,13 @@ class BlockGenerator(blockchainStorages: BlockchainStorages, blockchainConfig: B
         val blockTimestamp = blockTimestampProvider.getEpochSecond
         val header: BlockHeader = prepareHeader(blockNumber, ommers, beneficiary, parent, blockTimestamp)
 
-        val transactionsForBlock = transactions
-          .filter(_.tx.gasLimit < header.gasLimit)
-          //if we have 2 transactions from same address we want first one with lower nonce
-          .sortBy(_.tx.gasPrice)
-          .sortBy(_.tx.nonce)
+        val sortedTransactions = transactions.groupBy(_.senderAddress).values.toList.flatMap { txsFromSender =>
+          val ordered = txsFromSender.sortBy(_.tx.gasPrice).reverse.sortBy(_.tx.nonce)
```
ntallar: Could we maybe use only one tx per sender and nonce? This is for example when we receive 2 txs from a peer with the same sender and nonce (if someone is overriding one of it's txs). I think that in `Ledger.prepare` we already remove txs whose execution fails so this is up to discussion.

src/test/scala/io/iohk/ethereum/mining/BlockGeneratorSpec.scala
adamsmo: will add

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
adamsmo: yes we remove failing transaction from list there, what you propose would split validation logic into 2 places

rtkaczyk: We already have the validation split, because of the gas limit check. What @ntallar proposes is rather an optimisation, we know this TX will fail, so why bother executing it? 

But I think the optimisation should go even further. Currently when we have a failing TX we start over, executing the block from scratch with the failing TX filtered out. But we should be able to revert to the state after previous TX and just continue from there. This definitely a separate task though.

adamsmo: ok, will add that task

PR #244, opened 2017-07-07
[EC-238] Fixed FIXMEs in the codebase

ntallar:
## Description

Includes fixes to several `FIXME`s in the codebase. The remaining `FIXME`s were analyzed (with conclusions in task). The fixed `FIXME`s in this PR are:
- *BlockValidator: Can we avoid encoding ommers again? -> Removed FIXME*: This would require rethinking how messages are received from other peers, so as to leave the bytes of the messages cached, as Parity does. It doesn't seem worth the changes needed, as it would only improve the encoding of the ommers (for example, Geth doesn't implement this optimization), so the FIXME was removed.
- *InMemorySimpleMapProxy: We can cache retrieved values too -> Removed FIXME*: The optimization makes sense but it would require either: making the cache a var or returning a new InMemorySimpleMapProxy but it would unnecesary overcomplicate the interface. Due to this the `FIXME` was removed.
- *ObjectGenerators: Add use of tx generator when we are able to sign txs -> Fixed FIXME*: Changed `NewBlock` generator to include txs.
- *PeerActor: This message decoder should be configurable -> Fixed FIXME*: Made `MessageDecoder` configurable on `PeerActor` and when creating `PeerManagerActor`.
- *PeerEventBusActor: Remove both var -> Removed FIXME*: Removing both var doesn't fit well into the interface used by ActorEventBus (for example `ActorEventBus` requires `subscribe` being as `def subscribe(subscriber: ActorRef, to: Classifier): Boolean`). Due to this the `FIXME` was removed.
- *RLPSuite: this is used to test nested objects encoding. We can think of replacing this with eth real implementation -> Removed FIXME*: Removed `FIXME` as it makes sense to have a simple test for nested objects encoding, real block encoding should be tested somewhere else if it already isn't.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/db/components/StoragesComponent.scala
+++ b/src/main/scala/io/iohk/ethereum/db/components/StoragesComponent.scala
@@ -19,7 +19,7 @@ trait StoragesComponent {
 
     val mptNodeStorage: MptNodeStorage
 
-    val nodeStorage: NodeStorage //FIXME This storage is similar to MPTNodesStorage but
+    val nodeStorage: NodeStorage //FIXME This storage is similar to MPTNodesStorage, should we keep only one? [EC-147]
 
     val evmCodeStorage: EvmCodeStorage
 
--- a/src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
+++ b/src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
@@ -361,7 +361,9 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
     * The contract storage should be cleared during pruning as nodes could be used in other tries.
     * The contract code is also not deleted as there can be contracts with the exact same code, making it risky to delete
     * the code of an account in case it is shared with another one.
-    * FIXME: Should we keep track of this for deletion? Maybe during pruning we can also prune contract code.
+    * FIXME: [EC-242]
+    *   Should we delete the storage associated with the deleted accounts?
+    *   Should we keep track of duplicated contracts for deletion?
     *
     * @param addressesToDelete
     * @param worldStateProxy
--- a/src/test/scala/io/iohk/ethereum/ObjectGenerators.scala
+++ b/src/test/scala/io/iohk/ethereum/ObjectGenerators.scala
@@ -1,6 +1,7 @@
 package io.iohk.ethereum
 
 import java.math.BigInteger
+import java.security.SecureRandom
 
 import akka.util.ByteString
 import io.iohk.ethereum.mpt.HexPrefix.bytesToNibbles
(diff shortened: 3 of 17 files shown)


src/main/scala/io/iohk/ethereum/db/components/StoragesComponent.scala
```diff
@@ -19,7 +19,7 @@ trait StoragesComponent {
 
     val mptNodeStorage: MptNodeStorage
 
-    val nodeStorage: NodeStorage //FIXME This storage is similar to MPTNodesStorage but
+    val nodeStorage: NodeStorage //FIXME This storage is similar to MPTNodesStorage, should we keep only one? [EC-147]
```
adamsmo: they are pointing underneath to the same namespace `Namespaces.NodeNamespace`, but they are used for 2 different purpose, one is for fast sync it operates on objects defined for wire protocol 
second is for our internal implementation of mpt it is used when executing transactions and it uses objects defined for mpt

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -361,7 +361,9 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
     * The contract storage should be cleared during pruning as nodes could be used in other tries.
     * The contract code is also not deleted as there can be contracts with the exact same code, making it risky to delete
     * the code of an account in case it is shared with another one.
-    * FIXME: Should we keep track of this for deletion? Maybe during pruning we can also prune contract code.
+    * FIXME: [EC-242]
+    *   Should we delete the storage associated with the deleted accounts?
+    *   Should we keep track of duplicated contracts for deletion?
```
adamsmo: storage can be shared between many accounts
it is just mpt identified by hash, if 2 accounts are storing value `1` under key `name` they will share the same storage, so this should be part of pruning

what is duplicated contract?

src/test/scala/io/iohk/ethereum/ObjectGenerators.scala
```diff
@@ -101,12 +102,20 @@ trait ObjectGenerators {
     }
   }
 
-  //FIXME: Add use of tx generator when we are able to sign txs
-  def newBlockGen: Gen[NewBlock] = for {
+  def signedTxSeqGen(length: Int, secureRandom: SecureRandom, chainId: Option[Byte]): Gen[Seq[SignedTransaction]] = {
```
adamsmo: 👌

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
ntallar: Duplicated contracts are generated when deploying the same contract code but with different accounts. So when deleting an account we can't easily delete it's contract code as it could be shared with other contract (and we currently index it by code hash).

adamsmo: but they are 2 different contracts, which have same code, it is the same issue as with other parts of state
in my opinion deletion of code should be governed by pruning

src/main/scala/io/iohk/ethereum/db/components/StoragesComponent.scala
ntallar: Yes, currently two different classes are used for the different purposes of the nodes (LeafNode , ExtensionNode , BranchNode  with MPTLeaf , MPTExtension , MPTBranch), but apart from the `..Node` having hash configurable they are exactly the same, even in the way they are encoded with RLP. So task `EC-147` would involve seeing if we can remove this duplicated code and while doing so also unifying the storages.

adamsmo: 👍

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
ntallar: I agree, IMO we should also count references for code :+1: 
So we could take a look at this and doing deletion of storage after some basic pruning is implemented (as it seems that Geth doesn't yet implement deleting contract code)

src/main/scala/io/iohk/ethereum/db/components/StoragesComponent.scala
AlanVerbner: Yes...maybe we need to remove the "configurable hash" on `Mpt` . I think the unification and storage removal is a good idea but it's a task that will take some hours. I would create a ticket in order to address this in the future tho.

ntallar: I agree, there's already a task created for looking at this [EC-147](https://iohk.myjetbrains.com/youtrack/issue/EC-147) :+1:

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
AlanVerbner: In basic pruning only mpt nodes will be deleted (this includes account and account's storage) but not code. Maybe we can address that later if needed (haven't checked yet if contract code storage in relation with others)

PR #245, opened 2017-07-11
Fix calcMemCost

adamsmo:


src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -80,13 +80,13 @@ case class EvmConfig(
...
 
-    val memNeeded = if (dataSize.isZero) UInt256.Zero else offset + dataSize
+    val memNeeded = if (dataSize.isZero) BigInt(0) else offset.toBigInt + dataSize.toBigInt
     if (memNeeded > MaxMemory)
       UInt256.MaxValue / 2
     else if (memNeeded <= memSize)
       0
     else
-      c(memNeeded) - c(memSize)
+      c(UInt256(memNeeded)) - c(memSize)
```
rtkaczyk: Converting back to `UInt256` is weird. Wouldn't it be simpler to change the signatures of the involved functions: `calcMemCost`,`c`, `wordsForBytes`?

adamsmo: unfortunately it has to be wrapped into UInt256 and converted to big int with wordsForBytes otherwise test are failing

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -75,18 +75,18 @@ case class EvmConfig(
...
   def calcMemCost(memSize: UInt256, offset: UInt256, dataSize: UInt256): BigInt = {
     /** See YP H.1 (222) */
-    def c(m: UInt256): BigInt = {
+    def c(m: BigInt): BigInt = {
       val a = wordsForBytes(m)
       G_memory * a + a * a / 512
     }
 
-    val memNeeded = if (dataSize.isZero) UInt256.Zero else offset + dataSize
+    val memNeeded = if (dataSize.isZero) BigInt(0) else offset.toBigInt + dataSize.toBigInt
```
rtkaczyk: We're still mixing the types `UInt256` and `BigInt`, especially `memNeeded <= memSize`. Now, I can't see anything wrong with, but I also didn't see anything wrong all `UInt256` implementation until the bug manifested. Why don't you want to simply change the types in the signature?

src/main/scala/io/iohk/ethereum/vm/package.scala
```diff
@@ -5,6 +5,6 @@ package object vm {
   /**
     * Number of 32-byte UInt256s required to hold n bytes (~= math.ceil(n / 32))
     */
-  def wordsForBytes(n: UInt256): BigInt =
-    if (n.isZero) 0 else (n - 1) / UInt256.Size + 1
+  def wordsForBytes(n: BigInt): BigInt =
+    if (n.compare(0) == 0) 0 else (n - 1) / UInt256.Size + 1
```
rtkaczyk: In `n.compare(0)` the argument is still converted to a `BigInt`, so I don't see how that's better than the natural `n == 0`

adamsmo: it does not work with `n == 0`

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
adamsmo: we are now using BigInt internaly

rtkaczyk: I guess that's ok, but could you please tell me why you didn't want to change the signature?

adamsmo: to keep this change small

rtkaczyk: changing the signature would not affect other files, because uints are automatically converted to bigints

adamsmo: you convinced me, this also makes one of our tests failing so I fixed it as well

src/test/scala/io/iohk/ethereum/transactions/PendingTransactionsManagerSpec.scala
```diff
@@ -115,14 +115,17 @@ class PendingTransactionsManagerSpec extends FlatSpec with Matchers with ScalaFu
     pendingTransactionsManager ! AddOrOverrideTransaction(firstTx)
     peerManager.expectMsg(PeerManagerActor.GetPeers)
     peerManager.reply(Peers(Map(peer1 -> Handshaked)))
+    Thread.sleep(Timeouts.normalTimeout.toMillis)
```
mcsherrylabs: Really necessary? :/

adamsmo: I think yes, I was debugging this test for few hours, after changes in EvmConfig.scala and in package.scala this test was failing

rtkaczyk: It is better to have an expectation with a timeout than a `Thread.sleep`. Looking at the code I don't believe the waiting between the individual messages is required. My guess is it's the total waiting time since the first one that did the trick, so I would prefer a `longTimeout` in the expectation.

However, given how crazy this has been, I don't want to spend too much time on this, especially as part of this task...

adamsmo: I think it is required I tested case with waiting only after last `peerManager.reply(Peers(Map(peer1 -> Handshaked)))`

rtkaczyk: Did you wait `3 * Timeouts.normalTimeout`?

adamsmo: no I waited `1 *  Timeouts.normalTimeout` but I tested it now and after waiting `3 * Timeouts.normalTimeout` it fails

rtkaczyk: Well, I tried [this](https://github.com/input-output-hk/etc-client/commit/51c0861f63add7fe33c7a41ffecb60b087bb5849) and it [passed](https://circleci.com/gh/input-output-hk/etc-client/2299) on Circle

adamsmo: as I said it was it is not always failing but it fails

rtkaczyk: Thanks, I'm convinced 😁

PR #248, opened 2017-07-13
[EC-241] Add configuration to enable EIP-155 at specific block

adamsmo:
this PR adds configuration for EIP-155
it will reject chain specific transactions before fork block


src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
```diff
@@ -157,7 +157,12 @@ class PersonalService(
       val maybeCurrentNonce = getCurrentAccount(request.from).map(_.nonce.toBigInt)
       val maybeNextTxNonce = maybeLatestPendingTxNonce.map(_ + 1) orElse maybeCurrentNonce
       val tx = request.toTransaction(maybeNextTxNonce.getOrElse(blockchainConfig.accountStartNonce))
-      val stx = wallet.signTx(tx)
+
+      val stx = if(blockchain.getBlockByNumber(blockchainConfig.eip155BlockNumber).isDefined){
+        wallet.signTx(tx, Some(blockchainConfig.chainId))
+      }else{
+        wallet.signTx(tx, None)
+      }
```
rtkaczyk: Is this right? However unlikely this may be, say we're at a fork have the `X` block (where the EIP155 changeover happens), but a chain with a top block `X - 2` is better, so this TX will go into block `X - 1`...

adamsmo: I think yes, and it should work like this in that case transaction will fail if it is chain specific, we can change that and issue chain specific transactions after x blocks after for to be sure that it will not fail in case of block reorganisation since global transactions are still supported

ntallar: It also seems that Geth implements sending tx as we do (https://github.com/ethereumproject/go-ethereum/blob/master/eth/api.go#L497) but we could add what @adamsmo recommended as an optimization.

rtkaczyk: Well, not exactly. While the current method is perhaps not a problem, it makes more sense to me to do `appStateStorage.getBestBlockNumber() +1 >= blockchainConfig.eip155BlockNumber` (it may also be faster?). And that, I think, would be more similar to what geth does.

adamsmo: but what if branch is longer than 1, in that case your previous argument will apply, this is why I proposed to set this to arbitrary configured value instead of `1` which may be as good as 
checking for exact bestBlockNumber / checking if fork block exists

rtkaczyk: Yes, it doesn't address the chain re-org scenario, but after some thought, I don't think that issue is of any practical importance. Worst case, we get some failed TXs in a very short and unlikely timeframe, and only on a test/private network. If you think it's worth the effort and complication in our code then go for it. I doubt it, though.

Nonetheless, I would still prefer using `getBestBlockNumber()` in that condition, just because it's more logical and possibly faster.

src/main/resources/blockchain/morden.json
```diff
@@ -0,0 +1,17 @@
+{
```
rtkaczyk: This is already in `universal`, please remove.

src/main/resources/morden-application.conf
```diff
@@ -0,0 +1,147 @@
+grothendieck {
```
rtkaczyk: This is already in `universal`, please remove.

src/main/resources/application.conf
```diff
@@ -139,6 +139,12 @@ grothendieck {
     # Doc: https://github.com/ethereum/EIPs/issues/150
     eip150-block-number = "2500000"
 
+    # EIP-155 fork block number
+    # Doc: https://github.com/ethereum/eips/issues/155
+    # 3 000 000 because of
+    #https://github.com/paritytech/parity/blob/b50fb71dd1d29dfde2a6c7e1830447cf30896c31/ethcore/res/ethereum/classic.json#L15
```
rtkaczyk: While this may be useful for us, we have to keep in mind that is meant to be read by users and we don't want to form an impression like: "Hey, we don't know what we're doing, we're just copying from other guys". So I would remove the reference to parity.

This is somewhat political so calling @mcsherrylabs to have a final say.

Once done with this description, please copy it to `universal/blockchain.conf`.

rtkaczyk: Alternatively, we could have those description differ slightly here and in `blockchain.conf`

mcsherrylabs: It's ok to say 'Following lead of existing clients implementation to maintain compatibility <link>'. Which I think is what we did here in the absence of a clearly agreed spec....?

src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
```diff
@@ -157,7 +157,12 @@ class PersonalService(
       val maybeCurrentNonce = getCurrentAccount(request.from).map(_.nonce.toBigInt)
       val maybeNextTxNonce = maybeLatestPendingTxNonce.map(_ + 1) orElse maybeCurrentNonce
       val tx = request.toTransaction(maybeNextTxNonce.getOrElse(blockchainConfig.accountStartNonce))
-      val stx = wallet.signTx(tx)
+
+      val stx = if (appStateStorage.getBestBlockNumber() >= blockchainConfig.eip155BlockNumber) {
```
rtkaczyk: No `- 1`? I'd say that if we're at the block `X - 1` we already want to change the signing method for the new TXs, cause the will (most likely) end up in block X.

adamsmo: technically yes
I wrote it like this to produce chain specific transactions after fork block it is more logical to me and not chain specific transactions are valid after fork

PR #249, opened 2017-07-13
[EC-239] Fixed TODOs in the codebase

ntallar:
## Description

Includes fixes to several `TODO`s in the codebase. The remaining `TODO`s were analyzed (with conclusions in task). The fixed `TODO`s in this PR are:
- *InMemoryWorldStateProxy: change to MerklePatriciaTree[Address, Account] and use an Address seraializer that is defined in one place -> Fixed TODO*.
- *EthService: Configure timeout -> Fixed TODO*
- *SyncController: refactor to provide ActorRefs differently -> Removed TODO*: The idea of this TODO is reducing the number of parameters for the props function, however there's not much relation between those ActorRefs so it's not worth grouping actors.
- *InMemoryWorldStateProxy: why do we create blockchain every time we are calling this function? -> Fixed TODO*: Blockchain is now being created before creating funcion for accessing blocks, so as to fix this problem.
- *DumpChainActor: ask for block headers -> Removed TODO*: outdated TODO already fixed.
- *Ledger: why do we require in executeBlockTransactions separate blockchain and blockchain storage -> Left TODO*: Task [EC-253] was created to investigate this.
- *SyncControllerSpec: Investigate long timeouts -> Removed TODO*: This problems seems to have been caused by grouping several unconnected messages. Removed awaiting for unnecessary messages that were not being tested in these tests, which allowed lowering the timeout.
- *BlockHeaderValidator: should we report this? -> Removed TODO*: Reported difference between clients and YP. The use of 5000 as MinGasPrice can be seen for example in geth-classic: https://github.com/ethereumproject/go-ethereum/blob/master/core/block_validator.go#L36 and in parity: https://github.com/paritytech/parity/blob/master/ethcore/res/ethereum/classic.json#L29. Created issue https://github.com/ethereum/yellowpaper/issues/317.


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -57,8 +57,6 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
     val blockchain = BlockchainImpl(storages)
 
     val blockExecResult = for {
-      //todo why do we require in executeBlockTransactions separate blockchain and blockchain storage
-      //todo blockchain storage is part of blockchain
```
adamsmo: should we refactor this part to pass only 1 thing?

rtkaczyk: Maybe `Blockchain` should `extend BlockchainStorages` ?

ntallar: :thinking: 
That would be practical, however I'm not sure whether we should do because so far `Blockchain` is more of an abstract trait without implementation details such as the storages used.
Maybe we should probably add a function like `getWorldState` to the `Blockchain`, to solve this.

ntallar: While looking at this I also realized that we're using `BlockchainStorages` and `BlockchainImpl` in several places of our code, maybe it should be better to rather work with `Blockchain`, as we are mixing the implementation and the abstraction of our `Blockchain` as in here? If so, we should create a task to look into this, and re-add this TODO.

AlanVerbner: As we are currently using `Blockchain` to get state related data, what about if blockchain also allows to retrieve the WorldStateProxy, something like `def getWorldStateProxy(stateRootHash: Option[ByteString] = None): WS` and let BlockchainImple to create the WorldStateProxy using the storages it has.

I have a working version [here](https://github.com/input-output-hk/etc-client/commit/fdcb131512ac3b66871acad6e259363739662f00#diff-abb3f698c5418831dcb5f183fbf61937R243) as I needed to do some work in that code (If you think it's not ok, I can do a rollback tho).

adamsmo: @ntallar maybe we can create another task for this refactor and close this PR?

rtkaczyk: @adamsmo EC-253

PR #250, opened 2017-07-14
[EC-245] Fixed how node key is restored

ntallar:
## Description
In the scenario that the the node keys were already generated (in a previous execution of our client), when loaded and used to sign any message (using `ECDSASignature.sign(forSig, keyPair, None)`), a `null` exception happened.

## Fix
This exception was caused by not restoring properly the node keys, as the public key was being used as the private key, which was fixed in this PR.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/network/package.scala
+++ b/src/main/scala/io/iohk/ethereum/network/package.scala
@@ -31,8 +31,8 @@ package object network {
       val keysValuePair = generateKeyPair(secureRandom)
 
       //Write keys to file
-      val (_, priv) = keyPairToByteArrays(keysValuePair)
-      file.getParentFile.mkdirs()
+      val (priv, _) = keyPairToByteArrays(keysValuePair)
+      require(file.getParentFile.exists() || file.getParentFile.mkdirs(), "Key's file parent directory creation failed")
       val writer = new PrintWriter(filePath)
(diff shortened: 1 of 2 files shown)


src/main/scala/io/iohk/ethereum/network/package.scala
```diff
@@ -31,8 +31,8 @@ package object network {
       val keysValuePair = generateKeyPair(secureRandom)
 
       //Write keys to file
-      val (_, priv) = keyPairToByteArrays(keysValuePair)
-      file.getParentFile.mkdirs()
+      val (priv, _) = keyPairToByteArrays(keysValuePair)
+      require(file.getParentFile.exists() || file.getParentFile.mkdirs(), "Key's file parent directory creation failed")
```
rtkaczyk: Minor: `.mkdirs()` already checks for existence, so `.exists()` is redundant.

ntallar: I agree, `.mkdirs()` will work even if the dir already exits, however it will return `false`. So I added the `.exists()` check so as to distinguish the case when `.mkdirs()` failed and when the dir already existed

rtkaczyk: Ahh, that's right 👍

PR #252, opened 2017-07-17
[EC-246] Optimize transactions execution when preparing block

adamsmo:


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -104,6 +97,25 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
...
+    signedTransactions: Seq[SignedTransaction], world: InMemoryWorldStateProxy,
+    blockHeader: BlockHeader, signedTransactionValidator: SignedTransactionValidator,
+    acumGas: BigInt = 0, acumReceipts: Seq[Receipt] = Nil, executed: Seq[SignedTransaction] = Nil)
+  : Either[TxsExecutionError, (BlockResult, Seq[SignedTransaction])] = {
+
+    val result = executeTransactions(signedTransactions, world, blockHeader, signedTransactionValidator, acumGas, acumReceipts)
+
+    result match {
+      case Left(TxsExecutionError(stx, Some(StateBeforeFailure(worldState, gas, receipts)), reason)) if signedTransactions.length > 1 =>
+        log.debug(s"failure while preparing block because of $reason in transaction with hash ${}")
```
ntallar: Minor detail: Should we complete this log with the tx hash?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -104,6 +97,25 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
     blockTxsExecResult
   }
 
+
+  private[ledger] def executePreparedTransactions(
```
ntallar: Minor detail: As this function is tail recursive, maybe we could use the `@tailrec` annotation?

src/test/scala/io/iohk/ethereum/mining/BlockGeneratorSpec.scala
```diff
@@ -111,6 +116,43 @@ class BlockGeneratorSpec extends FlatSpec with Matchers with PropertyChecks with
...
+    val nextTransaction: SignedTransaction = SignedTransaction.sign(transaction.copy(nonce = signedTransaction.tx.nonce + 1), keyPair, Some(0x3d.toByte))
+
+    val privateKeyWithNoEthere = BigInt(1, Hex.decode("584a31be275195585603ddd05a53d16fae9deafba67213b6060cec9f16e44cae"))
+
+
+//    val gParam = new ECKeyGenerationParameters(curve, new SecureRandom())
+//    val eGen = new ECKeyPairGenerator
+//    eGen.init(gParam)
+//    val ephemPair = eGen.generateKeyPair
+//    val v = Hex.toHexString(ephemPair.getPrivate.asInstanceOf[ECPrivateKeyParameters].getD.toByteArray)
```
ntallar: Minor detail: Should we remove this commented code or maybe it's needed for something?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -104,6 +97,25 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
...
+  : Either[TxsExecutionError, (BlockResult, Seq[SignedTransaction])] = {
+
+    val result = executeTransactions(signedTransactions, world, blockHeader, signedTransactionValidator, acumGas, acumReceipts)
+
+    result match {
+      case Left(TxsExecutionError(stx, Some(StateBeforeFailure(worldState, gas, receipts)), reason)) if signedTransactions.length > 1 =>
+        log.debug(s"failure while preparing block because of $reason in transaction with hash ${}")
+        val txIndex = signedTransactions.indexWhere(tx => tx.hash == stx.hash)
+        executePreparedTransactions(signedTransactions.drop(txIndex + 1),
+          worldState, blockHeader, signedTransactionValidator, gas, receipts, signedTransactions.take(txIndex))
```
ntallar: Should we include the `executed` txs in the `executed` parameter for the recursive call here (doing something like `executed++signedTransactions.take(txIndex)`)?
For example if we are executing: `stx1, stx2, stx3, stx4, stx5`, and both `stx2` and `stx4` fail, when `stx4` fails, could it be the case that we are not including `stx1` as executed?

adamsmo: 👍 good catch

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -389,7 +401,8 @@ sealed trait BlockExecutionError{
 
 object BlockExecutionError {
   case class ValidationBeforeExecError(reason: String) extends BlockExecutionError
-  case class TxsExecutionError(stx: SignedTransaction, reason: String) extends BlockExecutionError
+  case class StateBeforeFailure(worldState: InMemoryWorldStateProxy, acumGas: BigInt, acumReceipts: Seq[Receipt])
+  case class TxsExecutionError(stx: SignedTransaction, stateBeforeError: Option[StateBeforeFailure], reason: String) extends BlockExecutionError
```
ntallar: In which scenario would `stateBeforeError` be `None`? Maybe I missed something but it seems that apart from the use in a test it's used as non empty.

src/main/scala/io/iohk/ethereum/mining/BlockGenerator.scala
```diff
@@ -76,6 +64,32 @@ class BlockGenerator(blockchainStorages: BlockchainStorages, blockchainConfig: B
     result
   }
 
+  private def prepareTransactions(transactions: Seq[SignedTransaction], blockGasLimit: BigInt) = {
+    val sortedTransactions = transactions.groupBy(_.senderAddress).values.toList.flatMap { txsFromSender =>
+      val ordered = txsFromSender
+        .sortBy(-_.tx.gasPrice)
+        .sortBy(_.tx.nonce)
+        .scanLeft(None: Option[BigInt], None: Option[SignedTransaction]) { case ((prevNonce, _), stx) =>
```
AlanVerbner: Even this is not wrong at all, maybe it would be more clear to replace the `scanLeft` and `flatMap` by a `foldLeft` with a condition whether to add or not at tx on last accumulated item. Something like (I haven't run it so it might not worn :disappointed: , read it as a reference impl )

```scala
.foldLeft(Seq.empty[SignedTransaction]) { (acc, stx) => {
   if(acc.lastOption.exists(_.nonce != stx.tx.nonce)) acc :+ stx
   else acc
}
```

adamsmo: I would go with
```
.foldLeft(Seq.empty[SignedTransaction]) { case (txs, tx) =>
  if (txs.exists(_.tx.nonce == tx.tx.nonce)) {
    txs
  } else {
    tx +: txs
  }
}
.reverse
```
to not append to Seq since it defaults to List
do you think it will be easier to read?

AlanVerbner: yes, I think it will be easier to read but it's not mandatory your change. If you consider it's better, go ahead.

adamsmo:
@ntallar / @AlanVerbner updated

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -104,6 +97,25 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
...
   }
 
+  @tailrec
+  private[ledger] final def executePreparedTransactions(
+    signedTransactions: Seq[SignedTransaction], world: InMemoryWorldStateProxy,
+    blockHeader: BlockHeader, signedTransactionValidator: SignedTransactionValidator,
+    acumGas: BigInt = 0, acumReceipts: Seq[Receipt] = Nil, executed: Seq[SignedTransaction] = Nil)
+  : Either[TxsExecutionError, (BlockResult, Seq[SignedTransaction])] = {
+
+    val result = executeTransactions(signedTransactions, world, blockHeader, signedTransactionValidator, acumGas, acumReceipts)
```
AlanVerbner: Have you considered executing transactions one by one? Do you see a problem with that? The only benefit I can find is that `StateBeforeFailure` wouldn't be necessary so it might be interesting to consider it.

adamsmo: I did not want to repeat steps from executeTransactions()
also if I do it in recursive fashion I would need `StateBeforeFailure`

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -104,6 +97,25 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
...
+
+    result match {
+      case Left(TxsExecutionError(stx, StateBeforeFailure(worldState, gas, receipts), reason)) if signedTransactions.length > 1 =>
```
AlanVerbner: What about if `signedTransactions.length == 1` and fails? I think it won't be able to create an empty block, do we want that behavior ?

ntallar: Also, in relation to what @AlanVerbner said, in which scenario would we want `executePreparedTransactions` to return something of type `Left`?
If the execution of all the pased txs failed, maybe we could return something of type `(BlockResult, Seq[SignedTransaction])` with the sequence of txs empty?

PR #253, opened 2017-07-18
[EC-228] Prunning - Basic Implementation 

AlanVerbner:
# Description

The implementation is based on:
- Each time an mpt node is inserted or deleted, the references it has are increased ir decreased respectively
- When a node reaches 0 references, it's tagged as to be removed candidate within that block
- Everytime we update the bestBlockNumber we check X blocks back and remove all nodes with 0 references.
- Genesis generated nodes are not going to be cleaned
- Mining mpt data is not stored in DB

The following results were found when running the client with and without pruning to fully sync ETC chain. Around block 300k the ratio was 4:1 (_X-axis = block number, Y-xis = leveldb folder size in kb_)

archive

pruning
    
Things to improve:
- Pruning should be moved to other plase than where the best block number is updated. This was made in order to make it easier to synchronize concurrent reads and pruning but it's not the best idea
- Genesis is a fixed amount of data and we can live without pruning it but we should take care of mined blocks

### TODO
- [x] Tests
- [x] Improve interfaces and beautify code
- [x] Add some code documentation
- [x] Think about how to deal with mining and pruning


src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  import ReferenceCountNodeStorage._
+
+  override def get(key: ByteString): Option[NodeEncoded] = nodeStorage.get(key).map(_.toStoredNode).map(_.nodeEncoded.toArray)
+
+  override def update(toRemove: Seq[NodeHash], toUpsert: Seq[(NodeHash, NodeEncoded)]): NodesKeyValueStorage = {
+
+    require(blockNumber.isDefined)
+
+    val bn = blockNumber.get
```
LukasGasior1: what to you think about making it non-optional in the constructor? is there a reason not to do so?

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -0,0 +1,132 @@
...
+    val upsertChanges = toUpsert.foldLeft(Map.empty[NodeHash, StoredNode]) { (storedNodes, toUpsertItem) =>
+      val (nodeKey, nodeEncoded) = toUpsertItem
+      val storedNode: StoredNode = storedNodes.get(nodeKey)
```
LukasGasior1: why do we need to first check `storedNodes.get(nodeKey)`? can `toUpsert` contain duplicates?

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -0,0 +1,132 @@
...
+        .orElse(nodeStorage.get(nodeKey).map(_.toStoredNode))
+        .getOrElse(StoredNode(ByteString(nodeEncoded), 0))
+        .incrementReferences(1)
+
+      storedNodes + (nodeKey -> storedNode)
+    }
+
+    val toPruneInThisBlockKey = pruneKey(bn)
```
LukasGasior1: what is a pruneKey?

LukasGasior1: what do we store under it? a list of nodes marked as deleted for this block? sorry if that's obvious, but I'm having hard time understanding this whole method :smile:

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -0,0 +1,132 @@
...
+
+    val key = pruneKey(blockNumber.get)
+
+    nodeStorage.get(key)
+      .map(_.toPruneCandidates)
+      .map { pruneCandidates: PruneCandidates =>
+        // Get Node from storage and filter ones which have references = 0 now (maybe they were added again after blockNumber)
+        pruneCandidates.nodeKeys.map(nodeStorage.get).zip(pruneCandidates.nodeKeys)
+          .filter(n => n._1.isDefined && n._1.get.toStoredNode.references == 0)
+          .map(_._2)
```
LukasGasior1: we can make it a little bit more readable with named variables and `collect` instead of map+filter

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -0,0 +1,132 @@
...
+      }.map(nodesToDelete => {
```
LukasGasior1: can we make it an intermediate val? like
```
val nodesToDelete = ...

nodesToDelete.foreach { nodeStorage.update }

nodesToDelete.map(_.size).getOrElse(0)
```

src/main/scala/io/iohk/ethereum/db/storage/pruning/Basic.scala
```diff
@@ -0,0 +1,12 @@
+package io.iohk.ethereum.db.storage.pruning
+
+import io.iohk.ethereum.db.storage.{NodeStorage, ReferenceCountNodeStorage}
+import io.iohk.ethereum.utils.Logger
+
+case class Basic(history: Int) extends PruningMode with RangePrune with Logger {
```
LukasGasior1: just a random thought, do we need those pruning modes. I mean do we need logic in them? Can't it be handled just as a part of `ReferenceCountNodeStorage`?

src/main/scala/io/iohk/ethereum/db/storage/pruning/RangePrune.scala
```diff
@@ -0,0 +1,15 @@
+package io.iohk.ethereum.db.storage.pruning
+
+import io.iohk.ethereum.utils.Logger
+
+trait RangePrune extends Logger {
+  def pruneBetween(start: BigInt, end: BigInt, pruneFn: BigInt => Int): PruneResult = {
```
LukasGasior1: what's the difference between pruneFn and PruneFn from PruningMode? Can't we have a single pruneFn (with the same signature?)

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncNodesRequestHandler.scala
```diff
@@ -77,7 +77,7 @@ class FastSyncNodesRequestHandler(
       val evm = n.getAccount.codeHash
       val storage = n.getAccount.storageRoot
 
-      mptNodeStorage.put(n)
+      nodesKeyValueStorage.update(Nil, Seq(n.hash -> n.toBytes))
```
AlanVerbner: The storage trait is defined as dependency from the `MerklePatriciaTrie` and the clases to be stored here are from Network. 

I do agree it seems a little bit low level but the reason behind this is to remove the `MptNodeStorage` so I think the trade off was positive. If you want to, I can give you more background.

src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
```diff
@@ -2,27 +2,36 @@ package io.iohk.ethereum.db.storage
...
+  protected def apply(dataSource: DataSource): AppStateStorage = new AppStateStorage(dataSource, pruneFn)
 
   def getBestBlockNumber(): BigInt =
     BigInt(get(Keys.BestBlockNumber).getOrElse("0"))
 
-  def putBestBlockNumber(bestBlockNumber: BigInt): AppStateStorage =
+  def putBestBlockNumber(bestBlockNumber: BigInt): AppStateStorage = {
+    // FIXME We need to decouple pruning from best block number storing in this fn
+    pruneFn(getLastPrunedBlock(), bestBlockNumber) match {
+      case PruneResult(_, 0) =>
```
AlanVerbner: In Archive mode you will still be calling `prune` but it will always return 0 and I wanted to avoid to call a "store" in that case. When using Basic mode, it will eventually update the lastPrunedBlock (even if iterates over the same block tags several times)

Your comment is totally correct but as it probably will always need to delete something (because each new block has a different root hash because of miner reward), I think it won't be a problem and we won't cal that extra save (`putLastPrunedBlock`)

if you think this is confusing (i'm starting to think it is), I can change it.

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
AlanVerbner: I was expecting this coment :smile: 

- If you are using Basic pruning mode, it will always store alongside with the mpt node, the reference count. That means that we are "changing" what data is being stored by MPT and it's the reason why you cant switch between archive and basic without reseting the database
- Because of :arrow_up: , everytime you want to access to the mpt nodes, you need to use the corresponding `NodesKeyValueStorage` or it won't work.
- Sometimes, when we need to access the state just to read some value (for example: `io.iohk.ethereum.domain.BlockchainImpl#getAccountStorageAt`) we don't have (and don't need) to tag the changes under a certain block number. 
- Other alternatives were considered but I wanted to avoid adding block taging changes within MPT (because it doesn't care if you are using a pruning mechanism or whatever) and that's why it's on `ReferenceCountNodeStorage` constructor and not just in update (which would be more logical).

Sorry for the long explanation, let me know if you need some clarification. Any other ideas are welcome :smile:

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
AlanVerbner: Yes it can. For example, in the [following test](https://github.com/input-output-hk/etc-client/blob/master/src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala#L436) (that was created before this test) the same node is being stored twice.

Maybe it's an issue with MPT implementation but I don't think it's a bug. will discuss it with @ntallar

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
AlanVerbner: Under that key you store a sequence of blockHashes you detected they have 0 references. It's derived from blockNumber so basically you are tagging mpt nodes likely to delete based on a certain block tag.

ps: You having hard time understanding it is no good at all. Maybe I need to improve the code, the description or both. Please let me know if you think so, I'm really open to do those changes if necessary.

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
AlanVerbner: After adding tests, I will.

src/main/scala/io/iohk/ethereum/db/storage/pruning/Basic.scala
AlanVerbner: maybe, let me check because I think you are right.

src/main/scala/io/iohk/ethereum/db/storage/pruning/RangePrune.scala
AlanVerbner: - `pruneFn` is a function that given a block tag, prunes and returns how many nodes has deleted
- `io.iohk.ethereum.db.storage.pruning.PruningMode#PruneFn` is a function that based on the last pruned block number tag and the latest block number, prunes the data and returns new last pruned block and the amount of nodes removed.

Maybe, after the refactor suggested [here](https://github.com/input-output-hk/etc-client/pull/253/files#r128592298) I can improve that.

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
AlanVerbner: @LukasGasior1 This is an example @ntallar made ... basically, when splitting a key when storing it in a branch node resulting in two leafs that are the same

img_20170720_162238

src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
AlanVerbner: nvm, i think this will be a problem after `FastSync` so I'm going to change it as you suggested.

src/main/scala/io/iohk/ethereum/db/storage/pruning/Basic.scala
AlanVerbner: Done!

LukasGasior1:
@AlanVerbner I'm going through it one more time :) also testing locally. Will get back to your comments

AlanVerbner:
Thanks @LukasGasior1. Please ask me all the questions you want

 I know is a big and ugly one 😞

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -0,0 +1,178 @@
...
+          else toDeleteInBlock
+        (storedNodes + (nodeKey -> storedNode.get), pruneCanditatesUpdated)
+      }
+      else acc
+    }
+
+    val toUpsertUpdated = changes._1.map {
+      case (nodeKey: NodeHash, storedNode: StoredNode) =>
+        Seq(nodeKey -> storedNode.toBytes)
+    }.toSeq.flatten
```
ntallar: Minor detail: Why is the mapping of each element to a `Seq` of one element needed?
Maybe we could just map it to the element of the `Seq` and remove the `flatten`?

AlanVerbner: Let me finish with the tests and will check that :+1:

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
```diff
@@ -90,20 +94,22 @@ class GenesisDataLoader(
...
-      MerklePatriciaTrie[Array[Byte], Account](ephemNodeStorage, (input: Array[Byte]) => crypto.kec256(input))
-    val stateMpt = genesisData.alloc.foldLeft(initialStateMpt) { case (mpt, (address, AllocAccount(balance))) =>
+    val initalRootHash = MerklePatriciaTrie.calculateEmptyRootHash((input: Array[Byte]) => crypto.kec256(input))
+
+    val stateMpt = genesisData.alloc.zipWithIndex.foldLeft(initalRootHash) { case (rootHash, (((address, AllocAccount(balance)), idx))) =>
+      val ephemNodeStorage =  PruningMode.nodesKeyValueStorage(pruningMode, nodeStorage)(Some(idx - genesisData.alloc.size))
+      val mpt = MerklePatriciaTrie[Array[Byte], Account](rootHash, ephemNodeStorage, (input: Array[Byte]) => crypto.kec256(input))
       val paddedAddress = address.reverse.padTo(addressLength, "0").reverse.mkString
-      mpt.put(crypto.kec256(Hex.decode(paddedAddress)), Account(blockchainConfig.accountStartNonce, UInt256(BigInt(balance)), emptyTrieRootHash, emptyEvmHash))
+      mpt.put(crypto.kec256(Hex.decode(paddedAddress)), Account(0, UInt256(BigInt(balance)), emptyTrieRootHash, emptyEvmHash)).getRootHash
```
ntallar: Why does the nonce of the initial accounts be always 0? Maybe I'm missing something but it seems from [parity morden configuration](https://github.com/paritytech/parity/blob/master/ethcore/res/ethereum/morden.json#L62) that the starting nonce also applies to the initial accounts.

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -357,8 +357,8 @@ trait FastSync {
       val (nonMptNodesToGet, remainingNonMptNodes) = nonMptNodesQueue.splitAt(nodesPerRequest)
       val (mptNodesToGet, remainingMptNodes) = mptNodesQueue.splitAt(nodesPerRequest - nonMptNodesToGet.size)
       val nodesToGet = nonMptNodesToGet ++ mptNodesToGet
-      val handler = context.actorOf(
-        FastSyncNodesRequestHandler.props(peer, etcPeerManager, peerEventBus, nodesToGet, blockchain, blockchainStorages.mptNodeStorage))
+      val handler = context.actorOf(FastSyncNodesRequestHandler.props(peer, etcPeerManager, peerEventBus, nodesToGet, blockchain,
+        blockchainStorages.nodesKeyValueStorageFor(Some(targetBlockOffset))))
```
ntallar: If I understood correctly the parameter of `nodesKeyValueStorageFor` should be the block number to which the nodes stored belong, should then it be the target block instead? (which I think is `Some(initialSyncState.targetBlock)`)

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
AlanVerbner: Done, please check if it's more clear now

src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
mcsherrylabs: Q: (which might not make sense) `putBestBlockNumber` is called in FastSync, FastSyncReceiptsHandler, and twice in RegularSync. Does it make sense to call 'prune' only in 'RegularSync'? If so, does that negate the need to pass `pruneFn` into `AppStateStorage`?

src/main/scala/io/iohk/ethereum/db/storage/MptNodeStorage.scala
```diff
@@ -1,36 +0,0 @@
-package io.iohk.ethereum.db.storage
```
mcsherrylabs: :+1:

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
AlanVerbner: I think you are right! I think it was introduced when merging :(

AlanVerbner: yes! I introduced the error when merging :disappointed:

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -156,7 +159,8 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
     val gasLimit = stx.tx.gasLimit
     val config = EvmConfig.forBlock(blockHeader.number, blockchainConfig)
 
-    val world1 = InMemoryWorldStateProxy(storages, blockchainConfig.accountStartNonce, Some(stateRoot))
+    //FIXME When mining we are setting block tag as -1 but this info won't be pruned
+    val world1 = BlockchainImpl(storages).getWorldStateProxy(-1, blockchainConfig.accountStartNonce, Some(stateRoot))
```
ntallar: As we are only simulating a tx here, maybe it would be better to use the read only `WorldStateProxy` from the blockchain?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -83,17 +81,22 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
...
   Either[BlockExecutionError, BlockResult] = {
     val parentStateRoot = blockchain.getBlockHeaderByHash(block.header.parentHash).map(_.stateRoot)
-    val initialWorld = InMemoryWorldStateProxy(storages, blockchainConfig.accountStartNonce, parentStateRoot)
+    val getWorldStateProxyFn =
+        if(readOnly) blockchain.getReadOnlyWorldStateProxy _
+        else blockchain.getWorldStateProxy _
+    val initialWorld: blockchain.WS = getWorldStateProxyFn(block.header.number, blockchainConfig.accountStartNonce, parentStateRoot)
+
+      //if(readOnly) blockchain.getReadOnlyWorldStateProxy(block.header.number, blockchainConfig.accountStartNonce, parentStateRoot)
+      //else blockchain.getWorldStateProxy(block.header.number, blockchainConfig.accountStartNonce, parentStateRoot)
```
ntallar: Minor detail: Is this commented code needed or maybe we could remove it?

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
```diff
@@ -122,14 +128,12 @@ class GenesisDataLoader(
...
         Success(())
-
       case Some(_) =>
         Failure(new RuntimeException("Genesis data present in the database does not match genesis block from file." +
           " Use different directory for running private blockchains."))
-
       case None =>
         // using empty namespace because ephemDataSource.storage already has the namespace-prefixed keys
-        dataSource.update(IndexedSeq(), Nil, ephemDataSource.storage.toSeq)
+        ephemDataSource.storage.toSeq.grouped(dbConfig.batchSize).foreach(toStore => dataSource.update(IndexedSeq(), Nil, toStore))
```
ntallar: As the use of `dbConfig.batchSize` was introduced due to `LevelDB` problems, maybe we should start using this batch size in other updates to the database?

src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
AlanVerbner: I think the ones we can avoid is `FastSync`and `FastSyncReceiptsHandler` as when doing fast sync it makes no sense to do pruning.

Are suggesting to place the `prune` call outside `putBestBlockNumber` ?

src/test/scala/io/iohk/ethereum/blockchain/sync/BlockchainHostActorSpec.scala
```diff
@@ -223,11 +226,9 @@ class BlockchainHostActorSpec extends FlatSpec with Matchers {
     etcPeerManager.expectMsg(EtcPeerManagerActor.SendMessage(NodeData(Seq(extensionNode.toBytes)), peerId))
   }
 
-  trait TestSetup {
+  trait TestSetup extends EphemBlockchainTestSetup {
```
ntallar: :+1:

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
AlanVerbner: good catch!

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
AlanVerbner: @ntallar I do agree with you but I wanted to avoid hiding it underneath `io.iohk.ethereum.db.dataSource.LevelDBDataSource#update`. It would be misleading because instead of doing an atomic update we would be creating several batches to avoid this issue. But if it's better, I can change it tho. @mcsherrylabs @LukasGasior1 what do you think?

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -0,0 +1,178 @@
...
+      val storedNode: StoredNode = storedNodes.get(nodeKey)
+        .orElse(nodeStorage.get(nodeKey).map(_.toStoredNode))
+        .getOrElse(StoredNode(ByteString(nodeEncoded), 0))
+        .incrementReferences(1)
+
+      storedNodes + (nodeKey -> storedNode)
+    }
+
+    val toPruneInThisBlockKey = pruneKey(bn)
+    val pruneCandidates = nodeStorage.get(toPruneInThisBlockKey).map(_.toPruneCandidates).getOrElse(PruneCandidates(Nil))
```
ntallar: Minor detail: I'm not sure about storing the `pruneCandidates` in the `NodeStorage`, maybe we could create separate storage or rename `NodeStorage`?

AlanVerbner: I avoided that in terms of simplicity (as this is only used by ReferenceCountNodeStorage) but maybe, if we want to be consistent, I should have done that. @mcsherrylabs @LukasGasior1 WDYT ?

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
LukasGasior1: got it

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
LukasGasior1: I think I'd just leave it as is

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -73,17 +72,19 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
...
     * @param signedTransactionValidator
     */
   private[ledger] def executeBlockTransactions(
     block: Block,
-    blockchain: Blockchain,
-    storages: BlockchainStorages,
-    signedTransactionValidator: SignedTransactionValidator):
+    blockchain: BlockchainImpl,
+    signedTransactionValidator: SignedTransactionValidator,
+    readOnly: Boolean = false):
```
ntallar: Minor detail: As the only usage of this function with `readOnly` as `true` was removed, maybe we should remove this parameter as well?

mcsherrylabs: I can only see `executeBlockTransactions` called from one place.... so readonly is *always* false ?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -164,7 +165,8 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
     val gasLimit = stx.tx.gasLimit
     val config = EvmConfig.forBlock(blockHeader.number, blockchainConfig)
 
-    val world1 = InMemoryWorldStateProxy(storages, blockchainConfig.accountStartNonce, Some(stateRoot))
+    //FIXME When mining we are setting block tag as -1 but this info won't be pruned
+    val world1 = BlockchainImpl(storages).getReadOnlyWorldStateProxy(-1, blockchainConfig.accountStartNonce, Some(stateRoot))
```
mcsherrylabs: Same comment as -1 above. Should this be an option None?

src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -116,13 +117,19 @@ object Config {
     val blockResolveDepth: Int = syncConfig.getInt("block-resolving-depth")
   }
 
-  object Db {
+  trait DbConfig {
```
mcsherrylabs: :+1:

src/test/scala/io/iohk/ethereum/db/storage/ReadOnlyNodeStorageSpec.scala
```diff
@@ -0,0 +1,36 @@
...
+    val key1 = ByteString("key1")
+    val val1 = ByteString("Value1").toArray
+    referenceCountNodeStorage.put(key1, val1)
+
+    dataSource.storage.size shouldEqual 1
+    readOnlyNodeStorage.get(key1).get shouldEqual val1
+
+    readOnlyNodeStorage.remove(key1)
+
+    dataSource.storage.size shouldEqual 1
```
mcsherrylabs: Optional: I would have checked the value here as well, just because it's more complete.

src/universal/conf/storage.conf
```diff
@@ -15,4 +15,14 @@ grothendieck {
       # path = ${grothendieck.datadir}"/leveldb/"
     }
   }
+
+  pruning {
+    # Pruning mode tha app should start with:
```
mcsherrylabs: "Pruning mode that the application will use"? Or just fix the spelling of 'that'

src/universal/conf/storage.conf
```diff
@@ -15,4 +15,14 @@ grothendieck {
...
+    # - archive: No pruning is performed
+    # - basic: reference count based pruning
+    # mode = "basic"
+
+    # The amount of blocks history is kept before pruning
```
mcsherrylabs: "The amount of block history kept before pruning"

src/universal/conf/storage.conf
```diff
@@ -15,4 +15,14 @@ grothendieck {
       # path = ${grothendieck.datadir}"/leveldb/"
     }
   }
+
```
mcsherrylabs: Good to see this wasn't forgotten, Radek will be pleased ;)

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
AlanVerbner: Let me check because it shouldn't :thinking:

AlanVerbner: Yes, it seems after last merge new method was added so the call is splited in two, one does need it and once not

src/main/scala/io/iohk/ethereum/db/storage/pruning/package.scala
```diff
@@ -0,0 +1,43 @@
...
+    /**
+      * Create a NodesKeyValueStorage to be used within MerklePatriciaTrie
+      *
+      * @param blockNumber block number to be used as tag when doing update / removal operations. None can be sent if read only
+      * @return Storage to be used
+      */
+    def nodesKeyValueStorage(pruningMode: PruningMode, nodeStorage: NodeStorage)(blockNumber: Option[BigInt]): PruningNodesKeyValueStorage =
+      pruningMode match {
+        case ArchivePruning => new ArchiveNodeStorage(nodeStorage)
+        case BasicPruning(history) => new ReferenceCountNodeStorage(nodeStorage, history, blockNumber) //FIXME
```
AlanVerbner: It's already fixed...forgot to remove the ``//FIXME``. That's why it doesn't have any comment :disappointed:

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -57,7 +56,7 @@ class LedgerImpl(vm: VM, blockchainConfig: BlockchainConfig) extends Ledger with
 
     val blockchain = BlockchainImpl(storages)
     val parentStateRoot = blockchain.getBlockHeaderByHash(block.header.parentHash).map(_.stateRoot)
-    val initialWorld = InMemoryWorldStateProxy(storages, blockchainConfig.accountStartNonce, parentStateRoot)
+    val initialWorld = blockchain.getReadOnlyWorldStateProxy(-1, blockchainConfig.accountStartNonce, parentStateRoot)
```
AlanVerbner: Yes! after adding `ReadOnlyNodeStorage` I'm able to change this to optional. Good catch!

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -0,0 +1,178 @@
...
+
+  private implicit class StoredToBytes(val storedNode: StoredNode) extends AnyVal {
+    def toBytes: Array[Byte] = storedNodeEncDec.encode(storedNode)
+  }
+
+  private implicit class PruneCandidatesToBytes(val pruneCanidates: PruneCandidates) extends AnyVal {
+    def toBytes: Array[Byte] = pruneCandidatesEncDec.encode(pruneCanidates)
+  }
+
+  private implicit class BytesToStoredNode(val array: Array[Byte]) extends AnyVal {
```
AlanVerbner: Done!

src/main/scala/io/iohk/ethereum/db/storage/pruning/package.scala
```diff
@@ -33,11 +33,13 @@ package object pruning {
...
-        case BasicPruning(history) => new ReferenceCountNodeStorage(nodeStorage, history, blockNumber) //FIXME
+        case BasicPruning(history) => new ReferenceCountNodeStorage(nodeStorage, history, blockNumber)
       }
 
   }
 
-  case class PruneResult(lastPrunedBlockNumber: BigInt, pruned: Int)
+  case class PruneResult(lastPrunedBlockNumber: BigInt, pruned: Int) {
+    override def toString: String = s"Number of mpt nodes deleted: $pruned. Last Pruned Block: $lastPrunedBlockNumber"
+  }
```
mcsherrylabs: :+1:

PR #254, opened 2017-07-18
[EC-234][EC-235] Basic setup for ETH support

rtkaczyk:


rtkaczyk:
I'm getting:

```
16:06:02.349 i.i.e.blockchain.sync.SyncController - Peers: 1 (0 blacklisted).
16:06:02.899 i.i.e.blockchain.sync.SyncController - no peers to download from
16:06:03.919 i.i.e.blockchain.sync.SyncController - no peers to download from
16:06:04.351 i.i.e.blockchain.sync.SyncController - Peers: 1 (0 blacklisted).
16:06:04.939 i.i.e.blockchain.sync.SyncController - no peers to download from
16:06:05.961 i.i.e.blockchain.sync.SyncController - no peers to download from
```

when trying to sync (regular) from a local geth node. This may require more work.

rtkaczyk:
Syncing appears to be working correctly (just the begining of the chain) once the local geth node got past the DAO fork block.

src/main/scala/io/iohk/ethereum/network/ForkResolver.scala
```diff
@@ -17,17 +17,17 @@ object ForkResolver {
 
   class EtcForkResolver(blockchainConfig: BlockchainConfig) extends ForkResolver {
```
ntallar: Should we also rename this resolver to `EthereumForkResolver` maybe?
There's also various classes with an `Etc-` prefix (as the `EtcHandshaker`), should we rename them as well? We could create a separate task to do all this renaming.

rtkaczyk: As long as it's clear that in those names `Ethereum` means general Ethereum - all chains, not just Ethereum HF, then I think that's the way to go. As usual I like to call @mcsherrylabs on naming related issues.

Separate task would be useful so as not to inhibit other dependent ETH tasks.

adamsmo:
needs merge

adamsmo:
needs merge

PR #255, opened 2017-07-19
Add nodes persistence

LukasGasior1:


src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
```diff
@@ -82,6 +82,7 @@ object DumpChainApp extends App with NodeKeyBuilder with SecureRandomBuilder wit
       nodeStatusHolder = nodeStatusHolder,
       peerConfiguration = peerConfig,
       peerMessageBus = peerMessageBus,
+      knownNodesManager = actorSystem.deadLetters, // TODO: fixme
```
LukasGasior1: There's already a task to fix DumpChainApp (if we're still going to need it)

src/main/resources/application.conf
```diff
@@ -101,7 +109,7 @@ grothendieck {
       max-mpt-components-per-message = 400
 
       # Maximum connected peers
-      max-peers = 10
+      max-peers = 20
```
LukasGasior1: usually most of them are the ones that we keep trying to connect, not really connected peers. It seems to make sense to increase it to default 20 (maybe even 30)?

adamsmo:
scalastyle

src/main/scala/io/iohk/ethereum/network/KnownNodesManager.scala
```diff
@@ -0,0 +1,88 @@
...
+  private def scheduler = externalSchedulerOpt getOrElse context.system.scheduler
+
+  var knownNodes: Set[URI] = knownNodesStorage.getKnownNodes()
+
+  var toAdd: Set[URI] = Set.empty
+
+  var toRemove: Set[URI] = Set.empty
+
+  scheduler.schedule(config.persistInterval, config.persistInterval) {
+    persistChanges()
```
adamsmo: is it safe?
should we send message to ourself here and process message in main actor thread?

LukasGasior1: It's not, good catch! I'll fix it

src/main/scala/io/iohk/ethereum/db/storage/KnownNodesStorage.scala
```diff
@@ -0,0 +1,32 @@
...
+  def valueDeserializer: IndexedSeq[Byte] => Set[String] = (valueBytes: IndexedSeq[Byte]) => new String(valueBytes.toArray).split(' ').toSet
+
+  protected def apply(dataSource: DataSource): KnownNodesStorage = new KnownNodesStorage(dataSource)
+
+  def getKnownNodes(): Set[URI] = synchronized {
+    get(key).getOrElse(Set.empty).filter(_.nonEmpty).map(new URI(_))
+  }
+
+  def updateKnownNodes(toAdd: Set[URI] = Set.empty, toRemove: Set[URI] = Set.empty): KnownNodesStorage = {
+    val updated = getKnownNodes() -- toRemove ++ toAdd
```
adamsmo: what if node is in `toRemove` and `toAdd` should it be added or removed?

LukasGasior1: It shouldn't happen at all. See how we handle add/remove requests: if node is requested to be added it's removed from `toRemove` set and vice versa, so if the node is the same, the most recent add/remove message will work.

src/main/scala/io/iohk/ethereum/db/storage/KnownNodesStorage.scala
```diff
@@ -0,0 +1,32 @@
...
+  val key = "KnownNodes"
+
+  val namespace: IndexedSeq[Byte] = Namespaces.KnownNodesNamespace
+  def keySerializer: String => IndexedSeq[Byte] = _.getBytes
+  def valueSerializer: Set[String] => IndexedSeq[Byte] = _.mkString(" ").getBytes
+  def valueDeserializer: IndexedSeq[Byte] => Set[String] = (valueBytes: IndexedSeq[Byte]) => new String(valueBytes.toArray).split(' ').toSet
+
+  protected def apply(dataSource: DataSource): KnownNodesStorage = new KnownNodesStorage(dataSource)
+
+  def getKnownNodes(): Set[URI] = synchronized {
```
LukasGasior1: as discussed on VSU, forgot to remove it after testing something locally

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -179,9 +193,16 @@ class PeerActor[R <: HandshakeResult](
     case RLPxConnectionHandler.MessageReceived(_: Ping) => rlpxConnection.sendMessage(Pong())
   }
 
-  def handleDisconnectMsg: Receive = {
+  def handleDisconnectMsg(rlpxConnection: RLPxConnection): Receive = {
     case RLPxConnectionHandler.MessageReceived(d: Disconnect) =>
+      import Disconnect.Reasons._
+      d.reason match {
+        case IncompatibleP2pProtocolVersion | UselessPeer | Other =>
```
LukasGasior1: I'll add those too

src/main/scala/io/iohk/ethereum/db/storage/KnownNodesStorage.scala
adamsmo: 👍 
but that is in separate class, so technically it may happen?

LukasGasior1: technically yes, do you think it's worth preventing that with and assert or something?

adamsmo: I do not like assertions since they may produce runtime error, but we can decide what to do if node is in `toRemove` and `toAdd` at the same time

LukasGasior1: what's your proposition?

LukasGasior1:
You're right @mcsherrylabs. With peer discovery in place we no longer need tryDiscardPeersToFreeUpLimit

LukasGasior1:
@mcsherrylabs Actually on second thoughts I'm not sure if we should remove it. Without this logic other peers may have problems connecting to us. Say we have a limit of 10 peers. 5 handshaked and 5 slots continuously taken by nodes we keep trying to connect to. Without `tryDiscardPeersToFreeUpLimit` we will just reject all incoming connections (as peer limit is reached). With `tryDiscardPeersToFreeUpLimit` however, we'll try to disconnect one of the nodes that we keep trying to connect to.
Another solution could be to allow to have more than `maxPeers` connections for a short period and then, from time to time remove some peers so that they fit the limit.

src/main/scala/io/iohk/ethereum/network/KnownNodesManager.scala
```diff
@@ -46,6 +44,9 @@ class KnownNodesManager(
 
     case GetKnownNodes =>
       sender() ! KnownNodes(knownNodes)
+
+    case PersistChanges =>
+      persistChanges()
```
mcsherrylabs: :+1:

LukasGasior1:
but then retrying connections can succeed and we end up with more connected peers than the limit

src/main/scala/io/iohk/ethereum/db/storage/KnownNodesStorage.scala
adamsmo: to do `(getKnownNodes()  ++ toAdd) -- toRemove`
if I am deleting node it is probably because it failed somehow so there is no reason to keep it

adamsmo:
other than my last comment LGTM

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -75,27 +73,10 @@ class PeerManagerActor(
...
-        context become(tryingToConnect, false)
-        tryDiscardPeersToFreeUpLimit()
-          .map(_ => (msg, Success(())))
-          .recover { case ex => (msg, Failure(ex)) }
-          .pipeTo(self)
-      } else {
-        log.info("Another connection with {} is already opened. Disconnecting.", msg.remoteAddress)
-        msg.connection ! PoisonPill
-      }
+      handleConnection(msg.connection, msg.remoteAddress)
```
mcsherrylabs: More idomatic to pattern match `connection` and `remoteAddress` in `HandlePeerConnection`?

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -44,10 +45,9 @@ class PeerActor[R <: HandshakeResult](
 
   def scheduler: Scheduler = externalSchedulerOpt getOrElse system.scheduler
 
-  val P2pVersion = 4
```
mcsherrylabs: Glad this is gone.

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -98,12 +100,12 @@ class PeerManagerActor(
 
   def handleConnection(connection: ActorRef, remoteAddress: InetSocketAddress): Unit = {
     if (!isConnectionHandled(remoteAddress)) {
-      val peer = createPeer(remoteAddress)
+      val peer = createPeer(remoteAddress, incomingConnection = true)
       peer.ref ! PeerActor.HandleConnection(connection, remoteAddress)
 
-      if (peers.size >= peerConfiguration.maxPeers) {
+      if (incomingPeers.size >= peerConfiguration.maxPeers) {
```
mcsherrylabs: `peerConfiguration.maxIncomingPeers`?

adamsmo:
change LGTM

PR #256, opened 2017-07-20
Fix bugs related to code deposit gas check within CREATE opcode

rtkaczyk:
There were 2 bugs:
- wrong value of gas was used as a base for code deposit gas check (value available in VM @ depth0, instead of the capped value in VM @ depth1 - init code)
- OOG was thrown, instead of consuming all gas passed to the VM @ depth1 (it's that VM that throws OOG, and in the parent VM it simply manifests as all gas used)

Additionally, JVM stack size was increased for the distribution.


src/universal/conf/application.ini
```diff
@@ -1 +1 @@
--Dconfig.file=./conf/grothendieck.conf -Dlogback.configurationFile=./conf/logback.xml
+-Dconfig.file=./conf/grothendieck.conf -Dlogback.configurationFile=./conf/logback.xml -Xss10M
```
adamsmo: 👍

PR #257, opened 2017-07-20
[EC-233] Implemented EIP 170

ntallar:
## Description

Includes implementing [EIP-170](https://github.com/ethereum/EIPs/issues/170). This disables the use of contracts larger than a fixed limit (24576 in the Eth mainnet), causing their creation to fail.


src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -191,6 +191,20 @@ object MiningConfig {
   }
 }
 
+trait VMConfig {
+  val maxCodeSize: Option[BigInt]
+}
+
+object VMConfig {
```
rtkaczyk: I'm not sure about a completely new trait and config section. To me this belongs under `blockchain` (with all the fork block numbers) and should end up in `EvmConfig`.

adamsmo:
I see it is not part of EIP but should we add block height configuration to enable this limit at specific I see condition `block.number >= FORK_BLKNUM`?

adamsmo:
sorry accidentally closed

ntallar:
> I see it is not part of EIP but should we add block height configuration to enable this limit at specific I see condition block.number >= FORK_BLKNUM?

ntallar:
> I see it is not part of EIP but should we add block height configuration to enable this limit at specific I see condition block.number >= FORK_BLKNUM?

It could be easily added, however AFAIK the block height isn't configurable on Parity or Geth (as for example in https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L278), maybe you've seen otherwise?

adamsmo:
but if it has no configuration how it can be turned on?
is it enabled in geth and parity?

ntallar:
> but if it has no configuration how it can be turned on?
is it enabled in geth and parity?

In Geth and Parity it is always turned on, it seems that the reason why that's the case is that it was activated before the block gas limit was big enough to allow having txs that created contract with code this large (due to the gas cost per contract code byte).

adamsmo:
needs merge

PR #259, opened 2017-07-25
[EC-259] Remove misleading message shown when keystore directory already exists

AlanVerbner:
## Description

The message to be removed is `Could not initialise keystore directory (/home/alverbner/.grothendieck/keystore): java.util.NoSuchElementException: Predicate does not hold for false`

Also, before this commit, when an error occurred when creating the directory the client didn't start and continued working until it was needed again, for example in `io.iohk.ethereum.keystore.KeyStoreImpl#listAccounts`.

Now, if the dir needs to be created and it fails, the client execution is finalized.


src/main/scala/io/iohk/ethereum/keystore/KeyStore.scala
```diff
@@ -66,8 +66,8 @@ class KeyStoreImpl(keyStoreDir: String, secureRandom: SecureRandom) extends KeyS
 
   private def init(): Unit = {
     val dir = new File(keyStoreDir)
-    val res = Try(dir.mkdirs()).filter(identity)
-    res.failed.foreach(ex => log.error(s"Could not initialise keystore directory ($dir): $ex"))
+    val res = Try(dir.exists() || dir.mkdirs()).filter(identity)
```
mcsherrylabs: I forgot to mention you might need to check it's a folder and not a file (dir.isDirectory) This should encompass the `exists`.  Sorry for wrong steer. Also `isDirectory` has no side effects so no `()`.

AlanVerbner: Done + more tests!

PR #261, opened 2017-07-27
[EC-256] Regression test for fast sync

adamsmo:
this PR contains configuration tweaks and additional scheduler to assign work if child actor died before reporting, this is also used when new peer are discovered to schedule requests for them


src/main/resources/application.conf
```diff
@@ -257,7 +257,7 @@ grothendieck {
     sync-retry-interval = 5.seconds
 
     # Response time-out from peer during sync. If a peer fails to respond within this limit, it will be blacklisted
-    peer-response-timeout = 10.seconds
+    peer-response-timeout = 3.minutes
```
LukasGasior1: 3 minutes is a lot. I doubt someone will respond in more than 1 minute. Maybe 1 then?

src/main/resources/application.conf
```diff
@@ -96,7 +96,7 @@ grothendieck {
 
       # Maximum number of reconnect attempts after the connection has been initiated.
       # After that, the connection will be dropped until its initiated again (eg. by peer discovery)
-      connect-max-retries = 1
+      connect-max-retries = 5
```
LukasGasior1: maybe decrease it and increase connect-retry-delay ?  trying 5 times in 5 seconds intervals doesn't seem like good approach

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -222,11 +224,20 @@ trait FastSync {
...
+    }
+
+    private def printStatus() = {
+      val totalNodesCount = downloadedNodesCount + mptNodesQueue.size + nonMptNodesQueue.size
+      val formatPeer: (Peer) => String = peer => s"${peer.remoteAddress.getAddress.getHostAddress}:${peer.remoteAddress.getPort}"
+      log.info(
+        s"""|Block: ${appStateStorage.getBestBlockNumber()}/${initialSyncState.targetBlock.number}.
+            |Peers waiting_for_response/connected: ${assignedHandlers.size}/${handshakedPeers.size} (${blacklistedPeers.size} blacklisted).
+            |State: $downloadedNodesCount/$totalNodesCount nodes.
+            |Details: connected(${assignedHandlers.values.map(formatPeer).toSeq.sorted})/
```
LukasGasior1: maybe let's put this info in separate log entry (possibly debug)?

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
```diff
@@ -138,7 +138,7 @@ class RLPxConnectionHandler(
         context.parent ! MessageReceived(message)
 
       case Failure(ex) =>
-        log.error(ex, s"Cannot decode message from $peerId")
```
LukasGasior1: doesn't it also show the error?

LukasGasior1:
@adamsmo I think you also need to update the defaults (in comments) in conf files inside `universal/conf`

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -222,11 +224,20 @@ trait FastSync {
...
+            |handshaked(${handshakedPeers.keys.map(formatPeer).toSeq.sorted})
+            | blacklisted(${blacklistedPeers.map { case (id, _) => id }})
```
ntallar: Very minor detail: Maybe it would be better to print out `id.value` and use `mkString(", ")` and do something like `blacklistedPeers.map { case (id, _) => id.value }.mkString(", ")`, so as to not print the use of `List` and `PeerId`?
We could also do something similar with `connected` and `handshaked`.

src/main/resources/application.conf
AlanVerbner: I needed to set it to 2 and it worked all the time, but cannot remember if tried with 1 first :thinking:

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
adamsmo: yes but with stack trace which makes logs not readable

src/main/resources/application.conf
adamsmo: @AlanVerbner did you tested 2 min on EC2?

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
mcsherrylabs: Can we do an `if debugEnabled` so as not to lose the trace when the LEVEL is `debug`?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
mcsherrylabs: Isn't this the most important status printout? The one that lets the user know what's going on? If this was `debug` where would the user the notified of the status?

adamsmo:
@LukasGasior1 updated universal conf

PR #262, opened 2017-07-27
[EC-255] Ethereum test suite - VM

rtkaczyk:
Attempts to run & pass `VMTests` from: https://github.com/ethereum/tests
The JSON test files are linked via a submodule. To download them run:

```
$ git submodule init
$ git submodule update
```

Run from `sbt`:

```
> ets:test
```

To exclude groups (files):

```
> ets:testOnly *VMSuite -- -DexcludeGroups=vmPerformanceTest,*Limits*
```

To include single scenarios:

```
> ets:testOnly *VMSuite -- -DincludeGroups=vmBlockInfoTest -DincludeScenarios=blockhashInRange,blockhashNotExistingBlock
```

By default groups and scenarios are matched by comma-separated simplified globbing patterns (only * is supported). This can be changed by adding `-Dmatch=plain` or `-Dmatch=regex` flag (see `TestOptions` for details).

The tests, with exception for `vmPerformanceTest`, are integrated into our Circle CI build. If for any reason we need to disable them, it can be done by going to https://circleci.com/gh/input-output-hk/etc-client/edit#env-vars and setting an environment variable `SKIP_ETS_TESTS` with any value.


adamsmo:
need merge from `phase/daedalus`

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -177,7 +177,7 @@ object OpCode {
   * @param delta number of words to be popped from stack
   * @param alpha number of words to be pushed to stack
   */
-sealed abstract class OpCode(val code: Byte, val delta: Int, val alpha: Int, val constGasFn: FeeSchedule => BigInt) {
+abstract class OpCode(val code: Byte, val delta: Int, val alpha: Int, val constGasFn: FeeSchedule => BigInt) extends Product with Serializable {
```
ntallar: Why is `extends Product with Serializable` needed?

rtkaczyk: `VMSuite`, line 82:

```
if (!Set(TestCREATE, TestCALL, TestCALLCODE).contains(itx.opcode))
```

fails to compile without this. To me it looks like a bug in the type inferencer. If I change this line to:

```
if (!Set[OpCode](TestCREATE, TestCALL, TestCALLCODE).contains(itx.opcode))
```

then it compiles.

But adding `extends Product with Serializable` also works, and in practice it also means that we're requiring implementations of `OpCode` to be case objects/classes - which makes sense to do.

src/main/scala/io/iohk/ethereum/vm/InternalTransaction.scala
```diff
@@ -0,0 +1,14 @@
+package io.iohk.ethereum.vm
+
+import akka.util.ByteString
+import io.iohk.ethereum.domain.Address
+
+//TODO: scaladoc
```
rtkaczyk: scaladoc added

src/ets/scala/io/iohk/ethereum/ets/vm/VMSuite.scala
```diff
@@ -0,0 +1,93 @@
...
+
+class VMSuite extends FreeSpec with Matchers with Logger {
+
+  override def run(testName: Option[String], args: Args): Status = {
+
+    val options = TestOptions(args.configMap)
+    val scenarios = ScenarioLoader.load(options)
+
+    scenarios.foreach { group =>
+      group.name - {
```
LukasGasior1: what's this syntax?

src/ets/scala/io/iohk/ethereum/ets/vm/testopcodes.scala
```diff
@@ -0,0 +1,79 @@
+package io.iohk.ethereum.ets.vm
+
+import akka.util.ByteString
+import io.iohk.ethereum.domain.Address
+import io.iohk.ethereum.vm._
+
+case object TestCREATE extends CreateOp {
```
LukasGasior1: how is it different from the real `CREATE`?

src/ets/scala/io/iohk/ethereum/ets/vm/testopcodes.scala
```diff
@@ -0,0 +1,79 @@
...
+      state
+        .withStack(stack2)
+        .withMemory(memory1)
+        .withInternalTxs(internalTx :: Nil)
+        .step()
+    }
+  }
+}
+
+abstract class TestCallOp(code: Int) extends CallOp(code, 7, 1) {
```
LukasGasior1: same as above?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
LukasGasior1: I'd personally go with changing the line that failed to compile rather than adding this requirement here, but it's not a blocker

src/ets/scala/io/iohk/ethereum/ets/vm/VMSuite.scala
rtkaczyk: [FreeSpec](http://doc.scalatest.org/3.0.0/index.html#org.scalatest.FreeSpec)

src/ets/scala/io/iohk/ethereum/ets/vm/testopcodes.scala
rtkaczyk: `TestCREATE` never creates new accounts and does not execute any initialisation code. It does however consume some amount of gas and can result in errors for certain argument values. It always replaces the real `CREATE` (probably because cpp-ethereum VM cannot operate in in-memory mode) and we only care about tracking them via `InternalTransaction` / `CallCreate`.

src/ets/scala/io/iohk/ethereum/ets/vm/testopcodes.scala
rtkaczyk: As with `TestCreate` this never executes any code of the called account.

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
rtkaczyk: I need an non-personal reason for changing this 😁

PR #263, opened 2017-07-27
Logging update

LukasGasior1:


src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -79,17 +79,17 @@ class PeerActor[R <: HandshakeResult](
         processHandshakerNextMessage(initHandshaker, rlpxConnection.copy(uriOpt = Some(uri)), numRetries)
 
       case RLPxConnectionHandler.ConnectionFailed =>
-        log.warning("Failed to establish RLPx connection")
+        log.debug("Failed to establish RLPx connection")
```
AlanVerbner: as this class will only be logging in debug, isn't it better to set level OFF in  `universal/conf/logback.xml` which, if i'm not mistaken, is the config dist will use?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -44,11 +44,12 @@ trait FastSync {
         peerEventBus ! Subscribe(MessageClassifier(Set(BlockHeaders.code), PeerSelector.WithId(peer.id)))
         etcPeerManager ! EtcPeerManagerActor.SendMessage(GetBlockHeaders(Right(status.bestHash), 1, 0, reverse = false), peer.id)
       }
-      log.info("Asking {} peers for block headers", peersUsedToChooseTarget.size)
+      log.debug("Asking {} peers for block headers", peersUsedToChooseTarget.size)
       val timeout = scheduler.scheduleOnce(peerResponseTimeout, self, BlockHeadersTimeout)
       context become waitingForBlockHeaders(peersUsedToChooseTarget.keySet, Map.empty, timeout)
     } else {
-      log.warning("Cannot start fast sync, not enough peers to download from. Scheduling retry in {}", startRetryInterval)
+      log.info("Block synchronization (fast mode) not started. Need at least {} peers, but there are only {} available at the moment. Retrying in {}",
```
AlanVerbner: :+1:

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
```diff
@@ -63,7 +63,7 @@ class RLPxConnectionHandler(
       context become new ConnectedHandler(connection).waitingForAuthHandshakeResponse(handshaker, timeout)
 
     case CommandFailed(_: Connect) =>
-      log.warning("[Stopping Connection] Connection to {} failed", uri)
+      log.debug("[Stopping Connection] Connection to {} failed", uri)
```
AlanVerbner: Same here, maybe setting `error` as min level of this class

src/main/resources/logback.xml
```diff
@@ -25,13 +25,9 @@
         </encoder>
     </appender>
 
-    <root level="DEBUG">
+    <root level="INFO">
```
AlanVerbner: this will change our dev config, right? do we want that? :thinking:

ntallar:
Maybe we should also change the logging level of the failure to decode a message from [RLPxConnectionHandler](https://github.com/input-output-hk/etc-client/blob/phase/beta1/src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala#L141) to debug? Or maybe we could at least hide the exception message, so that the frequent `Tx with invalid signature` doesn't appear on the console.

src/main/resources/logback.xml
mcsherrylabs: It's not biggie one way or the other but probably best to leave it at `debug` for development.

src/main/resources/application.conf
```diff
@@ -260,7 +260,7 @@ grothendieck {
     peer-response-timeout = 10.seconds
 
     # Interval for logging syncing status info
-    print-status-interval = 2.seconds
+    print-status-interval = 10.seconds
```
mcsherrylabs: Can I suggest 30 seconds for this? Trying to lower the noise.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
mcsherrylabs: After discussion, `ConnectionFailed` is a regular common occurrence and probably not a `warning` level log and if we turn OFF logging we never see any future `info` messages. We could go either way with this but I think getting our logging into a consistent state  will be less of a "dark art" if we push the common occurrences into the `debug` category.

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
mcsherrylabs: Assuming this is a very common occurrence - `debug` works.

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
```diff
@@ -137,11 +137,8 @@ class RLPxConnectionHandler(
       case Success(message) =>
         context.parent ! MessageReceived(message)
 
-      case Failure(ex) => if (log.isDebugEnabled) {
-        log.error(ex, s"Cannot decode message from $peerId")
-      } else {
-        log.error(s"Cannot decode message from $peerId because ${ex.getMessage}")
-      }
+      case Failure(ex) =>
+        log.debug(s"Cannot decode message from $peerId, because of ${ex.getMessage}")
```
mcsherrylabs: `getMessage` will print the stack trace anyway.

LukasGasior1: really, will it? In fact I only changed error -> debug (and removed log.isDebugEnabled as it's not needed)

src/main/resources/logback.xml
```diff
@@ -30,8 +30,4 @@
         <appender-ref ref="FILE" />
     </root>
 
-    <logger name="io.iohk.ethereum.network.rlpx.RLPxConnectionHandler" level="INFO" />
-    <logger name="io.iohk.ethereum.blockchain.sync.SyncController" level="INFO" />
-    <logger name="io.iohk.ethereum.network.PeerActor" level="INFO" />
```
AlanVerbner: @LukasGasior1 I've just noticed that after removing this, my terminal gets full of 

```
13:41:04 Received message: NodeData{
values: Queue(f8518080a0d16cddfbb65f4589e94d9a8e5174de5ee58c4ba9cbc5cf947659e4150fe474e68080808080808080a068e86d6b2296a760f3d89468b0bf0ba185f32a9f041cb61c2fb62afc999be9a28080808080, 606060405236156100405760e060020a600035046335faa416811461004d57806355a373d61461006e5780638da5cb5b14610080578063b69ef8a814610092575b6101026000610107610096565b61017760008054819033600160a060020a039081169116146101c857610002565b610190600154600160a060020a031681565b610190600054600160a060020a031681565b6101ad5b6000600160009054906101000a9004600160a060020a0316600160a060020a03166370a08231306040518260e060020a0281526004018082600160a060020a031681526020019150506020604051808303816000876161da5a03f1156100025750506040515191505090565b005b50565b604080516001546000805460e060020a63a9059cbb028452600160a060020a03908116600485015260248401869052935194955092169263a9059cbb926044808401936020939290839003909101908290876161da5a03f115610002575050604051511515905061010457610002565b6040805192835260208301919091528051918290030190f35b60408051600160a060020a03929092168252519081900360200190f35b60408051918252519081900360200190f35b600091505b9091565b6101d0610096565b600154600080546040805160e060020a63a9059cbb028152600160a060020a03928316600482015260248101869052905194955092169263a9059cbb9260448181019360209392839003909101908290876161da5a03f115610002575050604051511590506101bf57600191506101c456, f838a120290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e5639594d6c9710363cdb12eb585e817f431d44cfa442dde, 6060604052361560275760e060020a600035046341c0e1b58114606e578063e5225381146096575b60d56000341115606c57346060908152605890600160a060020a033316907f90890809c654f11d6e72a28fa60149770a0d11ec6c92319d6ceb2bb0a4ea1a1590602090a35b565b60d5600054600160a060020a0390811633919091161415606c57600054600160a060020a0316ff5b60d5600054600160a060020a0390811633919091161415606c5760008054600160a060020a039081169190301631606082818181858883f150505050...
```

LukasGasior1: I've added it back

PR #264, opened 2017-07-28
[EC-256] Regression test for fast sync

adamsmo:
this PR is to not blacklist peers which are not responding with mpt nodes, instead we mark them as blockchain only, this PR also changes order in which we schedule fast sync element to get mpt as fast as possible

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
@@ -171,7 +171,7 @@ trait FastSync {
 
     syncStateStorageActor ! fastSyncStateStorage
 
-    private var blockChainOnlyPeers = Set.empty[Peer]
+    private var blockChainOnlyPeers = Seq.empty[Peer]
 
     private val syncStatePersistCancellable = scheduler.schedule(persistStateSnapshotInterval, persistStateSnapshotInterval, self, PersistSyncState)
     private val heartBeat = scheduler.schedule(syncRetryInterval, syncRetryInterval * 2, self, ProcessSyncing)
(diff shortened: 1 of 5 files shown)


adamsmo:
Yes, I will add limit in conf

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -171,11 +171,13 @@ trait FastSync {
 
     syncStateStorageActor ! fastSyncStateStorage
 
+    private var blockChainOnlyPeers = Set.empty[Peer]
```
LukasGasior1: should we update this Set when peer disconnects and so on? (see: `handlePeerUpdates`)

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -432,6 +453,7 @@ object FastSync {
 
   private case object ProcessSyncing
   private case object PersistSyncState
+  case class BlockChainOnlyDownload(peer: Peer)
```
LukasGasior1: maybe `MarkPeerBlockchainOnly`?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
adamsmo: it is additional logic, we have limit on this pool set to 100 in conf, new entries will naturally remove olde ones (including disconnected)

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -237,9 +237,10 @@ trait FastSync {
         bestBlockHeaderNumber)
     }
 
-    private def handleFailingMptPeers: Receive ={
-      case BlockChainOnlyDownload(peer) =>
-        blockChainOnlyPeers = (blockChainOnlyPeers + peer).take(blockChainOnlyPeersPoolSize)
+    private def handleFailingMptPeers: Receive = {
+      case MarkPeerBlockchainOnly(peer) => if (!blockChainOnlyPeers.contains(peer)) {
+        blockChainOnlyPeers = peer +: blockChainOnlyPeers.dropRight(1)
```
mcsherrylabs: If this is the only place the `blockChainOnlyPeers` seq grows then doesn't this mean it will always be size 1?

adamsmo: yes my bad, will fix it

PR #265, opened 2017-07-28
Handle invalid MAC exception on RLPxConnectionHandler

ntallar:
## Description
In case incoming connections are enabled (having `server-address.interface="0.0.0.0"`), when a client is restarted with a new node key, many incoming connection attempts were received that caused both an invalid MAC and increasing the number of connected nodes well beyond the limit (ex: `Discovered 0 nodes, connected to 94/45. Trying to connect to 0 more nodes.`).

## Fix
This PR includes a fix to the huge number of connected nodes, which was caused by the [exception restarting the RLPxConnectionHandler](http://doc.akka.io/docs/akka/current/scala/fault-tolerance.html#default-supervisor-strategy), as no handling of this exception was done. This was fixed by closing the connection in case any exception happened during the MAC check.

**Note:** In the depicted scenario the invalid MAC was caused by peers trying to connect to our node using the previous node key in the RLPx handshake, as the MAC incorporates the node key, so it is an expected exception.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/network/PeerActor.scala
+++ b/src/main/scala/io/iohk/ethereum/network/PeerActor.scala
@@ -2,11 +2,9 @@ package io.iohk.ethereum.network
 
 import java.net.{InetSocketAddress, URI}
 
-import akka.actor.SupervisorStrategy.Stop
+import akka.actor.SupervisorStrategy.Escalate
 import io.iohk.ethereum.network.PeerManagerActor.PeerConfiguration
 import akka.actor._
-import akka.agent.Agent
-import akka.util.ByteString
 import io.iohk.ethereum.network.p2p._
 import io.iohk.ethereum.network.p2p.messages.WireProtocol._
 import io.iohk.ethereum.network.p2p.messages.Versions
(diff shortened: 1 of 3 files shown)


src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
```diff
@@ -74,39 +76,51 @@ class RLPxConnectionHandler(
...
-              val size = ByteUtils.bigEndianToShort(data.take(2).toArray)
-              val (packetData, remainingData) = data.splitAt(size + 2)
-              val result = handshaker.handleResponseMessageV4(packetData)
-              processHandshakeResult(result, remainingData)
+          val maybePreEIP8Result = Try {
+            val result = handshaker.handleResponseMessage(data.take(ResponsePacketLength))
+            val remainingData = data.drop(ResponsePacketLength)
+            (result, remainingData)
+          }
+          val maybePostEIP8Result = Try {
```
adamsmo: this looks similar to line 84, is it possible to replace both with a method that returns Try?

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
```diff
@@ -42,9 +42,11 @@ class RLPxConnectionHandler(
 
   override def receive: Receive = waitingForCommand
 
+  def tcpActor: ActorRef = IO(Tcp)
+
   def waitingForCommand: Receive = {
     case ConnectTo(uri) =>
-      IO(Tcp) ! Connect(new InetSocketAddress(uri.getHost, uri.getPort))
+      tcpActor ! Connect(new InetSocketAddress(uri.getHost, uri.getPort))
```
adamsmo: 👍

LukasGasior1:
Hmm that sounds like we may need supervisor strategy implemented in PeerActor, do we?
```
  override val supervisorStrategy: OneForOneStrategy =
    OneForOneStrategy() {
      case _ => Stop
    }
```
maybe something like this instead of default actor restart on exception

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -43,6 +44,11 @@ class PeerActor[R <: HandshakeResult](
   import PeerActor._
   import context.{dispatcher, system}
 
+  override val supervisorStrategy: OneForOneStrategy =
+    OneForOneStrategy() {
+      case _ => Stop
```
LukasGasior1: hmm actually.. we should probably also stop PeerActor here. Currently if this happens we end up with a Peer that doesn't have a child connection running. We can just use Escalate here (and PeerManager will stop the whole Peer - see its supervisorStrategy)

LukasGasior1: sorry for wrong suggestion about using `Stop` :)

ntallar: You are right! Fixed :)

LukasGasior1:
@ntallar needs merge

PR #266, opened 2017-07-31
RPC timeouts fixes

LukasGasior1:


src/main/scala/io/iohk/ethereum/jsonrpc/NetService.scala
```diff
@@ -42,7 +42,7 @@ class NetService(nodeStatusHolder: Agent[NodeStatus], peerManager: ActorRef) {
 
   def peerCount(req: PeerCountRequest): ServiceResponse[PeerCountResponse] = {
     import akka.pattern.ask
-    implicit val timeout = Timeout(2.seconds)
+    implicit val timeout = Timeout(5.seconds)
```
adamsmo: should we add configuration for this?

mcsherrylabs: Because we've had to tune it then yes, yes we should.

LukasGasior1:
@adamsmo @mcsherrylabs fixed

src/main/resources/application.conf
```diff
@@ -152,6 +152,10 @@ grothendieck {
       # Enabled JSON-RPC APIs over the HTTP endpoint
       # Available choices are: eth, web3, net, personal
       apis = "eth,web3,net"
+
+      net {
+        peer-manager-timeout = 5.seconds
```
adamsmo: should we add this to `universal` and add description?

LukasGasior1: I wouldn't add it, it's internal thing

mcsherrylabs: (We're leaving out config that we don't really expect the user to change without having to talk to a developer. i.e. it's legit to leave it out of the universal file under certain conditions)

adamsmo: ok

PR #269, opened 2017-08-02
[EC-275] InvalidJump when destination value exceeds 32-bit integer

rtkaczyk:


src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -508,24 +508,26 @@ case object SSTORE extends OpCode(0x55, 2, 0, _.G_zero) {
...
     val (pos, stack1) = state.stack.pop
-    val dest = pos.intValue
-    if(state.program.validJumpDestinations.contains(dest))
+    val dest = pos.toInt // fail with InvalidJump if convertion to Int is lossy
+
+    if (pos == dest && state.program.validJumpDestinations.contains(dest))
       state.withStack(stack1).goto(dest)
     else
-      state.withError(InvalidJump(dest))
+      state.withError(InvalidJump(pos))
```
LukasGasior1: shouldn't we return original dest in the error?

rtkaczyk: this is the original uncoverted one

PR #270, opened 2017-08-02
Add windows ini support

mcsherrylabs:
- fix the stack size setting on Linux, it was not being picked up
- allow the client to run from the command line on windows, it could not get at the settings file (application.ini) it's apparently called 'etc_client_config.txt' on windows.
- allow the client to use SecureRandom() by default on all platforms. The use may still override this value.


src/universal/conf/application.ini
```diff
@@ -1 +1 @@
--Dconfig.file=./conf/grothendieck.conf -Dlogback.configurationFile=./conf/logback.xml -Xss10M
+-Dconfig.file=./conf/grothendieck.conf -Dlogback.configurationFile=./conf/logback.xml -J-Xss10M
```
AlanVerbner: good catch!

AlanVerbner:
Tried it on a Windows 10 machine and it worked perfectly

PR #272, opened 2017-08-03
[EC-249] Block bodies validation during FastSync

adamsmo:
this PR adds validation for block bodies in fast sync

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
@@ -1,17 +1,18 @@
 package io.iohk.ethereum.blockchain.sync
 
-import scala.concurrent.duration.FiniteDuration
-import scala.concurrent.ExecutionContext.Implicits.global
 import akka.actor._
 import akka.util.ByteString
 import io.iohk.ethereum.domain.BlockHeader
-import io.iohk.ethereum.network.{EtcPeerManagerActor, Peer}
-import io.iohk.ethereum.network.PeerEventBusActor.PeerEvent.MessageFromPeer
-import io.iohk.ethereum.network.PeerEventBusActor.{PeerSelector, Subscribe, Unsubscribe}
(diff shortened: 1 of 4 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -266,10 +276,16 @@ trait FastSync {
       requestedReceipts = requestedReceipts - handler
     }
 
+    private def validateBlocks(requestedHashes: Seq[ByteString], blockBodies: Seq[BlockBody]): Boolean = (requestedHashes zip blockBodies)
```
mcsherrylabs: I really really hate to throw away good blocks. They are hard to come by. Is there a compelling reason to?

adamsmo: this way we can reduce check time, If we encounter failing one we just stop checking (forall) if we want to filter out failing we have to check all of them
If we get failure it means that all block bodies may fail validation because they are from malicious peer or because of peer failure they are returned in incorrect order
those are the case I can think of, in my opinion failing fast in this case is better, but I can change that

mcsherrylabs: Yes! Given the integrated nature of the response any bad block means we don't trust the sender at all and so we fail the response in it's entirety immediately.

ntallar: Minor comment up to discussion, should check whether the length of the `requestedHashes` and `blockBodies` is the same?

LukasGasior1: I think we should also check that

LukasGasior1: hmm actually maybe not necessarily, as you say we enqueue the remaining blocks in `insertBlocks`

PR #273, opened 2017-08-03
Change morden conf so it is easy to run

adamsmo:
after this change you can start morden with
`./bin/mantis -Dconfig.file=conf/morden.conf`

### DIFF
--- a/src/universal/conf/morden.conf
+++ b/src/universal/conf/morden.conf
@@ -29,6 +29,6 @@ mantis {
 
     chain-id = "3e"
 
-    custom-genesis-file = "src/universal/conf/morden.json"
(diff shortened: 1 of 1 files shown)


src/universal/conf/morden.conf
```diff
@@ -27,6 +29,6 @@ mantis {
 
     chain-id = "3e"
 
-    custom-genesis-file = "conf/morden.json"
+    custom-genesis-file = "src/universal/conf/morden.json"
```
mcsherrylabs: But it this now broken when run from unzipped 'dist' ?

adamsmo:
not needed

PR #275, opened 2017-08-04
Missing block headers

adamsmo:
This PR handle case when we have requested block bodies by hashes for block headers that are not in our DB, this may happen because of killing client while persisting data

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
@@ -232,6 +232,7 @@ trait FastSync {
         case DbError =>
           blockBodiesQueue = Seq.empty
           receiptsQueue = Seq.empty
+          //todo adjust the formula to minimize redownloaded block headers
           bestBlockHeaderNumber = bestBlockHeaderNumber - 2 * blockHeadersPerRequest
           log.debug("missing block header for known hash")
           self ! ProcessSyncing
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
@@ -2,7 +2,7 @@ package io.iohk.ethereum.blockchain.sync
 
 import akka.actor._
 import akka.util.ByteString
-import io.iohk.ethereum.domain.BlockHeader
+import io.iohk.ethereum.domain.{Block, BlockHeader}
 import io.iohk.ethereum.network.EtcPeerManagerActor.PeerInfo
 import io.iohk.ethereum.network.PeerEventBusActor.PeerEvent.MessageFromPeer
 import io.iohk.ethereum.network.PeerEventBusActor.SubscriptionClassifier.MessageClassifier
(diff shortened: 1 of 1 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -276,14 +281,24 @@ trait FastSync {
...
-        case (Some(header), body) =>
-          val result = validators.blockValidator.validateHeaderAndBody(header, body)
-          result.isRight
-        case _ => false
-      }
+    private def validateBlocks(requestedHashes: Seq[ByteString], blockBodies: Seq[BlockBody]): ValidationResult = {
+      var result: ValidationResult = Valid
+      (requestedHashes zip blockBodies)
+        .map { case (hash, body) => (blockchain.getBlockHeaderByHash(hash), body) }
+        .forall {
```
adamsmo: to stop on first error

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -473,6 +488,11 @@ object FastSync {
   private case object PersistSyncState
   case class MarkPeerBlockchainOnly(peer: Peer)
 
+  private sealed trait ValidationResult
```
LukasGasior1: maybe BlockBodyValidationResult?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -276,14 +281,24 @@ trait FastSync {
...
+      var result: ValidationResult = Valid
+      (requestedHashes zip blockBodies)
+        .map { case (hash, body) => (blockchain.getBlockHeaderByHash(hash), body) }
+        .forall {
+          case (Some(header), body) =>
+            val validationResult: Either[BlockValidator.BlockError, Block] = validators.blockValidator.validateHeaderAndBody(header, body)
+            result = validationResult.fold(_ => Invalid, _ => Valid)
+            validationResult.isRight
+          case _ =>
+            blockBodiesQueue = Seq.empty
```
LukasGasior1: can we move this side effects out of this function?

LukasGasior1: actually I'm not sure if it's good idea to clear the queues completely, maybe we can only remove the bodies/receipts for the headers we don't have?

adamsmo: side effects - good idea
filtering failure - I think clearing all is simpler than checking which block header are missing and deciding base on that, also i we clear we do not have to check every block hash for missing block header - if we find first we fail

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -226,6 +222,22 @@ trait FastSync {
...
+      validateBlocks(requestedHashes, blockBodies) match {
+        case Valid =>
+          insertBlocks(requestedHashes, blockBodies)
+        case Invalid =>
+          blacklist(peer.id, blacklistDuration, s"responded with block bodies not matching block headers, blacklisting for $blacklistDuration")
+          self ! FastSync.EnqueueBlockBodies(requestedHashes)
+        case DbError =>
+          blockBodiesQueue = Seq.empty
+          receiptsQueue = Seq.empty
+          bestBlockHeaderNumber = bestBlockHeaderNumber - 2 * blockHeadersPerRequest
```
LukasGasior1: what kind of magic formula is this? :P
maybe we can just set it to the first missing header num?

adamsmo: just rough estimate of safety buffer ;) , but yes we can go with first missing header

mcsherrylabs: (And then it won't need precedence brackets for readability)

adamsmo:
@mcsherrylabs  will add TODO

PR #279, opened 2017-08-09
[EC-269] Fixed net_version on Morden

ntallar:
## Description

When using the JSON RPC method `net_version` on the Morden testnet the value `"1"` was returned, while both the [specification](https://github.com/ethereum/wiki/wiki/JSON-RPC#net_version) and Geth and Parity clients return `"2"`.

## Fix

Fixed `net_version` to use `network-id` configuration instead of `protocol-version` (which never changed from `"1"`).
The value of `protocol-version` was also fixed according to [specification](https://github.com/ethereum/wiki/wiki/Ethereum-Wire-Protocol#ethereum-sub-protocol) and it was added to be used on `eth_protocolVersion`

### DIFF
--- a/src/main/scala/io/iohk/ethereum/utils/Config.scala
+++ b/src/main/scala/io/iohk/ethereum/utils/Config.scala
@@ -38,7 +38,7 @@ object Config {
   object Network {
     private val networkConfig = config.getConfig("network")
 
-    val protocolVersion = networkConfig.getString("protocol-version")
+    val protocolVersion = networkConfig.getInt("protocol-version")
 
     object Server {
       private val serverConfig = networkConfig.getConfig("server-address")
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -24,7 +24,7 @@ mantis {
 
   network {
     # Ethereum protocol version
-    protocol-version = "1"
+    protocol-version = 63
 
     server-address {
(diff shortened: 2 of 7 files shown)


src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -38,7 +38,7 @@ object Config {
   object Network {
     private val networkConfig = config.getConfig("network")
 
-    val protocolVersion = networkConfig.getString("protocol-version")
+    val protocolVersion = networkConfig.getInt("protocol-version")
```
rtkaczyk: Any reason this now `Int`, but network Id remains a `String`?

ntallar: Maybe I misunderstood something, but it seems the network id is an `Int` in the [config file](https://github.com/input-output-hk/etc-client/pull/279/files#diff-6bb9146203d6ddeb4f8d799d28054c96R62).
Maybe it would be better to have them both as a `String` instead? AFAIK there's no strong reason for having them being `Int` or `String`, so they could be changed either way.

rtkaczyk: I misread that, either way I was just curious, we can easily change this at anytime.

PR #280, opened 2017-08-09
[EC-278] Missing block headers when client is restarted

adamsmo:
This PR is to ask for block bodies and receipts only if we are sure that block header was inserted into DB
This also removes block and receipts scheduling from `SyncBlockHeadersRequestHandler` which should be handled by `FastSync`

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
@@ -313,6 +313,10 @@ trait FastSync {
           bestBlockHeaderNumber = lastHeader.number
         }
       }
+
+      val blockHashes = blockHeadersObtained.map(_.hash)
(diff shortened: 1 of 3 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -333,6 +333,10 @@ trait FastSync {
           bestBlockHeaderNumber = lastHeader.number
         }
       }
+
+      val blockHashes = blockHeadersObtained.map(_.hash)
+      self ! FastSync.EnqueueBlockBodies(blockHashes)
+      self ! FastSync.EnqueueReceipts(blockHashes)
```
rtkaczyk: Is this tested?

adamsmo: yes it is, we are testing if FastSync finishes and switches to regular and this is required (downloading block bodies and receipts)

PR #282, opened 2017-08-09
[EC-283] Ignore letter case in key file json keys

LukasGasior1:


rtkaczyk:
I think the appropriate course of action would be to leave it and report the bug on MyEtherWallet. Correct me if I'm wrong but JSON, by specification, is case-sensitive. But since it's that easy we may as well support it.

Should we add a youtrack ticket to keep track of such changes?

PR #283, opened 2017-08-10
[EC-268] Fix pending transactions not forwarded to new peers

LukasGasior1:


src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
```diff
@@ -54,9 +54,13 @@ class PendingTransactionsManager(txPoolConfig: TxPoolConfig, peerManager: ActorR
 
   implicit val timeout = Timeout(3.seconds)
 
-  peerMessageBus ! Subscribe(MessageClassifier(Set(SignedTransactions.code), PeerSelector.AllPeers))
+  peerEventBus ! Subscribe(SubscriptionClassifier.PeerHandshaked)
+  peerEventBus ! Subscribe(MessageClassifier(Set(SignedTransactions.code), PeerSelector.AllPeers))
 
   override def receive: Receive = {
+    case PeerEvent.PeerHandshakeSuccessful(peer, _) =>
+      self ! NotifyPeer(pendingTransactions.map(_.stx), peer)
```
adamsmo: should we add limit here? 
to not send all pending transactions
I am thinking about some strange situation when we have a lot of pending transactions

LukasGasior1: @adamsmo hmm maybe we should have a limit for pending transactions in general?

adamsmo: we have limit for the pool

LukasGasior1: lol, true :man_facepalming: why do we need another limit then? Just so that we don't send all pending tx we have at once?

adamsmo: yes, but as I said, it may not be worth implementing because we already have limit for the pool

LukasGasior1: Yes I think it would be very unlikely situation, and even if, the workaround would be very simple, just restart the node.

PR #284, opened 2017-08-10
[EC-253] Removed use of BlockchainStorages outside Blockchain

ntallar:
## Description

Reduced the passing of `BlockchainStorages` as a parameter and the construction of `BlockchainImpl` through out the codebase, so as to work just with `Blockchain` whenever possible.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncNodesRequestHandler.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncNodesRequestHandler.scala
@@ -4,8 +4,8 @@ import akka.actor.{ActorRef, Props, Scheduler}
 import akka.util.ByteString
 import io.iohk.ethereum.blockchain.sync.FastSync._
 import io.iohk.ethereum.crypto._
+import io.iohk.ethereum.db.storage.NodeStorage.{NodeEncoded, NodeHash}
 import io.iohk.ethereum.domain.{Account, Blockchain}
-import io.iohk.ethereum.mpt.NodesKeyValueStorage
 import io.iohk.ethereum.network.Peer
 import io.iohk.ethereum.network.p2p.messages.PV63._
 import org.spongycastle.util.encoders.Hex
@@ -16,7 +16,7 @@ class FastSyncNodesRequestHandler(
     peerMessageBus: ActorRef,
     requestedHashes: Seq[HashType],
     blockchain: Blockchain,
-    nodesKeyValueStorage: NodesKeyValueStorage)(implicit scheduler: Scheduler)
+    saveNodeFn: (NodeHash, NodeEncoded) => Unit)(implicit scheduler: Scheduler)
   extends SyncRequestHandler[GetNodeData, NodeData](peer, etcPeerManager, peerMessageBus) {
 
   override val requestMsg = GetNodeData(requestedHashes.map(_.v))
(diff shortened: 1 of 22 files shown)


src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
```diff
@@ -116,7 +116,7 @@ trait NodeStatusBuilder {
 trait BlockChainBuilder {
   self: StorageBuilder =>
 
-  lazy val blockchain: Blockchain = BlockchainImpl(storagesInstance.storages)
+  lazy val blockchain: BlockchainImpl = BlockchainImpl(storagesInstance.storages)
```
adamsmo: is this necessary

rtkaczyk:
@ntallar do you think we could also address this:

```
class GenesisDataLoader(
    dataSource: DataSource,
    blockchain: Blockchain,
    pruningMode: PruningMode,
    blockchainConfig: BlockchainConfig,
    dbConfig: DbConfig)
```

Could be a separate task...

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
ntallar: With our current codebase it is, as the `LedgerImpl` requires a `BlockchainImpl` as it is currently coupled with the `InMemoryWorldStateProxy` (which probably should be changed, I added a [FIXME](https://github.com/input-output-hk/etc-client/pull/284/files#diff-1b4e47475b368c25eb5141199a4289d0R25) regarding that).

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncNodesRequestHandler.scala
```diff
@@ -77,7 +77,7 @@ class FastSyncNodesRequestHandler(
       val evm = n.getAccount.codeHash
       val storage = n.getAccount.storageRoot
 
-      nodesKeyValueStorage.put(n.hash, n.toBytes)
+      saveNodeFn(n.hash, n.toBytes)
```
ntallar: :thinking: 
The purpose of doing so would be to avoid the usage of the DB on the `SyncRequestActor`s, so that it is only accessed by the `FastSync` actor, right? (which BTW is also used in `FastSyncReceiptsRequestHandler`)

ntallar:
@rtkaczyk I agree that we should remove that use of `DataSource` from `GenesisDataLoader` (probably the using something like `InMemorySimpleMapProxy` would help). However I also prefer creating a separate task for looking into that, will create one now :+1:

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncNodesRequestHandler.scala
ntallar: I think @LukasGasior1 addressed on the VSU that he will be working on an improvement that involves fixing this, so I would leave it as it is for now.

adamsmo: 👍

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
adamsmo: FIXME is good, do we have ticket for that as well?

ntallar: :+1:
Created [EC-288](https://iohk.myjetbrains.com/youtrack/issue/EC-288) for looking into it.

PR #285, opened 2017-08-11
[EC-282] Randomise peer selection / throttling of fast sync per peer

adamsmo:


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -163,6 +167,7 @@ trait FastSync {
     private var bestBlockHeaderNumber: BigInt = initialSyncState.bestBlockHeaderNumber
 
     private var assignedHandlers: Map[ActorRef, Peer] = Map.empty
+    private var peerRequestsTime: Map[Peer, Date] = Map.empty
```
rtkaczyk: Perhaps `java.time.Instant` or simply `Long` would be more appropriate than `Date`?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -14,6 +17,7 @@ import io.iohk.ethereum.validators.BlockValidator
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration.FiniteDuration
+import scala.util.Random
```
rtkaczyk: Is it used?

src/main/resources/application.conf
```diff
@@ -303,6 +303,9 @@ mantis {
     # size of the list that keeps track of peers that are failing to provide us with mpt node
     # we switch them to download only blockchain elements
     fastsync-block-chain-only-peers-pool = 100
+
+    #time between 2 consecutive requests to peer when doing fast sync, this is to prevent flagging us as spammer
```
rtkaczyk: Add space after #

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
adamsmo: not any more

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
adamsmo: good point

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
```diff
@@ -121,19 +121,27 @@ class SyncControllerSpec extends FlatSpec with Matchers {
     peerMessageBus.reply(MessageFromPeer(NodeData(Seq(stateMptLeafWithAccount)), peer2.id))
     peerMessageBus.expectMsg(Unsubscribe(MessageClassifier(Set(NodeData.code), PeerSelector.WithId(peer2.id))))
 
+    //wait for peers throttle
+    Thread.sleep(2.second.toMillis)
+    //trigger scheduling
+    time.advance(2.second)
```
rtkaczyk: If it's set to 100.millis for tests why are we waiting 2 seconds? Shouldn't we be using that config value here? Also doesn't the scheduler time already advance when we sleep?

src/main/resources/application.conf
LukasGasior1: minor comment: you can also change `2.second` to `2.seconds`

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
LukasGasior1: :+1: for long

src/main/resources/application.conf
rtkaczyk: Or `2 seconds` might be even better :)

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
```diff
@@ -121,19 +121,27 @@ class SyncControllerSpec extends FlatSpec with Matchers {
     peerMessageBus.reply(MessageFromPeer(NodeData(Seq(stateMptLeafWithAccount)), peer2.id))
     peerMessageBus.expectMsg(Unsubscribe(MessageClassifier(Set(NodeData.code), PeerSelector.WithId(peer2.id))))
 
+    //wait for peers throttle
+    Thread.sleep(200.milliseconds.toMillis)
```
rtkaczyk: I meant that we could refer to the value from configuration rather than hard-coding it.
I also pointed out that we probably don't need `time.advance` if we already do the `sleep`.

adamsmo: 👍 
we need `time.advance` because first we wait so timestamp for peer would pass validation, and `time.advance` is to start scheduled task that will ask peer for blockchain elements

rtkaczyk: ok, is the value in `time.advance` related to the one in `Thread.sleep`?

adamsmo: no it is not related

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -241,6 +246,8 @@ trait FastSync {
 
     private def handleActorTerminate(ref: ActorRef) = {
       context unwatch ref
+      val peer = assignedHandlers(ref)
+      peerRequestsTime -= peer
```
ntallar: Minor detail: As we remove the terminated peer from `peerRequestsTime`, should we do the same when handling `SyncRequestHandler.Done`?

adamsmo: no done means that we finished with request and peer is free for next round (after throttle period)

PR #286, opened 2017-08-11
[EC-248] Limit blockchain branch resolving depth

adamsmo:


src/main/resources/application.conf
```diff
@@ -303,6 +303,11 @@ mantis {
     # size of the list that keeps track of peers that are failing to provide us with mpt node
     # we switch them to download only blockchain elements
     fastsync-block-chain-only-peers-pool = 100
+
+    #limit of length of branch in chain that will be resolved while doing regular sync
+    #if branch is longer resolution will be aborted because of risk of out of memory error
+    #too long branch may also indicate malicious peer
```
rtkaczyk: Pls add spaces after #

src/main/resources/application.conf
```diff
@@ -303,6 +303,11 @@ mantis {
...
+    branch-max-depth-resolving = 100
```
rtkaczyk: How did you arrive at this value?

rtkaczyk:
Do we need a test for this?

adamsmo:
yes we do
@rtkaczyk I forgot to put wip on this PR

src/test/scala/io/iohk/ethereum/blockchain/sync/FastSyncBlockBodiesRequestHandlerSpec.scala
```diff
@@ -83,6 +83,7 @@ class FastSyncBlockBodiesRequestHandlerSpec extends FlatSpec with Matchers {
     val fastSyncBlockBodiesRequestHandler: ActorRef =
       parent.childActorOf(SyncBlockBodiesRequestHandler.props(
         peer,
+        1.second,
```
ntallar: Minor detail: Should we maybe use the timeouts defined in `Timeouts.scala` and change it for `Timeouts.normalTimeout`?

adamsmo: it is not test timeout, it is more of configuration value for SyncBlockBodiesRequestHandler actor, I will put it in named constant

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncReceiptsRequestHandler.scala
```diff
@@ -113,12 +116,14 @@ class FastSyncReceiptsRequestHandler(
   }
 }
 
+// todo move inserting receipts to FastSync
```
adamsmo: this todo is no longer relevant because of fast sync refactoring, it will be removed after merge with refactoring branch

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncRequestHandler.scala
```diff
@@ -8,11 +8,12 @@ import io.iohk.ethereum.network.PeerEventBusActor.PeerEvent.{MessageFromPeer, Pe
...
 import io.iohk.ethereum.network.p2p.{Message, MessageSerializable}
-import io.iohk.ethereum.utils.Config.Sync._
+
+import scala.concurrent.duration.FiniteDuration
 
 abstract class SyncRequestHandler[RequestMsg <: Message,
-                                  ResponseMsg <: Message : ClassTag](peer: Peer, etcPeerManagerActor: ActorRef,
-                                                                     peerEventBus: ActorRef)
+                                  ResponseMsg <: Message : ClassTag](peer: Peer, peerResponseTimeout: FiniteDuration,
+                                                                     etcPeerManagerActor: ActorRef, peerEventBus: ActorRef)
```
rtkaczyk: What is the `ClassTag` for?

src/main/resources/application.conf
```diff
@@ -298,14 +298,20 @@ mantis {
...
     fastsync-block-chain-only-peers-pool = 100
 
     # time between 2 consecutive requests to peer when doing fast sync, this is to prevent flagging us as spammer
     fastsync-throttle = 2 seconds
+
+    # limit of length of branch in chain that will be resolved while doing regular sync
+    # if branch is longer resolution will be aborted because of risk of out of memory error
+    # too long branch may also indicate malicious peer, 24 because confirmation depth in ethereum is ~12 blocks
+    # so we allow branches of 2x confirmation depth
+    branch-max-depth-resolving = 24
```
rtkaczyk: What is the correlation, if any, between this and `block-resolving-per-request`?

adamsmo: `block-resolving-per-request` has to be lower than `branch-max-depth-resolving`

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncRequestHandler.scala
adamsmo: it is for
```
SyncRequestHandler.scala:46: abstract type pattern ResponseMsg is unchecked since it is eliminated by erasure
```

src/main/resources/application.conf
rtkaczyk: I would also rename them, cause they may seem unrelated now. How about: `branch-resolution-batch-size` and `branch-resolution-max-depth`. Also wouldn't hurt to make them adjacent in the config file.

As for the values I would say max-depth has to be a multiple of batch-size. Otherwise we will never reach the max depth (when trying to resolve).

Finally, it would be better if the check for headers queue size (>=) would come after the attempt to resolve. Otherwise, on failed resolution, we will always be making 1 redunant requests (the downloaded headers will always be discarded because max-depth was reached)

adamsmo: about `>=` I think we will not be making 1 redundant request, because we start resolving with 1 block header in the queue
we start resolving branches when new block header that we get do not have a parent in our chain

adamsmo:
@ntallar / @rtkaczyk could you take a look?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -104,18 +103,22 @@ trait RegularSync extends BlockBroadcast {
...
         scheduleResume()
     }
   }
 
   private def handleBlockBranchResolution(peer: Peer, message: Seq[BlockHeader]) =
-    //todo limit max branch depth? [EC-248]
     if (message.nonEmpty && message.last.hash == headersQueue.head.parentHash) {
       headersQueue = message ++ headersQueue
-      processBlockHeaders(peer, headersQueue)
+      if (headersQueue.length > branchResolutionMaxDepth) {
```
rtkaczyk: So please correct me if I'm misreading the code here:
- start with 1 header which is missing the parent
- ask for 12 additional headers
- we have `13 < maxDepth` headers
- try to resolve and still parent not found
- ask for 12 more
- we have `25 > maxDepth`, we give up

And let's say the non-orphan header was the 20th in line. But we haven't checked it because the `maxDepth` condition short-circuited.

adamsmo: yes, you are right it will fail in that case, this should be:
((headersQueue.length - 1) > branchResolutionMaxDepth), and if it is equal it should go to `processBlockHeaders()` and check if branch is resolved

adamsmo: after more thinking I decided to change max depth to number of batches to process while resolving since it is less prone to errors when changing configuration

rtkaczyk: I also had that thought, but it seemed that depth expressed in number of blocks is more natural/intuitive. But as long as it's well documented I think we're fine.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -155,9 +153,14 @@ trait RegularSync extends BlockBroadcast {
             scheduleResume()
           }
         } else {
-          val request = GetBlockHeaders(Right(headersQueue.head.parentHash), blockResolveDepth, skip = 0, reverse = true)
-          waitingForActor = Some(context.actorOf(
-            SyncBlockHeadersRequestHandler.props(peer, etcPeerManager, peerEventBus, request, resolveBranches = true)))
+          if ((headersQueue.length - 1) / branchResolutionBatchSize >= branchResolutionMaxRequests) {
```
rtkaczyk: Yep, I think this should be better

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -104,18 +103,17 @@ trait RegularSync extends BlockBroadcast {
...
         scheduleResume()
     }
   }
 
   private def handleBlockBranchResolution(peer: Peer, message: Seq[BlockHeader]) =
-    //todo limit max branch depth? [EC-248]
     if (message.nonEmpty && message.last.hash == headersQueue.head.parentHash) {
       headersQueue = message ++ headersQueue
-      processBlockHeaders(peer, headersQueue)
+        processBlockHeaders(peer, headersQueue)
```
rtkaczyk: I hate to be such pain, but please remove this indent

PR #287, opened 2017-08-11
[EC-167] Fix error while running regular sync without all MPT nodes

ntallar:
## Description

In the scenario that:
* The client is started with fast sync enabled but it is canceled it before it completes
* The client is restarted with regular sync enabled.

currently it resulted in the client throwing the exception `Node not found, trie is inconsistent`.

## Fix
The exception is caused by trying to retrieve the MPT for the `BestBlockNumber` saved during fast sync, for which no MPT node is stored.
This was fixed by continuing with fast sync and logging this, even though regular sync was selected.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
@@ -176,7 +176,10 @@ trait FastSync {
 
     private var blockChainOnlyPeers = Seq.empty[Peer]
 
-    private val syncStatePersistCancellable = scheduler.schedule(0.seconds, persistStateSnapshotInterval, self, PersistSyncState)
+    //Delay before starting to persist snapshot. It should be 0, as the presence of it marks that fast sync was started
+    private val persistStateSnapshotDelay: FiniteDuration = 0.seconds
+
+    private val syncStatePersistCancellable = scheduler.schedule(persistStateSnapshotDelay, persistStateSnapshotInterval, self, PersistSyncState)
     private val heartBeat = scheduler.schedule(syncRetryInterval, syncRetryInterval * 2, self, ProcessSyncing)
 
     // scalastyle:off cyclomatic.complexity
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
@@ -14,6 +14,7 @@ import io.iohk.ethereum.validators.BlockValidator
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration.FiniteDuration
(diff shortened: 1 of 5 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -175,7 +176,7 @@ trait FastSync {
 
     private var blockChainOnlyPeers = Seq.empty[Peer]
 
-    private val syncStatePersistCancellable = scheduler.schedule(persistStateSnapshotInterval, persistStateSnapshotInterval, self, PersistSyncState)
+    private val syncStatePersistCancellable = scheduler.schedule(0.seconds, persistStateSnapshotInterval, self, PersistSyncState)
```
adamsmo: why 0 instead of named constant/configuration ?

src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -94,7 +94,31 @@ object Config {
 
   }
 
-  object Sync {
+  trait SyncConfig {
```
adamsmo: 👍

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
ntallar: In this case this value shouldn't be configurable as while checking whether to allow starting regular sync or not, it is needed to [check whether fast sync had previously started](https://github.com/input-output-hk/etc-client/pull/287/files#diff-8c13c0426c84bb54f28f3662b267aa16R70).
For this it is checked whether any SyncState was saved, so right after `FastSync` starts, it's state should be saved to prevent it not being saved before `BestBlockNumber` is updated.

adamsmo: ok, so maybe maned constant?

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
```diff
@@ -65,8 +66,12 @@ class SyncController(
         case (true, false) =>
           startRegularSync()
         case (false, false) =>
-          fastSyncStateStorage.purge()
-          startRegularSync()
+          //Check whether fast sync was started before
+          if (fastSyncStateStorage.getSyncState().isDefined) {
```
AlanVerbner: Even if this is ok, it's confusing to me to retrieve if the fastSync has ever started doing `fastSyncStateStorage.getSyncState()` but to check if it has finished we need to do it based on  `appStateStorage.isFastSyncDone()`

adamsmo: we may need new flag for this, but I would do it in separate PR

PR #290, opened 2017-08-15
[EC-281] Give more meaningful message if user inputs badly formed peers

ntallar:
## Description

Parse bootstrap nodes so as to give more meaningful message if user inputs badly formed peers. This also avoids throwing exception when an invalid peer is found, logging it instead.


src/main/scala/io/iohk/ethereum/network/discovery/Node.scala
```diff
@@ -0,0 +1,61 @@
...
+    require(maybeHost.isSuccess, s"Invalid host '${Option(uri.getHost).getOrElse("")}'")
+    val nodeId = maybeNodeId.get
+    val host = maybeHost.get
+
+    require(scheme == NodeScheme, s"Invalid node scheme $scheme, only $NodeScheme is supported")
+    require(nodeId.size == NodeIdSize, s"Invalid nodeId size ${nodeId.size}, it should be $NodeIdSize bytes long")
+
+    val address = new InetSocketAddress(host, port)
+    //FIXME: We currently don't support IPv6 nodes
+    require(host.isInstanceOf[Inet4Address], s"Invalid host $host, only IPv4 addresses are currently supported")
```
AlanVerbner: :thinking:

src/main/scala/io/iohk/ethereum/network/discovery/Node.scala
```diff
@@ -0,0 +1,61 @@
...
+    require(maybeNodeId.isSuccess, s"Invalid nodeId '${Option(uri.getUserInfo).getOrElse("")}'")
+    require(maybeHost.isSuccess, s"Invalid host '${Option(uri.getHost).getOrElse("")}'")
+    val nodeId = maybeNodeId.get
+    val host = maybeHost.get
+
+    require(scheme == NodeScheme, s"Invalid node scheme $scheme, only $NodeScheme is supported")
+    require(nodeId.size == NodeIdSize, s"Invalid nodeId size ${nodeId.size}, it should be $NodeIdSize bytes long")
+
+    val address = new InetSocketAddress(host, port)
+    //FIXME: We currently don't support IPv6 nodes
```
AlanVerbner: Is there any task for for this?

ntallar: You are right, there wasn't any. Just created [EC-295](https://iohk.myjetbrains.com/youtrack/issue/EC-295) for looking into it.

src/main/scala/io/iohk/ethereum/network/discovery/Node.scala
```diff
@@ -0,0 +1,67 @@
...
+    new URI(s"enode://${Hex.toHexString(id.toArray[Byte])}@$host:$port")
+  }
+}
+
+case class NodeImpl(id: ByteString, addr: InetSocketAddress) extends Node
+
+object NodeImpl {
+  def fromUri(uri: URI): NodeImpl = {
+    val nodeId = ByteString(Hex.decode(uri.getUserInfo))
+    NodeImpl(nodeId, new InetSocketAddress(uri.getHost, uri.getPort))
```
adamsmo: can this fail, I mean if URI is badly formatted
what happens if `getUserInfo` returns something that is not node Id?
should we use here similar logic as in `parseNode()`

ntallar: Maybe it should be checked, but currently it's only used for the known nodes stored on the storage, which are properly formatted as their URI is generated by our client.
However that made me realize that we weren't checking whether the node information received over the network was properly formatted, so it was good thinking that  :+1:

adamsmo:
CircleCI failed

src/main/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManager.scala
```diff
@@ -113,22 +111,18 @@ object PeerDiscoveryManager {
...
+  object NodeWithTimestamp {
 
-    def fromUri(uri: URI): Node = {
-      val nodeId = ByteString(Hex.decode(uri.getUserInfo))
-      Node(nodeId, new InetSocketAddress(uri.getHost, uri.getPort), System.currentTimeMillis())
-    }
-  }
+    def fromUri(uri: URI): NodeWithTimestamp = fromNode(NodeImpl.fromUri(uri))
+
+    def fromNode(node: Node): NodeWithTimestamp = NodeWithTimestamp(node.id, node.addr, System.currentTimeMillis())
```
LukasGasior1: to be honest, I'd really prefer to use just a single `case class Node()` (without splitting it into trait and impl). And use that class here, like: `DiscoveryNodeInfo(node: Node, timestamp)`, alternatively we could add another map: `addTimestamps: Map[NodeId, timestamp]` and just use the `Node` in original map.

src/main/scala/io/iohk/ethereum/network/discovery/Node.scala
```diff
@@ -0,0 +1,62 @@
...
+  val NodeIdSize = 64
+
+  /**
+    * Parse a node string, for it to be valid it should have the format:
+    * "enode://[128 char (64bytes) hex string]@[IPv4 address]:[port]"
+    *
+    * @param node to be parsed
+    * @return the parsed node, or the error detected during parsing
+    */
+  def parseNode(node: String): Try[Node] = {
```
KonradStaniec: Should 't we accumulate errors during parsing (for one node there can be 3 of them), and at the end inform a user about all errors in the node, not to force them to correct them one by one? I would imagine return signature would be something like `Either[Set[Throwable], Node]`. what do you think?

src/main/scala/io/iohk/ethereum/network/discovery/DiscoveryConfig.scala
```diff
@@ -30,4 +34,21 @@ object DiscoveryConfig {
...
       messageExpiration = discoveryConfig.getDuration("message-expiration").toMillis.millis)
   }
+
+  /**
+    * Parses all the bootstrap nodes, logging the invalid ones and returning the valid ones
+    *
+    * @param unParsedBootStrapNodes, with the bootstrap nodes to be parsed
+    * @return set of parsed and valid bootstrap nodes
+    */
+  private def parseBootstrapNodes(unParsedBootStrapNodes: Set[String]): Set[Node] = unParsedBootStrapNodes.foldLeft[Set[Node]](Set.empty) {
```
KonradStaniec: minor, shouldn't it belong to `NodeParser`?

PR #295, opened 2017-08-22
[EC-286] Investigate use of DataSource on GenesisDataLoader

adamsmo:
This PR removes `DataSource` from `GenesisDataLoader` which is replaced by usage of `Blockchain`


src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
```diff
@@ -134,15 +118,33 @@ class GenesisDataLoader(
         Failure(new RuntimeException("Genesis data present in the database does not match genesis block from file." +
           " Use different directory for running private blockchains."))
       case None =>
-        // using empty namespace because ephemDataSource.storage already has the namespace-prefixed keys
-        ephemDataSource.storage.toSeq.grouped(dbConfig.batchSize).foreach(toStore => dataSource.update(IndexedSeq(), Nil, toStore))
+        // key.tail to remove namespace-prefix keys
```
adamsmo: there should be option to remove prefix from key so GenesisDataLoader.scala does not depend on NodeStorage

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
```diff
@@ -134,15 +118,33 @@ class GenesisDataLoader(
         Failure(new RuntimeException("Genesis data present in the database does not match genesis block from file." +
           " Use different directory for running private blockchains."))
       case None =>
-        // using empty namespace because ephemDataSource.storage already has the namespace-prefixed keys
-        ephemDataSource.storage.toSeq.grouped(dbConfig.batchSize).foreach(toStore => dataSource.update(IndexedSeq(), Nil, toStore))
+        ephemDataSource.getAll(nodeStorage.namespace)
+          .foreach { case (key, value) => blockchain.saveNode(ByteString(key.toArray[Byte]), value.toArray[Byte], 0) }
```
LukasGasior1: as far as I remember that grouping was added to (significantly?) improve performance. Don't we need it anymore?

adamsmo: good catch, will add it back

adamsmo:
I tested performance on MacOs and it looks like version without direct access to data source is slower
original version loads genesis data in 5 seconds
version without direct access to data source loads genesis data in 7 seconds
there is no difference if we use batching or not in newer version
I was testing loading main net genesis file

adamsmo:
for some reason this is failing when doing regular sync

LukasGasior1:
is it failing only on regular sync? how so?

adamsmo:
We are checking MPT state root because we are executing blocks

AlanVerbner:
@adamsmo havent checks the code...but...be sure you are storing data as pruning mode would expect it (if basic pruning is enabled references count is stored alongside with mptnode)

adamsmo:
I think issue is with loading genesis data through `blockchain.saveNode` with enabled pruning
I think it results in incorrect MPT state root hash

LukasGasior1:
oh, I see. I thought it's the actual genesis init code that's failing

LukasGasior1:
just a random guess: maybe something with prefixes (doubled or missing)?

adamsmo:
@AlanVerbner good point, I think this is the issue that reference count is added twice, I am not removing this info when accessing mpt nodes direct in data source

AlanVerbner:
Well....storages are dealing with serialization of data so if writing directly through the data source you need to be careful with that. I would avoid using data source directly

adamsmo:
I think it is fixed, now since I am saving nodes with `blockchain.saveNode` I do not need to know what pruning is set for the app and I can set my internal pruning for checking MPT root hash to Archive which will not add any data that I will have to remove later

PR #296, opened 2017-08-23
[EC-292] Run Ethereum Test suite for 'blockchain'

KonradStaniec:
Current status is 89 tests failing, most of them - 55,  in `GeneralStateTests` package.


src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
```diff
@@ -0,0 +1,143 @@
...
+      val emptyStorage = worldWithAccountAndCode.getStorage(address)
+      val updatedStorage = accountState.storage.foldLeft(emptyStorage) { case (storage, (key, value)) =>
+        storage.store(key, value)
+      }
+      worldWithAccountAndCode.saveStorage(address, updatedStorage)
+    }
+  }
+
+  // Function copied from RegularSync.scala, as ledger itself do not persis schanges in blockchain
+  // TODO We need to take forks and branches into account
```
ntallar: Should we fix this TODO? Does it affect the implementation in `RegularSync.scala`?

KonradStaniec:
Quick summary of found and fixed bugs:

- Check if `v` (recovery id) sent to precompiled contract has correct format (only last byte set) https://github.com/input-output-hk/mantis/pull/296/commits/84a0ad937ca752835e578529f7d194062216a4bd
- Proper handling of address collisions in the creation of a new account (in Ledger and in EVM). https://github.com/input-output-hk/mantis/pull/296/commits/a9ea51586170289a6ffe289197b84a8879b634a6 and https://github.com/input-output-hk/mantis/pull/296/commits/1c41be329d57bacf620bf169d1d48668601170f6
- Proper handling of ether during account self-destruction and refunding to self (ether should be destroyed) https://github.com/input-output-hk/mantis/pull/296/commits/1c41be329d57bacf620bf169d1d48668601170f6
- Expanding memory in case of CALLop failure https://github.com/input-output-hk/mantis/pull/296/commits/95304171c3083950997a7cd45cdf18bbeaffe463 
- Including memory cost in `gasCap`, important in gas cost calculation in case of EIP150 https://github.com/input-output-hk/mantis/pull/296/commits/9d02b3aa5df3ea286a4940a178a66a89ed74ff1a

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
```diff
@@ -0,0 +1,92 @@
...
+    "ChainAtoChainB_blockorder1.json",
+    "ChainAtoChainBtoChainA.json",
+    "lotsOfBranches.json",
+    "lotsOfBranchesOverrideAtTheMiddle.json",
+    "lotsOfLeafs.json",
+    "sideChainWithMoreTransactions.json",
+    "uncleBlockAtBlock3afterBlock4.json",
+    "blockChainFrontierWithLargerTDvsHomesteadBlockchain.json",
+    "blockChainFrontierWithLargerTDvsHomesteadBlockchain2.json"
+  )
```
rtkaczyk: Proposed changes:
- select individual tests (scenarios), not whole files (`Map[group-name, Set[scenario-name]`?)
- full group names rather than just file names (in case of duplicates)
- explicitly `cancel` these tests

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  // scalastyle:off
```
rtkaczyk: Why?

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
```diff
@@ -0,0 +1,143 @@
...
+    case "Frontier"   => new FrontierConfig
+    case "Homestead"  => new HomesteadConfig
+    case "FrontierToHomesteadAt5" => new FrontierToHomesteadAt5
+    case "HomesteadToEIP150At5" => new HomesteadToEIP150At5
+
+    // Some default config, test will fail or be canceled
+    case _ => new FrontierConfig
+  }
+
+  // Block rlp hex decoding, should be done during test and not parsing json as it can fail
```
rtkaczyk: So should we do something about that?

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
rtkaczyk: [EC-303](https://iohk.myjetbrains.com/youtrack/issue/EC-303)
@KonradStaniec can you please add the ticket number to the comment?

src/main/scala/io/iohk/ethereum/crypto/ECDSASignature.scala
```diff
@@ -111,6 +111,12 @@ object ECDSASignature {
     curve.getCurve.decodePoint(compEnc)
   }
 
+  def hasOnlyLastByteSet(v: ByteString): Boolean = {
```
rtkaczyk: How about `v.dropWhile(_ == 0).size == 1`? No `if`.
Does this function have anything to do `ECDSASignature`?

src/main/scala/io/iohk/ethereum/domain/Account.scala
```diff
@@ -42,6 +42,9 @@ case class Account(
   def withStorage(storageRoot: ByteString): Account =
     copy(storageRoot = storageRoot)
 
+  def clearAccount: Account =
+    copy(nonce = UInt256.Zero, storageRoot = Account.EmptyStorageRootHash, codeHash = Account.EmptyCodeHash)
```
rtkaczyk: How is it different from `Account.empty()`?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -303,17 +303,28 @@ class LedgerImpl(vm: VM, blockchain: BlockchainImpl, blockchainConfig: Blockchai
...
-        ProgramContext(stx, address,  Program(stx.tx.payload), blockHeader, world1, config)
+  private[ledger] def prepareProgramContext(stx: SignedTransaction, blockHeader: BlockHeader, worldStateProxy: InMemoryWorldStateProxy,
+                                            config: EvmConfig): PC = {
+    def prepareContext: PC = {
+      val address = worldStateProxy.createAddress(creatorAddr = stx.senderAddress)
+      val worldAfterTransfer = worldStateProxy.transfer(stx.senderAddress, address, UInt256(stx.tx.value))
+
+      worldAfterTransfer.getAccount(address) match {
+        case None => ProgramContext(stx, address,  Program(stx.tx.payload), blockHeader, worldAfterTransfer, config)
+        case Some(acc) =>
```
rtkaczyk: So either way we create a new empty account, right? Do we then need to check for its prior existence?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -303,17 +303,28 @@ class LedgerImpl(vm: VM, blockchain: BlockchainImpl, blockchainConfig: Blockchai
...
 
-  private[ledger] def prepareProgramContext(stx: SignedTransaction, blockHeader: BlockHeader, worldStateProxy: InMemoryWorldStateProxy, config: EvmConfig): PC =
-    stx.tx.receivingAddress match {
-      case None =>
-        val address = worldStateProxy.createAddress(creatorAddr = stx.senderAddress)
-        val world1 = worldStateProxy.transfer(stx.senderAddress, address, UInt256(stx.tx.value))
-        ProgramContext(stx, address,  Program(stx.tx.payload), blockHeader, world1, config)
+  private[ledger] def prepareProgramContext(stx: SignedTransaction, blockHeader: BlockHeader, worldStateProxy: InMemoryWorldStateProxy,
+                                            config: EvmConfig): PC = {
+    def prepareContext: PC = {
```
rtkaczyk: What is the point of this closure?

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
KonradStaniec: forgot to take it off after some initial work

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
KonradStaniec: Hex decoding is already moved to test case. I will remove comment as it seems it is misleading.

src/main/scala/io/iohk/ethereum/crypto/ECDSASignature.scala
KonradStaniec: cool one liner definitely an improvement over `if`. 
As for reason of putting it in `ECDSASignature` , theoretically, we could use this check in many contexts, but for now, we are using it only in the context of checking `ECDSASignature` recovery id.

src/main/scala/io/iohk/ethereum/domain/Account.scala
KonradStaniec: `Account.empty()` creates account with `balance = 0`, `clearAccount` has somewhat unintuitive behaviour that it leaves account ether balance intact, but clears other fields.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
KonradStaniec: I like to have logic in separate function if logic in it is over 5LOC. After simplification (introduced in previous comment) it won't be necessary i think

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
KonradStaniec: It is not strictly speaking empty (its ether balance is not cleared).
We can simplify this function to avoid pattern matching, as `transfer` already creates an account if one is not existent, but nevertheless, we need to `getAccount(address)` to clear it. 
We could theoretically make some special transfer function which would clear account during transfer, but I don't think it is necessary as this code path has almost 0 probability of execution.

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
```diff
@@ -0,0 +1,143 @@
...
+      }
+      worldWithAccountAndCode.saveStorage(address, updatedStorage)
+    }
+  }
+
+  // Function copied from RegularSync.scala, as ledger itself do not persis schanges in blockchain
+  // TODO We need to take forks and branches into
+  // https://iohk.myjetbrains.com/youtrack/issue/EC-303
+  @tailrec
+  final def processBlocks(blocks: Seq[Block], blockParentTd: BigInt,
```
rtkaczyk: Do we need all of this functionality? Saving best block number, total difficulty, deleting blocks?

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
```diff
@@ -0,0 +1,143 @@
...
+
+  private def getBlocks(blocks: List[BlockDef]): List[Block] = {
+    blocks.flatMap(blockDef => decodeBlock(blockDef.rlp))
+  }
+
+  def getAllBlocks: List[Block] = {
+    getBlocks(scenario.blocks)
+  }
+
+  def getInvalidBlocks: List[Block] = {
```
rtkaczyk: Is it worth creating new functions for something that is used only once and is one-liner anyway?

src/main/scala/io/iohk/ethereum/crypto/ECDSASignature.scala
rtkaczyk: But it doesn't belong in this class/object. As a new user of this class I would have no idea why I might need this method. And even if you were to explain it in scaladoc, it's still very poor design to require a 2-step interaction with this class.

The primary problem here is that the API of `ECDSASignature` is wrong. It tells us it accepts any `ByteString` for the value of `v` but it does not handle it correctly - and this has been the source of the bug in the precompiled contract. We need to change this API so that either:
- it accepts `ByteString`, but handles _any possible value_ correctly, returning an error for invalid `v`
- it only accepts `Byte` type for `v`, and this should be already handled correctly

The latter should be easier.

src/main/scala/io/iohk/ethereum/domain/Account.scala
rtkaczyk: The name of this method is misleading (think about a bank). Also the nonce should not be just set to zero but to the configured value (as with `Account.empty`).

Also are we sure we need to reset the nonce? In the event of a collision this is very likely to lead to more collisions.

src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
```diff
@@ -76,6 +90,14 @@ trait WorldStateProxy[WS <: WorldStateProxy[WS, S], S <: Storage[S]] { self: WS
...
-    updatedWorld.createAddress(creatorAddr) -> updatedWorld
+    val newAddress = updatedWorld.createAddress(creatorAddr)
+
+    updatedWorld.getAccount(newAddress) match {
+      case None =>  newAddress -> updatedWorld
+      case Some(acc@Account(_, _, _, _)) =>
+        val clearedAcc = acc.clearAccount
+        val newWorld = updatedWorld.saveAccount(newAddress, clearedAcc)
+        newAddress -> newWorld
+    }
```
rtkaczyk: I believe this method was only meant to create the address, not save the new account.

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -902,7 +906,11 @@ case object SELFDESTRUCT extends OpCode(0xff, 1, 0, _.G_selfdestruct) {
     val (refund, stack1) = state.stack.pop
     val refundAddr: Address = Address(refund)
     val gasRefund: BigInt = if (state.addressesToDelete contains state.ownAddress) 0 else state.config.feeSchedule.R_selfdestruct
-    val world = state.world.transfer(state.ownAddress, refundAddr, state.ownBalance)
+
+    val world = if (state.ownAddress == refundAddr) {
+      state.world.removeEther(state.ownAddress, state.ownBalance)
+    } else
```
rtkaczyk: Minor, style: please be consistent with the braces. Try to adhere to these guidelines: https://docs.scala-lang.org/style/

src/main/scala/io/iohk/ethereum/vm/WorldStateProxy.scala
```diff
@@ -49,12 +49,26 @@ trait WorldStateProxy[WS <: WorldStateProxy[WS, S], S <: Storage[S]] { self: WS
...
+    val credited = getAccount(to).getOrElse(getEmptyAccount).increaseBalance(value)
+    saveAccount(from, debited).saveAccount(to, credited)
+  }
+
+  /**
+    * In case of transfer to self, during selfdestruction the ether is actually destroyed
+    * see https://github.com/ethereum/wiki/wiki/Subtleties/d5d3583e1b0a53c7c49db2fa670fdd88aa7cabaf#other-operations
+    * and https://github.com/ethereum/go-ethereum/blob/ff9a8682323648266d5c73f4f4bce545d91edccb/core/state/statedb.go#L322
+    */
+  def removeEther(from: Address, value: UInt256): WS = {
```
rtkaczyk: Maybe this should be `removeAllEther` with `address` parameter only? No other usages for it, are there?

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
```diff
@@ -0,0 +1,97 @@
...
+
+
+class BlockchainSuite extends FreeSpec with Matchers with Logger {
+
+  val unsupportedNetworks = Set("Byzantium","Constantinople", "EIP158", "EIP158ToByzantiumAt5", "HomesteadToDaoAt5")
+  val supportedNetworks = Set("EIP150", "Frontier", "FrontierToHomesteadAt5", "Homestead", "HomesteadToEIP150At5")
+
+  type GroupName = String
+  type TestName = String
+  type IgnoredTestNames = Set[String]
```
rtkaczyk: What's the purpose of these types?

src/main/scala/io/iohk/ethereum/domain/Account.scala
KonradStaniec: Actually, after EIP161 it should be changed to `1`, but now `0` is correct.  (https://github.com/ethereum/cpp-ethereum/issues/4225)

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
KonradStaniec: maybe, in that case, it is a little bit too much

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
KonradStaniec: actually, they help with understanding what is this map :
` val ignoredTests: Map[GroupName, IgnoredTestNames]`
but i don't stick to them if you think it is unnecessary

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
KonradStaniec: I wanted to to be as most similar to this one from `RegularSync`, but it can definitely be simplified only to saving block and best block number

src/ets/scala/io/iohk/ethereum/ets/blockchain/scenario.scala
```diff
@@ -0,0 +1,66 @@
...
+  gasLimit: BigInt,
+  to: Option[Address],
+  value: BigInt,
+  data:ByteString,
+  r: BigInt,
+  s: BigInt,
+  v: ByteString
+)
+
+case class BlockHeaderDef(
```
ntallar: Minor comment: Could we use `BlockHeader` instead, as it is defined in the same way?

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
```diff
@@ -0,0 +1,129 @@
...
+        storage.store(key, value)
+      }
+      worldWithAccountAndCode.saveStorage(address, updatedStorage)
+    }
+  }
+
+  // TODO We need to take forks and branches into
+  // https://iohk.myjetbrains.com/youtrack/issue/EC-303
+  @tailrec
+  final def processBlocks(blocks: Seq[Block],
```
rtkaczyk: Were my comments about this function addressed? Also what about the return type?

src/ets/scala/io/iohk/ethereum/ets/blockchain/scenario.scala
rtkaczyk: We could, but I was rather insistent on not doing so. These classes implement a similar model, but in totally different contexts. The fact the implementation happen to be the same shouldn't drive the design. For reference/rationale see _bounded contexts_ principle in DDD.

src/test/scala/io/iohk/ethereum/vm/MockWorldState.scala
```diff
@@ -62,4 +62,8 @@ case class MockWorldState(
   def isAccountDead(address: Address): Boolean =
     getAccount(address).forall(_ == Account.empty())
 
+  def resetAccount(address: Address): MockWorldState = {
+    val account = getGuaranteedAccount(address).resetAccount()
+    saveAccount(address, account)
+  }
```
rtkaczyk: Minor: since it's written in terms of other methods defined in this class, it could be moved to `WorldStateProxy`, and real and mock implementations would not need to override.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -303,17 +303,19 @@ class LedgerImpl(vm: VM, blockchain: BlockchainImpl, blockchainConfig: Blockchai
...
+  private[ledger] def prepareProgramContext(stx: SignedTransaction, blockHeader: BlockHeader, worldStateProxy: InMemoryWorldStateProxy,
+                                            config: EvmConfig): PC = {
     stx.tx.receivingAddress match {
       case None =>
         val address = worldStateProxy.createAddress(creatorAddr = stx.senderAddress)
-        val world1 = worldStateProxy.transfer(stx.senderAddress, address, UInt256(stx.tx.value))
-        ProgramContext(stx, address,  Program(stx.tx.payload), blockHeader, world1, config)
-
+        val worldAfterTransfer = worldStateProxy.transfer(stx.senderAddress, address, UInt256(stx.tx.value))
+        val savedClearedAccountWorld = worldAfterTransfer.resetAccount(address)
```
rtkaczyk: This looks better! But I would go with a single method `initialiseAccount(address: Address, value: UInt256)` - that finds the balance of the possibly existing account (or zero), and adds `value` to it. Otherwise, just like `resetAccount`. The read/write vs 2 writes, is not that big of an issue since these should be cached. But the important thing is we have a single method that captures the logic, instead of duplicating it here and in `CREATE`

src/main/scala/io/iohk/ethereum/domain/Account.scala
```diff
@@ -42,6 +42,9 @@ case class Account(
   def withStorage(storageRoot: ByteString): Account =
     copy(storageRoot = storageRoot)
 
+  def resetAccount(startNonce: UInt256 = UInt256.Zero): Account =
+    copy(nonce = startNonce, storageRoot = Account.EmptyStorageRootHash, codeHash = Account.EmptyCodeHash)
```
rtkaczyk: If we really need this method then how about `resetAccountPreservingBalance`? I'm not fond of long names but I feel `reset` may still be misleading.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
rtkaczyk: Also please add a comment/scaladoc that it handles address conflicts.

src/main/scala/io/iohk/ethereum/vm/PrecompiledContracts.scala
```diff
@@ -65,15 +65,23 @@ object PrecompiledContracts {
...
+        ByteString.empty
+
     }
 
     def gas(inputDataSize: UInt256): BigInt =
       3000
+
+    private def hasOnlyLastByteSet(v: ByteString): Boolean = {
+      v.dropWhile(_ == 0).size == 1
+    }
```
rtkaczyk: Minor, I suggest we skip the braces: https://docs.scala-lang.org/style/declarations.html#body

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
KonradStaniec: Yes they were, function now only saves blocks in blockchain and current bestblock number. Both are needed, as we later need bestblock to check its state root.
What you mean by return type? It is `(Seq[NewBlock], Seq[BlockExecutionError])`

KonradStaniec:
@ntallar I think this StackOverflow is because you need to run those test with `-Xss10m` option. You can do this, adding `-J-Xss10M` line to your `/usr/local/etc/sbtopts` file for example.

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
rtkaczyk: 👍 I must have looked at an old commit.

So why does it return `Seq[NewBlock]`? Where is it used?

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
```diff
@@ -0,0 +1,76 @@
...
+
+    val blocksToProcess = getAllBlocks
+
+    val invalidBlocks = getInvalidBlocks
+
+    val newBlocks: (Seq[NewBlock], Seq[BlockExecutionError]) = processBlocks(blocksToProcess, genesisBlock.header.difficulty)
+
+    // If there is more block execution errors than expected invalidblocks, it means we rejected block which should
+    // pass validations, and the final state will not be correct
+    if(newBlocks._2.size > invalidBlocks.size)
```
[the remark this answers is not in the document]
rtkaczyk: Was this comment addressed?

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
```diff
@@ -0,0 +1,94 @@
...
+
+    val blocksToProcess = getBlocks(scenario.blocks)
+
+    val invalidBlocks = getBlocks(getInvalid)
+
+    val newBlocks: (Seq[NewBlock], Seq[BlockExecutionError]) = processBlocks(blocksToProcess)
+
+    // If there is more block execution errors than expected invalidblocks, it means we rejected block which should
+    // pass validations, and the final state will not be correct
+    if(newBlocks._2.size > invalidBlocks.size)
```
rtkaczyk: > Maybe it's not considered in the BlockchainSuite scenarios, but maybe we should also consider the case that newBlocks._2.size < invalidBlocks.size?
I'm not sure how this scenario is handled but maybe the test should fail in case this size comparison is invalid?

Was this comment from @ntallar addressed?

rtkaczyk:
Do we have unit tests for the changes in the `main` code?

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
KonradStaniec: ahh sorry @ntallar if forgot to properly address it, during comments barrage from Radek. Actually I tested it now and I think it is worth having assertion `newBlocks._2.size > invalidBlocks.size` as it helps to catch to false positives (both connected to branching and forking)

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
KonradStaniec: It is not necessary, actually only block errors can add some value.

KonradStaniec:
Actually, I did not add unit tests for changes in `main` as they are not new features but bugs covered by other tests so there would be IMO not much value added by unit testing those several rare cases. 
As for now, unit tests would only prove that there isn't a bug in the place where earlier it has been.
Of course, if you think it is worth it, I could add those tests as there is never too much testing :)

rtkaczyk:
@KonradStaniec unit tests are a must. We always add at least one test for any bug we uncover, that covers the specific edge case. Note that the fact that they may be covered by `ETS` is not enough because not all of those tests will be run for every build. Also we do try to observe the coverage statistics and get them as high as possible (they only work for unit tests).

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
```diff
@@ -59,6 +59,10 @@ class CallOpcodesSpec extends WordSpec with Matchers with PropertyChecks {
       SELFDESTRUCT
     )
 
+    val selfDestructSelfCode = assemblyFromBytes (
```
rtkaczyk: `Assembly` is meant to be a simple DSL, that enables to write EVM code that resembles it's actual assembly language and get the actual bytecode as a `ByteString`. Here we have a function that takes a `ByteString` and ultimately gives the same `ByteString` back, which is a bit pointless 😉 

`Assembly` can be improved to handle byte sequences. See the branch `fix/improvedAssemblyDSL`, and if you like the changes merge the latest commit into this branch.

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
```diff
@@ -330,6 +342,12 @@ class CallOpcodesSpec extends WordSpec with Matchers with PropertyChecks {
         call.stateOut.gasRefund shouldBe 0
       }
 
+      "destruct ether if own address equals refund address" in {
```
rtkaczyk: It's a good test :+1:
But the primary place for unit tests for opcodes other than *CALL* and CREATE is `OpCodeFunSpec`. Could we add a test case there? (Note these are property tests with generators for general cases, and tables for specific ones)

src/test/scala/io/iohk/ethereum/vm/PrecompiledContractsSpec.scala
```diff
@@ -46,6 +47,30 @@ class PrecompiledContractsSpec extends FunSuite with Matchers with PropertyCheck
     gasUsedFailedRecover shouldEqual 3000
   }
 
+  test("ECDSARECOVER_vPrefixedf0") {
```
rtkaczyk: What does it mean?

src/test/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxySpec.scala
```diff
@@ -141,12 +141,60 @@ class InMemoryWorldStateProxySpec extends FlatSpec with Matchers {
...
+    acc2.nonce shouldEqual UInt256.Zero
+    acc2.balance shouldEqual balanceAfterTransfer
+    acc2.codeHash shouldEqual Account.EmptyCodeHash
+    acc2.storageRoot shouldEqual Account.EmptyStorageRootHash
+  }
+
+  "InMemoryWorldStateProxy" should "remove all ather from existing account" in new TestSetup {
+    val startValue = 100
+    val transferValue = 50
+    val balanceAfterTransfer = startValue + transferValue
```
rtkaczyk: Unused vals

src/test/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxySpec.scala
```diff
@@ -141,12 +141,60 @@ class InMemoryWorldStateProxySpec extends FlatSpec with Matchers {
...
+    acc1.balance shouldEqual UInt256.Zero
+  }
+
   trait TestSetup extends EphemBlockchainTestSetup {
 
     val worldState = BlockchainImpl(storagesInstance.storages).getWorldStateProxy(-1, UInt256.Zero, None)
 
     val address1 = Address(0x123456)
     val address2 = Address(0xabcdef)
+    val address3 = Address(0x789101)
```
rtkaczyk: Unused

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
```diff
@@ -349,13 +367,36 @@ class CallOpcodesSpec extends WordSpec with Matchers with PropertyChecks {
         CallResult(op = CALL, context = context, gas = UInt256.MaxValue / 2)
       }
 
+      def callVarMemCost(config: EvmConfig): CallResult = {
+        val gasFailingBeforeEIP150Fix = 141072
```
rtkaczyk: How did you arrive at this number? Maybe you could express it with function calls that calculate it?

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
```diff
@@ -0,0 +1,97 @@
...
+    if(newBlocksErrors.size > invalidBlocks.size)
+      newBlocksErrors.foreach(err => log.info(err.toString))
+
+    newBlocksErrors.size shouldEqual invalidBlocks.size
+
+    val lastBlock = getBestBlock()
+
+    val expectedWorldStateHash = getFinalWorld().stateRootHash
+
+    lastBlock shouldBe defined
```
rtkaczyk: Should we add assertions on the post-state? While comparing the root hash should be enough to catch errors, the post state could be used to provide better information on what wrong in the event of a failure.

src/test/scala/io/iohk/ethereum/vm/PrecompiledContractsSpec.scala
KonradStaniec: Hmm I have just copied it from `blockchainsuite`, I will rename it as someone without context may not decipher it

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
KonradStaniec: I have arrived at it by trial and error, as there is quite a narrow margin that `gasCap(state, gas, gExtra)` fails to execute because of OOG, but `gasCap(state, gas, gExtra + memCost)` executes normally.
Maybe a comment would be enough?

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
KonradStaniec: hmm I think checking the whole state helped me once or twice, so I can add it as it will maybe help debugging future multi-chain test cases

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
rtkaczyk: Comment is fine

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
```diff
@@ -0,0 +1,100 @@
...
+    newBlocksErrors.size shouldEqual invalidBlocks.size
+
+    val lastBlock = getBestBlock()
+
+    val expectedWorldStateHash = finalWorld.stateRootHash
+
+    lastBlock shouldBe defined
+
+    val expectedState = getExpectedState()
+    val resultState = getResultState()
```
rtkaczyk: Account class does not have an address, maybe this should be a `List[Address -> Option[Account]]`?

src/test/scala/io/iohk/ethereum/vm/OpCodeFunSpec.scala
```diff
@@ -701,7 +701,25 @@ class OpCodeFunSpec extends FunSuite with OpCodeTesting with Matchers with Prope
...
+          .halt
+        stateOut shouldEqual expectedState
+      }
+    }
+
+    // Refunding gas to self should destroy all account ether
+    val stateGenSelf = getProgramStateGen(
+      stackGen = getStackGen(elems = 2, uint256Gen = getUInt256Gen(ownerAddr.toUInt256, ownerAddr.toUInt256))
+    )
+    forAll(stateGenSelf) { stateIn =>
```
rtkaczyk: No need to run 10 additional tests (that's what `forAll` with a generator will do) to test this edge case. Anyway this test wasn't great from the start. See the branch `fix/selfdestructUnitTest`, and like last time, merge if you like it.

PR #297, opened 2017-08-25
[EC-87] BlockBroadcast propagation algorithm

ntallar:
## Description

Modifies the block propagation algorithm from sending `NewBlock` messages to all peers, to sending `NewBlockHashes` messages to all peers but sending `NewBlock` messages to only a randomly selected `sqrt(|peers|)` as both Parity and Geth propagate blocks that way.

### DIFF
--- a/src/test/scala/io/iohk/ethereum/blockchain/sync/BlockBroadcastSpec.scala
+++ b/src/test/scala/io/iohk/ethereum/blockchain/sync/BlockBroadcastSpec.scala
@@ -68,29 +68,21 @@ class BlockBroadcastSpec extends FlatSpec with Matchers  {
     val peer4 = Peer(new InetSocketAddress("127.0.0.1", 0), peer4Probe.ref, false)
 
     //when
-    val peersWithInfo = Map(
-      peer -> initialPeerInfo,
-      peer2 -> initialPeerInfo,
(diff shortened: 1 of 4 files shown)


src/test/scala/io/iohk/ethereum/blockchain/sync/BlockBroadcastSpec.scala
```diff
@@ -23,30 +23,74 @@ class BlockBroadcastSpec extends FlatSpec with Matchers  {
...
+    //when
+    val peersWithInfo = Map(
+      peer -> initialPeerInfo,
+      peer2 -> initialPeerInfo,
+      peer3 -> initialPeerInfo,
+      peer4 -> initialPeerInfo)
+    val blockBroadcastFirstTwoPeers = new BlockBroadcast {
+      override val etcPeerManager: ActorRef = etcPeerManagerProbe.ref
+
+      override def obtainRandomPeerSubset(peers: Set[Peer]): Set[Peer] = Set(peer, peer2)
```
rtkaczyk: Wouldn't it be better to test the actual implementations? Just check that there are 2 receiving peers within a subset of `peersWithInfo`. `etcPeerManagerProbe.expectMsgPF` may come handy.

PR #298, opened 2017-08-29
[EC-298] Node encoding and hash caching

ntallar:
## Description

Based on the reported issue of Merkle Patricia Trie use being slow, we started to some some benchmarks, which resulted in finding that getting the `Node.capped` was one of the non DB access functions that took more time (almost 20% of the total execution time, with DB access taking 70% of it).

## Fix

A simple improvement was made to node encoding and hashing, based on one found on the [Geth client](https://github.com/ethereumproject/go-ethereum/blob/master/trie/node.go#L159), which is not having to calculate the hash or the encoding of a node if it's already known.
It seems the main effort that's being saved is the [capping of all the nodes to be removed after a `put`](https://github.com/input-output-hk/mantis/blob/phase/daedalus/src/main/scala/io/iohk/ethereum/mpt/MerklePatriciaTrie.scala#L86). This is due to the fact that many of the nodes to be removed were obtained from the DB, so the caching will optimize capping them.

## Benchmarks

The performance test uploaded on the report issue was run both with the LevelDB database and in-memory. The following performance improvement was obtained on the `EphemDataSource` comparing using the caching improvement and not:

mptcachingimprovement


src/it/scala/io/iohk/ethereum/mpt/MerklePatriciaTreeIntegrationSuite.scala
```diff
@@ -147,6 +147,29 @@ class MerklePatriciaTreeIntegrationSuite extends FunSuite
       else assert(rootHash.take(4) == "da8a" && rootHash.drop(rootHash.length - 4) == "0ca4")
     }
   }
+
+  ignore("MPT benchmark") {
```
AlanVerbner: I'm not about sure the `ignore`. I would prefer creating a new sbt task `benchmark` (or `longrunning`) and don't run them on CI. There are other tests that should be included in this category too.

@mcsherrylabs what do you think?

src/main/scala/io/iohk/ethereum/mpt/Node.scala
```diff
@@ -25,11 +31,26 @@ object Node {
   val hashFn: (Array[Byte]) => Array[Byte] = (input: Array[Byte]) => crypto.kec256(input)
 }
 
-case class LeafNode(key: ByteString, value: ByteString) extends MptNode
+case class LeafNode(key: ByteString, value: ByteString,
+                    cachedHash: Option[Array[Byte]] = None, cachedRlpEncoded: Option[Array[Byte]] = None) extends MptNode {
+  def withCachedHash(cachedHash: Array[Byte]): MptNode = copy(cachedHash = Some(cachedHash))
```
AlanVerbner: This seems to be repeated on each Node type...

ntallar: Yes, however it seems that repeated code can't be easily removed as the `copy` function called is different for all the node case classes :disappointed:

LukasGasior1: macros could help, but by no means I suggest to use them

src/it/scala/io/iohk/ethereum/mpt/MerklePatriciaTreeIntegrationSuite.scala
mcsherrylabs: If this is the third (or more than third) 'benchmark' task, then yes, we should move them all out to another grouping. I further assume moving this won't lower the unit test coverage.

circle.yml
```diff
@@ -29,6 +29,8 @@ test:
         fi
       fi
 
+    # As the benchmark tests don't test functionality and should be manually ran, having them compile is enough
+    - sbt benchmark:compile
```
AlanVerbner: :+1:

PR #299, opened 2017-08-29
[EC-250] eth_getWork returns error if called while syncing

ntallar:
## Description

Changes `eth_getWork` implementation so that no block is generated and returned if the client is syncing (as the mined block won't be accepted in RegularSync), as is done in [Parity](https://github.com/paritytech/parity/blob/master/rpc/src/v1/impls/eth.rs#L534).
The `EthService.lastActive` time is not updated in that case, so the `eth_mining` response is not affected by this `eth_work` call.

*Note*: The returned error code is `-32001` as it also is it in Parity.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
@@ -433,26 +433,38 @@ class EthService(
     }
   }
 
+  /**
+    * Implements the eth_getWork method, which returns the data needed for mining: header hash, DAG seed and
+    * target difficulty.
+    * In case the client is mining a NoWork error is returned (and the eth_mining response is not affected)
+    *
+    * @param req
+    * @return mining data or an error
+    */
   def getWork(req: GetWorkRequest): ServiceResponse[GetWorkResponse] = {
-    reportActive()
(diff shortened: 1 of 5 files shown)


src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -498,22 +510,30 @@ class EthService(
...
    Right(SyncingResponse(maybeSyncStatus))
  }
 
+  /**
+    * Returns whether the client is syncing
+    * The node is assumed to be syncing only if there's any block that other peers have and this peer doesn't
+    *
+    * @return if the node is syncing
+    */
+  private def isSyncing: Boolean = {
```
AlanVerbner: So...this will only mine if we have all the blocks the latest block we know (for example defined in `io.iohk.ethereum.network.EtcPeerManagerActor#handleReceivedMessage`), right?

Could it happen that another dishonest peer send us a really big block to prevent us from mining?

ntallar: It seems that could be the case with the current implementation (Parity seems to have ways of preventing that), changing the state of the PR to WIP to look into handling this.

PR #300, opened 2017-09-04
[EC-296] JSON-RPC HTTPS support

ntallar:
## Description

Enables using HTTPS for the JSON-RPC requests.

## Example usage

After creating the certificate and having started a Mantis client, the certificate can be used for requests by exporting the certificate as `pem`:
```
keytool -export -v \
  -alias mantisCA \
  -file mantisCA.pem \
  -keypass:env PW \
  -storepass:env PW \
  -keystore mantisCA.jks \
  -rfc
```

Then requests can be done using the obtained certificate and curl with:
```
curl --cacert mantisCA.pem -H 'Content-Type: application/json' -X POST --data '{"jsonrpc":"2.0","method":"web3_clientVersion","params":[],"id":67}' https://127.0.0.1:8546
```


rtkaczyk:
The code looks good overall, but if one of the path settings is missing or invalid, there is no error message logged. Probably the client should not even start. Could we add such validation?

ntallar:
@rtkaczyk - I just added more validations of the path settings, with the current implementation any error detected is logged and the HTTPS server is not started, as in:
```
10:52:14 Cannot start JSON HTTPS RPC server due to: Certificate keystore path and password file configuration required
```
Currently the same is done when detecting an error and using HTTP, do you think we should maybe change that in both cases our client doesn't start?

rtkaczyk:
@ntallar Not sure.. I think I'd prefer if the client failed to start if there's _anything_ wrong with the config. Maybe that should be a separate task? As for this one it's good enough for me.

AlanVerbner:
I also think starting the client after an error is not a good idea. For example, today we had a wrong value in rpc port and the server was starting and just showing a line with the error that was hidden by other messages (like "block received") ones. I would do this in another task tho.

ntallar:
@rtkaczyk @AlanVerbner Just created task [EC-304](https://iohk.myjetbrains.com/youtrack/issue/EC-304) to look into having our client fail when there's anything wrong with the configuration

src/universal/HTTPS-README.md
```diff
@@ -0,0 +1,26 @@
+For using JSON-RPC through HTTPS a certificate has to be created. For this first a password has to be created:
```
AlanVerbner: Kudos for the explanation :heart: . That being said, I would put this within the wiki and a link to it in `src/universal/conf/network.conf`

PR #301, opened 2017-09-07
[EC-302] Adding maximum gasLimit validation

KonradStaniec:
Validating that block header gas limit is lower than 2**63-1

### DIFF
--- a/src/main/scala/io/iohk/ethereum/utils/Config.scala
+++ b/src/main/scala/io/iohk/ethereum/utils/Config.scala
@@ -282,7 +282,7 @@ object BlockchainConfig {
 
       override val monetaryPolicyConfig = MonetaryPolicyConfig(blockchainConfig.getConfig("monetary-policy"))
 
-      override val eip106BlockNumber: BigInt = Try(blockchainConfig.getString("eip106-block-number")).map(BigInt(_)).getOrElse(Long.MaxValue)
+      override val eip106BlockNumber: BigInt = BigInt(blockchainConfig.getString("eip106-block-number"))
     }
   }
 }
--- a/src/main/scala/io/iohk/ethereum/utils/Config.scala
+++ b/src/main/scala/io/iohk/ethereum/utils/Config.scala
@@ -239,6 +239,7 @@ trait BlockchainConfig {
   val eip150BlockNumber: BigInt
   val eip155BlockNumber: BigInt
   val eip160BlockNumber: BigInt
+  val eip106BlockNumber: BigInt
   val difficultyBombPauseBlockNumber: BigInt
   val difficultyBombContinueBlockNumber: BigInt
 
@@ -280,6 +281,8 @@ object BlockchainConfig {
       }
 
       override val monetaryPolicyConfig = MonetaryPolicyConfig(blockchainConfig.getConfig("monetary-policy"))
+
+      override val eip106BlockNumber: BigInt = Try(blockchainConfig.getString("eip106-block-number")).map(BigInt(_)).getOrElse(Long.MaxValue)
     }
   }
 }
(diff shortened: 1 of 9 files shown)


rtkaczyk:
Could we add a unit test for this?

src/main/scala/io/iohk/ethereum/validators/BlockHeaderValidator.scala
```diff
@@ -101,15 +102,21 @@ class BlockHeaderValidatorImpl(blockchainConfig: BlockchainConfig) extends Block
...
     * @param blockHeader BlockHeader to validate.
     * @param parentHeader BlockHeader of the parent of the block to validate.
     * @return BlockHeader if valid, an [[HeaderGasLimitError]] otherwise
     */
   private def validateGasLimit(blockHeader: BlockHeader, parentHeader: BlockHeader): Either[BlockHeaderError, BlockHeader] = {
-    val gasLimitDiff = (blockHeader.gasLimit - parentHeader.gasLimit).abs
-    val gasLimitDiffLimit = parentHeader.gasLimit / GasLimitBoundDivisor
-    if(gasLimitDiff < gasLimitDiffLimit && blockHeader.gasLimit >= MinGasLimit) Right(blockHeader)
-    else Left(HeaderGasLimitError)
+    if(blockHeader.gasLimit > MaxGasLimit)
```
rtkaczyk: style: space after `if`

src/main/scala/io/iohk/ethereum/validators/BlockHeaderValidator.scala
```diff
@@ -101,15 +102,21 @@ class BlockHeaderValidatorImpl(blockchainConfig: BlockchainConfig) extends Block
...
+      Left(HeaderGasLimitError)
+    else {
+      val gasLimitDiff = (blockHeader.gasLimit - parentHeader.gasLimit).abs
+      val gasLimitDiffLimit = parentHeader.gasLimit / GasLimitBoundDivisor
+      if(gasLimitDiff < gasLimitDiffLimit && blockHeader.gasLimit >= MinGasLimit) Right(blockHeader)
```
rtkaczyk: style: space after `if`

rtkaczyk: It would be more concise to just extend this conditional with `MaxGasLimit`

KonradStaniec: I did not want to calculate unnecessary `gasLimitDiff` and `gasLimitDiffLimit`, but this is probably such small thing that for me both versions are ok

ntallar:
Code LGTM!
Do you know if Geth/Parity implement this EIP or if it is implemented in Ethereum Classic? I know it's needed for passing the tests but I wasn't able to find any information on when it was accepted and included into ETH/ETC.

KonradStaniec:
It is implemented definitely in geth and pytheruem.
https://github.com/ethereum/pyethereum/blob/8d69d808096fb599b8b7b6716c1b3e8df6d4a307/ethereum/blocks.py#L449
https://github.com/ethereum/go-ethereum/blob/27a5622e995d762683dd1a79423d83fcf3e62ccf/consensus/ethash/consensus.go#L247

Also under ticket ( https://github.com/ethereum/EIPs/issues/106 ) it seems that java and c++ client also have it.

KonradStaniec:
After research added a possibility to configure eip106 block number similar way to parity. If the `eip106-block-number` is not defined, it will have default value `Long.maxValue` so basically it will be unused.

src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -280,6 +281,8 @@ object BlockchainConfig {
       }
 
       override val monetaryPolicyConfig = MonetaryPolicyConfig(blockchainConfig.getConfig("monetary-policy"))
+
+      override val eip106BlockNumber: BigInt = Try(blockchainConfig.getString("eip106-block-number")).map(BigInt(_)).getOrElse(Long.MaxValue)
```
rtkaczyk: I'd rather avoid optional settings in the config. Can we make this required and use high values for disabling? Just like other clients do.

KonradStaniec: Actually parity uses optional values https://github.com/paritytech/parity/blob/246b5282e546333c086b2e725de024fdfab0b8d9/json/src/spec/ethash.rs#L101, but ok that way we will be consistent with our other configuration values.

rtkaczyk:
Please copy the new setting (commented out) with doc to `src/universal/conf/blockchain.conf`. Those files serve as config documentation for the users.

PR #302, opened 2017-09-12
[EC-128][EC-244] Handle NewBlock messages

ntallar:
## Description

Includes handling of `NewBlock` messages so that they are inserted into the blockchain or the client restarts syncing if necessary. This fixes the [issue](https://iohk.myjetbrains.com/youtrack/issue/EC-244) that the Mantis client had on [etcstats](http://etcstats.net/) where the Mantis node was usually out-dated in comparison with the other nodes.

**Note:** Due to the refactoring to be done in [EC-303](https://iohk.myjetbrains.com/youtrack/issue/EC-303) the handling of the `NewBlockHashes` message was not included in this PR.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -293,7 +293,7 @@ mantis {
     target-block-offset = 500
 
     # How often to query peers for new blocks after the top of the chain has been reached
-    check-for-new-block-interval = 1.seconds
+    check-for-new-block-interval = 10.seconds
 
     # size of the list that keeps track of peers that are failing to provide us with mpt node
     # we switch them to download only blockchain elements
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
@@ -7,6 +7,9 @@ import io.iohk.ethereum.domain._
 import io.iohk.ethereum.ledger.{BlockExecutionError, Ledger}
 import io.iohk.ethereum.network.{EtcPeerManagerActor, Peer}
 import io.iohk.ethereum.network.EtcPeerManagerActor.PeerInfo
+import io.iohk.ethereum.network.PeerEventBusActor.PeerEvent.MessageFromPeer
+import io.iohk.ethereum.network.PeerEventBusActor.{PeerSelector, Subscribe}
+import io.iohk.ethereum.network.PeerEventBusActor.SubscriptionClassifier.MessageClassifier
 import io.iohk.ethereum.network.p2p.messages.CommonMessages.NewBlock
 import io.iohk.ethereum.network.p2p.messages.PV62._
 import io.iohk.ethereum.transactions.PendingTransactionsManager
@@ -37,8 +40,11 @@ class RegularSync(
   private var headersQueue: Seq[BlockHeader] = Nil
   private var waitingForActor: Option[ActorRef] = None
   private var resolvingBranches: Boolean = false
+  private var resumeRegularSyncTimeout: Option[Cancellable] = None
 
   scheduler.schedule(printStatusInterval, printStatusInterval, self, PrintStatus)
+
+  peerEventBus ! Subscribe(MessageClassifier(Set(NewBlock.code), PeerSelector.AllPeers))
 
   def handleCommonMessages: Receive = handlePeerListMessages orElse handleBlacklistMessages
 
@@ -52,10 +58,56 @@ class RegularSync(
       askForHeaders()
   }
 
-  def running: Receive = handleCommonMessages orElse {
-    case ResumeRegularSync =>
-      askForHeaders()
+  private def resumeRegularSync(): Unit = {
+    resumeRegularSyncTimeout.foreach(_.cancel)
+    resumeRegularSyncTimeout = None
+    askForHeaders()
+  }
 
+  /**
+    * Handles broadcasted blocks, processing them and inserting them to the blockchain if necessary
+    * FIXME: Current implementation is naive in that NewBlock msgs that can't be immediately processed are discarded
+    *        and requested again to the peers
+    */
+  def handleBroadcastedBlockMessages: Receive = {
+    case MessageFromPeer(NewBlock(newBlock, _), peerId) =>
+      //we allow inclusion of new block only if we are not syncing / reorganising chain
+      if (headersQueue.isEmpty && waitingForActor.isEmpty) {
+        val currentBestBlock = appStateStorage.getBestBlockNumber()
+        val maybeBlockParentTd = blockchain.getBlockHeaderByHash(newBlock.header.parentHash)
+          .flatMap(b => blockchain.getTotalDifficultyByHash(b.hash))
+        val maybeCurrentBlockTd = blockchain.getTotalDifficultyByNumber(currentBestBlock)
+
+        (maybeBlockParentTd, maybeCurrentBlockTd) match {
+          case (_, None) =>
+            //Something went wrong, we don't have the total difficulty of the latest block
+            //TODO: Investigate if we can recover from this error [EC-165]
+            throw new IllegalStateException(s"No total difficulty for the latest block with number $currentBestBlock")
+          case (None, Some(_)) =>
+            //The received block should be imported, however we don't have it's parent.
+            //This could be caused by our client not being up-to-date or a fork in the blockchain
+            //RegularSync is resumed which will lead to the needed reorganization
+            log.debug("Resumed regular sync due to receiving NewBlock and not having it's parent")
+            resumeRegularSync()
+          case (Some(parentTd), Some(currentTd)) if currentTd >= parentTd + newBlock.header.difficulty =>
+            //The received block is not imported as it's total difficulty is too low, do nothing
+            log.debug("Not inserting NewBlock due to having too low total difficulty")
+          case (Some(parentTd), Some(_)) =>
+            val blockProcessResult = processBlock(newBlock, parentTd)
+            blockProcessResult match {
+              case Right(_) =>
+                //Delete old blocks no longer needed
+                if(newBlock.header.number < currentBestBlock)
+                  ((newBlock.header.number + 1) to currentBestBlock).foreach(blockchain.removeBlock)
+                log.debug(s"Added new block ${newBlock.header.number} received from $peerId")
+              case Left(err) =>
+                blacklist(peerId, blacklistDuration, err.toString)
+            }
+        }
+      }
+  }
+
+  def handleResponseToRequest: Receive = {
     case ResponseReceived(peer: Peer, BlockHeaders(headers), timeTaken) =>
       log.info("Received {} block headers in {} ms", headers.size, timeTaken)
       waitingForActor = None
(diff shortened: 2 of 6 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -52,10 +58,56 @@ class RegularSync(
...
+        val currentBestBlock = appStateStorage.getBestBlockNumber()
+        val maybeBlockParentTd = blockchain.getBlockHeaderByHash(newBlock.header.parentHash)
+          .flatMap(b => blockchain.getTotalDifficultyByHash(b.hash))
+        val maybeCurrentBlockTd = blockchain.getTotalDifficultyByNumber(currentBestBlock)
+
+        (maybeBlockParentTd, maybeCurrentBlockTd) match {
+          case (_, None) =>
+            //Something went wrong, we don't have the total difficulty of the latest block
+            //TODO: Investigate if we can recover from this error [EC-165]
+            throw new IllegalStateException(s"No total difficulty for the latest block with number $currentBestBlock")
```
LukasGasior1: throw here will just restart the actor, do we need that?

src/main/resources/application.conf
```diff
@@ -293,7 +293,7 @@ mantis {
     target-block-offset = 500
 
     # How often to query peers for new blocks after the top of the chain has been reached
-    check-for-new-block-interval = 1.seconds
+    check-for-new-block-interval = 10.seconds
```
KonradStaniec: Why this ten fold increase? (i don't tell it is bad just want some justification)

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -52,10 +58,56 @@ class RegularSync(
...
+      if (headersQueue.isEmpty && waitingForActor.isEmpty) {
+        val currentBestBlock = appStateStorage.getBestBlockNumber()
+        val maybeBlockParentTd = blockchain.getBlockHeaderByHash(newBlock.header.parentHash)
+          .flatMap(b => blockchain.getTotalDifficultyByHash(b.hash))
+        val maybeCurrentBlockTd = blockchain.getTotalDifficultyByNumber(currentBestBlock)
+
+        (maybeBlockParentTd, maybeCurrentBlockTd) match {
+          case (_, None) =>
+            //Something went wrong, we don't have the total difficulty of the latest block
+            //TODO: Investigate if we can recover from this error [EC-165]
```
KonradStaniec: We have exactly same TODO error in line `302`, is this really the same case ? As i understand here we don't have the difficulty of currently best block saved in storage (is this even really possible?), and in line `302` we don't have the difficulty of the parent of some incoming block in the queue.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -52,10 +58,56 @@ class RegularSync(
...
+    resumeRegularSyncTimeout = None
+    askForHeaders()
+  }
 
+  /**
+    * Handles broadcasted blocks, processing them and inserting them to the blockchain if necessary
+    * FIXME: Current implementation is naive in that NewBlock msgs that can't be immediately processed are discarded
+    *        and requested again to the peers
+    */
+  def handleBroadcastedBlockMessages: Receive = {
```
KonradStaniec: Minor: I found this function with all its comments, cases, `_` marks really hard to decipher what is the intent here. It maybe just my opinion as I am still new to this regular sync, so feel free ignore this comment. (maybe as the bigger refactor of regular sync is coming it not have much sense)

src/main/resources/application.conf
ntallar: I should probably add a comment on this to the code, but I changed to reflect the [value set in Geth](https://github.com/ethereumproject/go-ethereum/blob/master/eth/sync.go#L137). This change also makes sense as we are processing new blocks in two different ways (handling NewBlock messages and by requesting them with `check-for-new-block-interval` frequency) instead of only requesting them.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
ntallar: > We have exactly same TODO error in line 302, is this really the same case ?

You are right, it's not exactly the same case, however it might be the case that both can be handled in a similar way, [as done in FastSync](https://github.com/input-output-hk/mantis/blob/phase/daedalus/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala#L335). I'll add a comment on task EC-165 to split it into two if needed.

> is this even really possible?

AFAIK it shouldn't be, it should only be caused by data corruption or a bug on our client.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
ntallar: Maybe we change that the actor is not restarted in this case or just log the error? I decided to throw the exception as it was also handled this way when [receiving a block and not having it's parent total difficulty](https://github.com/input-output-hk/mantis/blob/phase/daedalus/src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala#L243), as discussed in https://github.com/input-output-hk/mantis/pull/148#discussion_r112447736.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -52,10 +58,56 @@ class RegularSync(
...
+    case MessageFromPeer(NewBlock(newBlock, _), peerId) =>
+      //we allow inclusion of new block only if we are not syncing / reorganising chain
```
rtkaczyk: What exactly does it mean? We asked for a blocks and waiting for response, right? Any other circumstances? What would happen if we ignored that? 

I'm asking because I'm still not sure as to what logic exactly should be implemented in [Ledger.importBlock](https://github.com/input-output-hk/mantis/pull/305/files#diff-1b4e47475b368c25eb5141199a4289d0R34). Ideally we should be able to just throw any blocks at it without a care in the world.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: I have a similar impression. Maybe at least adding some vertical spacing would improve readability?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: I'd say we should just crash. Any rationale for restarting the actor?

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -123,6 +126,8 @@ trait Blockchain {
 
   def removeBlock(hash: ByteString): Unit
 
+  def removeBlock(blockNumber: BigInt): Unit = getHashByBlockNumber(blockNumber).foreach(removeBlock)
```
rtkaczyk: I don't think this removes the entry from `blockNumberMappingStorage`. Should it?

src/test/scala/io/iohk/ethereum/blockchain/sync/BlockBroadcastSpec.scala
```diff
@@ -15,42 +15,60 @@ import org.scalatest.{FlatSpec, Matchers}
...
     //then
-    etcPeerManagerProbe.expectMsg(EtcPeerManagerActor.SendMessage(secondBlock, peer.id))
-    etcPeerManagerProbe.expectMsg(EtcPeerManagerActor.SendMessage(secondBlockNewHashes, peer.id))
+    etcPeerManagerProbe.expectMsg(EtcPeerManagerActor.SendMessage(newBlock, peer.id))
+    etcPeerManagerProbe.expectMsg(EtcPeerManagerActor.SendMessage(newBlockNewHashes, peer.id))
     etcPeerManagerProbe.expectNoMsg()
   }
 
-  it should "send a new block only when it is not known by the peer (known by comparing max block number)" in new TestSetup {
+  it should "not send a new block when it is not known by the peer (known by comparing total difficulties)" in new TestSetup {
```
rtkaczyk: `"send a new block when it is not known by the peer (known by comparing total difficulties)"`
`"not send a new block when it is not known by the peer (known by comparing total difficulties)"`

Having a cake and eating it too? 😉

src/test/scala/io/iohk/ethereum/blockchain/sync/BlockBroadcastSpec.scala
```diff
@@ -15,42 +15,60 @@ import org.scalatest.{FlatSpec, Matchers}
...
+    //when
+    blockBroadcast.broadcastBlock(newBlock, Map(peer -> initialPeerInfo))
+
+    //then
+    etcPeerManagerProbe.expectMsg(EtcPeerManagerActor.SendMessage(newBlock, peer.id))
+    etcPeerManagerProbe.expectMsg(EtcPeerManagerActor.SendMessage(newBlockNewHashes, peer.id))
+    etcPeerManagerProbe.expectNoMsg()
+  }
+
+  it should "not send a new block only when it is not known by the peer (known by comparing max block number)" in new TestSetup {
```
rtkaczyk: Similar as above. I don't understand the intent of these tests.

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
ntallar: You are right! :+1:
I will add the removal of that entry to `def removeBlock(hash: ByteString): Unit` implementation.

src/test/scala/io/iohk/ethereum/blockchain/sync/BlockBroadcastSpec.scala
ntallar: You are right, I missed this when splitting the test `"send a new block only when it is not known by the peer (known by comparing total difficulties)"` into one for when the block is broadcasted and one when it's not.
This was changed due to changing `BlockBroadcast.broadcastBlock` receiving only a `NewBlock` instead of a `Seq[NewBlock]`.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
ntallar: I also think that ideally we should do so. However, due to the refactoring being done next, I would leave the current implementation as is, as removing this condition could have unexpected results as it could cause our client to ask for headers, causing it to possibly have two pending requests at once (if any `resumeRegularSync()` is called).

>What exactly does it mean? We asked for some blocks and we're waiting for response, right? Any other circumstances? What would happen if we ignored that?

That's right, IMO this only includes when the node is syncing and when it periodically asks it's peers for the latest blocks (due to [resumeRegularSyncTimeout](https://github.com/input-output-hk/mantis/pull/302/files#diff-65b863fb327f34212f7a48e87ccdec13R43)).

rtkaczyk: Sure, no changes required here, I was just asking in relation to my task.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -52,10 +59,65 @@ class RegularSync(
...
+        val maybeBlockParentTd = blockchain.getBlockHeaderByHash(newBlock.header.parentHash)
+          .flatMap(b => blockchain.getTotalDifficultyByHash(b.hash))
+        val maybeCurrentBlockTd = blockchain.getTotalDifficultyByNumber(currentBestBlock)
+
+        (maybeBlockParentTd, maybeCurrentBlockTd) match {
+          case (_, None) =>
+            //Something went wrong, we don't have the total difficulty of the latest block
+            //TODO: Investigate if we can recover from this error [EC-165]
+            log.error(s"No total difficulty for the latest block with number $currentBestBlock")
+            context stop self
```
LukasGasior1: maybe we should close the whole app then? is there a point of having the process running when it's no longer syncing?

PR #305, opened 2017-09-14
[EC-303] RegularSync refactor

rtkaczyk:
The aim of this PR is to provide is to remove the logic of importing blocks from `RegularSync` and have single entry point in the `Ledger`, thus providing better decoupling between components. This single entry point - `importBlock` should handle chain re-org and be suitable for use with ETS blockchain tests: #296 

TODOs:

- [ ] correctly handle branch switches wrt pending transactions and ommers
- [ ] ignore blocks that are too far behind
- [ ] delete/prune old stale blocks
- [ ] integrate with blockchain tests
- [ ] reorganize packages (`blockchain.sync`,  `domain.Blockchain`)
- [ ] try to reduce dependencies (`Blockchain`)

So far I have used this code to successfully sync the first 150k blocks.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/NewRegularSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/NewRegularSync.scala
@@ -4,7 +4,7 @@ import akka.actor._
 import io.iohk.ethereum.blockchain.sync.PeerRequestHandler.ResponseReceived
 import io.iohk.ethereum.db.storage.AppStateStorage
 import io.iohk.ethereum.domain._
-import io.iohk.ethereum.ledger.{BlockExecutionError, Ledger}
+import io.iohk.ethereum.ledger._
 import io.iohk.ethereum.network.{EtcPeerManagerActor, Peer}
 import io.iohk.ethereum.network.EtcPeerManagerActor.PeerInfo
 import io.iohk.ethereum.network.p2p.messages.CommonMessages.NewBlock
@@ -18,20 +18,21 @@ import org.spongycastle.util.encoders.Hex
 import scala.annotation.tailrec
 import scala.concurrent.ExecutionContext.Implicits.global
 
-class RegularSync(
-    val appStateStorage: AppStateStorage,
-    val blockchain: Blockchain,
-    val validators: Validators,
-    val etcPeerManager: ActorRef,
-    val peerEventBus: ActorRef,
-    val ommersPool: ActorRef,
-    val pendingTransactionsManager: ActorRef,
-    val ledger: Ledger,
-    val syncConfig: SyncConfig,
-    implicit val scheduler: Scheduler)
+class NewRegularSync(
+  val appStateStorage: AppStateStorage,
+  // FIXME: can we make indepent of Blockchain?
+  val blockchain: Blockchain,
+  val validators: Validators,
+  val etcPeerManager: ActorRef,
+  val peerEventBus: ActorRef,
+  val ommersPool: ActorRef,
+  val pendingTransactionsManager: ActorRef,
+  val ledger: Ledger,
+  val syncConfig: SyncConfig,
+  implicit val scheduler: Scheduler)
   extends Actor with ActorLogging with PeerListSupport with BlacklistSupport with SyncBlocksValidator with BlockBroadcast {
 
-  import RegularSync._
+  import NewRegularSync._
   import syncConfig._
 
   private var headersQueue: Seq[BlockHeader] = Nil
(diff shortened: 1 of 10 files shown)


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -15,26 +15,122 @@ import scala.annotation.tailrec
...
-    validators: Validators): Either[BlockExecutionError, Seq[Receipt]] = {
+  def importBlock(block: Block): BlockImportResult = {
+    // TODO: ignore blocks that are too far behind the head
+    if (blockchain.getBlockByHash(block.header.hash).isDefined)
+      ImportFailedBlockExists
+
+    else
+      blockchain.getBlockByHash(block.header.parentHash) match {
+        case None =>
+          ImportFailedMissingParent
```
rtkaczyk: Should we accept orphaned blocks and keep them in a queue of some sort for later execution? It looks to me like other clients do something of such sort (that's as specific as I can get ATM).

src/main/scala/io/iohk/ethereum/blockchain/sync/NewRegularSync.scala
```diff
@@ -0,0 +1,311 @@
...
+    Props(new NewRegularSync(appStateStorage, blockchain, validators, etcPeerManager, peerEventBus, ommersPool, pendingTransactionsManager,
+      ledger, syncConfig, scheduler))
+
+  private case object ResumeRegularSync
+  private case class ResolveBranch(peer: ActorRef)
+  private case object PrintStatus
+
+  case object Start
+  case class MinedBlock(block: Block)
+}
```
LukasGasior1: is it the original RegularSync with your changes? why do we need a copy (and not modify the existing `RegularSync`)? Can you tell which areas in this file has been changed?

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -140,6 +142,10 @@ trait Blockchain {
 
   def saveNode(nodeHash: NodeHash, nodeEncoded: NodeEncoded, blockNumber: BigInt): Unit
 
+  def saveBlockNumber(number: BigInt, hash: ByteString): Unit
+
+  def saveBestBlockNumber(number: BigInt): Unit
```
LukasGasior1: I wonder if we really need this method? We already have `saveBlockNumber`, maybe it could also update best block num automatically?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -15,26 +15,122 @@ import scala.annotation.tailrec
...
 
-  def executeBlock(
-    block: Block,
-    validators: Validators): Either[BlockExecutionError, Seq[Receipt]] = {
+  def importBlock(block: Block): BlockImportResult = {
+    // TODO: ignore blocks that are too far behind the head
+    if (blockchain.getBlockByHash(block.header.hash).isDefined)
+      ImportFailedBlockExists
+
+    else
```
LukasGasior1: i'd suggest to use `{` and `}` in this if statement

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -15,26 +15,122 @@ import scala.annotation.tailrec
...
+          }
+      }
+  }
+
+  private def isTopOfChain(block: Block, td: BigInt): Boolean = {
+    val bestBlock = blockchain.getBlockByNumber(blockchain.getBestBlockNumber()).get // TODO: safe?
+    val isChild = bestBlock.header.hash == block.header.parentHash
+    lazy val isHigherTd = td > blockchain.getTotalDifficultyByHash(bestBlock.header.hash).getOrElse(0)
+
+    isChild || isHigherTd
```
LukasGasior1: why isn't `isHigherTd` sufficient?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -15,26 +15,122 @@ import scala.annotation.tailrec
...
+
+      parentByNumber match {
+        case None =>
+          UnknownBranch
+
+        case Some(parent) =>
+          // find blocks with same numbers in the current chain
+          val oldBranch = getOldBlocks(headers)
+          val currentBranchTotalDifficulty = oldBranch.map(_.header.difficulty).sum
+          val newBranchTotalDifficulty = headers.map(_.difficulty).sum
```
LukasGasior1: is it really the branch totalDifficulty? Not just sum of these headers difficulties?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -15,26 +15,122 @@ import scala.annotation.tailrec
...
+
+          if (currentBranchTotalDifficulty < newBranchTotalDifficulty)
+            NewBranch(oldBranch)
```
LukasGasior1: it seems a bit unintuitive, I mean the naming. How about something like: BranchReplaced or ChainReorganized or similar.
Same for `OldBranch` maybe rename to `ChainUntouched` or something like that

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -15,26 +15,122 @@ import scala.annotation.tailrec
...
+  private def checkHeadersChain(headers: Seq[BlockHeader]): Boolean =
+    if (headers.length > 1)
+      headers.zip(headers.tail).forall {
+        case (parent, child) =>
+          parent.hash == child.parentHash && parent.number + 1 == child.number
+      }
+    else
+      true
+
+  private def getOldBlocks(headers: Seq[BlockHeader]): List[Block] = headers match {
```
LukasGasior1: isn't it just `getBlocks` or `getBlocksForHeaders`?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -15,26 +15,122 @@ import scala.annotation.tailrec
...
+    case Seq(h, tail @ _*) =>
+      blockchain.getBlockByNumber(h.number).map(_ :: getOldBlocks(tail)).getOrElse(Nil)
```
LukasGasior1: also I think it can easily be changed to use tailrec (or can't it?)

src/main/scala/io/iohk/ethereum/blockchain/sync/NewRegularSync.scala
```diff
@@ -0,0 +1,311 @@
...
+
+    case PeerRequestHandler.RequestFailed(peer, reason) if waitingForActor.contains(sender()) =>
+      waitingForActor = None
+      if (handshakedPeers.contains(peer)) {
+        blacklist(peer.id, blacklistDuration, reason)
+      }
+      scheduleResume()
+  }
+
+  private def insertMinedBlock(block: Block, parentTd: BigInt) = {
```
ntallar: Should we also start using `Ledger.importBlock` here?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
ntallar: Regarding what @LukasGasior1 said, couldn't it be the case that the "old" branch has more blocks than the "new" one and a block is not obtained with `getOldBlocks`?
Maybe I'm missing something but it seems `getOldBlocks` only retrieves blocks at the same height as the `headers` parameter and not at a greater height, could this affect the `resolveBranch` function?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -15,26 +15,122 @@ import scala.annotation.tailrec
...
   def simulateTransaction(stx: SignedTransaction, blockHeader: BlockHeader): TxResult
+
+  def importBlock(block: Block): BlockImportResult
+
+  def resolveBranch(headers: Seq[BlockHeader]): BranchResolutionResult
 }
 
 //FIXME: Make Ledger independent of BlockchainImpl, for which it should become independent of WorldStateProxy type
-class LedgerImpl(vm: VM, blockchain: BlockchainImpl, blockchainConfig: BlockchainConfig) extends Ledger with Logger {
+class LedgerImpl(vm: VM, blockchain: BlockchainImpl, blockchainConfig: BlockchainConfig, validators: Validators) extends Ledger with Logger {
```
ntallar: This is up to discussion, but maybe we could separate the block execution and branch resolution aspects of the Ledger into separate traits?

rtkaczyk:
Continued in #313

PR #306, opened 2017-09-18
[EC-232] Adding support for eip 161 eth

KonradStaniec:
Docs of EIP161: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-161.md


src/main/resources/application.conf
```diff
@@ -224,6 +224,10 @@ mantis {
     # Doc: https://github.com/ethereum/EIPs/issues/160
     eip160-block-number = "3000000"
 
+    # EIP-161 fork block number (ETH Only)
+    # Doc: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-161.md
+    eip161-block-number = "9223372036854775807"
```
LukasGasior1: what's that number?

src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala
```diff
@@ -172,6 +175,23 @@ class InMemoryWorldStateProxy private(
   override def saveStorage(address: Address, storage: InMemoryWorldStateProxyStorage): InMemoryWorldStateProxy =
     copyWith(contractStorages = contractStorages + (address -> storage.wrapped))
 
+  override def touchAccounts(addresses: Address*): InMemoryWorldStateProxy =
+    copyWith(touchedAccounts = touchedAccounts.map(oldAddresses => oldAddresses ++ addresses.toSet))
```
LukasGasior1: what if `touchedAccounts` is None? is it expected that it remains `None` in this case?

LukasGasior1: oh I see now, `touchedAccounts` also holds the boolean flag `noEmptyAccounts`. What do you think about making it more explicit and adding this flag separately? (and changing `touchedAccounts` to just `Set`)

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -194,18 +199,22 @@ class LedgerImpl(vm: VM, blockchain: BlockchainImpl, blockchainConfig: Blockchai
...
+    val world2 = (deleteAccountsFn andThen deleteTouchedAccountsFn andThen persistStateFn)(worldAfterPayments)
 
     log.debug(
       s"""Transaction ${stx.hashAsHexString} execution end. Summary:
          | - Error: ${result.error}.
          | - Total Gas to Refund: $totalGasToRefund
          | - Execution gas paid to miner: $executionGasToPayToMiner""".stripMargin)
 
-    TxResult(world2, gasLimit - totalGasToRefund, resultWithErrorHandling.logs, result.returnData)
+    TxResult(world2, executionGasToPayToMiner, resultWithErrorHandling.logs, result.returnData)
```
LukasGasior1: :+1:

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -351,9 +360,13 @@ class LedgerImpl(vm: VM, blockchain: BlockchainImpl, blockchainConfig: Blockchai
...
 
-  private def pay(address: Address, value: UInt256)(world: InMemoryWorldStateProxy): InMemoryWorldStateProxy = {
-    val account = world.getAccount(address).getOrElse(Account.empty(blockchainConfig.accountStartNonce)).increaseBalance(value)
-    world.saveAccount(address, account)
+  private[ledger] def pay(address: Address, value: UInt256)(world: InMemoryWorldStateProxy): InMemoryWorldStateProxy = {
+    if (world.isZeroValueTransferToNonExistentAccount(address, value)) {
+      world
+    } else {
+      val account = world.getAccount(address).getOrElse(Account.empty(blockchainConfig.accountStartNonce)).increaseBalance(value)
+      world.saveAccount(address, account).touchAccounts(address)
```
LukasGasior1: hm, sorry it it's obvious but why are we touching `address` here. It can still be "normal", non-empty account, can't it?

LukasGasior1: ok, I see. `deleteTouchedAccounts` does not delete all touched accounts, it checks for emptiness first.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -194,18 +199,22 @@ class LedgerImpl(vm: VM, blockchain: BlockchainImpl, blockchainConfig: Blockchai
...
     val payMinerForGasFn = pay(Address(blockHeader.beneficiary), (executionGasToPayToMiner * gasPrice).toUInt256) _
+
+    val worldAfterPayments = (refundGasFn andThen payMinerForGasFn)(resultWithErrorHandling.world)
+
     val deleteAccountsFn = deleteAccounts(resultWithErrorHandling.addressesToDelete) _
+    val deleteTouchedAccountsFn = deleteTouchedAccounts _
     val persistStateFn = InMemoryWorldStateProxy.persistState _
 
-    val world2 = (refundGasFn andThen payMinerForGasFn andThen deleteAccountsFn andThen persistStateFn)(resultWithErrorHandling.world)
+    val world2 = (deleteAccountsFn andThen deleteTouchedAccountsFn andThen persistStateFn)(worldAfterPayments)
```
LukasGasior1: a random thought: do you think it would make sense to always have a list of touched accounts in WorldStateProxy (and not consider noEmptyAccounts at all there), and instead have a single condition here, that would delete them (if noEmptyAccounts=true or else do nothing)?

LukasGasior1: that way we wouldn't have to maintain the `Option` logic there, and we wouldn't need the additional flag. Of course I might miss something obvious here :)

src/main/resources/application.conf
KonradStaniec: It is in: https://github.com/input-output-hk/mantis/pull/306/files#diff-1494bba019381b0106c673078b722f7cL21, and it is 2675000.
So it is only enabled in ETH network, setting it `Long.maxValue` here disables it for other networks.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
KonradStaniec: hmm i will change name to ` deleteEmptyTouchedAccounts`, to be more precise

src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala
KonradStaniec: `touchedAccounts` is set based on `noEmptyAccounts` flag in https://github.com/input-output-hk/mantis/pull/306/files#diff-7632a397485d58b59d376a5eb28a40c8R36. How do you think it could be made more explicit ?

And  Actually, `touchedAccounts`  could be just `Set` always, and we could always just add accounts to it, delete empty accounts only if a flag is set, and clear it at the end of the transaction. But i thought that it would be quite a wasteful approach as: 
1. `oldAddresses ++ addresses.toSet` operations on `Set` are quite expensive. 
2. This eip is only enabled for ETH net, so we would be wasting resources thorugh all ETC chain. 

My logic here was that making it `Option[Set]`, we : 
explicitly declare that sometimes (on ETC chain, before fork block), we don't need those touched accounts.
Save ourselves quite a computation that is not necessary, and `Option` take care of all `if .. else` comparisons.

src/main/resources/application.conf
LukasGasior1: :+1:

src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala
LukasGasior1: Ok, let's leave it as an Option

LukasGasior1:
needs merge

src/main/resources/application.conf
AlanVerbner: What about setting it to `null` and when reading config mapping it to an `Option` instead of `Long.maxValue`?

KonradStaniec: Actually, other clients use both of these strategies (empty values or large block numbers) to disable correct EIPs. During pr to EIP106 (https://github.com/input-output-hk/mantis/blob/ea26ee798ed73b9973a2d8eca3abca1db410c9b9/src/main/resources/application.conf#L211)  we agreed with @rtkaczyk to use higher block number and here it is just a matter of consistency.

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
```diff
@@ -28,9 +28,12 @@ class BlockchainSuite extends FreeSpec with Matchers with Logger {
     "bcTotalDifficultyTest/uncleBlockAtBlock3afterBlock4" -> Set.empty,
     "TransitionTests/bcFrontierToHomestead/blockChainFrontierWithLargerTDvsHomesteadBlockchain"  -> Set.empty,
     "TransitionTests/bcFrontierToHomestead/blockChainFrontierWithLargerTDvsHomesteadBlockchain2"  -> Set.empty,
-    "TransitionTests/bcFrontierToHomestead/HomesteadOverrideFrontier" -> Set.empty
+    "TransitionTests/bcFrontierToHomestead/HomesteadOverrideFrontier" -> Set.empty,
```
rtkaczyk: Please add a comment explaining why DAO test these are being ignored. @AlanVerbner do you remember why they weren't working for older versions of the tests repository?

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainTestConfig.scala
```diff
@@ -8,6 +8,7 @@ import org.spongycastle.util.encoders.Hex
 trait BlockchainTestConfig extends BlockchainConfig {
 
   val frontierBlockNumber: BigInt = Long.MaxValue
+  val eip161BlockNumber: BigInt = Long.MaxValue
```
rtkaczyk: Minor, could we make them ordered by EIP number?

src/main/resources/application.conf
rtkaczyk: Yeah, but `9223372036854775807` is really random looking number and as we can see it raises confusion. Could we use 10000...000 instead, or accept hexadecimal values and use 0x7fffffffffffffff (parity or geth uses that).

src/main/scala/io/iohk/ethereum/domain/Account.scala
```diff
@@ -45,6 +45,12 @@ case class Account(
   def resetAccountPreservingBalance(startNonce: UInt256 = UInt256.Zero): Account =
     copy(nonce = startNonce, storageRoot = Account.EmptyStorageRootHash, codeHash = Account.EmptyCodeHash)
 
+  /**
+    * According to EIP161: An account is considered empty when it has no code and zero nonce and zero balance.
```
rtkaczyk:
```
... zero nonce and zero balance. An account's storage is not relevant when determining emptiness.
```

Let's be explicit, it's confusing stuff.

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
AlanVerbner: @rtkaczyk I was only running `bcHomesteadToDao` [tests](https://github.com/ethereum/tests/tree/develop/BlockchainTests/TransitionTests/bcHomesteadToDao) and those weren't working because the name and the comments didn't match (it was saying the transition was at Block#5 but the doc was referring to Block#8)

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -402,10 +415,37 @@ class LedgerImpl(vm: VM, blockchain: BlockchainImpl, blockchainConfig: Blockchai
...
+    */
+  private[ledger] def deleteTouchedAccounts(worldStateProxy: InMemoryWorldStateProxy): InMemoryWorldStateProxy = {
+    def deleteEmptyAccounts(addressesToDelete: Set[Address]) =
+      addressesToDelete.foldLeft(worldStateProxy){ case (world, address) => deleteEmptyAccount(world, address) }
 
+    def deleteEmptyAccount(world: InMemoryWorldStateProxy, address: Address) = {
+      if (world.getAccount(address).exists(_.isEmpty))
+        world.deleteAccount(address)
+      else
+        world
```
rtkaczyk: Do we need to check for existence? Isn't `deleteAccount` idempotent when account does not exist?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -802,9 +803,12 @@ abstract class CallOp(code: Int, delta: Int, alpha: Int) extends OpCode(code, de
       val gasAdjustment: BigInt = if (validCall) 0 else -startGas
       val mem2 = mem1.expand(outOffset, outSize)
 
+      val updateWorld = if (validCall) state.world.combineTouchedAccounts(result.world) else state.world
```
rtkaczyk: `updatedWorld` tells us nothing, either use the convention and name it `world1` or think of something clever :)

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -926,13 +944,23 @@ case object SELFDESTRUCT extends OpCode(0xff, 1, 0, _.G_selfdestruct) {
...
+        state.world.isDead(refundAddress)
 
+    def preEip161CostCondition: Boolean =
+      state.config.chargeSelfDestructForNewAccount && !state.world.accountExists(refundAddress)
+
+
+    if (state.config.noEmptyAccounts)
+      if (postEip161CostCondition) state.config.feeSchedule.G_newaccount else 0
+    else
+      if (preEip161CostCondition) state.config.feeSchedule.G_newaccount else 0
```
rtkaczyk: So... `if (noEmptyAccounts && postEIP161 || preEIP161) G_newaccount else 0` ?

Also has this logic has gotten quite more complex it might be desirable to share it between `varGas` and `gasExtra`

src/ets/scala/io/iohk/ethereum/ets/vm/VMSuite.scala
```diff
@@ -57,12 +57,12 @@ class VMSuite extends FreeSpec with Matchers with Logger {
...
 
       val postWorld = ScenarioBuilder.prepareWorld(post, scenario.env.currentNumber)
-      val deadAccounts = postWorld.accounts.keys.filter(postWorld.isAccountDead)
+      val deadAccounts = postWorld.accounts.keys.filter(postWorld.isDead)
       val expectedWorld = deadAccounts.foldLeft(postWorld)(_ deleteAccount _)
       val actualWorldNoDead = deadAccounts.foldLeft(result.world)(_ deleteAccount _)
 
       actualWorldNoDead shouldEqual expectedWorld
-      deadAccounts.foreach(addr => result.world.isAccountDead(addr) shouldBe true)
+      deadAccounts.foreach(addr => result.world.isDead(addr) shouldBe true)
```
rtkaczyk: Is `isDead` a better name than `isAccountDead`? I feel like sometimes it reads as if the world was dead (which may be quite depressing).

src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala
rtkaczyk: Well, I don't know... I like @LukasGasior1 's idea of separating those aspects. And while it's a valid point that want to avoid unnecessary `setA ++ setB` - why would we have to do that?

```
override def touchAccounts(addresses: Address*): InMemoryWorldStateProxy =
  if (noEmptyAccounts)
    copyWith(touchedAccounts = touchedAccounts ++ addresses.toSet)
  else 
    this
```

That should simplify the related code, and I don't think you'd need another conditional on `noEmptyAccounts` flag.

src/main/scala/io/iohk/ethereum/vm/Program.scala
```diff
@@ -34,7 +35,7 @@ case class Program(code: ByteString) {
     if(pos < 0 || pos >= length) accum
     else {
       val byte = code(pos)
-      val opCode = EvmConfig.FrontierConfigBuilder(None).byteToOpCode.get(byte) // we only need to check PushOp and JUMPDEST, they are both present in Frontier
+      val opCode = frontierConfig.byteToOpCode.get(byte) // we only need to check PushOp and JUMPDEST, they are both present in Frontier
```
rtkaczyk: 👍

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -402,10 +415,37 @@ class LedgerImpl(vm: VM, blockchain: BlockchainImpl, blockchainConfig: Blockchai
...
+    * Account is the source or newly-creation of a CREATE operation or contract-creation transaction endowing zero or more value;
+    * as the block author ("miner") it is recipient of block-rewards or transaction-fees of zero or more.
+    *
+    * Deletion of touched account should be executed immediately following the execution of the suicide list
+    *
+    * @param worldStateProxy world after execution of all potentially state-changing operations
+    * @return a worldState equal worldStateProxy except that the accounts touched during exection are deleted and touched
+    *         Set is cleared
+    */
+  private[ledger] def deleteTouchedAccounts(worldStateProxy: InMemoryWorldStateProxy): InMemoryWorldStateProxy = {
```
rtkaczyk: This is to be renamed right?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
KonradStaniec: I have tried the condition you propose but it doesn't work (blockchain test fails). it is because these conditions are separate i.e we don't want to check `preEIP161` if we are `postEIP161`. There are some cases where `noEmptyAccounts && postEIP161 ` will be false when we post EIP161, but 
`preEIP161` will be true and we would charge an incorrect fee.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
KonradStaniec: This check is for the sake of `_.isEmpty` method as it is called on retrieved Account.  And we cant call here `getGuaranteedAccount` as Account could be earlier deleted in suicide list (as it is executed before touched accounts )

src/main/resources/application.conf
KonradStaniec: Oh we should expect that users that will look at configuration should at first glance recognize that
`9223372036854775807` is really `Long.maxValue`  😄  and more seriously I would go with
`10000...000` because most of our config are just decimal values, and this hex would like out of place for me

src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala
KonradStaniec: You have valid point with that there would be only conditional needed, but I am still not entirely convinced by that approach as: 
1. We have other methods like `combineTouchedAccounts`,   `clearTouchedAccounts`, we would not need to have conditionals in them, but they would operate on empty sets mostly. ( not a deal breaker) 
2.  In `InMemoryWorldStateProxy` we would have something like 
`  val touchedAccounts: Set[Address],
  val noEmptyAccount: Boolean`
in a constructor and it is not immediately obvious that these values are correlated with each other.
Only look at `touchAccounts` would reveal that these values exist together.
now this dependence is explicitly declared in `apply` method

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
rtkaczyk: So maybe setting the configs to block 8 would work with the version we currently use? Anyway, we can just re-enable them after EIP-684.

src/universal/conf/ethereumHF.conf
```diff
@@ -18,7 +18,7 @@ mantis {
     eip150-block-number = "2463000"
     eip155-block-number = "2675000"
     eip160-block-number = "2675000"
-    # TODO: EIP-161 a,b,c,d = 2675000
+    eip161-block-number = "2675000"
     # TODO: EIP-170
```
rtkaczyk: Oh, could you remove the `TODO` apparently we don't need a block number for that (should have been done earlier) ?

src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala
rtkaczyk: 1. Yes, and btw `EmptySet` is optimised
2. I agree that without any documentation it's not obvious, just as it's not obvious why `Option[Set[..]]` is used. Code comments should clarify that

KonradStaniec: 1. This is not a matter of optimisation, but a matter of some silent assumption make by other functions that they are operating on empty sets, `Option` takes this logic out in open
2. Good point with adding some comments. Nevertheless `if (noEmptyAccounts) Some(Set.empty) else None` constructor makes it obvious when need or dont need the `Set[..]`. ( though i can be biased here as i am implenter and you and @LukasGasior1  did not see it that way )

src/universal/conf/ethereumHF.conf
KonradStaniec: hmm Good catch. I think there should be `max-code-size` specified, in EIP170 pr. Am I right @ntallar ?

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainSuite.scala
KonradStaniec: Setting config to block 8 did not help. We will re-enable this test after updating blockchain suite

src/universal/conf/ethereumHF.conf
KonradStaniec: Added here `max-code-size =  24576` and removed todo

src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala
KonradStaniec: The more I think about it, the more it feels as the matter of preference of representation 
(`Option[Set] vs  Set and Boolean`). 
So let's have it your way as there is 2 vs 1, in terms who prefers what

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
rtkaczyk: Ok, I didn't read too carefully, but my point here was that this function seems more verbose than it needs to be. What about something like:

```scala
val addressesToDelete = world.touchedAccounts.filter(addr => world.getAccount(addr).exists(_.isEmpty))
addressesToDelete.foldLeft(world)(_ deleteAccount _).clearTouchedAccounts
```

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
rtkaczyk: Pardon, that should be `if (noEmptyAccounts && postEIP161 || !noEmptyAccount && preEIP161) G_newaccount else 0` - which is a bit less concise but at least we don't repeat the branches.

Also my argument about sharing (not repeating) in `varGas` and `gasExtra` still stands.

src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpec.scala
```diff
@@ -10,7 +10,7 @@ import io.iohk.ethereum.vm.MockWorldState.PS
 
 class OpCodeGasSpec extends FunSuite with OpCodeTesting with Matchers with PropertyChecks {
 
-  override val config = EvmConfig.PostEIP160ConfigBuilder(None)
+  override val config = EvmConfig.PostEIP161ConfigBuilder(None)
```
rtkaczyk: Not sure if this a good idea. After all our priority is ETC. These tests were designed before hard-fork/EIP support was added and they don't offer much support for those :(

I would say they should test the latest in-effect ETC hard-fork. What do you think?
Unless maybe you can think of a clean way to test for both ETC and ETH?

src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpec.scala
```diff
@@ -477,19 +477,19 @@ class OpCodeGasSpec extends FunSuite with OpCodeTesting with Matchers with Prope
...
     }
 
-    // Sending refund to an already existing account
+    // Sending refund to an already existing account not dead account
     forAll(stateGen) { (stateIn) =>
       val (refund, _) = stateIn.stack.pop
       val world = stateIn.world.saveAccount(
         Address(refund),
-        Account.empty())
+        Account.empty().increaseNonce())
```
rtkaczyk: If we were to leave that EVM config - why is this needed?

src/test/scala/io/iohk/ethereum/vm/CallOpcodesSpec.scala
```diff
@@ -13,7 +14,7 @@ import org.scalatest.prop.PropertyChecks
 // scalastyle:off file.size.limit
 class CallOpcodesSpec extends WordSpec with Matchers with PropertyChecks {
 
-  val config = EvmConfig.PostEIP160ConfigBuilder(None)
+  val config = EvmConfig.PostEIP161ConfigBuilder(None)
```
rtkaczyk: See https://github.com/input-output-hk/mantis/pull/306/files#r141736062

src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpec.scala
rtkaczyk: Or maybe we could add an EIP-161 specific opcode test? I.e. move these changes to a separate spec, and leave these tests unmodified, WDYT?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
KonradStaniec: This condition is little less readable for me but we can go with it.

This condition (not really the same but similar) is used in `gasExtra` of `CallOp` and in `varGas` of `SelfDestruct`. You propose to share it between opCodes ?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
KonradStaniec: I am not convinced, we are saving some space definitely, but we lose readability ( i can definitely imagine myself staring in some time at `_ deleteAccount _` and struggling to decipher it) and we create unnecessary transition collection of `addressesToDelete`.

src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpec.scala
KonradStaniec: It is only needed because of EIP161 config, as an empty account is considered dead and there are different fees.

src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpec.scala
KonradStaniec: You definitely have valid point here. I think in the scope of this pr most appropriate would be to move it to other specs like `OpCodeGasSpecPostEIP161`, an additional bonus of that would be that we will see what op codes has been modified in this eip changes.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
rtkaczyk: I'm fine with explicitly named params. Also `clearTouchedAccounts` can be a separate line.

The main point here is that this logic consists of 3 easy steps:

1. determine the addresses to delete
2. delete them
3. clear touched set

IMO, using closures does not make it so obvious (it's also slightly less efficient).

As for efficiency, yes, we could do away with the set transformations by checking account inside `foldLeft`'s inner body, but do we really have a lot to gain here?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
rtkaczyk: Sorry, I misread. I thought it was `CallOp#varGas`. Nevermind

src/main/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxy.scala
rtkaczyk: @KonradStaniec don't get me wrong. There was nothing fundamentally wrong with the `Option` solution, and the proposed change is only cosmetic. But I disagree about it being a matter of preference. Separate set and flag are slightly, but clearly better because:

- philosophical argument: understanding 2 simple things is easier than 1 thing composed of 2 things. Terms _separation of concerns_ applies to variables as well.
- practical and more important argument: in most places you don't need to deal with both aspects - you're only interested in the elements of the set, and not whether they are collected at all - which results in cleaner simpler code.

src/test/scala/io/iohk/ethereum/vm/CallOpFixture.scala
```diff
@@ -0,0 +1,157 @@
...
+  val worldWithSstoreWithClearProgram = worldWithoutExtAccount.saveAccount(extAddr, accountWithCode(sstoreWithClearProgram.code))
+    .saveCode(extAddr, sstoreWithClearCode.code)
+
+  val worldWithReturnSingleByteCode = worldWithoutExtAccount.saveAccount(extAddr, accountWithCode(returnSingleByteProgram.code))
+    .saveCode(extAddr, returnSingleByteProgram.code)
+
+  val env = ExecEnv(ownerAddr, callerAddr, callerAddr, 1, ByteString.empty, 123, Program(ByteString.empty), null, 0)
+  val context: PC = ProgramContext(env, ownerAddr, 2 * requiredGas, worldWithExtAccount, config)
+
+  case class CallResult(
```
rtkaczyk: I like this separation. One thing though: does it need to be a path-dependent type? If not, it would better to move it to `CallOpFixture` companion and import it from there.

KonradStaniec: with the current design, I am afraid it needs to be as all default parameters depend on values in  `CallOpFixture`. Maybe we would refactor this in a separate task as I don't like the idea of messing with those tests fixture more than necessary or duplicating `CallResult` in two test classes .

rtkaczyk: >  default parameters depend on values in CallOpFixture

Of course.. That's ok then, nevermind

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
KonradStaniec: I don't have big this address set could be but doing some unnecessary work when we can easily avoid it is not good option I think. As for readability it is I think really subjective.

Maybe we can meet in the middle with something like:
`    world.touchedAccounts
      .foldLeft(world){ case (actualWorld, address) => deleteEmptyAccount(actualWorld, address) }
      .clearTouchedAccounts`

In there we abandon one closure, see that we are operating on touched accounts and deleting only empty accounts from that set, and clearing set at the end.

rtkaczyk: I would be fine with that, but I still think my proposition is better. At any rate we don't need the outer closure, cause that can simply be a `val`. I can see some usefulness of the other one - `deleteEmptyAccount`, but I would still skip cause the body of `foldLeft` would be small enough.

Regarding efficiency, all the work of checking accounts and deleting accounts has to be done anyway, and creating the intermediate collections seems small enough compared to that. But if we're concerned, we can still avoid that by using a view: `touchedAccounts.view.flatMap(..).filter(..)`

As for readability: the problem can be formulated in English in these 2 ways:

  - find empty accounts in touched accounts
  - delete them

vs

  - go through all the touched accounts, and if they're empty, delete them

I know the difference seems negligible with such a small problem, but which one sounds clearer?

KonradStaniec: I don't like this `.view` idea as it brings another complication in code which supposes to do a simple thing and can be implemented without it.
As for English readability, the two-step style has its merits (two clearly separated steps), the second is really close and trumps it a little with performance without additional combinators.
Also, I think `deleteEmptyAccount` closure is necessary imo, as this `if .. else` combined with``{ case (actualWorld, address)` just look bad.

Summarising;
I would stay with current approach, but if you insist I can go your way as problem at hand is not some big data set transformation and I would like to close this pr before more merge conflicts arise

mcsherrylabs: Both are reasonable. I'm slightly in favor of the 2 liner, but lets go with the authors approach in this case. (Now fun's over - back to work ;) )

rtkaczyk: > I don't like this .view idea as it brings another complication in code

How is:

```scala
val addressesToDelete = world.touchedAccounts.view.filter(addr => world.getAccount(addr).exists(_.isEmpty))
```

or 

```scala
val addressesToDelete = world.touchedAccounts.withFilter(addr => world.getAccount(addr).exists(_.isEmpty))
```

more complicated than:

```scala
val addressesToDelete = world.touchedAccounts.filter(addr => world.getAccount(addr).exists(_.isEmpty))
```
?

Usage of those methods simply implies lazy evaluation. Should we avoid the keyword `lazy` because that increases complexity?

Anyway, all my points were made. If you want to go with the closures, that's good enough for me. Just please get rid off that outer one, like you proposed above.

Also, consider these pointers:
- .foldLeft(z){ (a, x) => ... } - no need for the `case` keyword
- you can write `touchedAccounts.foldLeft(world)(deleteEmptyAccount)`

I'm ready to approve this one, pending your decision here.

KonradStaniec: > Usage of those methods simply implies lazy evaluation. Should we avoid the keyword lazy because that increases complexity <

I am not saying to avoid it, but to use it with care e.g not use `lazy val` when `val` will suffice. I just think that lazy evaluation in collections (e.g `Streams`, `withFilter`)is great when we have many operations stacked up onto large collections. In the current situation, it seems overkill as we have just one condition and operation.

>you can write touchedAccounts.foldLeft(world)(deleteEmptyAccount)

👍

PR #307, opened 2017-09-19
[EC-231] ECIP 1017 Geth compatibility fixes and tests

ntallar:
## Description

Includes integration tests that executes blocks generated by Geth, for which the `eraDuration` was reduced to 3. The purpose of these was to test the compatibility of Mantis with Geth, for which several changes were done to the `BlockRewardCalculator`, based on the formulas used on Geth.

During testing it was found that Geth and Parity aren't [compatible](https://github.com/paritytech/parity/issues/6523), which is why this test doesn't check compatibility with Parity.


src/main/scala/io/iohk/ethereum/ledger/BlockRewardCalculator.scala
```diff
@@ -41,12 +41,9 @@ class BlockRewardCalculator(config: MonetaryPolicyConfig) {
...
   def calcBlockMinerReward(blockNumber: BigInt, ommersCount: Int): BigInt = {
     val era = eraNumber(blockNumber)
-    val eraMultiplier = rewardReductionRateNumer.pow(era)
-    val eraDivisor = rewardReductionRateDenom.pow(era)
 
-    val baseReward = (firstEraBlockReward * eraMultiplier) / eraDivisor
-    val ommersReward = (firstEraBlockReward * ommersCount * ommerInclusionRewardNumer * eraMultiplier) /
-      (ommerInclusionRewardDenom * eraDivisor)
+    val baseReward = calcMinerBaseReward(era)
+    val ommersReward = calcMinerRewardPerOmmer(era) * ommersCount
```
rtkaczyk: Take a note of https://github.com/ethereumproject/go-ethereum/issues/352 and https://github.com/ethereumproject/ECIPs/issues/15#issuecomment-330660976. Maybe we should at least add comments with those links until the issue is resolved. Ideally we should hold on merging this until we get some feedback (hopefully soon)

src/main/scala/io/iohk/ethereum/ledger/BlockRewardCalculator.scala
```diff
@@ -56,13 +53,31 @@ class BlockRewardCalculator(config: MonetaryPolicyConfig) {
     if (era == 0) {
       val numer = firstEraOmmerMiningRewardMaxNumer - (blockNumber - ommerNumber - 1)
       (firstEraBlockReward * numer) / firstEraOmmerMiningRewardDenom
-    } else {
-      val eraMultiplier = rewardReductionRateNumer.pow(era)
-      val eraDivisor = rewardReductionRateDenom.pow(era)
-      (firstEraBlockReward * ommerMiningRewardNumer * eraMultiplier) / (ommerMiningRewardDenom * eraDivisor)
-    }
+    } else
+      calcMinerBaseReward(era) * ommerMiningRewardNumer / ommerMiningRewardDenom
```
rtkaczyk: You're multiplying the result of division, thus introducing rounding errors

src/main/scala/io/iohk/ethereum/ledger/BlockRewardCalculator.scala
rtkaczyk: If we go with this implementation, comments need to be updated to reflect that these are separate rewards that have already been rounded down.

src/it/scala/io/iohk/ethereum/txExecTest/ECIP1017Test.scala
```diff
@@ -0,0 +1,59 @@
...
+    val blockHeaderValidator: BlockHeaderValidator = new BlockHeaderValidatorImpl(blockchainConfig)
+    val ommersValidator: OmmersValidator = new OmmersValidatorImpl(blockchainConfig)
+    val signedTransactionValidator: SignedTransactionValidator = new SignedTransactionValidatorImpl(blockchainConfig)
+  }
+
+  "Ledger" should "execute blocks with respect to block reward changed by ECIP 1017" in {
+    val fixtures: FixtureProvider.Fixture = FixtureProvider.loadFixtures("/txExecTest/ecip1017Test")
+
+    val startBlock = 1
+    val endBlock = 100
```
rtkaczyk: Could we have more blocks in this test? The further the era, the more chance of catching a rounding error. Ideally we should reach era #170 - IIRC, that's when rewards drop to zero.

rtkaczyk: Era <span>#</span>193 actually:

```
x = 1 - floor(ln(a) / ln(1 - m))

where:
  x - the era number where rewards drop to zero
  a - base reward (5 Ether)
  m - reduction rate (0.2)
```

rtkaczyk: Can we make it an even 200 to be sure? 😃

ntallar: Changed `endBlock` to 602, which is the first block that belongs to era 200.
No other compatibility error was found :smile:

src/main/scala/io/iohk/ethereum/ledger/BlockRewardCalculator.scala
ntallar: Yes, this could cause rounding error, however as this seems to be the more closer approach to how it is [calculated in Geth](https://github.com/ethereumproject/go-ethereum/pull/220/files#diff-5b5942d1b7feac47d38b879796555098R197) and `ommerMiningRewardNumer=1`, I decided to leave it this way.
Do you think it would be better to change this?

rtkaczyk: No, if geth does it then it's fine. I thought about it and it's not devoid of logic to have it rounded down prior to another division.

src/it/scala/io/iohk/ethereum/txExecTest/ECIP1017Test.scala
```diff
@@ -0,0 +1,59 @@
...
+    val blockHeaderValidator: BlockHeaderValidator = new BlockHeaderValidatorImpl(blockchainConfig)
+    val ommersValidator: OmmersValidator = new OmmersValidatorImpl(blockchainConfig)
+    val signedTransactionValidator: SignedTransactionValidator = new SignedTransactionValidatorImpl(blockchainConfig)
+  }
+
+  "Ledger" should "execute blocks with respect to block reward changed by ECIP 1017" in {
+    val fixtures: FixtureProvider.Fixture = FixtureProvider.loadFixtures("/txExecTest/ecip1017Test")
+
+    val startBlock = 1
+    val endBlock = 602
```
rtkaczyk: Could we maybe add a comment about era 193 going zero, and how we want to go past that point.

rtkaczyk:
Could we maybe add some test cases to `BlockRewardSpec` that capture those tiny differences in rounding?

PR #309, opened 2017-09-21
Static analysis tools fixes

ntallar:
## Description

Includes fixes to several issues detected while running various static analysis tools.
The problems fixed were:

* Removed unused `DbConfig` config trait and `targetBlockChoose` parameter for `waitingForTargetBlock`
* Replaced empty case classes with case objects
* Removed unnecessary and duplicated imports


KonradStaniec:
@ntallar actually this ticket got me thinking, maybe it would be worthwhile to turn on more compiler safety flags and let the compiler do work for us?  
We have got only `  "-unchecked",
  "-deprecation",
  "-feature",
  "-Xfatal-warnings"`, but full list is much longer.
The most comprehensive summary I could find is in https://tpolecat.github.io/2017/04/25/scalac-flags.html,
WDYT?

ntallar:
@KonradStaniec, I just added several of those flags that seem to be useful, is there any of those you think we should also add?
`-Ywarn-unused` is also useful, however it detects a couple of false positives so I didn't added it to the list (the same happens with the static analysis tools whose fixes were included in this PR https://github.com/HairyFotr/linter, https://github.com/sksamuel/scapegoat and http://www.wartremover.org/doc/warts.html)

KonradStaniec:
@ntallar Actually I think I will be enough ( it is not worth adding checks which could yield false positives imo). 
Hmh there are many unused imports in tests, that's why `CiricleCI` is failing.

ntallar:
Hi @zjhmale, we are using [HairyFotr](https://github.com/HairyFotr/linter), [scapegoat](https://github.com/sksamuel/scapegoat) and [Wartremover](http://www.wartremover.org/doc/warts.html).

zjhmale:
@ntallar Cool, thanks!

PR #313, opened 2017-09-28
[EC-303] block import

rtkaczyk:


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -1,59 +1,225 @@
...
+    else {
+      val parentByHash = blockchain.getBlockHeaderByHash(headers.head.parentHash)
+
+      parentByHash match {
+        case None =>
+          UnknownBranch
+
+        case Some(parent) =>
+          // find blocks with same numbers in the current chain
+          val oldBranch = getBlocksForHeaders(headers)
```
rtkaczyk: On #305 @ntallar writes:

> Regarding what @LukasGasior1 said, couldn't it be the case that the "old" branch has more blocks than the "new" one and a block is not obtained with getOldBlocks?
Maybe I'm missing something but it seems getOldBlocks only retrieves blocks at the same height as the headers parameter and not at a greater height, could this affect the resolveBranch function?

nb. `getOldBlocks` has been renamed to `getBlocksForHeaders`

rtkaczyk: That's a good point but this whole function is meant only to move the logic previously found in `RegularSync#processBlockHeaders`. 

We would have to compare branches from a common to ancestor down to the chain head. But how do we know we receive all headers that form a parallel branch in the peer?

ntallar: In that case the parallel branch would be the "new" branch? It seems that comparing `from a common to ancestor down to the chain head` is as much as we can do. Also as we are saving blocks from different blockchain branches, if there are better blocks, it will be easier to process them when receiving them later.

However I'm not sure if that's changing the `Ledger#resolveBranch` implementation is needed. Due to how `RegularSync` works, AFAIK the `headers` parameter to `Ledger#resolveBranch` always has greater or equal height than the "old" branch (as it is always equal to `headersQueue`). Maybe it would be better to at least include a comment regarding that.

rtkaczyk: > AFAIK the headers parameter to Ledger#resolveBranch always has greater or equal height than the "old" branch (as it is always equal to headersQueue)

Why is that?

ntallar: I might be missing something, but I think this is caused by the fact that we always start asking for headers (in `RegularSync#askForHeaders`) for the `appStateStorage.getBestBlockNumber() + 1` block, and we accumulate the blocks received in `headersQueue` as we are going "back" in the blockchain (while `resolvingBranches=true`). As this process is not stopped while importing new blocks, the "new" branch will have at least 1 more block than the "old" branch.

I don't like so much about just leaving a comment that this could be easily altered, for example if we enable importing new blocks while syncing.

rtkaczyk: @ntallar I have added a requirement that the oldest header must have number at least equal to current best block number. Otherwise `InvalidBranch` is returned. WDYT?

ntallar: @rtkaczyk That should work :+1:
Having it documented like this would be useful in case any other refactoring is needed :+1:

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainTestConfig.scala
```diff
@@ -24,6 +24,9 @@ trait BlockchainTestConfig extends BlockchainConfig {
   override val monetaryPolicyConfig: MonetaryPolicyConfig = MonetaryPolicyConfig(5000000, 0.2, BigInt("5000000000000000000"))
   override val daoForkConfig: Option[DaoForkConfig] = None
   override val accountStartNonce: UInt256 = UInt256.Zero
+
+  // TODO: only place where this was supposed to be used but now it seems it's not, remove?
```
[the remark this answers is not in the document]
rtkaczyk: Yeah, funny thing. So when I was running those tests initially I determined that one of the failures was due to 2 chains having the same total difficulty and that this was needed to resolve the tie. This does not seem to be the case anymore and I can think of 2 logical explanations:
- AFAIR the problem test was `bcMultiChainTest/ChainAtoChainB_difficultyB`, and looking at it again I can see the chains have different difficulties (I might have misread an `8` for a `0`)
- maybe it's a different test and maybe I was running a different version of the subproject than currently committed (it's likely cause I was investigating the DAO tests at around that time)

Interestingly setting the flag to `true` *does* cause some tests to fail. I think the safest approach here might be to leave it for now and a task to investigate it once we update to the latest version of the tests. WDYT?

N.b. AFAIK no clients other than cpp-ethereum have this flag. But then again cpp-ethereum seems to be the only one capable of running the whole suite.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -276,78 +259,58 @@ class RegularSync(
...
-                context.self ! ResumeRegularSync
-              }
-          }
-        case None =>
-          //TODO: Investigate if we can recover from this error (EC-165)
-          val parentHash = Hex.toHexString(blocks.head.header.parentHash.toArray)
-          log.error(s"No total difficulty for the latest block with number ${blocks.head.header.number - 1} (and hash $parentHash)")
-          shutdownAppFn()
+      @tailrec
+      def importBlocks(blocks: List[Block], importedBlocks: List[Block] = Nil): (List[Block], Option[BlockImportFailed]) =
```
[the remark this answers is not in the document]
rtkaczyk: 🤔

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -1,59 +1,285 @@
...
+    *        (oldBranch, newBranch) as lists of blocks
+    */
+  private def reorganiseChainFromQueue(queuedLeaf: ByteString): Either[BlockExecutionError, (List[Block], List[Block])] = {
+    val newBranch = blockQueue.removeBranch(queuedLeaf)
+    val parent = newBranch.head.header.parentHash
+    val bestNumber = blockchain.getBestBlockNumber()
+
+    val staleBlocksWithReceiptsAndTDs = removeBlocksUntil(parent, bestNumber).reverse
+    val staleBlocks = staleBlocksWithReceiptsAndTDs.map(_._1)
+    val staleTd = (for (block <- staleBlocks) yield blockQueue.enqueueBlock(block))
```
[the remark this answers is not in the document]
rtkaczyk: I went through some effort to calculate it... I wonder what I wanted it for.. 🤔

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -232,9 +243,15 @@ class BlockchainImpl(
   override protected def getHashByBlockNumber(number: BigInt): Option[ByteString] =
     blockNumberMappingStorage.get(number)
 
-  private def saveBlockNumberMapping(number: BigInt, hash: ByteString): Unit = blockNumberMappingStorage.put(number, hash)
+  private def saveBlockNumberMapping(number: BigInt, hash: ByteString): Unit = {
+    blockNumberMappingStorage.put(number, hash)
+    appStateStorage.putBestBlockNumber(number) // FIXME: mother of consistency?!?!
```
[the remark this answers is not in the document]
rtkaczyk: Yeah, so I wanted avoid having yet another save (in the series of three), so I went with this. But this is terrible because it expects blocks to be saved and removed in specific order - I don't think we'll ever want to save them in a different order but it's still bad.

What if the `save`/`remove` methods took another optional param, that would determine whether the best block number should be updated and how?

(I also thought adding a bit more logic to the best number updates in those methods, but I think that would be wrong, not only because of an extra interaction with the storage)

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -189,6 +194,12 @@ class BlockchainImpl(
 
   override def getTotalDifficultyByHash(blockhash: ByteString): Option[BigInt] = totalDifficultyStorage.get(blockhash)
 
+  override def getBestBlockNumber(): BigInt =
+    appStateStorage.getBestBlockNumber()
+
+  override def getBestBlock(): Block =
+    getBlockByNumber(getBestBlockNumber()).get // TODO: safe?
```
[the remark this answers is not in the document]
rtkaczyk: Yes, consistency should not be an issue here, but if the DB fails? Will the actors just restart and try that again?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -1,59 +1,285 @@
...
+        Left(error)
+    }
+  }
+
+  /**
+    * Executes a list blocks, storing the results in the blockchain
+    * @param blocks block to be executed
+    * @return a list of blocks that were correctly executed and an optional [[BlockExecutionError]]
+    */
+  private def executeBlocks(blocks: List[Block]): (List[Block], Option[BlockExecutionError]) = {
```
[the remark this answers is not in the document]
rtkaczyk: Yes, this one can be converted quite easily. Not that it's needed because the number of blocks is also bounded by `BlockQueue`, but I can do it if it makes anyone more comfortable (I realise there is common sense of alarm against non-tail recursive functions, which I do not share 😉 )

As for your question, yes, at least I'm not concerned 😉. Besides we're operating on a tree here, and recursive functions are a natural way of doing that.

src/main/scala/io/iohk/ethereum/ledger/BlockQueue.scala
```diff
@@ -0,0 +1,187 @@
...
+
+      case None =>
+        val parentTd = blockchain.getTotalDifficultyByHash(parentHash)
+
+        parentTd match {
+
+          case Some(_) =>
+            addBlock(block, parentTd)
+            log.debug(s"Enqueued new block (${blockId(block)}) with parent on the main chain")
+            updateTotalDifficulties(hash)
```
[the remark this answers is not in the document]
rtkaczyk: No, this is the case where the parent lives on the main chain.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -67,53 +77,45 @@ class RegularSync(
   }
 
   /**
-    * Handles broadcasted blocks, processing them and inserting them to the blockchain if necessary
-    * FIXME: Current implementation is naive in that NewBlock msgs that can't be immediately processed are discarded
-    *        and requested again to the peers
+    * Handles broadcasted blocks, importing them if we're not currently syncing
     */
   def handleBroadcastedBlockMessages: Receive = {
     case MessageFromPeer(NewBlock(newBlock, _), peerId) =>
```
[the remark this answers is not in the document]
rtkaczyk: 🤔

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainTestConfig.scala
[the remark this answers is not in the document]
KonradStaniec: >Interestingly setting the flag to true does cause some tests to fail. I think the safest approach here might be to leave it for now and a task to investigate it once we update to the latest version of the tests. WDYT?

👍

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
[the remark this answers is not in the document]
KonradStaniec: default strategy for dealing with this kind of failures is to restart the child actor so most probably that will happen. 
Also another thing is that here we can get `None` only if the value is not in DB, if the db fails ( throw some kind of exception),  we must deal with it anyway as code for db.get is: 
`Option(db.get((namespace ++ key).toArray))` so we will get this exception even without calling `.get` on option.

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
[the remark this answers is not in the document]
KonradStaniec: >What if the save/remove methods took another optional param, that would determine whether the best block number should be updated and how?

It sounds good

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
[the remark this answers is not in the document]
KonradStaniec: The part `for (block <- staleBlocks) yield blockQueue.enqueueBlock(block)` is needed to enque staleblocks into BlockQueue but the rest of computation (`.lastOption.flatten.map(_.totalDifficulty).getOrElse(0)` ) looks weird

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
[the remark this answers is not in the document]
rtkaczyk: Well, I'm happy to remove this TODO, worst case we'll get a stacktrace pointing to this code line and it should be immediately obvious what went wrong.

There are also a few examples of `Option#get` in the `Ledger`. Have you had a chance to look at them?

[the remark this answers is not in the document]
LukasGasior1: :+1: for removing this comment, it's safe (and if not, we have a bigger problem anyway :P )

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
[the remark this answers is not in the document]
LukasGasior1: IMO, in the end you still need to write multiple keys to db, does it matter which part of code handles that?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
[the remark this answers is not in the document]
rtkaczyk: I don't think it's a good idea. There is no logic related to importing blocks here, it just iterates over the blocks and calls `importBlocks` on them, the only logic is dealing with the import result (`BlockImportedToTop`, `ChainReorganised`, etc.). That's what is meant for `Ledger#importBlocks` - it should be a general use single entry point, and the callers should decide what to do with the result of a single block import. Different callers might want to handle it differently.

src/main/scala/io/iohk/ethereum/ledger/BlockQueue.scala
```diff
@@ -0,0 +1,187 @@
...
+
+
+object BlockQueue {
+  case class QueuedBlock(block: Block, totalDifficulty: Option[BigInt])
+  case class Leaf(hash: ByteString, totalDifficulty: BigInt)
+}
+
+class BlockQueue(blockchain: Blockchain, syncConfig: SyncConfig) extends Logger {
+
+  private val blocks = mutable.Map[ByteString, QueuedBlock]()
```
LukasGasior1: what do you think about making it synchronized in case someone invokes Ledger from multiple threads/actors?

src/main/scala/io/iohk/ethereum/ledger/BlockQueue.scala
```diff
@@ -0,0 +1,187 @@
...
+  private val parentToChildren = mutable.Map[ByteString, Set[ByteString]]()
+
+  /**
+    * Enqueue a block for optional later inclusion into the blockchain
+    *
+    * @param block the block to be enqueued
+    * @return if the newly enqueued block is part of a known branch (rooted somewhere on the main chain), return
+    *         the leaf hash and its total difficulty, otherwise None
+    */
+  def enqueueBlock(block: Block, bestBlockNumber: BigInt = blockchain.getBestBlockNumber()): Option[Leaf] = {
```
LukasGasior1: can you add a comment explaining what's the logic for queueing algorithm?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -1,59 +1,285 @@
...
   def simulateTransaction(stx: SignedTransaction, blockHeader: BlockHeader): TxResult
+
+  def importBlock(block: Block): BlockImportResult
+
+  def resolveBranch(headers: Seq[BlockHeader]): BranchResolutionResult
 }
 
 //FIXME: Make Ledger independent of BlockchainImpl, for which it should become independent of WorldStateProxy type
-class LedgerImpl(vm: VM, blockchain: BlockchainImpl, blockchainConfig: BlockchainConfig) extends Ledger with Logger {
+//TODO: This has grown a bit large, consider splitting the aspects block import, block exec and TX exec
```
LukasGasior1: maybe add a task in YT?

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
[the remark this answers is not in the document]
KonradStaniec: @rtkaczyk you mean that one https://github.com/input-output-hk/mantis/pull/313/files#diff-1b4e47475b368c25eb5141199a4289d0R67 ? I think it should be safe, afaik we are always saving total difficulty when we are saving block, and we already retrieved block in the previous line.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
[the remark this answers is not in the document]
KonradStaniec: sounds reasonable, but if we would add this as function to ledger, i think we would be able to test that logic in blockchain testsuite instead of just doing `blocksToProcess.foreach(ledger.importBlock)` do you think it would be valuable? (if not lets leave it as it is)

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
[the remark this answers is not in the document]
rtkaczyk: @KonradStaniec yes, but also lines 168, 197, 198.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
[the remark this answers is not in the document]
rtkaczyk: In general there are imaginably different strategies on handling `BlockImportResult` for individual blocks, and therefore there cannot be a general case `importBlocks` function.

Even currently, usages of `importBlock` on sequences of blocks are different in `RegularSync` and `BlockchainSuite`. In the former, when handling `BlockBodies` message, we expect the blocks to form a single ordered sub-chain and we give up on execution on the first error. In the latter we are dealing with different sub-chains and `importBlocks` as defined `RegularSync` would simply not work.

Note that `importBlocks` is not even a private method. It's a nested function because it's applicable only to the case of `BlockBodies` message. The function itself is just a functional **loop** and nothing else.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -14,34 +14,31 @@ import io.iohk.ethereum.network.p2p.messages.CommonMessages.NewBlock
...
-  extends Actor with ActorLogging with PeerListSupport with BlacklistSupport with SyncBlocksValidator with BlockBroadcast {
+  extends Actor with ActorLogging with PeerListSupport with BlacklistSupport {
 
   import RegularSync._
   import syncConfig._
 
   private var headersQueue: Seq[BlockHeader] = Nil
   private var waitingForActor: Option[ActorRef] = None
-  private var resolvingBranches: Boolean = false
+  var resolvingBranches: Boolean = false
```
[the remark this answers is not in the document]
rtkaczyk: It's used in the tests

src/main/scala/io/iohk/ethereum/ledger/BlockQueue.scala
rtkaczyk: `synchronized` sounds good as I believe actorising this would be an overkill. But do you think we should do it now?

Would there be any practical downsides of having it synchronized?

Even if do we it, the `Ledger` would still not be thread-safe, we would not to also synchronize the `Blockchain`, right?

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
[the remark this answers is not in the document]
rtkaczyk: It's about making it clear what the function does. For example currently `removeBlock` would set the best block number to `n - 1`, but what if someone at some point wants to remove a block which is not the best block? He would be surprised, at the least.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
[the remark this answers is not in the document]
rtkaczyk: I made some changes in this area, have a look

src/main/scala/io/iohk/ethereum/ledger/BlockQueue.scala
LukasGasior1: as I'm thinking now, maybe it would be enough to add comments on these Ledger methods that they should only be called from one thread (i.e RegularSync). wdyt?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -1,59 +1,319 @@
...
+    * Tries to import the block as the new best block in the chain or enqueue it for later processing
+    * @param block - block to be imported
+    * @return One of:
+    *         - [[BlockImportedToTop]] - if the block was added as the new best block
+    *         - [[BlockEnqueued]] - block is stored in the [[BlockQueue]]
+    *         - [[ChainReorganised]] - a better new branch was found causing chain reorganisation
+    *         - [[DuplicateBlock]] - block already exists either in the main chain or in the queue
+    *         - [[BlockImportFailed]] - block failed to execute (when importing to top or reorganising the chain)
+    */
+  def importBlock(block: Block): BlockImportResult = {
```
AlanVerbner: Is this block being validated before queuing? If not, is there an chance a malicious peer keep sending us invalid nodes to fill up our memory/queue? I know there is a `cleanUp` by block number but not by queued block count.

_PS: sorry if this doesn't make any sense, I was doing a quick pass over this PR_

rtkaczyk: Good point 🤔

rtkaczyk: Even if we do initial validation, the malicious peer could send us blocks that pass this validation and fail to execute later. Then we're not even able to tell which peer sent us this block and blacklist it.

Unless full PoW would be sufficiently prevent a malicious peer from generating such blocks, @LukasGasior1 ?

AlanVerbner: I need to take a deeper look but what about `io.iohk.ethereum.validators.BlockHeaderValidatorImpl#validatePoW` ?

AlanVerbner: ps: I mean, is it enough?

rtkaczyk: Yep, anyway, I will use it now (it will be later improved), but as we discussed with @LukasGasior1, we'll also need to validate the blocks for the accepted difficulty range (we may not have a parent for enqueued block, but we know it's within certain accepted number range). That should take care of spam attacks.

Apart from that do you still think I should implement a size limit to the `BlockQueue`? If so, which blocks should be discarded when reaching the limit? (This smells like a follow-up task 😉 )

rtkaczyk: Maybe this even this validation should be a follow-up task? Pinging @mcsherrylabs

AlanVerbner: Regarding queue limits, I need to take a second look at the code to answer your question. I haven't gone through reorg logic yet (I'm wondering if it can switch to another branch where might be some missing parents but we know it's the longest chain for some reason).  **I think doing this in a separate task is better tho**

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -1,59 +1,319 @@
...
+    }
+  }
+
+  /**
+    * Remove blocks from the [[Blockchain]] along with receipts and total difficulties
+    * @param parent remove blocks until this hash (exclusive)
+    * @param fromNumber start removing from this number (downwards)
+    * @return the list of removed blocks along with receipts and total difficulties
+    */
+  private def removeBlocksUntil(parent: ByteString, fromNumber: BigInt): List[(Block, Seq[Receipt], BigInt)] = {
```
AlanVerbner: I think we might have an issue here. In this function we are removing blocks from blockchain but we are not "undoing" state changes.

Also, while reviewing the code, @ntallar discovered (and I agree after discussion) that current pruning implementation has a bug that will cause to remove valid state nodes after reorg (specific details are no worth to be mentioned here imo). I think we will need to define a way to let the pruning know that we are doing a blockchain reorg.

src/main/resources/application.conf
```diff
@@ -6,7 +6,7 @@ mantis {
   client-version = "mantis/v0.1"
 
   # Base directory where all the data used by the node is stored, including blockchain data and private keys
-  datadir = ${user.home}"/.mantis"
+  datadir = "/Users/radek/scalac/iohk/synctest-refactor"
```
LukasGasior1: needs a fix

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
LukasGasior1: @rtkaczyk are you adding validating if difficulty is within required range in this PR?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
rtkaczyk: Yes, that was intentional. Even if we didn't prune those extra nodes, thy shouldn't be a problem.
But if valid nodes may be removed that's a different story... I'm keen on getting to know the details, and what we can do about it...

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
rtkaczyk: Yes, I am

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
AlanVerbner: Currently we have a super naive pruning approach based on MptNode reference count: 
- We increment when someone inserts it and decrement when it's removed
- We track the block number when mpt node reference count reaches to zero, lets call it `Bn`
- We define a parameter `Offset` we will "hold" that node
- if at `Bn` + `Offset` references are still 0, we remove mpt node as it might not be used by anyone else

**The problem is that we are not updating reference count when dropping block executions after a reorg (actually we would need to change internal implementation but that shouldn't be a problem from the Ledger point of view)**

The following example demonstrates why we need to let know pruning algo that we are undoing a block:
- `Offset` = 5
- Let's suppose that we execute the blocks in the following order: b1, b2, b3, b4, b5', [Reorg], b5, b6, b7, b8, b9, b10
- At b2, NodeX references is 1
- At b5' NodeX references is 0 and marked as prune candidate
- At b10 we will remove NodeX as it still has 0 references but we shouldn't because b5' changes do not count after doing a reorg so reference count should be 1

prune_example

rtkaczyk: Thanks for this explanation, indeed we have a problem. However it seems it's not a new problem, related to this PR, as we already are doing re-orgs [by removing blocks](https://github.com/input-output-hk/mantis/blob/phase/daedalus/src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala#L182) 🙁

AlanVerbner: @rtkaczyk yes, the bug was present before this PR. The good thing is:
- It forced us to rethink about pruning in case of reorgs <3
- We need to agree on an interface that should be called from `Ledger` when doing a reorg (something _like_ `undoBlock(block)` )

src/main/scala/io/iohk/ethereum/ledger/BlockQueue.scala
```diff
@@ -0,0 +1,193 @@
...
+      children.foreach(removeSubtree)
+      blocks -= block.header.hash
+      parentToChildren -= block.header.hash
+    }
+
+  /**
+    * Removes stale blocks - too old or too young in relation the current best block number
+    * @param bestBlockNumber - best block number of the main chain
+    */
+  private def cleanUp(bestBlockNumber: BigInt): Unit = {
```
AlanVerbner: Shouldn't we remove the whole subtree when removing a block because it's old?

rtkaczyk: Well, we could but there is tiny chance that after asking a peer for headers those "abandoned" children may get usable again. Other than that, they will eventually be removed anyway, and it's just slightly easier to remove them one by one.

I can change it to `removeSubtree` though, if you're convinced we should be doing it that way.

AlanVerbner: My question was in case I was missing something, ie: keeping the children for a little bit longer had a specific use case. I think it's ok leaving as it (a small comment wont hurt tho :wink:)

AlanVerbner (approved):
great piece of code :rocket:

KonradStaniec (approved):
LGTM

PR #314, opened 2017-09-28
[EC-307] Feature/timeout account unlock

KonradStaniec:
Unlocking of account only for a specified amount of time, implemented as passive expiration map.


ntallar:
It seems both [Geth|https://github.com/ethereum/go-ethereum/wiki/Management-APIs#personal_unlockaccount] and [Parity|https://github.com/paritytech/parity/wiki/JSONRPC-personal-module] support passing 0 duration to set an unlimited unlock, should we also allow that? I'm not sure whether it's needed so it's up to discussion.

KonradStaniec:
Good catch @ntallar, I think it is a good idea as users from other clients should be familiar with this possibility.

src/main/scala/io/iohk/ethereum/jsonrpc/ExpiringMap.scala
```diff
@@ -0,0 +1,60 @@
...
+import scala.collection.mutable
+import scala.util.Try
+
+object ExpiringMap {
+
+ case class ValueWithDuration[V](value: V, expiration: Instant)
+
+  def empty[K, V](defaultElementRetentionTime: Duration,
+                  clock: Clock = Clock.systemUTC()): ExpiringMap[K, V] =
+    new ExpiringMap(mutable.Map.empty, clock, defaultElementRetentionTime)
```
AlanVerbner: I know this wasn't introduced in this PR but I'm not sure using this map is the best alternative here as, If i'm not mistaken, our Http server accepts multiple concurrent requests.

rtkaczyk: I wonder if using `Clock` is the way to go. Theoretically an attacker could keep the key unlocked indefinitely by adjusting the system clock.

KonradStaniec: @AlanVerbner yep I pondered on this one a little bit, but eventually, I decided that if the map was good enough till now I will leave it that way. If we are starting to worry about concurrency issues the simplest and most foolproof way would be to wrap this `ExpiringMap` in Actor to ensure proper comunication.  What do you think ? 

@rtkaczyk interesting point, but isn't it that the case, that if an attacker could adjust our system clock we are already so compromised that keeping our key unlocked indefinitely is the least of our problems? I mean `Instant.now(Clock.systemUTC)` under the hood just calls `System.currentTimeMillis()`, and to the attacker to tinker with that, he would have to have full admin rights over the machine. Of course, I am not cybersec specialist so I am maybe missing something. Anyhow, we need some notion of passing time.

rtkaczyk: I'm not sure how much of a security issue this is. The only plausible scenario I can think of is that once someone took control of the machine running Mantis, they can buy themselves some time to figure out what do about unlocked keys (if any).

Anyway, if we could use `System.nanoTime()` here that should make us resilient to any time changes, and that also includes some potentially confusing behaviour in the event of DST or geographical time changes.

As for the `Map`, given how this API is used , this bug is very unlikely to manifest, so maybe a todo comment / separate task?

KonradStaniec: 👍 for both suggestions

AlanVerbner: Todo comment works for me

src/main/scala/io/iohk/ethereum/jsonrpc/ExpiringMap.scala
```diff
@@ -9,34 +10,35 @@ import scala.util.Try
...
+    new ExpiringMap(mutable.Map.empty, defaultElementRetentionTime)
 }
 /**
   * Simple wrapper around mutable map which enriches each element with expiration time (specified by user or default)
-  * Map is passive which means it only check for expiration and remove expired element during get function
+  * Map is passive which means it only check for expiration and remove expired element during get function.
+  * Duration in all calls is relative to current System.nanoTime()
   */
+//TODO: Make class thread safe
 class ExpiringMap[K, V] private (val underLaying: mutable.Map[K, ValueWithDuration[V]],
```
rtkaczyk: Typo:

`underLaying` -> `underlying`

src/main/scala/io/iohk/ethereum/jsonrpc/ExpiringMap.scala
```diff
@@ -9,34 +10,35 @@ import scala.util.Try
...
-                                 val clock: Clock,
                                  val defaultRetentionTime: Duration) {
+  private val maxHoldDuration = ChronoUnit.CENTURIES.getDuration
 
-  def addUntil(k: K, v: V, until: Instant): ExpiringMap[K, V] = {
-    underLaying += k -> ValueWithDuration(v, until)
+  def addUntil(k: K, v: V, duration: Duration): ExpiringMap[K, V] = {
+    underLaying += k -> ValueWithDuration(v, Try(currentPlus(duration)).getOrElse(currentPlus(maxHoldDuration)))
```
rtkaczyk: What can fail here?

src/main/scala/io/iohk/ethereum/jsonrpc/ExpiringMap.scala
```diff
@@ -55,6 +57,13 @@ class ExpiringMap[K, V] private (val underLaying: mutable.Map[K, ValueWithDurati
   }
 
   private def isNotExpired(value: ValueWithDuration[V]) =
-    Duration.between(value.expiration, Instant.now(clock)).isNegative
+    currentNanoDuration.minus(value.expiration).isNegative
+
+  private def currentPlus(duration: Duration) =
+    currentNanoDuration.plus(duration)
+
+  private def currentNanoDuration =
```
rtkaczyk: Style: it's suggested to use parenthesis for side-effecting function (`def currentNanoDuration()`). This form suggests pure function.

src/main/scala/io/iohk/ethereum/jsonrpc/ExpiringMap.scala
```diff
@@ -0,0 +1,69 @@
...
+  * Simple wrapper around mutable map which enriches each element with expiration time (specified by user or default)
+  * Map is passive which means it only check for expiration and remove expired element during get function.
+  * Duration in all calls is relative to current System.nanoTime()
+  */
+//TODO: Make class thread safe
+class ExpiringMap[K, V] private (val underLaying: mutable.Map[K, ValueWithDuration[V]],
+                                 val defaultRetentionTime: Duration) {
+  private val maxHoldDuration = ChronoUnit.CENTURIES.getDuration
+
+  def addUntil(k: K, v: V, duration: Duration): ExpiringMap[K, V] = {
```
rtkaczyk: _until_ means a point in time. We're adding _for_ a duration.

src/main/scala/io/iohk/ethereum/jsonrpc/ExpiringMap.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  def empty[K, V](defaultElementRetentionTime: Duration): ExpiringMap[K, V] =
+    new ExpiringMap(mutable.Map.empty, defaultElementRetentionTime)
+}
+/**
+  * Simple wrapper around mutable map which enriches each element with expiration time (specified by user or default)
+  * Map is passive which means it only check for expiration and remove expired element during get function.
+  * Duration in all calls is relative to current System.nanoTime()
+  */
+//TODO: Make class thread safe
```
rtkaczyk: 👍

src/main/scala/io/iohk/ethereum/jsonrpc/ExpiringMap.scala
```diff
@@ -0,0 +1,69 @@
...
+class ExpiringMap[K, V] private (val underLaying: mutable.Map[K, ValueWithDuration[V]],
```
rtkaczyk: Typo:

`underLaying` -> `underlying`

src/main/scala/io/iohk/ethereum/jsonrpc/ExpiringMap.scala
```diff
@@ -0,0 +1,69 @@
...
+                                 val defaultRetentionTime: Duration) {
+  private val maxHoldDuration = ChronoUnit.CENTURIES.getDuration
+
+  def addUntil(k: K, v: V, duration: Duration): ExpiringMap[K, V] = {
+    underLaying += k -> ValueWithDuration(v, Try(currentPlus(duration)).getOrElse(currentPlus(maxHoldDuration)))
```
rtkaczyk: What can fail here?

src/main/scala/io/iohk/ethereum/jsonrpc/ExpiringMap.scala
```diff
@@ -0,0 +1,69 @@
...
+}
+/**
+  * Simple wrapper around mutable map which enriches each element with expiration time (specified by user or default)
+  * Map is passive which means it only check for expiration and remove expired element during get function.
+  * Duration in all calls is relative to current System.nanoTime()
+  */
+//TODO: Make class thread safe
+class ExpiringMap[K, V] private (val underLaying: mutable.Map[K, ValueWithDuration[V]],
+                                 val defaultRetentionTime: Duration) {
+  private val maxHoldDuration = ChronoUnit.CENTURIES.getDuration
```
rtkaczyk: Is this to represent _forever_? `Duration.Inf`?

src/main/scala/io/iohk/ethereum/jsonrpc/ExpiringMap.scala
KonradStaniec: for example `Duration.ofNanos(System.nanoTime()).plus(Duration.ofSeconds(Long.MaxValue))` can overflow and give us`java.lang.ArithmeticException: long overflow`

src/main/scala/io/iohk/ethereum/jsonrpc/ExpiringMap.scala
rtkaczyk: Oh, I thought we were using `scala.concurrent.duration.Duration` rather than `java.time.Duration`. I guess that's ok, but is there any reason for this choice?

KonradStaniec: `java.time` is enough for our simple use case, and we are avoiding the middleman. (some implicit conversions beneath, additional object creation etc.)
For an external user of class, i think 100 years is enough to think of it as forever.
Also as docs states `This class is not meant as a general purpose representation of time, it is optimized for the needs of scala.concurrent.` and we are not using any `scala.concurrent` here.

PR #315, opened 2017-09-29
EC-185: Add timeout for pending transactions

LukasGasior1:
This PR adds a timeout for pending transactions in the pool. This is a simple mechanism that prevents some cases when transaction that is not going to be mined occupies a slot in the pool indefinitely.


src/test/scala/io/iohk/ethereum/transactions/PendingTransactionsManagerSpec.scala
```diff
@@ -155,6 +155,30 @@ class PendingTransactionsManagerSpec extends FlatSpec with Matchers with ScalaFu
...
+    override val pendingTransactionsManager = system.actorOf(
+      PendingTransactionsManager.props(txPoolConfig, peerManager.ref, etcPeerManager.ref, peerMessageBus.ref))
+
+    val stx = newStx()
+    pendingTransactionsManager ! AddTransactions(stx)
+
+    val pendingTxs = (pendingTransactionsManager ? GetPendingTransactions).mapTo[PendingTransactionsResponse].futureValue
+    pendingTxs.pendingTransactions.map(_.stx).toSet shouldBe Set(stx)
+
+    Thread.sleep(1100)
```
rtkaczyk: Is there a chance we can make this wait shorter? A second here, a second there...

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
```diff
@@ -99,12 +107,24 @@ class PendingTransactionsManager(txPoolConfig: TxPoolConfig, peerManager: ActorR
...
       signedTransactions.foreach(setTxKnown(_, peerId))
   }
 
+  private def setTimeout(stx: SignedTransaction): Unit = {
+    timeouts.get(stx.hash).map(_.cancel())
+    val cancellable = context.system.scheduler.scheduleOnce(txPoolConfig.transactionTimeout, self, RemoveTransactions(Seq(stx)))
+    timeouts += (stx.hash -> cancellable)
+  }
+
+  private def clearTimeout(stx: SignedTransaction): Unit = {
```
rtkaczyk: The name... I'd say what we do about the `stx` is more essential. Also `clearTimeout` may sound like we reset it. What about `forgetTransaction` or something like that?

src/main/resources/application.conf
```diff
@@ -180,6 +180,8 @@ mantis {
     tx-pool-size = 1000
 
     pending-tx-manager-query-timeout = 5.seconds
+
+    transaction-timeout = 2.minutes
```
AlanVerbner: Super minor: Why 2 minutes?

LukasGasior1: @AlanVerbner  I think it's a reasonable default, but I'm open to other suggestions

AlanVerbner: @LukasGasior1 actually your value seems to be quite good. Maybe we can think of this as a number of blocks function like RSK guys and leave a comment explaining the rationale like: https://github.com/rsksmart/rskj/blob/c739053e23ef26baa43f069afb20d32fff1710b7/rskj-core/src/main/resources/config/rsk-sample.conf#L89

wdyt?

LukasGasior1: 120/14 gives around 9 blocks, pretty similar

src/test/scala/io/iohk/ethereum/transactions/PendingTransactionsManagerSpec.scala
LukasGasior1: how about half a second?

src/main/resources/application.conf
AlanVerbner: yeah, I was suggesting to leave a comment behind the rationale of that value so it doesn't look arbitrary for someone that might look at it

PR #317, opened 2017-10-04
[EC-262] Add verify jars plugin

LukasGasior1:
It requires to build this plugin locally: https://github.com/input-output-hk/sbt-verify (and first merge this PR there https://github.com/input-output-hk/sbt-verify/pull/1 )


build.sbt
```diff
@@ -47,6 +47,61 @@ val dep = {
   )
 }
 
+val verifyDeps = Seq(
```
ntallar: There are a couple jars that are only used for testing, but their signature is not checked, should it be also checked?
```
"com.typesafe.akka" %% "akka-http-testkit" % akkaHttpVersion % "it,test"
"com.miguno.akka" %% "akka-mock-scheduler" % "0.5.1" % "it,test"
```

build.sbt
```diff
@@ -60,7 +115,18 @@ val Snappy = config("snappy") extend Test
 val root = project.in(file("."))
     .configs(Integration, Benchmark, Evm, Ets, Snappy)
     .settings(commonSettings: _*)
-    .settings(libraryDependencies ++= dep)
+    .settings(
```
AlanVerbner: I think it would be a good idea to make `dist` task dependant on `verify`. This will force to run checks before releasing a version. Something like `dist in Universal := ((dist in Universal) dependsOn verify).value`

What do you think?

LukasGasior1: Good idea

build.sbt
LukasGasior1: They don't go into the final package anyway, so I'm not sure if it's worth doing that (it would probably require some changes in the plugin code)

circle.yml
```diff
@@ -1,6 +1,7 @@
 dependencies:
   pre:
     - sudo add-apt-repository -y ppa:ethereum/ethereum; sudo apt-get update; sudo apt-get install -y solc
+    - git clone https://github.com/input-output-hk/sbt-verify.git; cd sbt-verify; sbt publishLocal
```
AlanVerbner: This is not a blocker at all but this way we are trusting that no one has compromised this repo and default branch. We need to keep an eye on it and check it's permissions.

> Good security means trying to limit the damage a Trusted role can do, even if you can't completely eliminate the Trusted role. [1](https://www.linkedin.com/pulse/id-known-what-we-were-starting-ray-dillinger/)

LukasGasior1: You're right...I wonder if it would be possible to include a jar with the plugin in our repository itself

PR #318, opened 2017-10-06
[EC-299] separate max incoming peers into connection and handshaked

KonradStaniec:
Done only for incoming connections as`geth` and `parity` seems doing it that way.


src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -227,4 +267,42 @@ object PeerManagerActor {
   }
 
   case class SendMessage(message: MessageSerializable, peerId: PeerId)
+
+  class PeerManagerState(var pendingIncomingPeers: Map[PeerId, Peer], var peers: Map[PeerId, Peer]) {
```
AlanVerbner: I know this is only accessed within the Actor but I would prefer not to use `var`

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -75,97 +73,138 @@ class PeerManagerActor(
...
       (peer.ref ? PeerActor.GetStatus)
         .mapTo[PeerActor.StatusResponse]
         .map { sr => Success((peer, sr.status)) }
         .recover { case ex => Failure(ex) }
     }.map(r => Peers.apply(r.collect { case Success(v) => v }.toMap))
   }
 
+
+  private def connectionAlreadyHandled(remoteAddress: InetSocketAddress, error: ConnectionError): Either[ConnectionError, InetSocketAddress] = {
+    if (!managerState.isConnectionHandled(remoteAddress))
```
AlanVerbner: What about `Either.cond(!managerState.isConnectionHandled(remoteAddress), remoteAddress, error)` ?

src/test/resources/application.conf
```diff
@@ -10,6 +10,10 @@ mantis {
 
   network.peer.max-peers = 3
```
AlanVerbner: This one then becomes `max-outgoing-peers` right? Maybe it's worth renaming it just to make it more clear

AlanVerbner: ps: Same comment for `src/test/resources/application.conf`

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
KonradStaniec: This class only purpose is to encapsulate state  operations of `PeerManagerActor` and we encode actor state operations in all our actors via var`s or mutable collections.
Do you prefer mutable collections or you mean to re-design this actor to mutate all state via `context.become` (theoretically it most pure way, but I am not convinced it is worth it for just one actor)

AlanVerbner: I was thinking about `context.become` but you have a point that might not be worthy. 

What do you think @LukasGasior1 ?

KonradStaniec: @AlanVerbner what do you think about adding `TODO` and task on youtrack to refactor this Actor to more pure style, as i think it would be good thing but also it exceeds the scope of this particular pr ?

AlanVerbner: @KonradStaniec That's a good idea.

PR #319, opened 2017-10-11
[EC-314] FIX: Pruning may fail if a node is updated after marked as prune candidate

AlanVerbner:
# Description

In some rare cases, we might delete an MPT node while doing pruning that might be used after a reorg.

**Important: This is nor related with randomly missing mpt nodes as this should be deterministic on a regular sync run**

# How to reproduce

1. If a mpt node is marked as prune candidate (which means it can be deleted after X blocks because it has no more references) at block 10
2. Mpt node is added again at block 11
3. Mpt node is deleted at block 12 

When pruning candidates for block 10, it will delete mpt node

# Fix

When looking for prune candidates, check by current references **and** last mpt node usage


AlanVerbner:
Closing this PR as pruning implementation will change

PR #321, opened 2017-10-15
Refactor DataSource.update API

AlanVerbner:
# Description

The following PR proposes a refactor on `io.iohk.ethereum.db.dataSource.DataSource#update` API in order to make it more clear. 

Currently it receives two collections: keys to remove and data to upsert. That way it will execute all the removals first and then the upserts. That doesn't feel natural for the user that might want to create an operations batch to do everything at once. 
For example, if we want to: `(Insert, k1, v1)` and then `(Remove, k1)` two calls to `DataSource#update` are neededed in order to prevent update to execute `Remove` and `Insert` leaving something in the DB.

**_Note: This refactor is not urgent as it's not fixing any bugs in our code_**

## Proposal
Update API in order to use `batchOperations: Seq[BatchOperation[Key, Value]]`. This is not a complex change and we have a bunch of tests to be sure we don't introduce any bugs.

## Todo
- [x] Update IODB DataSource
- [ ] Take a second look at the code before removing WIP
- [ ] Maybe defining an operand to avoid adding `Upsert` and `Removal`

### DIFF
--- a/src/it/scala/io/iohk/ethereum/db/DataSourceIntegrationTestBehavior.scala
+++ b/src/it/scala/io/iohk/ethereum/db/DataSourceIntegrationTestBehavior.scala
@@ -5,6 +5,7 @@ import java.nio.file.Files
 
 import akka.util.ByteString
 import io.iohk.ethereum.ObjectGenerators
+import io.iohk.ethereum.common.{Removal, Upsert}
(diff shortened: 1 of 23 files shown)


mirkoAlic:
[TOO OLD]

PR #322, opened 2017-10-16
[EC-88] Add PoW validation

LukasGasior1:
TODO ~(in separate PR)~:
- [x] - improve handling of already calculated cache for current epoch (currently stored in var in Validator)
- [x] - fix block generator tests (they seem to use invalid blocks)
- [x] - investigate how to improve performance (currently sync is significantly slower with PoW validation)


src/test/scala/io/iohk/ethereum/consensus/EthashSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+      Ethash.cacheSize(blockNumber) shouldBe referenceSize
+    }
+  }
+
+  it should "compute hash" in {
+    val hash = Array(0xf5, 0x7e, 0x6f, 0x3a, 0xcf, 0xc0, 0xdd, 0x4b, 0x5b, 0xf2, 0xbe, 0xe4, 0x0a, 0xb3, 0x35, 0x8a, 0xa6, 0x87, 0x73, 0xa8, 0xd0, 0x9f, 0x5e, 0x59, 0x5e, 0xab, 0x55, 0x94, 0x05, 0x52, 0x7d, 0x72).map(_.toByte)
+    val nonce = Array(0xd7, 0xb3, 0xac, 0x70, 0xa3, 0x01, 0xa2, 0x49).map(_.toByte)
+
+    val mixHash = Array(0x1f, 0xff, 0x04, 0xce, 0xc9, 0x41, 0x73, 0xfd, 0x59, 0x1e, 0x3d, 0x89, 0x60, 0xce, 0x6b, 0xdf, 0x8b, 0x19, 0x71, 0x04, 0x8c, 0x71, 0xff, 0x93, 0x7b, 0xb2, 0xd3, 0x2a, 0x64, 0x31, 0xab, 0x6d).map(_.toByte)
+    val boundary = Array(0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x3e, 0x9b, 0x6c, 0x69, 0xbc, 0x2c, 0xe2, 0xa2, 0x4a, 0x8e, 0x95, 0x69, 0xef, 0xc7, 0xd7, 0x1b, 0x33, 0x35, 0xdf, 0x36, 0x8c, 0x9a, 0xe9, 0x7e, 0x53, 0x84).map(_.toByte)
```
rtkaczyk: How do you know these are correct? 😁 
Could we maybe have more examples (eg. as a table - `forAll(table)`) ?

rtkaczyk:
From ethash spec:

> This spec is REVISION 23. ... in all implementations please include a spec revision number

Should we do that?

src/test/scala/io/iohk/ethereum/consensus/EthashSpec.scala
LukasGasior1: This one is copied from Parity :P I can grab it from few blocks and make a table

src/main/scala/io/iohk/ethereum/consensus/Ethash.scala
```diff
@@ -0,0 +1,234 @@
...
+    val sz = (CACHE_BYTES_INIT + CACHE_BYTES_GROWTH * epoch(blockNumber)) - HASH_BYTES
+    highestPrimeBelow(sz, HASH_BYTES)
+  }
+
+  def dagSize(blockNumber: Long): Long = {
+    val sz = DATASET_BYTES_INIT + DATASET_BYTES_GROWTH * epoch(blockNumber) - MIX_BYTES
+    highestPrimeBelow(sz, MIX_BYTES)
+  }
+
+  private def highestPrimeBelow(n: Long, len: Long): Long = {
```
ntallar: Minor detail: Maybe we could use `@tailrec` annotation here?

src/main/scala/io/iohk/ethereum/consensus/Ethash.scala
```diff
@@ -0,0 +1,234 @@
...
+    if (isPrime(n / len)) n
+    else highestPrimeBelow(n - 2 * len, len)
+  }
+
+  private def isPrime(n: BigInt): Boolean =
+    !((2 until n.toInt-1) exists (n % _ == 0))
```
ntallar: It seems that `isPrime(1)` returns true, could that bring any issues to the implementation?

ntallar: Could this `n.toInt` convertion bring any conversion issues as it is being used in `highestPrimeBelow` were all the number used are `Long`?
For example when `blockNumber=1000000` `(DATASET_BYTES_INIT + DATASET_BYTES_GROWTH * epoch(x) - MIX_BYTES) / MIX_BYTES` seems to be greater than `Int.MaxValue`.

rtkaczyk: Also, a simple optimisation: it's enough to iterate through numbers up to (and including) `floor(sqrt(n))`.

It this function used often? If yes, and if we know that `n < 2 ^ 31`, then [Fermat primality test](https://en.wikipedia.org/wiki/Primality_test#Fermat_primality_test) would be a simple-to-code, yet very efficient implementation. A TODO comment is also fine :)

LukasGasior1: Good catch, I'll make it `toLong`

LukasGasior1: I've changed the implementation a bit so that it works with BigInts

LukasGasior1:
I've updated this PR with a way of storing precomputed PoW cache (a cache for cache). `MaxPowCaches` recent caches are now stored in a hash map in BlockHeaderValidatorImpl.

src/test/scala/io/iohk/ethereum/mining/BlockGeneratorSpec.scala
```diff
@@ -24,7 +24,8 @@ import scala.concurrent.duration.FiniteDuration
 
 class BlockGeneratorSpec extends FlatSpec with Matchers with PropertyChecks with Logger {
 
-  "BlockGenerator" should "generate correct block with empty transactions" in new TestSetup {
+  // fixme: block used in this test does not pass PoW validation
+  ignore should "generate correct block with empty transactions" in new TestSetup {
```
rtkaczyk: If I understand correctly these tests validate a fully mined block. Does this mean, it's not just these tests, but actually our mining is broken?

LukasGasior1: we do not have mining (yet). These blocks were generated using ethminer and I think something's wrong with them (possibly becuase they were mined on priv network). What I noticed is that if I force validator to use epoch 1 instead of 0 (which should be used for block with num=1) they work.

src/main/scala/io/iohk/ethereum/validators/BlockHeaderValidator.scala
```diff
@@ -15,13 +15,19 @@ object BlockHeaderValidatorImpl {
   val GasLimitBoundDivisor: Int = 1024
   val MinGasLimit: BigInt = 5000 //Although the paper states this value is 125000, on the different clients 5000 is used
   val MaxGasLimit = Long.MaxValue // max gasLimit is equal 2^63-1 according to EIP106
+  val MaxPowCaches: Int = 2 // maximum number of epochs for which PoW cache is stored in memory
+
+  case class PowCacheData(cache: Array[Int], dagSize: Long)
 }
 
 class BlockHeaderValidatorImpl(blockchainConfig: BlockchainConfig) extends BlockHeaderValidator {
```
ntallar: Maybe I'm missing something but it seems another `BlockHeaderValidator` separate instance is [being created in the `OmmersValidator`](https://github.com/input-output-hk/mantis/blob/phase/daedalus/src/main/scala/io/iohk/ethereum/validators/OmmersValidator.scala#L28), won't each of them have to create their respective caches separately? This is up to discussion but maybe we could avoid having to create this duplicated caches.

LukasGasior1: wow this is a good catch! I was pretty sure we were using only a single instance of validator, totally missed that ommers validator creates a new one. I'll fix it.

src/test/scala/io/iohk/ethereum/mining/BlockGeneratorSpec.scala
LukasGasior1: I've mined these blocks again with ethminer, they now work as expected

src/main/scala/io/iohk/ethereum/validators/BlockHeaderValidator.scala
```diff
@@ -15,13 +15,19 @@ object BlockHeaderValidatorImpl {
   val GasLimitBoundDivisor: Int = 1024
   val MinGasLimit: BigInt = 5000 //Although the paper states this value is 125000, on the different clients 5000 is used
   val MaxGasLimit = Long.MaxValue // max gasLimit is equal 2^63-1 according to EIP106
+  val MaxPowCaches: Int = 2 // maximum number of epochs for which PoW cache is stored in memory
+
+  case class PowCacheData(cache: Array[Int], dagSize: Long)
```
rtkaczyk: Detail: I'd rather not make it a **case** class since it has a mutable element (default impls of `equals` and `hashCode` are invalid)

src/main/scala/io/iohk/ethereum/validators/BlockHeaderValidator.scala
```diff
@@ -15,13 +15,19 @@ object BlockHeaderValidatorImpl {
...
 }
 
 class BlockHeaderValidatorImpl(blockchainConfig: BlockchainConfig) extends BlockHeaderValidator {
 
   import BlockHeaderValidatorImpl._
   import BlockHeaderError._
 
+  // we need concurrent map since validators can be used from multiple places
+  val powCaches: java.util.Map[Long, PowCacheData] = new java.util.concurrent.ConcurrentHashMap[Long, PowCacheData]()
```
rtkaczyk: Detail: could we add `.asScala` here and have thereafter interactions more convenient?

LukasGasior1:
@rtkaczyk @ntallar I've updated the branch, I need your approvals/reviews again

PR #323, opened 2017-10-18
[EC-306] Feature/new block hashes

KonradStaniec:
Still, need to be tested on private network or on top of the main chain


src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -112,6 +123,56 @@ class RegularSync(
...
 
+  /**
+    * Handles NewHashesMessege, should only cover this message when we are top of the chain
+    */
+  def handleNewBlockHashesMessages: Receive = {
+    case MessageFromPeer(NewBlockHashes(hashes), peerId) =>
+      val maybePeer = peersToDownloadFrom.find(peer => peer._1.id == peerId)
+      //we allow asking for new hashes when we are not syncing and we can download from specified peer and we are
+      //top of the chain
+      if (notDownloading() && topOfTheChain && maybePeer.isDefined) {
```
ntallar: As they are similar, should we maybe also check that we are in the `topOfTheChain` when handling `NewBlock` messages?

In relation to that, could it be the case that our client stops being in the top of the chain (as if there is any network issues)? Could not detecting that affect properly handling `NewBlockHashes` messages?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -112,6 +123,56 @@ class RegularSync(
...
+        val (peer, _) = maybePeer.get
+        val hashesToCheck = hashes.take(syncConfig.maxNewHashes)
+
+        if (!containsAncientBlockHash(hashesToCheck)) {
+          val filteredHashes = getValidHashes(hashesToCheck)
+
+          if (filteredHashes.nonEmpty) {
+            val request = GetBlockHeaders(Right(filteredHashes.head.hash), filteredHashes.length, BigInt(0), reverse = false)
+            requestBlockHeaders(peer, request)
+            resolvingBranches = false
```
ntallar: Should we maybe also cancel the scheduled `ResumeRegularSync` if there is one? If I understood correctly, when processing the requested headers we should always end up scheduling a `ResumeRegularSync`.
Also, as in `scheduleResume()` we don't cancel the previously scheduled `ResumeRegularSync` we could end up with 2 of them scheduled (as would happen if the [request for the headers associated with the hashes fail](https://github.com/input-output-hk/mantis/blob/feature/new_block_hashes/src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala#L193)) and I'm not sure exactly if there could potentially be any issue with how we handle that.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -112,6 +123,56 @@ class RegularSync(
       }
   }
 
+  /**
+    * Handles NewHashesMessege, should only cover this message when we are top of the chain
+    */
+  def handleNewBlockHashesMessages: Receive = {
+    case MessageFromPeer(NewBlockHashes(hashes), peerId) =>
```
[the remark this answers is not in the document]
KonradStaniec: Actually I don't think that other clients do this. Also later we are requesting hashes with something like `GetBlockHeaders(Right(filteredHashes.head.hash), filteredHashes.length` ) so we should receive consecutive block headers, from first `UnknownHash` .
Another thing is that most of the time `hashes` have only element (when regular syncing about 1milion blocks and logging all hashes I have never seen it have more than one hash)

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
KonradStaniec: Yep we could definitely check it also for `NewBlock` message, as currently even if we are still far away from a top, we are receiving new blocks and try to import them. It ends up  in https://github.com/input-output-hk/mantis/blob/e6d14f5abbae5d1ad3257df39d8793a8c0997f67/src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala#L105
So it would be minor optimisation.

As for possible network issues, currently, i decided to go for most simple solution i.e once we reach top of the chain once, we are turning on handling `NewBlockHashes` permanently. It can have some downsides like, if we will be left behind by to many blocks because of this network issue, we will be receiving too fresh `NewBlockHashes` and handling them will slow down our regular `askForHeaders` cycle. (though I am still not sure how much, it needs to be tested in action)

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
KonradStaniec: I think you have valid point here and we should cancel it. 👍

src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -129,6 +129,10 @@ object Config {
 
     val maxQueuedBlockNumberAhead: Int
     val maxQueuedBlockNumberBehind: Int
+
+    // HardCoded values based on parity client
+    val maxNewBlockHashAge = 20
+    val maxNewHashes = 64
```
ntallar: Minor detail: Maybe we could make this values be obtained from the `application.conf` file?

src/main/resources/application.conf
```diff
@@ -379,6 +379,13 @@ mantis {
     # if: block_number - current_best_block_number > max-queued-block-number-ahead
     # then: the block will not be queued (such already queued blocks will be removed)
     max-queued-block-number-ahead = 10
+
+    # Maxium number of blocks, after which block hash from NewBlockHashes packet is considered ancient
+    # and peer sending it is blacklisted
```
rtkaczyk: What does _number of blocks_ mean? Is it how old a hash is in relation to our current known best block number?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -112,6 +127,57 @@ class RegularSync(
       }
   }
 
+  /**
+    * Handles NewHashesMessege, should only cover this message when we are top of the chain
+    */
+  def handleNewBlockHashesMessages: Receive = {
```
rtkaczyk: `NewBlockHashes` is a broadcast message, isn't it? So I'd say this should go into `handleBroadcastedBlockMessages` or `handleBroadcastedBlockMessages` should be renamed (eg. `handleNewBlockMessages`).

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -112,6 +127,57 @@ class RegularSync(
...
+    case MessageFromPeer(NewBlockHashes(hashes), peerId) =>
+      val maybePeer = peersToDownloadFrom.find(peer => peer._1.id == peerId)
+      //we allow asking for new hashes when we are not syncing and we can download from specified peer and we are
+      //top of the chain
+      if (notDownloading() && topOfTheChain && maybePeer.isDefined) {
+        val (peer, _) = maybePeer.get
+        val hashesToCheck = hashes.take(syncConfig.maxNewHashes)
+
+        if (!containsAncientBlockHash(hashesToCheck)) {
```
rtkaczyk: If I understand correctly, we're only blacklisting the sender of `NewBlockHashes` if sends us an `ancient` peer. What if it sends us some nonsense hashes (with sensible numbers)? How would it affect our syncing if malicious peers kept sending us such messages causing us to query other peers for blocks they do not know?

src/main/resources/application.conf
KonradStaniec: Yes, exactly it is it, in other words `distance (in number of blocks) of received hash from our current best block`

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
KonradStaniec: valid point, I would go with renaming ` handleBroadcastedBlockMessages` to `handleNewBlockMessages`

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
KonradStaniec: As I see it if some malicious peer would keep sending us some procured `newBlockHashMessage` (with some valid number )we would ask only this peer for this headers via `GetBlockHeaders` message. And it depends on his answer to this: 

If he would send us some fake headers we would go with our usual strategy ( handle headers and then bodies). It would probably finish with some validation fail during import block and we would blacklist peer. 

If he would respond to us with empty `BlockHeaders` message, we would simply reschedule our normal ask for headers cycle.

So we would never query other peers for some hashes that one malicious peer is sending to us.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -133,9 +133,9 @@ class RegularSync(
   def handleNewBlockHashesMessages: Receive = {
     case MessageFromPeer(NewBlockHashes(hashes), peerId) =>
       val maybePeer = peersToDownloadFrom.find(peer => peer._1.id == peerId)
-      //we allow asking for new hashes when we are not syncing and we can download from specified peer and we are
-      //top of the chain
-      if (notDownloading() && topOfTheChain && maybePeer.isDefined) {
+      //we allow asking for new hashes when we are not syncing and we can download from specified peer ,we are
+      //top of the chain and not resolving branches currently
+      if (notDownloading() && topOfTheChain && maybePeer.isDefined && !resolvingBranches) {
```
ntallar: Do we maybe need to add this check also for handling the `NewBlock` messages? Maybe this check could then be added to the `notDownloading()` function?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -112,6 +127,56 @@ class RegularSync(
...
 
+  /**
+    * Handles NewHashesMessege, should only cover this message when we are top of the chain
+    */
+  def handleNewBlockHashesMessages: Receive = {
+    case MessageFromPeer(NewBlockHashes(hashes), peerId) =>
+      val maybePeer = peersToDownloadFrom.find(peer => peer._1.id == peerId)
+      //we allow asking for new hashes when we are not syncing and we can download from specified peer ,we are
+      //top of the chain and not resolving branches currently
+      if (notDownloading() && topOfTheChain && maybePeer.isDefined && !resolvingBranches) {
```
ntallar: Do we maybe need to add this check also for handling the NewBlock messages? Maybe this check could then be added to the notDownloading() function?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -61,28 +63,41 @@ class RegularSync(
...
   private def resumeRegularSync(): Unit = {
+    cancelScheduledResume()
+
+    // The case that waitingForActor is defined (we are waiting for some response),
+    // can happen when we are on top of the chain and currently handling newBlockHashes message
+
+    if (waitingForActor.isEmpty)
+      askForHeaders()
+    else
+      resumeRegularSyncTimeout = Some(scheduler.scheduleOnce(checkForNewBlockInterval, self, ResumeRegularSync))
```
ntallar: Nice catch :+1:

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
KonradStaniec: actually, I think we don't need it in terms of safety ( newBlock don't start any download or schedule resume), but it may be good idea to just move this condition to `notDownloading()`

PR #325, opened 2017-10-23
[EC-317] Pruning error when doing chain reorg

AlanVerbner:
# Description

Current pruning implementation wasn't designed to allow chain reorgs leading to MPT node removal that will be used after picking a diferent fork. 
A full description of the issue can be found here: https://github.com/input-output-hk/mantis/pull/313#discussion_r144534162

# Fix

It's still a naive implementation that:

1. Keeps track of MPT node references count in order to avoid deleting a node that's being used by another branch of the trie.
2. Instead of delaying unused mpt nodes removal we do it immediately when we found it has no more references
3. In order to be able to rollback (a deletion, a reference increase or decrease) we store MPT node snapshot before doing any changes alongside with the block number where that happened. If a rollback is needed, we just need to replace current DB version with snapshot one.

Note that:
1. We will be storing more info (as we are keeping snapshots of the nodes before doing changes)
2. It's still super simple implementation

# TODO
- [x] Check (and run) if there ets tests to cover rollbacks
- [x] Based on ^, do more proper testing on rollbacks
- [ ] Run a a full chain and track DB size


src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -36,99 +36,106 @@ class ReferenceCountNodeStorage(nodeStorage: NodeStorage, pruningOffset: BigInt,
...
+        }
+        // Save snapshots and latest snapshot index
+        (snapshotCountKey -> snapshotsCountToBytes(blockNumberSnapshotsCount + snapshotsToSave.size)) +: snapshotsToSave
+      }
       else Nil
-
-    nodeStorage.update(Nil, toUpsertUpdated ++ toMarkAsDeleted)
-
+    }
+    nodeStorage.update(toRemoveUpdated, toUpsertUpdated ++ snapshotToSave)
```
KonradStaniec: As we are deleting some nodes here, shouldn't we somehow decrement the reference count of nodes this deleted node was referencing?  ( suppose we have node A referencing node B,  shouldn't we decrement reference count of B after deleting A)

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -36,99 +36,106 @@ class ReferenceCountNodeStorage(nodeStorage: NodeStorage, pruningOffset: BigInt,
...
     this
```
KonradStaniec: generally, i find this function pretty long and having much logic, wdyt about splitting it into few smaller ones?

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -36,99 +36,106 @@ class ReferenceCountNodeStorage(nodeStorage: NodeStorage, pruningOffset: BigInt,
...
-        val pruneCanditatesUpdated =
-          // if references is 0, mark it as prune candidate for current block tag
-          if (storedNode.get.references == 0) toDeleteInBlock.copy(nodeKeys = nodeKey +: toDeleteInBlock.nodeKeys)
-          else toDeleteInBlock
-        (storedNodes + (nodeKey -> storedNode.get), pruneCanditatesUpdated)
+    val changes = toRemove.foldLeft(upsertChanges) { (storedNodes, nodeKey) =>
+      val maybeStoredNode: Option[(StoredNode, StoredNodeSnapshot)] = storedNodes.get(nodeKey) // get from current changes
+        .orElse(nodeStorage.get(nodeKey).map(storedNodeFromBytes).map(s => s -> StoredNodeSnapshot(nodeKey, Some(s)))) // or db
+
+      if (maybeStoredNode.isDefined) {
```
KonradStaniec: this could be expressed as fold like:
`maybeStoredNode.fold(storedNodes){ case (storedNode, snapshot) => storedNodes + (nodeKey -> (storedNode.decrementReferences(1), snapshot))}`

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -36,99 +36,106 @@ class ReferenceCountNodeStorage(nodeStorage: NodeStorage, pruningOffset: BigInt,
...
+      case None => ()
+    }
   }
 
-  /**
-    * Model to be used to store, by block number, which block keys are no longer needed (and can potentially be deleted)
-    */
-  case class PruneCandidates(nodeKeys: Seq[ByteString])
+  private def snapshotKeysUpTo(blockNumber: BigInt, snapshotCount: BigInt): Seq[ByteString] =
+    (BigInt(0) to snapshotCount).map(snapshotIndex => getSnapshotKey(blockNumber, snapshotIndex))
```
KonradStaniec: possible performance optimization: in `getSnapshotKey` we call `blockNumber.toByteArray` for every iteration, but it doesn't change at all here. ( i dont know how much perf overhead it adds, but we could `blockNumber.toByteArray` in this function and change `getSnapshotKey` signature to take block number as `Array[Byte]`)

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
AlanVerbner: That a good question. In this context, references means that someone has "inserted" the node, which means, it's being used somewhere else. 

Think of a tree where nodes are chained by hash, so we might have, for example, two leafs with same value and those will be stored just once. When removing one of those leafs because we deleted the whole branch, we cannot physically remove it as there is another leaf equal to that one on a non deleted branch.

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
AlanVerbner: I think it's a good idea, I was thinking to that too.

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
AlanVerbner: And, specifically talking about your question, the storage only handles data being stored, not the trie logic.

So, if you have dependent nodes in a tree, the one using this storage (`io.iohk.ethereum.mpt.MerklePatriciaTrie`) will make sure references are consistent. Without getting too deep into the MPT, if you remove `A` and not `B` it will happen that:
- `A` references is decremented by 1
- `B` references is decremented by 1 
- `B` is linked to another node, let's call it `C` (which is either new or an existing one). It's is basically incrementing `B` references by 1
- `C` references is incremented by 1

So, `B` references are not changing in your scenario

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
AlanVerbner: Good catch. I'm about to push a solution with currying. Let me know if you like it

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -326,6 +336,13 @@ class BlockchainImpl(
       stateRootHash,
       noEmptyAccounts = false
     )
+
+  def nodesKeyValueStorageFor(blockNumber: Option[BigInt]): NodesKeyValueStorage =
+    PruningMode.nodesKeyValueStorage(pruningMode, nodeStorage)(blockNumber)
+
+  def pruneState(blockNumber: BigInt): Unit = PruningMode.prune(pruningMode, blockNumber, nodeStorage)
```
KonradStaniec: Currently, we are using `pruneState` only during block save in regularSync, shouldn't we also prune our state trie during fast sync?

AlanVerbner: No, because while doing fast sync you are downloading the whole state trie for a single block, the target block. Therefore there is no need to prune anything.

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -280,6 +289,7 @@ class BlockchainImpl(
     receiptStorage.remove(blockHash)
     maybeTxList.foreach(removeTxsLocations)
     maybeBlockHeader.foreach{ h =>
+      rollbackState(h.number)
```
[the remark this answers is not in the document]
AlanVerbner: No, because it rollback state changes, ie, undo whatever has been done. Maybe `rollbackState` is not a good name and `rollbackStateChanges` is more appropriate, what do you think?

src/main/scala/io/iohk/ethereum/db/storage/pruning/package.scala
```diff
@@ -4,42 +4,63 @@ import io.iohk.ethereum.mpt.NodesKeyValueStorage
...
         case ArchivePruning => new ArchiveNodeStorage(nodeStorage)
-        case BasicPruning(history) => new ReferenceCountNodeStorage(nodeStorage, history, blockNumber)
+        case BasicPruning(history) => new ReferenceCountNodeStorage(nodeStorage, blockNumber)
       }
 
-  }
+    /**
+      * Prunes node storage for a given pruning mode and block number
+      * @param pruningMode Pruning mode tobe used
+      * @param blockNumber Block number to prune
```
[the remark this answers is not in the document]
AlanVerbner: well...it will use pruning only if it's supported (`io.iohk.ethereum.db.storage.pruning.BasicPruning`)

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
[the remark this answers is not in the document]
KonradStaniec: So if I understood correctly it rollbacks state changes made by block we are currently deleting, and not rollback state changes to some block? (if yes then it could be good to rename function to sth like 
`rollbackStateChangesMadeByBlock` or something similar)

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
KonradStaniec: As I understand currently `PruningMode.prune(pruningMode, blockNumber, nodeStorage)` only delete stale snapshots from db.

so If during fastsync we are saving nodes using https://github.com/input-output-hk/mantis/blob/b022f9657fa8b76f9352992de2a3935763b7685f/src/main/scala/io/iohk/ethereum/domain/Blockchain.scala#L269

and we would not use `prune` anywhere,  wouldn't we create and save unnecessary snapshots for whole chain? 

I just did a little test (10000 blocks) of fastsync on this branch and `phase/deadalus`, and size of `phase/deadalus` datadir is 53,7 MB, and with new pruning it is 61,5 MB, there is a little increase.

AlanVerbner: We will run `fastSyncTarget` pruning when doing regular sync on the block `fastSyncTarget`+ `history`. So, for example if blockchain head is block `4745467`: 
- we will pick `4744967` as `fastSyncTarget` (checkout `target-block-offset` config, default to 500)
- When executing regular sync (as it's after `fastSyncTarget`) on block `4746467` (`history` is 1000 by default) we will prune `4744967`

I think the size increase might be because we are storing the Snapshots (which contains the whole node) instead of the previous buggy version (where we were only storing  block numbers to delete when we detected zero references)

PR #326, opened 2017-10-26
New deleteAccout endpoint

KonradStaniec:


LukasGasior1:
Can we call it DeleteWallet (and other places respectively)? Even though it does not matter on Daedalus side on our backend Wallet != Account.

PR #327, opened 2017-10-26
[EC-327] Fix minor linter issues

mirkoAlic:
Minor fixes:

- Unused method parameter: 'peerId'
- Unused imports


LukasGasior1:
I have mixed feelings about adding 'final' to our case classes, do we really need it? (hey, we all know we shouldn't extend them)

mirkoAlic:
I agree with you but as it's an open source project, is better follow the standard style checks, also the linter's report gave a better output (less verbose) and made the analysis easier. We can either do this now, leave it for later or even disable the rule, what do you think?

rtkaczyk:
I'm all for disabling that rule!

rtkaczyk:
> Implicit definitions must have an explicit type ascription

Why?

mirkoAlic:
You are right is not necessary, I rechecked the rule, and I noticed that for these situations it can't be possible that the compiler has troubles. I'll revert that commit too.

mirkoAlic:
Changes are ready!

PR #328, opened 2017-10-27
[EC-169] Changed  Validator functions return type

KonradStaniec:


AlanVerbner:
I think we also need to address this file, right? https://github.com/input-output-hk/mantis/blob/phase/daedalus/src/main/scala/io/iohk/ethereum/validators/BlockValidator.scala#L14

ntallar:
Maybe we should also address this file as well?

- https://github.com/input-output-hk/mantis/blob/refactor/validator_return_type/src/main/scala/io/iohk/ethereum/validators/BlockHeaderValidator.scala#L9

src/test/scala/io/iohk/ethereum/validators/BlockValidatorSpec.scala
```diff
@@ -46,12 +46,6 @@ class BlockValidatorSpec extends FlatSpec with Matchers {
     }
   }
 
-  "Block" should "return a block when verifing if a block body corresponds to a block header" in {
```
ntallar: Minor detail: Why was this test removed?

KonradStaniec: We are no longer returning `block` from `validateHeaderAndBody` and test was explicitly testing this. A valid case of a validator is covered by other tests.

PR #330, opened 2017-10-31
[EC-328] Add recent transactions api

LukasGasior1:


src/main/scala/io/iohk/ethereum/jsonrpc/TransactionResponse.scala
```diff
@@ -14,13 +14,15 @@ case class TransactionResponse(
     value: BigInt,
     gasPrice: BigInt,
     gas: BigInt,
-    input: ByteString)
+    input: ByteString,
+    pending: Option[Boolean])
```
ntallar: What's the difference between what `None` and `Some(false)` represent? Maybe it is enough to have a boolean representing whether the tx is pending or not?

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerSpec.scala
```diff
@@ -1323,6 +1323,39 @@ class JsonRpcControllerSpec extends FlatSpec with Matchers with PropertyChecks w
     ))
   }
 
+  it should "eth_getAccountRecentTransactions" in new TestSetup {
```
ntallar: Minor detail: Maybe we could also add a test `EthServiceSpec` testing how `eth_getAccountRecentTransactions` works without mocking it?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -723,4 +726,35 @@ class EthService(
...
+
+      val pendingReceived = pendingTransactions
+        .map(_.stx)
+        .filter(_.tx.receivingAddress.contains(request.address))
+        .map(TransactionResponse(_, pending = Some(true)))
+
+      val numRecentBlocks = Config.Network.Rpc.txNumRecentBlocks
+
+      val bestBlockNum = appStateStorage.getBestBlockNumber()
+      val recentTxs = (0 until numRecentBlocks)
```
AlanVerbner: Minor:

What about something like?

```scala
val start = (bestBlockNum - numRecentBlocks).max(0)
val recentTxs = (start until bestBlockNum)
```

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -723,4 +726,35 @@ class EthService(
     }
 
   }
+
+  def getAccountRecentTransactions(request: GetAccountRecentTransactionsRequest): ServiceResponse[GetAccountRecentTransactionsResponse] = {
```
AlanVerbner: I'm wondering, shouldn't we put this endpoint on a custom _"namespace"_ like `mantis_`? I'

src/main/scala/io/iohk/ethereum/jsonrpc/TransactionResponse.scala
LukasGasior1: I made it an Option, so that we don't include this flag in other endpoints (and so that they match spec)

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
LukasGasior1: good question, I don't know. @rtkaczyk I think you mentioned Parity has some custom endpoints, are they in separate namespace?

src/main/resources/application.conf
```diff
@@ -175,6 +175,9 @@ mantis {
       net {
         peer-manager-timeout = 5.seconds
       }
+
+      # How many blocks back do we consider when searching for account recent transactions (eth_getAccountRecentTransactions)
+      tx-num-recent-blocks = 1000
```
AlanVerbner: Maybe it's a good idea to allow users to send this value within the request instead. Maybe this config can be used as a default / max.

LukasGasior1: makes sense

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
rtkaczyk: Yes, Parity has `parity_` 😄

src/main/scala/io/iohk/ethereum/jsonrpc/TransactionResponse.scala
ntallar: Maybe it would be better then to return a `Some(false)` when the tx was obtained from a block in [`eth_getAccountRecentTransactions`](https://github.com/input-output-hk/mantis/pull/330/files#diff-4aa60425d2cdec6299a3815b98a8e30dR750) instead?

LukasGasior1: Yes, I missed that, good catch

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
LukasGasior1: looks good, but with `to` instead of `until` :)

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
LukasGasior1: I'll put it in another namespace, but still leave it in EthService class (as it seems tightly coupled, it uses the same dto classes etc) wdyt?

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerSpec.scala
LukasGasior1: done

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
AlanVerbner: I think its ok. (When I need to look for an implementation I start from `JsonRpcContoller` so it's easy to figure out that will be on `EthService`)

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -572,13 +568,20 @@ object EthJsonMethodsImplicits extends JsonMethodsImplicits {
...
+    input match {
+      case JNothing => Right(None)
+      case o => extractQuantity(o).map(Some(_))
+    }
+
   implicit val eth_getAccountRecentTransactions = new JsonDecoder[GetAccountRecentTransactionsRequest] with JsonEncoder[GetAccountRecentTransactionsResponse] {
     def decodeJson(params: Option[JArray]): Either[JsonRpcError, GetAccountRecentTransactionsRequest] =
       params match {
-        case Some(JArray((address: JString) :: Nil)) =>
+        case Some(jarray: JArray) =>
```
ntallar: Very minor detail: Maybe this could be changed to what's done in [`eth_sign`](https://github.com/input-output-hk/mantis/blob/9b55ae760559f60dc515a235af4b2b1a376ce72c/src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala#L541) for extracting the parameters?

LukasGasior1: maybe I'm missing something, but is this possible for variable-sized Jarray (either 1 or 2 elems)?

ntallar: You are right, it doesn't seem to allow doing so.
The only option seems to be something like https://github.com/input-output-hk/mantis/blob/phase/daedalus/src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala#L255, but I don't have any strong feelings regarding if doing so is necessary or not, we could leave it as is.

PR #331, opened 2017-10-31
[EC-290] Enable creation of package with embedded JVM

ntallar:
## Description

Enables the creation of a package including the Mantis codebase and an embedded JVM, which is obtained by running the command: `sbt jdkPackager:packageBin`. This will be obtained for running the Mantis client with the Daedalus wallet (in this [open PR](https://github.com/input-output-hk/mantis-daedalus/pull/1))

The package created was tested in Mac 10.13, Ubuntu 16.04 and Windows 10.

*Note*: Requires the 'ant-javafx.jar' that comes with Oracle JDK, which is detected automatically if Oracle JDK is installed, or can be configured by following [this](http://www.scala-sbt.org/sbt-native-packager/formats/jdkpackager.html#requirements).
*Note*: For creating the package for a desired OS, it has to be run on that OS.


AlanVerbner:
On windows, after starting it (even from the cmd) it runs it in background, is that the desired behavior? Also, I think if you run it twice, it returns a non-specific error.

AlanVerbner:
On linux seems to be running just fine (I had to do the `ant-javafx.jar` thing tho)

ntallar:
>On windows, after starting it (even from the cmd) it runs it in background, is that the desired behavior? 

Although I agree this isn't the preferable behavior for running the mantis client on it's own, it is the preferable one for running both the Daedalus wallet and the Mantis client (as is done with the Cardano client). As far as I was able to investigate, this can't be changed with the currently used plugin, so we should change it if it's needed. @mcsherrylabs what do you think about this?

LukasGasior1:
On my Linux machine it seems to work just fine :+1:

PR #333, opened 2017-11-01
[EC-335] CORS configuration

AlanVerbner:
## Description

In order to incrase HTTP Json RPC endpoint security, this PR allows to:
- disable CORS
- enable cors for a specific domains

### DIFF
--- a/src/main/scala/io/iohk/ethereum/utils/Config.scala
+++ b/src/main/scala/io/iohk/ethereum/utils/Config.scala
@@ -2,6 +2,7 @@ package io.iohk.ethereum.utils
 
 import java.net.InetSocketAddress
 
+import akka.http.scaladsl.model.headers.{HttpOrigin, HttpOriginRange}
 import akka.util.ByteString
 import com.typesafe.config.{ConfigFactory, Config => TypesafeConfig}
 import io.iohk.ethereum.db.dataSource.LevelDbConfig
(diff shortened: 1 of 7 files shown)


src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -98,6 +99,17 @@ object Config {
...
+      def parseMultipleOrigins(origins: Seq[String]): HttpOriginRange = HttpOriginRange(origins.map(HttpOrigin(_)):_*)
+      def parseSingleOrigin(origin: String): HttpOriginRange = origin match {
+          case "*" => HttpOriginRange.*
+          case s => HttpOriginRange.Default(HttpOrigin(s) :: Nil)
+        }
+
+      override val corsAllowedOrigins: Option[HttpOriginRange] =
+        (Try(parseMultipleOrigins(rpcConfig.getStringList("cors-allowed-origins").asScala)) recoverWith {
+          case _ => Try(parseSingleOrigin(rpcConfig.getString("cors-allowed-origins")))
+        }).toOption
```
rtkaczyk: It's not immediately clear whether not specifying this setting is equivalent to `cors-allowed-origins = []` or `cors-allowed-origins = *`. BTW, TypesafeConfig authors advise against implementing optional (default) settings.

Also, is `cors-allowed-origins = ""` a valid setting?

AlanVerbner: Well ... `*` is standard for _all_ (`Access-Control-Allow-Origin: *`) and  `[]` refers to no domains (as it's an empty seq). `""` Is not a valid domain so it will fail.

Can you think of a better alternative? I'm open to suggestions

rtkaczyk: Yes, but I'm saying that this setting does not need to be provided at all. And it's not clear what that means. IMO, best approach is avoiding defaults, so instead of `.toOption` just let it fail.

AlanVerbner: I've just removed the optional param. Btw, while taking a look at this I've found that if there is an error while loading RPC Server, the [app will start besides that](https://github.com/input-output-hk/mantis/blob/phase/daedalus/src/main/scala/io/iohk/ethereum/App.scala#L41). Maybe its a good idea to change that too, wdyt?

rtkaczyk: I agree, the client should not start. If you feel it's OK to change it on this PR, then you have my blessing 😉

PR #334, opened 2017-11-02
[EC-332][EC-329] changePassphrase + sorting accounts by date

rtkaczyk:


src/main/scala/io/iohk/ethereum/keystore/KeyStore.scala
```diff
@@ -99,6 +113,15 @@ class KeyStoreImpl(keyStoreDir: String, secureRandom: SecureRandom) extends KeyS
     }
   }
 
+  private def overwrite(name: String, encKey: EncryptedKey): Either[KeyStoreError, Unit] = {
+    val json = EncryptedKeyJsonCodec.toJson(encKey)
+    val path = Paths.get(keyStoreDir, name)
+    Try {
+      Files.write(path, json.getBytes(StandardCharsets.UTF_8))
+      ()
+    }.toEither.left.map(ioError)
```
AlanVerbner: if an error happens, can the previous key get lost?

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala
```diff
@@ -194,6 +194,9 @@ class JsonRpcController(
 
     case req @ JsonRpcRequest(_, "personal_deleteWallet", _, _) =>
       handle[DeleteWalletRequest, DeleteWalletResponse](personalService.deleteWallet, req)
+
+    case req @ JsonRpcRequest(_, "personal_changePassphrase", _, _) =>
```
AlanVerbner: I think we need to discuss if new endpoints will be under existing _namespaces_ or not. This is the second case (see https://github.com/input-output-hk/mantis/pull/330)

I don't have a strong opinion tho.

src/main/scala/io/iohk/ethereum/keystore/KeyStore.scala
rtkaczyk: If the error somehow occurs during write (some bytes written, but not all - probably also depends on FS) then I guess it's possible.

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala
rtkaczyk: > I don't have a strong opinion tho.

Me neither. I was hoping the review would clarify that 😄

src/main/scala/io/iohk/ethereum/keystore/KeyStore.scala
rtkaczyk: Of course trying to revert in case of error may also result in an error. The baseline is that storing these accounts does not relieve the user of responsibility to know his private key (or the passphrase to generate it). So I don't consider the unlikely event of losing the key this way to be a terrible thing.

Unless you see an easy solution for this, we can either schedule a follow-up task or simply ignore this issue (and I'm more inclined towards the latter).

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala
rtkaczyk: Changed to `daedalus`

src/main/scala/io/iohk/ethereum/keystore/KeyStore.scala
AlanVerbner: I don't see an easy solution but I would be really sad if a client loses my wallet because I changed the passphrase. What we can do is to let daedalus guys know so they can put a big warning sign.

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala
```diff
@@ -192,8 +192,11 @@ class JsonRpcController(
     case req @ JsonRpcRequest(_, "personal_ecRecover", _, _) =>
       handle[EcRecoverRequest, EcRecoverResponse](personalService.ecRecover, req)
 
-    case req @ JsonRpcRequest(_, "personal_deleteWallet", _, _) =>
+    case req @ JsonRpcRequest(_, "daedalus_deleteWallet", _, _) =>
```
ntallar: Minor detail: Should we maybe move this and the handling of `daedalus_changePassphrase` to a function like `handleDaedalusRequest`?

rtkaczyk: Good point, and also I think I should add it [here](https://github.com/input-output-hk/mantis/blob/68a5e642bf3056bbd179c2cd7fc03d7fc5b92164/src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala#L30) and make it configurable

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala
```diff
@@ -33,6 +33,7 @@ object JsonRpcController {
     val Net = "net"
     val Db = "db"
     val Personal = "personal"
+    val Daedalus = "daedalus"
```
AlanVerbner: :+1: This has already been done by Lukasz so after updating up to latest `phase/daedalus` it shouldn't be marked as a change

src/main/resources/application.conf
```diff
@@ -170,7 +170,7 @@ mantis {
 
       # Enabled JSON-RPC APIs over the JSON-RPC endpoint
       # Available choices are: eth, web3, net, personal, daedalus
-      apis = "eth,web3,net,daedalus"
+      apis = "eth,web3,net,personal,daedalus"
```
AlanVerbner: so, personal and daedalus will be enabled by default?

What about handling this in https://github.com/input-output-hk/mantis/pull/331 ?

rtkaczyk: This was discussed with @mcsherrylabs

PR #335, opened 2017-11-02
[EC-258] Implement mining

LukasGasior1:


src/main/scala/io/iohk/ethereum/mining/Miner.scala
```diff
@@ -0,0 +1,194 @@
...
+  private def dagFile(seed: ByteString): File = {
+    new File(s"${miningConfig.ethashDir}/full-R${Ethash.Revision}-${Hex.toHexString(seed.take(8).toArray[Byte])}")
+  }
+
+  private def generateDagAndSaveToFile(epoch: Long, dagSize: Long, seed: ByteString): Array[Array[Int]] = {
+    // scalastyle:off magic.number
+
+    val file = dagFile(seed)
+    if (file.exists()) file.delete()
+    file.createNewFile()
```
ntallar: Should we also create the parent directory if needed with `file.getParentFile.mkdirs()`? I'm getting an exception `No such file or directory` which I think is caused by that.

src/main/scala/io/iohk/ethereum/mining/Miner.scala
```diff
@@ -0,0 +1,194 @@
...
+
+    while (inputStream.read(buffer) > 0) {
+      if (index % 100000 == 0) log.info(s"Loading DAG from file ${((index / res.length.toDouble) * 100).toInt}%")
+      res(index) = Ethash.bytesToInts(buffer)
+      index += 1
+    }
+
+    Try(inputStream.close())
+
+    res
```
ntallar: Maybe we could check whether the loaded file has the expected size? Maybe I'm missing something but it seems that if we stop the client while generating the DAG file it will load an only partially generated DAG.

src/main/scala/io/iohk/ethereum/mining/Miner.scala
```diff
@@ -0,0 +1,194 @@
...
+
+  private def generateDagAndSaveToFile(epoch: Long, dagSize: Long, seed: ByteString): Array[Array[Int]] = {
+    // scalastyle:off magic.number
+
+    val file = dagFile(seed)
+    if (file.exists()) file.delete()
+    file.createNewFile()
+
+    val outputStream = new FileOutputStream(dagFile(seed).getAbsolutePath)
+    outputStream.write(Array(0xfe, 0xca, 0xdd, 0xba, 0xad, 0xde, 0xe1, 0xfe).map(_.toByte))
```
ntallar: This bytes seem to be skipped while loading the DAG file, why are they needed?

src/main/resources/application.conf
```diff
@@ -204,6 +204,13 @@ mantis {
 
     # Extra data to add to mined blocks
     header-extra-data = "mantis"
+
+    # If true this node will mine new blocks
+    mining-enabled = false
+
+    ethash-dir = ${user.home}"/.ethash"
```
KonradStaniec: I don't know how other clients handle it but maybe it should be a subdirectory of our `datadir` ? As a user I would probably be surprised seeing this unknown new folder in my home dir.

src/main/scala/io/iohk/ethereum/mining/Miner.scala
```diff
@@ -0,0 +1,194 @@
...
+
+    val cache = Ethash.makeCache(epoch)
+    val res = new Array[Array[Int]]((dagSize / Ethash.HASH_BYTES).toInt)
```
KonradStaniec: This expr `dagSize / Ethash.HASH_BYTES).toInt` seems to be used in 3 places, maybe we can precompute it earlier and pass it to function instead of `dagSize` ?

src/main/scala/io/iohk/ethereum/mining/Miner.scala
LukasGasior1: It's a magic number defined in: https://github.com/ethereum/wiki/wiki/Ethash-DAG

src/main/scala/io/iohk/ethereum/mining/Miner.scala
```diff
@@ -0,0 +1,194 @@
...
+    // scalastyle:off magic.number
+
+    val file = dagFile(seed)
+    if (file.exists()) file.delete()
+    file.createNewFile()
+
+    val outputStream = new FileOutputStream(dagFile(seed).getAbsolutePath)
+    outputStream.write(Array(0xfe, 0xca, 0xdd, 0xba, 0xad, 0xde, 0xe1, 0xfe).map(_.toByte))
+
+    val cache = Ethash.makeCache(epoch)
```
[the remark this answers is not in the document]
LukasGasior1: It could be done, but I'm not sure if it's worth (to add additional coupling). It doesn't take that long and it's only executed once per dag generation.

src/main/resources/application.conf
LukasGasior1: It's a standard dir for DAG, these files can be shared across clients, see: https://github.com/ethereum/wiki/wiki/Ethash-DAG

src/main/scala/io/iohk/ethereum/mining/Miner.scala
LukasGasior1: I'll add it as a constant and also check it when reading from file

PR #336, opened 2017-11-03
[EC-199]  estimate gas improvement

KonradStaniec:
Little clarification: 
from docs (https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_estimategas) It is not clear what should be the semantics of `estimateGas`, should we return gas used by transaction, or minimal gas limit required to send transaction. (those could be different) 

This pr changes the semantics of our implementation of `estimateGas` from returning gas used by the transaction, to returning minimal gas limit required to send transaction. Main motivation factor is to match up to other ethereum clients (`partity` and `geth` implement `estimateGas` that way)


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -469,6 +471,14 @@ class LedgerImpl(
     TxResult(result.world, gasLimit - totalGasToRefund, result.logs, result.returnData, result.error)
   }
 
+  def binarySearchGasEstimation(stx: SignedTransaction, blockHeader: BlockHeader): BigInt = {
+    val lowLimit = EvmConfig.forBlock(blockHeader.number, blockchainConfig).feeSchedule.G_transaction
+    val highLimit = stx.tx.gasLimit
+
+    LedgerUtils.binaryChop(lowLimit, highLimit)(gasLimit =>
```
ntallar: If I understood correctly, if the `tx.gasLimit` is less than `G_transaction` this would result in an exception, maybe we should just return the `highLimit` as is the case when `tx.gasLimit` isn't enough?

src/main/scala/io/iohk/ethereum/ledger/LedgerUtils.scala
```diff
@@ -0,0 +1,30 @@
...
+    * Function finds minimal value in some interval for which provided function do not return error
+    * If searched value is not in provided interval, function returns maximum value of searched interval
+    * @param min minimum of searched interval
+    * @param max maximum of searched interval
+    * @param f function which return error in case to little value provided
+    * @return minimal value for which provided function do not return error
+    */
+  @tailrec
+  private[ledger] def binaryChop[Error](min: BigInt, max: BigInt)(f: BigInt => Option[Error]): BigInt = {
+    assert(min <= max)
```
AlanVerbner: :+1:

src/main/scala/io/iohk/ethereum/ledger/LedgerUtils.scala
```diff
@@ -0,0 +1,30 @@
...
+
+    if (min == max)
+      max
+    else {
+      val mid = min + (max - min) / 2
+      val possibleError = f(mid)
+      if (possibleError.isEmpty)
```
AlanVerbner: I think the only error we should use to continue searching is `OutOfGas` as the other ones will return the same error even if narrowing the value. I think it will converge to `max` .. wdyt?

src/it/scala/io/iohk/ethereum/SimulateTransactionTest/SimulateTransactionTest.scala
```diff
@@ -0,0 +1,149 @@
...
+  val ledger = new LedgerImpl(VM, blockchain, blockchainConfig, syncConfig, validators)
+
+  val existingAddress = Address(10)
+  val existingAccount = Account(nonce = UInt256.Zero, balance = UInt256(10))
+
+  val existingEmptyAccountAddres = Address(20)
+  val existingEmptyAccount = Account.empty()
+
+  /*
+   * Failing code which mess up with gas estimation
```
AlanVerbner: :+1:

src/main/scala/io/iohk/ethereum/ledger/LedgerUtils.scala
KonradStaniec: Actually, there are cases with other kinds of errors that will eventually converge to some minimum, good example is this one in https://github.com/input-output-hk/mantis/pull/336/files/8345e90cb532910c9ab74abd446f4fa21fa5b9f1#diff-16527700d5f0bb6dd4d24c426f5c49a3R104. If we will give it to little gas it fails with `InvalidJump` and not `OOG`
Also, other clients do not look at a type of error so I assume this is trade-off which is accepted.  (that there will be some errors that will always end with maximum gas)

PR #337, opened 2017-11-08
Update transactions api parameters

LukasGasior1:
Added: fromBlock and toBlock params instead of max-num limit


src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -739,11 +737,7 @@ class EthService(
         .filter(_.tx.receivingAddress.contains(request.address))
         .map(TransactionResponse(_, pending = Some(true)))
 
-      val numRecentBlocks = request.numRecentBlocks.getOrElse(txMaxNumRecentBlocks) min txMaxNumRecentBlocks
-
-      val bestBlockNum = appStateStorage.getBestBlockNumber()
-      val start = (bestBlockNum - numRecentBlocks).max(0)
-      val recentTxs = (start to bestBlockNum)
+      val recentTxs = (request.fromBlock to request.toBlock)
         .flatMap { n => blockchain.getBlockByNumber(n) }
```
KonradStaniec: Wouldn't this step load all fetched blocks into memory?  
I mean when we had an explicit limit of 1000 blocks there were no risks with it, but currently, if user would like blocks for example from 1 to million, would we risk out of memory error?

rtkaczyk: Maybe leave `txMaxNumRecentBlocks ` and return an error when `toBlock - fromBlock > txMaxNumRecentBlocks`?

LukasGasior1: @KonradStaniec good point, how about making it a stream?
@rtkaczyk if the num of blocks is too high it will already return an error anyway

KonradStaniec: @LukasGasior1  I think `Stream` would be good, but maybe `Iterable` would be better as we probably do not need memoization here.

Also we can add a stage to to filter out blocks which can return empty tuples ( as i understand here https://github.com/input-output-hk/mantis/blob/d827964ffdbf6385df20a2a3cfeab1f6ac72f671/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala#L745, even if both collectios are empty we would return `(Seq.empty, Seq.empy)` for every block)

rtkaczyk: > if the num of blocks is too high it will already return an error anyway

Sorry, I may be missing something obvious, where does it happen?

LukasGasior1: I meant it will throw an error when it runs out of mem ;)

LukasGasior1: my point is, I'm not sure if this limit is needed. If user wants to wait and get all his txs, then why not?

rtkaczyk: I guess it's not a big problem when a user is running both Daedalus and Mantis locally. Then we can say: "you messed up, you nasty user". But what if it's a public node and any user can crash it so easily?

Besides adding a limit does not prevent a user from getting all his TXs. It's just that Daedalus would have to query for them using multiple requests, in a pagination-like fashion.

LukasGasior1: well web3 is not really suitable to be exposed publicly anyway, isn't it? There are way too many ways to crash it without additional protection.
But ok, I can add that limit, just to be a bit safer

rtkaczyk: > well web3 is not really suitable to be exposed publicly anyway, isn't it?

I don't see Ethereum (and the like) ever getting wider popularity if every user is expected to run a full node.

LukasGasior1: web3 is not the only way to share access to blockchain from a full node. There are some services that offer a limited-subset of web3-like endpoints. Or you can have a non-full node, and only get the data you need from other nodes on-demand (using regular peer protocol)

LukasGasior1: * web3 is really just a lib, I of course mean RPC api

rtkaczyk: But given how Daedalus integration works a public Mantis node with the necessary endpoints exposed is currently the only way a user can avoid running a full node.

> only get the data you need from other nodes on-demand (using regular peer protocol)

I am not familiar with that, can you shed more light? Surely Mantis doesn't support this mode, does it?

LukasGasior1: It's not a standard "mode", but it can be done. I've been playing with MPT implementation that, instead of querying local db, would use `GetNodeData` to get the nodes from other peers when needed - that way you can access state without a need to be fully synced (or syncing at all).

> But given how Daedalus integration works a public Mantis node with the necessary endpoints exposed is currently the only way a user can avoid running a full node.

I'd rather think that synced local node is necessary for Daedalus (especially that it uses personal_ endpoint to manage wallets).

rtkaczyk: Sounds interesting and promising. Are there any existing implementations that utilise such or similar method?

> I'd rather think that synced local node is necessary for Daedalus (especially that it uses personal_ endpoint to manage wallets).

I don't think it has to be considered unsafe, if done right ;)

LukasGasior1: > Sounds interesting and promising. Are there any existing implementations that utilise such or similar method?
I don't know if there are any, but for sure there are some customized nodes implementations

> I don't think it has to be considered unsafe, if done right ;)
You mean using personal_ on external/public RPC that is mean to be used by multiple users?

rtkaczyk: Well yeah, I guess the only we're missing would be some prevention against brute-forcing a password.
Anyway, the best solution would be not to use `personal_` API at all. A wallet could handle all of that internally.

LukasGasior1: @rtkaczyk limit added
@KonradStaniec I made it a Stream and added that filter

LukasGasior1: > Anyway, the best solution would be not to use personal_ API at all. A wallet could handle all of that internally.

:+1: :+1:

PR #339, opened 2017-11-15
Enable configuring certificate keystore type

ntallar:
## Description

Enables configuring the type of certificate keystore being used.

This allows recreating a certificate keystore from a certificate and key files (using `openssl pkcs12 -export ...` command), which is needed for obtaining a keystore for the Mantis client based on the `server.crt` and `server.key` from [Daedalus](https://github.com/input-output-hk/daedalus/tree/master/tls).

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -155,17 +155,21 @@ mantis {
       enabled = true
 
       # Listening address of JSON-RPC HTTP/HTTPS endpoint
-      interface = "127.0.0.1"
+      interface = "localhost"
 
       # Listening port of JSON-RPC HTTP/HTTPS endpoint
       port = 8546
 
-      # Path to the JKS keystore storing the certificate (used only for https)
-      # null value indicates no certificate is being used
+      # Path to the keystore storing the certificates (used only for https)
+      # null value indicates HTTPS is not being used
       certificate-keystore-path = null
 
-      # File with the password used for accessing the certificate (used only for https)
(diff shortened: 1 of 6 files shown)


src/main/resources/application.conf
```diff
@@ -155,17 +155,21 @@ mantis {
...
-      # null value indicates no certificate is being used
+      # Path to the keystore storing the certificates (used only for https)
+      # null value indicates HTTPS is not being used
       certificate-keystore-path = null
 
-      # File with the password used for accessing the certificate (used only for https)
-      # null value indicates no certificate is being used
+      # Type of certificate keystore being used
+      # null value indicates HTTPS is not being used
+      certificate-keystore-type = null
```
rtkaczyk: What are valid keystore types? Maybe that info could be included in the comment here?

ntallar:
> Any setup guidance anywhere?

The steps required for using this for setting this up for using Daedalus were added to the related [Youtrack ticket](https://iohk.myjetbrains.com/youtrack/issue/EC-339), I just updated (and moved some of the setup from the ticket) to the [configuring HTTPS wiki](https://github.com/input-output-hk/mantis/wiki/Configuring-HTTPS-for-JSON-RPC).

src/main/resources/application.conf
rtkaczyk: Detail: a clarification was added to `network.conf`. Could we also add it to this file `application.conf`?

PR #341, opened 2017-11-23
[EC-344] Pruning should keep some old MPT Nodes

AlanVerbner:
## Description

After latest pruning algorithm change in order to handle chain reorgs made in https://github.com/input-output-hk/mantis/pull/325, two issues were introduced:
- Mantis Fast Sync Server broken after new pruning implementation
- Miner reports and error if it requests work while applying a block

## Reason
Latest implementation only keeps latest version of MPT nodes in db. Deleted ones are moved to a stash area where they can be taken in case of a rollback but cannot be accessed in order to serve it to other services (miner, fast sync host actor).

## Fix
The following change delays MPT node removal from DB until it's pruned. In order to do so, it checks if it's no longer needed looking for:
- Zero references in the latest version
- It wasn't referenced by any other change in later blocks

### DIFF
--- a/src/main/scala/io/iohk/ethereum/db/storage/encoding/package.scala
+++ b/src/main/scala/io/iohk/ethereum/db/storage/encoding/package.scala
@@ -21,11 +21,11 @@ package object encoding {
 
   private val storedNodeEncDec = new RLPDecoder[StoredNode] with RLPEncoder[StoredNode] {
     override def decode(rlp: RLPEncodeable): StoredNode = rlp match {
-      case RLPList(nodeEncoded, references) => StoredNode(nodeEncoded, references)
+      case RLPList(nodeEncoded, references, lastUsedByBlock) => StoredNode(nodeEncoded, references, lastUsedByBlock)
       case _ => throw new RuntimeException("Error when decoding stored node")
     }
 
-    override def encode(obj: StoredNode): RLPEncodeable = RLPList(obj.nodeEncoded, obj.references)
+    override def encode(obj: StoredNode): RLPEncodeable = RLPList(obj.nodeEncoded, obj.references, obj.lastUsedByBlock)
   }
 
   private val snapshotEncDec = new RLPDecoder[StoredNodeSnapshot] with RLPEncoder[StoredNodeSnapshot] {
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -417,6 +417,8 @@ mantis {
     mode = "basic"
 
     # The amount of block history kept before pruning
+    # Note: if fast-sync clients choose target block offset greater than this value, mantis may not be able to
+    # correctly act as a fast-sync server
     history = 1000
   }
 
--- a/src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
+++ b/src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
@@ -10,18 +10,23 @@ import io.iohk.ethereum.utils.Logger
 /**
   * This class helps to deal with two problems regarding MptNodes storage:
   * 1) Define a way to delete ones that are no longer needed but allow rollbacks to be performed
-  * 2) Avoids removal of nodes that can be used in different trie branches because the hash it's the same
+  * 2) Avoids removal of nodes that can be used in different trie branches because the hash is the same
   *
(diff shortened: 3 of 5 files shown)


LukasGasior1:
it's LGTM but just one question, you said we were moving removed nodes to stashed area, couldn't we just do something like: `getNode = getFromDb orElse getFromStashArea` and keep the old implementation?

AlanVerbner:
@lgasior I used the term "stash area" in order to simplify my explanation. What it really does is to take a [snapshot of that node](https://github.com/input-output-hk/mantis/pull/341/files#diff-1e345fe7a1acef2b55aced59e24d122dR208) and store it based on block number. The reason behind this is that:

- Doing a block rolback it's just applying the changes in reverse order for that block number
- Pruning consists in dropping Snapshots for the prune target (relase space as we won't allow rollbacks)

The problem that we are using a key-value db. We need to index MPT nodes by hash to query them but in order to be able to delete them we need to access them by block number (as pruning is made by that field). This is the easiest way I found (and by that I mean, with less impact on the current code) to be able to keep the ease of use of the undo log but to be able to provide those nodes by hash while doing fast sync or mining.

Does it make sense?

LukasGasior1:
I think it does now, thanks! I was thinking about something simpler but didn't consider chain reorgs - this indeed makes the implementation tricky.

AlanVerbner:
Yes, chain reorgs are quite tricky and made the simpler solution a little bit more complex

src/main/scala/io/iohk/ethereum/db/storage/encoding/package.scala
```diff
@@ -21,11 +21,11 @@ package object encoding {
 
   private val storedNodeEncDec = new RLPDecoder[StoredNode] with RLPEncoder[StoredNode] {
     override def decode(rlp: RLPEncodeable): StoredNode = rlp match {
-      case RLPList(nodeEncoded, references) => StoredNode(nodeEncoded, references)
+      case RLPList(nodeEncoded, references, lastUsedByBlock) => StoredNode(nodeEncoded, references, lastUsedByBlock)
```
rtkaczyk: So DB-wise it's a backward incompatible change?

AlanVerbner: Yes :sad:

mcsherrylabs: :/

AlanVerbner: Let me add more info about this. In order to delay MPT node deletion we need to find a way to determine if the node can be deleted at certain point (ie, it wasn't used after that). In order to do so, the easiest way I found is to put that stamp based on block number.

rtkaczyk: `StoredNode#lastUsedBlock`? But what's the relation of this field with reference count? How can a node be used later if its ref count drops to zero?

AlanVerbner: Reference count has the current references the node has and last used block is the last time the block was used.

Think about the following scenario.

- `Block#0` -> `(Insert, Key1 -> V1)`
- `Block#1` -> `(Delete, Key1)`
- `Block#999` -> `(Insert, Key1 -> V1)`

Block deletion is delayed up to pruning process, so:
- Pruning `Block#0` -> Nothing done as no delete
- Pruning `Block#1` -> We have a delete but current references count is 1 as it was added at `Block#999` so block is not deleted

Now, think about this case:
- `Block#0` -> `(Insert, Key1 -> V1)`
- `Block#1` -> `(Delete, Key1)`
- `Block#998` -> `(Insert, Key1 -> V1)`
- `Block#999` -> `(Delete, Key1)`

Pruning will be:
- Pruning `Block#0` -> Nothing done as no delete
- Pruning `Block#1` -> We have a delete and current reference count is 0 because it was deleted at `Block#999` *but we cannot delete it*. Why? Because if we do so, we won't be able to provide that node to someone doing fast sync starting at `Block#998` or mining from the same one
- Pruning `Block#998` -> Nothing done as no delete
- Pruning `Block#999` -> Node has no more references and it wasn't used (999-1999), delete it from db

rtkaczyk: Great expl, thanks

PR #344, opened 2017-11-28
[EC-346] Add miner status reporting

LukasGasior1:


src/main/scala/io/iohk/ethereum/mining/Miner.scala
```diff
@@ -203,4 +216,11 @@ object Miner {
   val MaxNonce: BigInt = BigInt(2).pow(64)
 
   val DagFilePrefix: ByteString = ByteString(Array(0xfe, 0xca, 0xdd, 0xba, 0xad, 0xde, 0xe1, 0xfe).map(_.toByte))
+
+  sealed trait MineResult {
+    def triedHashes: Int
+  }
+  case class MineSuccessful(override val triedHashes: Int, pow: ProofOfWork, nonce: ByteString) extends MineResult
```
loverdos: We can omit `override val`.

src/main/scala/io/iohk/ethereum/mining/Miner.scala
```diff
@@ -203,4 +216,11 @@ object Miner {
...
+  case class MineUnsuccessful(override val triedHashes: Int) extends MineResult
```
loverdos: Same as previously, we can omit `override val`.

src/main/scala/io/iohk/ethereum/mining/Miner.scala
LukasGasior1: I personally prefer to always add these, but I can remove it if you insist

rtkaczyk: I'm with Christos on this, because it's a case class

src/main/scala/io/iohk/ethereum/mining/Miner.scala
```diff
@@ -203,4 +216,11 @@ object Miner {
   val MaxNonce: BigInt = BigInt(2).pow(64)
 
   val DagFilePrefix: ByteString = ByteString(Array(0xfe, 0xca, 0xdd, 0xba, 0xad, 0xde, 0xe1, 0xfe).map(_.toByte))
+
+  sealed trait MineResult {
```
rtkaczyk: Why is it `MineResult`, not `MiningResult`?  "Mine" is not an action, so this sounds a bit weird to me. Same for the subclasses.

mcsherrylabs: ('Mine' is an action. They mine ETC. Sounds ok to me.)

rtkaczyk: A verb is not an action, it describes one. We need a noun. When you compute something, do you get a _compute result_ or a _computation result_?

rtkaczyk: `MineResult` would be fine as a command. Maybe it sounds ok to you because you misconstrued it as such? This is why I think it's important to address this.

LukasGasior1: In case you haven't noticed, I changed it :)

rtkaczyk: Yeah, I only noticed after a while. GH pages are refreshing somewhat strangely when you keep them open.

mcsherrylabs: Oh boy. Glad I missed this.

PR #348, opened 2017-11-29
Fix fork resolver on private networks

LukasGasior1:
We don't want to check forks on private networks


AlanVerbner:
What about setting `mantis.blockchain.dao=null`? I'm doing that on my local tests and seems to be working. Am I missing a case?

rtkaczyk:
If the config setting does the job, I'd say this is not needed. A private network with an simulated fork over the DAO bailout could imaginably be a valid test scenario, and this would not allow it, right?

mcsherrylabs:
Sorry @LukasGasior1 it was on *my* list to update that doco.

PR #351, opened 2017-11-30
Fix nonce in miner

LukasGasior1:
It could happen that our miner generated a nonce that was longer (or shorter) than required 8 bytes. Mantis accepted such blocks just fine, but Geth was rejecting them. Actually, what do you think about also validating it (nonce length) as a part of block validation?

### DIFF
--- a/src/main/scala/io/iohk/ethereum/mining/Miner.scala
+++ b/src/main/scala/io/iohk/ethereum/mining/Miner.scala
@@ -21,6 +21,7 @@ import scala.concurrent.Future
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 import scala.util.{Failure, Random, Success, Try}
+import io.iohk.ethereum.utils.BigIntExtensionMethods._
 
 class Miner(
     blockchain: Blockchain,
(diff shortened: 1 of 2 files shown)


src/main/scala/io/iohk/ethereum/mining/Miner.scala
```diff
@@ -159,10 +160,11 @@ class Miner(
 
     (0 to numRounds).toStream.map { n =>
       val nonce = (initNonce + n) % MaxNonce
-      val pow = Ethash.hashimoto(headerHash, nonce.toByteArray, dagSize, dag.apply)
-      (Ethash.checkDifficulty(difficulty, pow), pow, nonce, n)
+      val nonceBytes = ByteUtils.padLeft(ByteString(nonce.toUnsignedByteArray), 8)
```
LukasGasior1: `nonce.toByteArray` (signed) was not what we wanted here. We need unsigned byte array padded to 8 bytes

rtkaczyk: Do you think it might have been more straightforward if `nonce` was a `BigInt` or `UInt256` (here and in `BlockHeader`)? I'm just wondering, it's not a right time for changes like that...

LukasGasior1: I was thinking more about `require(nonce.length == 8)` in BlockHeader, but yeah, it's not good time for it

PR #352, opened 2017-12-04
Ommers validation fix

LukasGasior1:
TODO: test case


rtkaczyk:
Yep, the fix obviously looks good. Test case in `BlockImportSpec` would be nice

LukasGasior1:
It turns out it's not that straightforward to test it. Currently we use mock validators impl in `BlockImportSpec`. How about a task in YT for that? @rtkaczyk @mcsherrylabs

rtkaczyk:
I'd be happy with a YT task

LukasGasior1:
EC-364

PR #353, opened 2017-12-04
Returning total difficulty in status message instead of best block difficulty

KonradStaniec:


src/main/scala/io/iohk/ethereum/network/handshaker/EtcNodeStatusExchangeState.scala
```diff
@@ -44,10 +44,11 @@ case class EtcNodeStatusExchangeState(handshakerConfiguration: EtcHandshakerConf
 
   private def createStatusMsg(): Status = {
     val bestBlockHeader = getBestBlockHeader()
+    val totalDifficulty = blockchain.getTotalDifficultyByHash(bestBlockHeader.hash).getOrElse(bestBlockHeader.difficulty)
```
rtkaczyk: Why `getOrElse(bestBlockHeader.difficulty)`? Wouldn't it better to return 0 or throw?

KonradStaniec: I have chosen that as: 
sending 0 to other peers because of our error condition seems weird 
getting here null pointer seems to have a low probability (if we have best block, we should have its total difficulty)
bestBlockHeader.difficulty was working good till now, so it seemed like a good default in case of this rare error.

Do you think that throwing here could have some value to us ?

mcsherrylabs: @rtkaczyk What is the requested change? Remove 'getOrElse(bestBlockHeader.difficulty)'?

rtkaczyk: > sending 0 to other peers because of our error condition seems weird

It may indicate to other peers that we're not capable of serving, which is true because our DB is clearly broken.

> Do you think that throwing here could have some value to us ?

We get an immediate and accurate error message, rather than a possibly lengthy debugging in the (unlikely) event it does go wrong.


Baseline is our best block difficulty is just a random number in relation to total difficulty.

KonradStaniec: As i am aware in other such cases we are just calling `.get` (e.g in ledger). Maybe we will just stick to it here ?

rtkaczyk: Good enough for me, as long as it works at the genesis block.

rtkaczyk: [Should be fine](https://github.com/input-output-hk/mantis/blob/phase/daedalus/src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala#L119)

src/test/scala/io/iohk/ethereum/network/p2p/PeerActorSpec.scala
```diff
@@ -401,7 +401,16 @@ class PeerActorSpec extends FlatSpec with Matchers {
       extraData = ByteString("0"),
       mixHash = ByteString("0"),
       nonce = ByteString("0"))
-    blockchain.save(testGenesisHeader)
+
+    val testGenesisBlockBody: BlockBody = BlockBody(
+      transactionList = Seq[SignedTransaction](
+      ),
+      uncleNodesList = Seq[BlockHeader](
+      ))
```
rtkaczyk: Minor: `BlockBody(Nil, Nil)` also does the job and looks much simpler

KonradStaniec: fair point

PR #354, opened 2017-12-05
[EC-350] Adapting ECIP1039 tables as test cases

KonradStaniec:
Test tables from: https://github.com/ethereumproject/ECIPs/blob/master/ECIPs/ECIP-1039.md


src/main/scala/io/iohk/ethereum/ledger/BlockRewardCalculator.scala
```diff
@@ -8,6 +8,8 @@ import io.iohk.ethereum.utils.MonetaryPolicyConfig
   * rounding errors.
   * A Geth issue (ethereumproject/go-ethereum#352) and a comment on the ECIP (https://github.com/ethereumproject/ECIPs/issues/15#issuecomment-330660976)
   * were created to check whether this should be changed
+  * Previous issues resulted in https://github.com/ethereumproject/ECIPs/blob/master/ECIPs/ECIP-1039.md which completely
```
rtkaczyk: Doc comments should reflect the current state so please remove the 2 lines above and reference to "Previous issues" in this line.

src/main/scala/io/iohk/ethereum/ledger/BlockRewardCalculator.scala
```diff
@@ -6,8 +6,7 @@ import io.iohk.ethereum.utils.MonetaryPolicyConfig
   * Calculates rewards for mining blocks and ommers.
   * To be compatible with Geth division was not defered to be a single and final operation, which produces expected
   * rounding errors.
-  * A Geth issue (ethereumproject/go-ethereum#352) and a comment on the ECIP (https://github.com/ethereumproject/ECIPs/issues/15#issuecomment-330660976)
-  * were created to check whether this should be changed
+  * https://github.com/ethereumproject/ECIPs/blob/master/ECIPs/ECIP-1039.md completely specifies eventual rounding issues.
```
rtkaczyk: Sorry, I wasn't specific enough, I meant to also remove this sentence:

> * To be compatible with Geth division was not defered to be a single and final operation, which produces expected
>  * rounding errors.

It addresses the previous erroneous approach to these calculations, and it only serves to confuse the reader. Could please remove it as well?

KonradStaniec: no problem, I did not know the whole earlier story and did not want to remove too much

PR #357, opened 2017-12-06
[EC-358] fix branch resolution depth

rtkaczyk:
Branch resolution has been simplified by making a single request (of configurable size) for additional headers. Therefore no additional constraints for branch length are required.

This has been tested by revoking the fix made in #352 and forcing the issue to occur. After re-applying the said fix, syncing resumed immediately - previously it wouldn't have happened unless a peer chose to ignore our request for 200 headers and responded instead with something between 1 and 13 headers (which probably would have happened eventually).

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
@@ -37,9 +37,15 @@ class RegularSync(
   import RegularSync._
   import syncConfig._
 
+
   private var headersQueue: Seq[BlockHeader] = Nil
   private var waitingForActor: Option[ActorRef] = None
   var resolvingBranches: Boolean = false
+  /** This is used as an optimisation to avoid handling broadcast messages when we haven't reached the top of the chain.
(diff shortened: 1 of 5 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -186,7 +192,7 @@ class RegularSync(
       log.info("Received {} block headers in {} ms", headers.size, timeTaken)
       waitingForActor = None
       if (resolvingBranches) handleBlockBranchResolution(peer, headers.reverse)
```
rtkaczyk: Somehow `headersQueue` can be empty here... resulting in `java.util.NoSuchElementException: head of empty list` 🤔

rtkaczyk: Cause found, fix will be pushed soon

rtkaczyk: ✅

PR #359, opened 2017-12-08
cancel branch resolution after request failure

rtkaczyk:
Fixes a bug introduced in #357

Branch resolution was not cancelled after the request for additional headers failed, which led to wrong assumption that `resolvingBranch` mode is still on and `headersQueue` is non-empty, resulting in: `java.util.NoSuchElementException: head of empty list`

Also adds more logging for `RegularSync`, and refactors `SyncConfig` into a case class (I finally had motivation to do it).


src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -376,6 +382,7 @@ class RegularSync(
 
   private def scheduleResume() = {
     headersQueue = Nil
+    resolvingBranches = false
```
mcsherrylabs: The original PR #357 was successfully tested manually. Was this test re run? 
Can we get a unit test for this and some explanatory text on the PR?

rtkaczyk: The PR title itself is an explanation, branch resolution mode was not turned off after the request for additional headers failed. 
This a rare occurrence so the manual test did not catch immediately, but I left it running and it happened overnight.
As for unit test, I'll give it another look, but I would have to simulate a failed request and it may not be trivial with the current design.

mcsherrylabs: It's not much of an explanation, what were the implications of the bug?

rtkaczyk: Description updated

PR #364, opened 2017-12-11
[EC-362] eth_call fails when executed in "pending" block

KonradStaniec:
Solved by keeping readOnlyWorldStateProxy of world changed by pending block in cache, and using it to simulate transactions.


ntallar:
In relation to this change, should we maybe use the pending state for other JSON-RPC methods? For example, `eth_getBalance` and other methods accept a `pending` tag, and at least in [Geth](https://github.com/ethereumproject/go-ethereum/blob/master/eth/api.go#L574) it seems to use a pending state as in this PR. If that's the case we should probably create a separate task to look into that.

KonradStaniec:
@ntallar it is a good point, I will create a task on yt to review other endpoints which need pending state.

PR #365, opened 2017-12-11
[EC-368] redownload missing state nodes

rtkaczyk:
~~Proof of concept - poor code quality~~
[EC-368](https://iohk.myjetbrains.com/youtrack/issue/EC-368)

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -403,7 +403,7 @@ mantis {
     # off - missing state node will result in an exception
     # on - missing state node will be redownloaded from a peer and block execution will be retried. This can repeat
     #      several times until block execution succeeds
-    redownload-missing-state-nodes = off
+    redownload-missing-state-nodes = on
   }
 
   pruning {
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -392,12 +392,18 @@ mantis {
     # then: the block will not be queued (such already queued blocks will be removed)
     max-queued-block-number-ahead = 10
 
-    # Maxium number of blocks, after which block hash from NewBlockHashes packet is considered ancient
+    # Maximum number of blocks, after which block hash from NewBlockHashes packet is considered ancient
     # and peer sending it is blacklisted
     max-new-block-hash-age = 20
 
-    # Maxium number of hashes processed form NewBlockHashes packet
+    # Maximum number of hashes processed form NewBlockHashes packet
     max-new-hashes = 64
+
+    # This a recovery mechanism for the issue of missing state nodes during blocks execution:
+    # off - missing state node will result in an exception
+    # on - missing state node will be redownloaded from a peer and block execution will be retried. This can repeat
+    #      several times until block execution succeeds
+    redownload-missing-state-nodes = off
   }
 
   pruning {
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
@@ -3,6 +3,7 @@ package io.iohk.ethereum.blockchain.sync
 import akka.actor._
 import akka.util.ByteString
 import io.iohk.ethereum.blockchain.sync.PeerRequestHandler.ResponseReceived
+import io.iohk.ethereum.crypto._
 import io.iohk.ethereum.db.storage.AppStateStorage
 import io.iohk.ethereum.domain._
 import io.iohk.ethereum.ledger._
@@ -26,6 +27,7 @@ import scala.concurrent.ExecutionContext.Implicits.global
 import scala.util.{Failure, Success, Try}
 
 //TODO Refactor to get rid of most of mutable state [EC-320]
+// scalastyle:off number.of.methods
 class RegularSync(
     val appStateStorage: AppStateStorage,
     val etcPeerManager: ActorRef,
@@ -213,7 +215,9 @@ class RegularSync(
     case ResponseReceived(peer, BlockBodies(blockBodies), timeTaken) =>
       log.info("Received {} block bodies in {} ms", blockBodies.size, timeTaken)
       waitingForActor = None
-      handleBlockBodies(peer, blockBodies)
+      matchBlockBodiesWithHeaders(peer, blockBodies).foreach { blocks =>
+        handleBlocks(peer, blocks)
+      }
 
     case ResponseReceived(peer, NodeData(nodes), timeTaken) =>
       log.info("Received {} missing state nodes in {} ms", nodes.size, timeTaken)
@@ -300,12 +304,22 @@ class RegularSync(
       log.debug(s"Did not receive missing state node from peer ($peer)")
       resumeWithDifferentPeer(peer, "did not receive missing state node")
     } else {
-      val MissingStateNodeRetry(nodeHash, blockPeer, nextBlocks) = missingStateNodeRetry.get
-      blockchain.saveNode(nodeHash, nodes.head.toArray, headersQueue.head.number)
-      missingStateNodeRetry = None
-      log.info(s"Inserted missing state node: ${Hex.toHexString(nodeHash.toArray)}. " +
-        s"Retrying execution starting with block ${headersQueue.head.number}")
-      handleBlockBodies(blockPeer, nextBlocks)
+      val MissingStateNodeRetry(requestedHash, blockPeer, blocksToRetry) = missingStateNodeRetry.get
+      val receivedHash = kec256(nodes.head)
+
+      if (receivedHash != requestedHash) {
+        log.debug(s"Received missing state node has different hash than requested")
+        resumeWithDifferentPeer(peer, "wrong state node hash")
+      } else {
+        val nextBlockNumber = blocksToRetry.head.header.number
+        // note that we do not analyse whether the node is a leaf, extension or a branch, thus we only
+        // handle one state node at a time and retry executing block - this may require multiple attempts
+        blockchain.saveNode(requestedHash, nodes.head.toArray, nextBlockNumber)
+        missingStateNodeRetry = None
+        log.info(s"Inserted missing state node: ${Hex.toHexString(requestedHash.toArray)}. " +
+          s"Retrying execution starting with block $nextBlockNumber")
+        handleBlocks(blockPeer, blocksToRetry)
+      }
     }
   }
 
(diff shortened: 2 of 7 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -261,6 +280,35 @@ class RegularSync(
...
+        scheduleResume()
+    }
+  }
+
+  private def handleRedownloadedStateNodes(peer: Peer, nodes: Seq[ByteString]): Unit = {
+    if (nodes.isEmpty) {
+      log.debug(s"Did not receive missing state node from peer ($peer)")
+      resumeWithDifferentPeer(peer, "did not receive missing state node")
+    } else {
+      val MissingStateNodeRetry(nodeHash, blockPeer, nextBlocks) = missingStateNodeRetry.get
```
LukasGasior1: do we need to validate the node first i.e check if hash(receivedNode) == requestedHash?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -20,6 +23,7 @@ import org.spongycastle.util.encoders.Hex
 
 import scala.annotation.tailrec
 import scala.concurrent.ExecutionContext.Implicits.global
+import scala.util.{Failure, Success, Try}
 
 //TODO Refactor to get rid of most of mutable state [EC-320]
 class RegularSync(
```
AlanVerbner: Minor: I'm getting a scalastyle error

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -361,11 +419,20 @@ class RegularSync(
         log.debug(s"got new blocks up till block: ${importedBlocks.last.header.number} " +
           s"with hash ${Hex.toHexString(importedBlocks.last.header.hash.toArray[Byte])}")
       }
+
       errorOpt match {
+        case Some(missingNodeEx: MissingNodeException) =>
+          log.error(missingNodeEx, "Requesting missing state nodes")
+          headersQueue = headersQueue.drop(blocks.length)
+          missingStateNodeRetry = Some(MissingStateNodeRetry(missingNodeEx.hash, peer, m))
```
AlanVerbner: shouldn't we set `MissingStateNodeRetry#bodies` the blocks that weren't executed (because it stopped when missing MPT error occurred) instead of `m`?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -334,25 +382,35 @@ class RegularSync(
       val blocks = headersQueue.zip(m).map{ case (header, body) => Block(header, body) }
 
       @tailrec
-      def importBlocks(blocks: List[Block], importedBlocks: List[Block] = Nil): (List[Block], Option[BlockImportResult]) =
+      def importBlocks(blocks: List[Block], importedBlocks: List[Block] = Nil): (List[Block], Option[Any]) =
```
AlanVerbner: Cant we create a `BlockImportResult` for the missing MPT node to avoid `Any`?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: yes, also `headersQueue` was wrong, and that caused the unnecessary re-download of block headers and bodies to be retried

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: I'd rather not do that. All errors extending `BlockImportResult` are logical errors, that are relevant to the nature of the blockchain. `MissingNodeException` is a result of our implementation and in the perfect world shouldn't even be handled.

I agree that `Option[Any]` does look a bit clumsy, but I would not be overly concerned with that as it's a nested function (well, it will be private now, but still).

rtkaczyk:
Improvements pushed. All pointed out issues addressed. Blocks shouldn't no longer be redownloaded in between state node requests. This needs to be retested on a synced instance - I will have to manually delete a node 😁

rtkaczyk:
Tested with manual node deletion:

```
13:22:36 Received 128 block bodies in 14808 ms
13:23:38 Requesting missing state nodes
io.iohk.ethereum.mpt.MerklePatriciaTrie$MissingRootNodeException: Root node not found f69abe4faa805697becaf033643892c528b813242755d83adcb326ab5279ab78
        at io.iohk.ethereum.mpt.MerklePatriciaTrie$.$anonfun$getRootNode$1(MerklePatriciaTrie.scala:85)
        at scala.Option.getOrElse(Option.scala:121)
  ...
13:23:38 Block: 5010123. Peers: 5 (0 blacklisted)
13:23:38 Block: 5010123. Peers: 5 (0 blacklisted)
13:23:38 Received 1 missing state nodes in 95 ms
13:23:38 Inserted missing state node: f69abe4faa805697becaf033643892c528b813242755d83adcb326ab5279ab78. Retrying execution starting with block 5010124
13:28:16 Block: 5010251. Peers: 5 (0 blacklisted)
```

Blocks and headers were **not** re-downloaded. The delay between blocks bodies download and reporting the missing node is unexpected, but oh well... it's just a patch.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -89,10 +98,17 @@ class RegularSync(
     // The case that waitingForActor is defined (we are waiting for some response),
     // can happen when we are on top of the chain and currently handling newBlockHashes message
 
-    if (waitingForActor.isEmpty)
-      askForHeaders()
-    else
+    if (waitingForActor.isEmpty) {
```
AlanVerbner: (super minor) is it worthy to replace those nested `if`that check for `isEmpty` with something else like map or pattern matching?

src/main/resources/application.conf
```diff
@@ -392,12 +392,18 @@ mantis {
...
 
-    # Maxium number of hashes processed form NewBlockHashes packet
+    # Maximum number of hashes processed form NewBlockHashes packet
     max-new-hashes = 64
+
+    # This a recovery mechanism for the issue of missing state nodes during blocks execution:
+    # off - missing state node will result in an exception
+    # on - missing state node will be redownloaded from a peer and block execution will be retried. This can repeat
+    #      several times until block execution succeeds
+    redownload-missing-state-nodes = off
```
mcsherrylabs: Please add to sync.conf also.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -261,6 +284,45 @@ class RegularSync(
...
+      case None =>
+        log.debug("Requesting missing state nodes: no peers to download from")
+        scheduleResume()
+    }
+  }
+
+  private def handleRedownloadedStateNodes(peer: Peer, nodes: Seq[ByteString]): Unit = {
+    if (nodes.isEmpty) {
+      log.debug(s"Did not receive missing state node from peer ($peer)")
+      resumeWithDifferentPeer(peer, "did not receive missing state node")
```
AlanVerbner: if our best peer doesn't have a specific MPT node (that might be old) we will be blacklisting it even if best peer is selected by total difficulty (nothing related to state nodes). That being said, if you think that should be handled in another task or even dismissed as it doesn't present any problems or for this patch is good enough, dismiss this comment.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -329,18 +391,60 @@ class RegularSync(
         responseMsgCode = BlockBodies.code)))
   }
 
-  private def handleBlockBodies(peer: Peer, m: Seq[BlockBody]) = {
-    if (m.nonEmpty && headersQueue.nonEmpty) {
-      val blocks = headersQueue.zip(m).map{ case (header, body) => Block(header, body) }
+  private def matchBlockBodiesWithHeaders(peer: Peer, m: Seq[BlockBody]): Option[Seq[Block]] = {
+    if (m.nonEmpty) {
+      assert(headersQueue.nonEmpty, "handling block bodies while not having any queued headers")
```
AlanVerbner: 👍

src/test/scala/io/iohk/ethereum/blockchain/sync/RegularSyncSpec.scala
```diff
@@ -352,6 +353,34 @@ class RegularSyncSpec extends TestKit(ActorSystem("RegularSync_system")) with Wo
...
+        inSequence {
+          (ledger.resolveBranch _).expects(Seq(newBlock.header)).returning(NewBetterBranch(Nil))
+          (ledger.importBlock _).expects(newBlock).throwing(new MissingNodeException(missingNodeHash))
+          (ledger.importBlock _).expects(newBlock).returning(BlockImportedToTop(List(newBlock), List(0)))
+        }
+
+        sendBlockHeadersFromBlocks(Seq(newBlock))
+        sendBlockBodiesFromBlocks(Seq(newBlock))
+
+        Thread.sleep(10000)
```
AlanVerbner: 🤔

src/main/resources/application.conf
LukasGasior1: why isn't it "on" by default?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -199,7 +215,14 @@ class RegularSync(
     case ResponseReceived(peer, BlockBodies(blockBodies), timeTaken) =>
       log.info("Received {} block bodies in {} ms", blockBodies.size, timeTaken)
       waitingForActor = None
-      handleBlockBodies(peer, blockBodies)
+      matchBlockBodiesWithHeaders(peer, blockBodies).foreach { blocks =>
```
LukasGasior1: it looks a bit strange, the logic is scattered. if the result is Some we handle the blocks here, if it's None we resume inside  `matchBlockBodiesWithHeaders`. How about moving side effects from this method here?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -261,6 +284,45 @@ class RegularSync(
...
+        scheduleResume()
+    }
+  }
+
+  private def handleRedownloadedStateNodes(peer: Peer, nodes: Seq[ByteString]): Unit = {
+    if (nodes.isEmpty) {
+      log.debug(s"Did not receive missing state node from peer ($peer)")
+      resumeWithDifferentPeer(peer, "did not receive missing state node")
+    } else {
+      val MissingStateNodeRetry(requestedHash, blockPeer, blocksToRetry) = missingStateNodeRetry.get
```
LukasGasior1: `.get` alert :smile: someone may just send us NodeData msg and cause this actor to restart

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -261,6 +284,45 @@ class RegularSync(
...
+      val receivedHash = kec256(nodes.head)
```
LukasGasior1: it probably doesn't matter anyway, but how about `receivedNodeOpt = nodes.find(kec256(_) == requestedHash)` and then match on `receivedNodeOpt`?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -199,7 +215,14 @@ class RegularSync(
     case ResponseReceived(peer, BlockBodies(blockBodies), timeTaken) =>
       log.info("Received {} block bodies in {} ms", blockBodies.size, timeTaken)
       waitingForActor = None
-      handleBlockBodies(peer, blockBodies)
+      matchBlockBodiesWithHeaders(peer, blockBodies).foreach { blocks =>
+        handleBlocks(peer, blocks)
+      }
+
+    case ResponseReceived(peer, NodeData(nodes), timeTaken) =>
```
LukasGasior1: maybe add `if missingStateNodeRetry.isDefined` (it'll solve my comment below about using .get).

src/main/resources/application.conf
rtkaczyk: @LukasGasior1 because it's a patch not a solution. I'd like for us to be aware the problem still occurs and have motivation to finally fix it.

@mcsherrylabs is it something we want advertise to the users?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: Good catch

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: Nb. `RegularSync` actor does not restart

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: I don't think I follow, what is the gain here? (for both suggestions)

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: It's not about about a peer not having this node, it simply about it refusing to give it to us. If the peer does not have the node we're most likely screwed anyway - we're too far behind, pruning has kicked in other nodes, and there's no way for us to recover.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: This method handles various responses so I'd like to keep it's body short. I have another idea...

src/test/scala/io/iohk/ethereum/blockchain/sync/RegularSyncSpec.scala
rtkaczyk: Good catch, this is unintentional

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: What would that look like? TBH I believe this whole class deserves a full rewrite. I have yet to write a single change in it without introducing a bug. There should be clearly defined state (mode of operation), that determines the way we handle all the mutables - currently it's the other way around.

src/main/resources/application.conf
mcsherrylabs: We need to have this 'on' for release. It can be 'off' in dev. But when this happens the user doesn't want to have to discover a problem (the wallet screen will just pause on 99.98 percent) and then go look online for a fix.

rtkaczyk: So:
- `on` by default, not shown in `sync.conf`
- `off` by default, overridden in `sync.conf`
- or... ?

LukasGasior1: I'd say "on by default, not shown in sync.conf"

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
LukasGasior1: what happens when uncaught exception is thrown? I thought default supervisorStrategy is to restart

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
LukasGasior1: for my suggestion, as I said it probably doesn't matter :) the gain is, that if we somehow receive 2 or more nodes we can still handle it. Anyway, you can disregard that comment

src/main/resources/application.conf
mcsherrylabs: :+1:

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
AlanVerbner: Peer doesn't have it or is refusing but it's giving us a lot of "good blocks". A good example is a good and responsive node that is running pruning (as the missing MPT node might be from any block within the history, right?). That being said, this is starting to become a philosophical discussion and this is meant to be a _patch_ so, lt's dismiss this comment.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
AlanVerbner: And I meant that we asked for a single MPT node, why would respond with more than 1?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: FWIW, blacklisting is only temporary, so by the time we receive the node and re-execute the blocks, the peer will most likely be un-blacklisted.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: > why would respond with more than 1

exactly, I chose not to deal with other situations.

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
rtkaczyk: My observation was that after `MPTException` is thrown, the `RegularSync` actor does not restart

LukasGasior1: strange...wasn't it wrapped in Try or something like that in Ledger?

AlanVerbner: For me it seems that after the exception, `RegularSync` actor stops asking for blocks, does that make sense or am I seeing something wrong?

rtkaczyk: There is no logging whatsoever in `RegularSync` after an exception is thrown, so I just assumed the actor died and was not restarted. Whatever the true nature of this behaviour is, it **was not changed** in this PR. I also don't see any reason to change it.

AlanVerbner: mmm...but this was changed. After this PR this would break if we didn't have an MPT node (our fault) but in this case, any malicious peer can start sending us MPT node messages to make our `RegularSync` to stop, can't they?

rtkaczyk: > mmm...but this was changed.

The way `RegularSync` actor (or its supervisor) deals with exceptions was not changed.

> peer can start sending us MPT node messages to make our RegularSync to stop, can't they?

They can't. That was fixed as per [this comment](https://github.com/input-output-hk/mantis/pull/365#discussion_r156730881)

AlanVerbner: you are going to kill me...I didn't pull that change.

PR #367, opened 2017-12-12
Changed Mantis default log location to datadir

ntallar:
## Description

Changed Mantis default log and log zips location to `path/to/home/logs`.


mcsherrylabs:
Had wrong base. merged and retargetted

PR #368, opened 2017-12-14
[EC-381] bootstrap cmd line to download bootstrap file to db folder.

mcsherrylabs:
This allows a commandline to download a bootstrap to levedb folder. 

Can be tested with 

sbt "run bootstrap https://s3-eu-west-1.amazonaws.com/iohk.mantis.bootstrap/downloadtest.zip a92cdbb166d9aa517546422334993e8b 47"

~~Needs unit tests.~~

~~(Ignore Radeks code, I branched off his)~~


src/main/scala/io/iohk/ethereum/BootstrapDownload.scala
```diff
@@ -0,0 +1,135 @@
...
+      val dis = new DigestInputStream(is, md5)
+      try {
+        val out = new FileOutputStream(outFile)
+        try {
+          val buffer = new Array[Byte](bufferSize)
+          Stream.continually(dis.read(buffer)).takeWhile(_ != -1).foreach(out.write(buffer, 0, _))
+        } finally (out.close())
+        md5.digest.map("%02x".format(_)).mkString
+      } finally(dis.close())
+    } finally(is.close())
```
rtkaczyk: How about using loan pattern to close the resources (akin to Java's ARM)? It might look something like:

```
def using[T <: { def close() }, U](resource: T)(body: T => U): U = {
  try {
    body(resource)
  } finally {
    resource.close()
  }
}
```

Then:

```
using(new URL(urlToDownloadFrom).openStream()) { is =>
  readStream(is)
}
```

src/main/scala/io/iohk/ethereum/BootstrapDownload.scala
```diff
@@ -0,0 +1,135 @@
...
+  * - clean files out of given location
+  * - unzip to a given location
+  */
+object BootstrapDownload extends Logger {
+
+  val bufferSize = 4 * 1024
+  val leveldbFolderName = "leveldb"
+
+  private def assertAndLog(cond: Boolean, msg: String): Unit = {
+    if(!cond) log.info(msg)
```
rtkaczyk: What's this for? The message will be shown anyway, won't it?

src/main/scala/io/iohk/ethereum/BootstrapDownload.scala
```diff
@@ -0,0 +1,135 @@
...
+    val is = new URL(urlToDownloadFrom).openStream()
+    try {
+      val dis = new DigestInputStream(is, md5)
+      try {
+        val out = new FileOutputStream(outFile)
+        try {
+          val buffer = new Array[Byte](bufferSize)
+          Stream.continually(dis.read(buffer)).takeWhile(_ != -1).foreach(out.write(buffer, 0, _))
+        } finally (out.close())
+        md5.digest.map("%02x".format(_)).mkString
```
rtkaczyk: Same as `Hex.toHexString(md5.digest)`

src/main/scala/io/iohk/ethereum/BootstrapDownload.scala
rtkaczyk: Also, we don't need as many `finally`s, `dis` will close `is`.

src/main/scala/io/iohk/ethereum/BootstrapDownload.scala
mcsherrylabs: The exe on windows will not shoe console messages. So it goes in the log too.

src/main/scala/io/iohk/ethereum/BootstrapDownload.scala
mcsherrylabs: I did think about using another pattern for resource closing, I like it cos it's neat, but it's very debatably clearer. A quick search of the codebase shows we use 'try'/'finally'. That might be a post 1.0 task. 

I just checked and the Digest constructor is pretty safe so we can put them together. I had them together but I separated them wondering if 

`val dis = new DigestInputStream(new URL(urlToDownloadFrom).openStream(), md5)`

... DigestInputStream throws in the constructor is that a resource leak? In this case it's not.

src/main/scala/io/iohk/ethereum/BootstrapDownload.scala
mcsherrylabs: :+1: (Right, but the SO I cut and paste from didn't know that .... )

rtkaczyk:
Maybe it's nothing of relevance, but it didn't quite work with a custom path. When I tried `/tmp/leveldb` it used `~/tmp/leveldb` instead, and still I couldn't find any contents in it. Anyway, worked fine with the suggested parameters.

rtkaczyk (dismissed):
Works! Minor style improvements possible.

rtkaczyk (approved):

KonradStaniec (approved):
LGTM

PR #369, opened 2017-12-15
[EC-374] Fix account tx ordering

LukasGasior1:


rtkaczyk:
As a test you could just add more than 1 TX as the expectation in the existing test

rtkaczyk:
And by the way, why this change? Am I to assume it's for daedalus? Is there a task?

KonradStaniec:
@rtkaczyk the task https://iohk.myjetbrains.com/youtrack/issue/EC-374

KonradStaniec:
Code looks good (although test case would be appreciated), but in private network, i am still getting some weird sorting of transactions in Daedalus.
summary

transactions

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -754,6 +754,7 @@ class EthService(
               receivedInBlock.map(TransactionResponse(_, Some(block.header), pending = Some(false))))
           }
           .filter { case (sent, received) => sent.nonEmpty || received.nonEmpty }
+          .map { case (sent, received) => (sent.reverse, received.reverse) }
```
rtkaczyk: Style/optional: I think it would more readable if it was inside the first `map`. Also would be great if that inner function returned something like `(a, b)`, not `(a.map(intoSomethingComplex), b.map(...))`

rtkaczyk (dismissed):
Looks good. Optional readability improvement

KonradStaniec (dismissed):
LGTM

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
KonradStaniec: I was thinking of replacing that `.filter.map` into sth like 
`          .collect {
            case (sent, received) if sent.nonEmpty || received.nonEmpty => (sent.reverse, received.reverse)
          }`
but i also consider it optional.

LukasGasior1: Both comments addressed. I used collect and added intermediate vals in the map above.

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -733,32 +733,27 @@ class EthService(
...
           .map(_.stx)
-          .filter(_.tx.receivingAddress.contains(request.address))
-          .map(TransactionResponse(_, pending = Some(true)))
+          .collect(collectTxs(None, pending = true))
 
-        val txsFromBlocks = (request.fromBlock to request.toBlock)
+        val txsFromBlocks = (request.toBlock to request.fromBlock by -1)
           .toStream
           .flatMap { n => blockchain.getBlockByNumber(n) }
           .map { block =>
```
KonradStaniec: Minor/style: Instead `map` and latter `flatten` you can use `flatMap`

KonradStaniec (dismissed):
One minor comment , LGTM

mcsherrylabs (dismissed):
Given you've both tested it (I didn't), LGTM. 
One query on the from to block - will putting those in backwards cause any issue?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -733,32 +733,27 @@ class EthService(
         s"""Maximum number of blocks to search is $accountTransactionsMaxBlocks, requested: $numBlocksToSearch.
            |See: 'network.rpc.account-transactions-max-blocks' config.""".stripMargin)))
```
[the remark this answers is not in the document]
LukasGasior1: (see below: if "to block" is smaller than "from block": `(50 to 100 by -1).toList ==  List()`

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
LukasGasior1: done

LukasGasior1:
@KonradStaniec @mcsherrylabs I've pushed a style fix, need your approvals again

KonradStaniec (approved):

mcsherrylabs (approved):

PR #375, opened 2017-12-29
[EC-391] Download bootstrap file to target dir

LukasGasior1:


src/main/scala/io/iohk/ethereum/BootstrapDownload.scala
```diff
@@ -33,7 +33,7 @@ object BootstrapDownload extends Logger {
     val leveldbFolder = pathToDownloadTo.toFile
     assertAndLog(leveldbFolder.isDirectory, s"${pathToDownloadTo} must be a folder.")
     assertAndLog(leveldbFolder.getName == leveldbFolderName, s"${pathToDownloadTo} must end in a folder named $leveldbFolderName")
-    leveldbFolder.listFiles().foreach(_.delete())
+    leveldbFolder.listFiles(pathname => !pathname.getName.endsWith(".zip")).foreach(_.delete())
```
rtkaczyk: Why was this added? Won't this prevent re-download (if the first download was cancelled)?

LukasGasior1: No, re-download will behave the same way (override existing file). I've added this because we now download zip to the datadir (instead of current directory, which may be on different location). This datadir is later cleared and bootstrap db unzipped. We don't want to delete the newly downloaded zip while clearing it.

rtkaczyk: Oh, I thought we were cleaning out the *prior to* downloading - wouldn't that make more sense? Anyway, good enough for me.

LukasGasior1: we could do that prior but now, in case download fails old db is still there

PR #377, opened 2018-01-03
[EC-395] Optimisation of ethash algorithm

KonradStaniec:
Main improvements by:
1. rewriting foreach as while loops, and avoiding allocation of `Range` object ( `scalac` compiler should do this under the hood, but as it seems it fails sometimes)

2. Avoiding Allocation new Keccak512 engine during each hash computation as it is quite expensive. And taking digest always reset engine to starting state. 

3. optimising array translations between byte and int array.  changing `scala` constructs `map` and `flatMaps` to operations on`ByteFuffers`.  

As it stands  current bottleneck is point `3`, but as I rewritten those to manually manipulating bytes on pre-allocated arrays,  point of contention just moved to https://github.com/input-output-hk/mantis/pull/377/files#diff-0dafe97cd157b1f1432980a711be1ebdR196, and there was only little speedup (1 - 2%) so i left it as it is, because `ByteBuffer` implementation is much simpler.

And I think we cant do much about https://github.com/input-output-hk/mantis/pull/377/files#diff-0dafe97cd157b1f1432980a711be1ebdR196, as it is just this algorithm complexity and it is implemented similarly in other clients.

During test of syncing from one master node number of blocks imported during `10 min` increased from `40448` to `71936`.
Attached recording shows as bottleneck moves from allocating `Range` objects, to byte array to int array conversion as mentioned in point `3`

[PerfNoPerfRecordings.zip](https://github.com/input-output-hk/mantis/files/1600299/PerfNoPerfRecordings.zip)

[perfNoPerfComp.txt](https://github.com/input-output-hk/mantis/files/1600295/perfNoPerfComp.txt)

### DIFF
--- a/src/main/scala/io/iohk/ethereum/consensus/Ethash.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/Ethash.scala
@@ -8,6 +8,7 @@ import io.iohk.ethereum.crypto.{kec256, kec512}
 import io.iohk.ethereum.utils.ByteUtils._
 import org.spongycastle.util.BigIntegers
 import org.spongycastle.util.encoders.Hex
+import java.lang.Integer.remainderUnsigned
 
 import scala.annotation.tailrec
 
@@ -123,11 +124,11 @@ object Ethash {
 
   def hashimoto(hashWithoutNonce: Array[Byte], nonce: Array[Byte], fullSize: Long, datasetLookup: Int => Array[Int]): ProofOfWork = {
     /* watch out, arrays are mutable here */
-
-    val w = MIX_BYTES / WORD_BYTES
+    val wHash = MIX_BYTES / WORD_BYTES
     val mixHashes = MIX_BYTES / HASH_BYTES
-    val s = bytesToInts(kec512(hashWithoutNonce ++ nonce.reverse))
     val mix = new Array[Int](MIX_BYTES / 4)
+
+    val s = bytesToInts(kec512(hashWithoutNonce ++ nonce.reverse))
 
     (0 until mixHashes).foreach { i =>
       System.arraycopy(s, 0, mix, i * s.length, s.length)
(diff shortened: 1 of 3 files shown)


src/main/scala/io/iohk/ethereum/consensus/Ethash.scala
```diff
@@ -8,6 +8,7 @@ import io.iohk.ethereum.crypto.{kec256, kec512}
 import io.iohk.ethereum.utils.ByteUtils._
 import org.spongycastle.util.BigIntegers
 import org.spongycastle.util.encoders.Hex
+import java.lang.Integer.remainderUnsigned
```
rtkaczyk: 😮 
Great find, maybe there are other functions we can replace with imports? 😛 

Have you compared both impls wrt functionality and speed?

src/main/scala/io/iohk/ethereum/crypto/package.scala
```diff
@@ -17,6 +17,8 @@ package object crypto {
   val curveParams: X9ECParameters = SECNamedCurves.getByName("secp256k1")
   val curve: ECDomainParameters = new ECDomainParameters(curveParams.getCurve, curveParams.getG, curveParams.getN, curveParams.getH)
 
+  val kec512engine = new Keccak512
```
rtkaczyk: Definitely looks like a speed improvement, but is it secure this way? If it is maybe we could do this for other digests as well?

src/main/scala/io/iohk/ethereum/consensus/Ethash.scala
KonradStaniec: As terms of seed, I only compared it globally (run ethash with our own implementation and with this import) and there was no speed difference and recording showed almost the same time in `fnv` function which uses it. 
As for other imports, I think there weren't any to use in `ethash` and I have not looked at it globally.

src/main/scala/io/iohk/ethereum/crypto/package.scala
KonradStaniec: Each digest resets engine state so it should be safe in one threaded environment.  That's why i have synchronized method which uses this engine ( `kec512` is only used in miner and ethash so this should not lead to performance decrease because of thread contention)

As for `kec256` it is used in several other actors so synchronising its method could lead to thread contention  and not much performance increase. It could possibly be negated by each Actor having its own engine and passing it to the function, but i am not sure it worth it. (maybe one place i can think of which would gain from it is possibly `DiscoveryListner` as it is alocating new engine for each message decoding and encoding and some recording were pointing to that methods)

PR #378, opened 2018-01-04
[EC-397 ]Improve logging format

KonradStaniec:
Former format required `String` concatenation and calling `.toString` methods, which were pretty costly for `BigInts` and `Uint256`, even when trace logging turned off (when there was a lot vm action going on it was taking 10% of our cpu time)

Current format incurs only cost of creating `Array[Object]`, `Integer.valueOf` are necessary as primitives cannot be cast to `Object`

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/VM.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/VM.scala
@@ -36,7 +36,7 @@ class VM extends Logger {
       case Some(opCode) =>
         val newState = opCode.execute(state)
         import newState._
-        log.trace(s"$opCode | pc: $pc | depth: ${env.callDepth} | gas: $gas | stack: $stack")
+        log.trace("{} | pc: {} | depth: {} | gas: {} | stack: {}", opCode, Integer.valueOf(pc), Integer.valueOf(env.callDepth), gas, stack)
         if (newState.halted)
           newState
         else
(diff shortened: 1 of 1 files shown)


rtkaczyk:
I'm actually surprised we're not using this: https://github.com/typesafehub/scala-logging

KonradStaniec:
@rtkaczyk sounds good, wdyt about creating ticket to include this library and change our logger ?

src/main/scala/io/iohk/ethereum/vm/VM.scala
```diff
@@ -36,7 +36,7 @@ class VM extends Logger {
       case Some(opCode) =>
         val newState = opCode.execute(state)
         import newState._
-        log.trace(s"$opCode | pc: $pc | depth: ${env.callDepth} | gas: $gas | stack: $stack")
+        log.trace("{} | pc: {} | depth: {} | gas: {} | stack: {}", opCode, Integer.valueOf(pc), Integer.valueOf(env.callDepth), gas, stack)
```
rtkaczyk: So this is fine, but while we don't have `scala-logging` how about we do what it does under the hood, meaning:

```
        if (log.isTraceEnabled) {
          log.trace(s"$opCode | pc: $pc | depth: ${env.callDepth} | gas: $gas | stack: $stack")
        }
```

That would be even more performant than the proposed solution.

KonradStaniec: yep, it will avoid the cost of array allocation

rtkaczyk (approved):

AlanVerbner (approved):

PR #382, opened 2018-01-08
[EC-382] Feature/rpc transaction validation

KonradStaniec:
- Original target: `phase/release1_1`


src/main/scala/io/iohk/ethereum/domain/Transaction.scala
```diff
@@ -13,6 +13,23 @@ object Transaction {
   def apply(nonce: BigInt, gasPrice: BigInt, gasLimit: BigInt, receivingAddress: Address, value: BigInt, payload: ByteString): Transaction =
     Transaction(nonce, gasPrice, gasLimit, Some(receivingAddress), value, payload)
 
+
+  /**
+    * v0 ≡ Tg (Tx gas limit) * Tp (Tx gas price). See YP equation number (68)
```
loverdos: Nice symbol `≡` :)

mirkoAlic:
This improvement could be at wallet level. Anyway, if at some point we want to provide this pre-check capability, we could base on this implementation.

PR #383, opened 2018-01-09
[EC-401] Handling import of genesis block

KonradStaniec:
Little test refactor


rtkaczyk:
Why would the genesis block be imported?

KonradStaniec:
As I understand the ticket https://iohk.myjetbrains.com/youtrack/issue/EC-401, it can happen when resolved new  better branch contains `genesisBlock`. In that case, during validation in import block, we will return `HeaderParentNotFoundError` instead of correctly `DuplicateBlock`

PR #384, opened 2018-01-16
[EC-415 ]Bug/discovery no nodes

KonradStaniec:
Pr to address 1 issue in ec-415 ticket


src/main/scala/io/iohk/ethereum/network/discovery/message.scala
```diff
@@ -1,16 +1,28 @@
...
 
+import scala.util.Try
+
 sealed trait Message {
   def packetType: Byte
 }
 
 object Endpoint {
 
+  def makeEndpoint(udpAddres: InetSocketAddress, tcpPort: Int): Endpoint =
```
LukasGasior1: small typo: udpAddres**s**

PR #387, opened 2018-01-19
[EC-421] Minimal passphrase length enforcement

KonradStaniec:
It would need changes also on Daedalus side.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -11,9 +11,6 @@ mantis {
   # The unencrypted private key of this node
   node-key-file = ${mantis.datadir}"/nodeId.keys"
 
-  # Keystore directory: stores encrypted private keys of accounts managed by this node
-  keystore-dir = ${mantis.datadir}"/keystore"
-
   # timeout for shutting down the ActorSystem
   shutdown-timeout = "15.seconds"
 
@@ -21,6 +18,18 @@ mantis {
   # https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#SecureRandom
   # Uncomment this to specify, otherwise use the default implementation
   # secure-random-algo = "NativePRNG"
+
+  keyStore {
+    # Keystore directory: stores encrypted private keys of accounts managed by this node
+    keystore-dir = ${mantis.datadir}"/keystore"
+
+    # Enforces minimal length for passphrase of this keystore
+    minimal-passphrase-length = 7
+
+    # Allows possibility for no passphrase
+    # If passphrase is set it must be greater than minimal-passphrase-length
+    allow-no-passphrase = true
+  }
 
   network {
     # Ethereum protocol version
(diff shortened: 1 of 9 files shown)


KonradStaniec:
As per Daedalus guys request, we leave the possibility to have empty passphrase (it is big change on frontend to make it impossible).
Option for empty passphrase is configurable, so CLI client who wants to have it enforced can disable this possiblity.

src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
```diff
@@ -59,6 +59,7 @@ object PersonalService {
   val InvalidAddress = InvalidParams("Invalid address, expected 20 bytes (40 hex digits)")
   val InvalidPassphrase = LogicError("Could not decrypt key with given passphrase")
   val KeyNotFound = LogicError("No key found for the given address")
+  val PassPhraseTooShort = LogicError("Provided passphrase must have at least 8 characters")
```
AlanVerbner: isn't it 7?

src/main/resources/application.conf
```diff
@@ -22,6 +19,18 @@ mantis {
   # Uncomment this to specify, otherwise use the default implementation
   # secure-random-algo = "NativePRNG"
 
+  keyStore {
```
ntallar: Should we also make this change to the [`storage.conf`](https://github.com/input-output-hk/mantis/blob/security/minimal_password_length/src/universal/conf/storage.conf#L10) file in the universal config?

src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
AlanVerbner: ps: this can be configured, it's not hardcoded at a certain number, right?

src/main/scala/io/iohk/ethereum/keystore/KeyStore.scala
```diff
@@ -40,15 +41,34 @@ trait KeyStore {
...
+
+  private def saveNewAccount(passphrase: String): Either[KeyStoreError, Address] = {
     val keyPair = generateKeyPair(secureRandom)
     val (prvKey, _) = keyPairToByteStrings(keyPair)
     val encKey = EncryptedKey(prvKey, passphrase, secureRandom)
     save(encKey).map(_ => encKey.address)
+
+  }
+
+  private def validateNewPassPhrase(passPhrase: String): Either[KeyStoreError, Unit] = {
```
ntallar: Should we also add this validation when importing pk keys? (on the `KeyStore.importPrivateKey` method)

KonradStaniec: I was thinking about it, but my conclusion was that it is better to allow to import already created key with an even too short passphrase  (therefore not scare user who wants to try mantis). All trouble he will have later is to change this passphrase (he will need to change it to empty and then to what he wants or mingle with configs.) wdyt?

src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
KonradStaniec: wooh embarrassing mistake, fixed in next commit

src/main/scala/io/iohk/ethereum/keystore/KeyStore.scala
KonradStaniec: as per discussion corrected :)

PR #388, opened 2018-01-19
external VM protocol - discussion

rtkaczyk:
This is a work in progress and not to be merged.

The point if this PR is to discuss the external VM [protocol definition](https://github.com/input-output-hk/mantis/blob/7fe93fcb2049d55f2dfbfa00bbeb05124abfe4d8/src/main/protobuf/extvm/msg.proto)


src/main/protobuf/extvm/msg.proto
```diff
@@ -0,0 +1,142 @@
...
+  bytes ownerAddr = 1;
+  bytes callerAddr = 2;
+  bytes originAddr = 3;
+  bytes contractCode = 4;
+  bytes inputData = 5;
+  bytes callValue = 6;
+  bytes gasPrice = 7;
+  bytes gasProvided = 8;
+  int32 callDepth = 9;
+  bytes receivingAddr = 10;
```
dwightguth: one more field will have to be added here, a string, containing the name of the function that is the entry point of the transaction on the contract that is being called. That field should be empty on transactions which create a new contract.

dwightguth (commented):
I've only looked at the changes to msg.proto, but I believe the comments I made there should represent all the differences in API needed between Ethereum and IELE. Feel free to comment if you have questions or foresee problems arising from any of my comments.

src/main/protobuf/extvm/msg.proto
```diff
@@ -0,0 +1,142 @@
+syntax = "proto3";
+package io.iohk.ethereum.extvm;
+
+message CallContext {
+  bytes ownerAddr = 1;
```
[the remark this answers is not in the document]
rtkaczyk: Owner's address determines the context account for state modifications, e.g. this is an address for which storage modifications will happen by executing `SSTORE`. 

The recipient's address is an implementation artifact, as for calls resulting from directly transactions (VM @ depth 0), it will always be the same as owner's address, thus will be removed from the final version.

(Internally we needed to distinguish owner's and recipient's addresses to correctly handle _precompiled contracts_, because of the possibility to call them using `DELEGATECALL` and `CALLCODE`, and how those instructions affect owner's address.)

src/main/protobuf/extvm/msg.proto
rtkaczyk: That's doable, the EVM can simply ignore that.. but how would Mantis ever provide anything relevant in that field? I guess we would have to modify the transaction object, but that may complicate things if we want to maintain support for Ethereum in the same codebase.

src/main/protobuf/extvm/msg.proto
```diff
@@ -0,0 +1,142 @@
...
+  bytes gasPrice = 7;
+  bytes gasProvided = 8;
+  int32 callDepth = 9;
+  bytes receivingAddr = 10;
+  BlockHeader blockHeader = 11;
+  BlockchainConfig blockchainConfig = 12;
+}
+
+message BlockHeader {
+  bytes parentHash = 1;
```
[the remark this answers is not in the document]
rtkaczyk: exactly

src/main/protobuf/extvm/msg.proto
```diff
@@ -0,0 +1,142 @@
...
+  bytes gasLimit = 10;
+  bytes gasUsed = 11;
+  int64 unixTimestamp = 12;
+  bytes extraData = 13;
+  bytes mixHash = 14;
+  bytes nonce = 15;
+}
+
+message BlockchainConfig {
+  bytes frontierBlockNumber = 1;
```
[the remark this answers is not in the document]
rtkaczyk: That depends on what IELE testnet is and whether there will be different configurable instances for it (like testnets for Ethereum mainnet). We don't need this at all, if we're only targeting a single network.

So, if we maintain ETC support, it might look like:

```
message BlockchainConfig {
  oneof config {
    EtcConfig etcConfig = 1;
    IELETestnet ieleTestnet = 2;
  }
}
```

where `IELETestnet` is empty.

src/main/protobuf/extvm/msg.proto
```diff
@@ -0,0 +1,142 @@
...
+  bytes nonce = 1;
+  bytes balance = 2;
+  //FIXME: the hashes required because of an implementation detail, should be gone in the final solution
+  bytes storageHash = 3;
+  bytes codeHash = 4;
+}
+
+message GetStorageData {
+  bytes address = 1;
+  bytes offset = 2;
```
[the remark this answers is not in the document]
rtkaczyk: Acknowledged. This will definitely affect many parts of the codebase. Whether those changes will be straightforward or will require a larger rework, I cannot tell at this moment.

src/main/protobuf/extvm/msg.proto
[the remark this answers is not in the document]
rtkaczyk: Yes, but again, I see no reason to keep it in the final version.

src/main/protobuf/extvm/msg.proto
```diff
@@ -0,0 +1,141 @@
...
+  repeated bytes touchedAccounts = 7;
+  repeated LogEntry logs = 8;
+}
+
+message ModifiedAccount {
+  bytes address = 1;
+  bytes nonce = 2;
+  bytes balance = 3;
+  repeated StorageUpdate storageUpdates = 4;
+  bytes code = 5;
```
dwightguth: I noticed something here. If the code of an account isn't modified, we need a way to indicate that unless we want to call GetCode on every modified account, which judging from the separate RPC call I suspect you don't want.

src/main/protobuf/extvm/msg.proto
```diff
@@ -0,0 +1,141 @@
...
+message GetAccount {
+  bytes address = 1;
+}
+
+message Account {
+  bytes nonce = 1;
+  bytes balance = 2;
+  //FIXME: the hashes required because of an implementation detail, should be gone in the final solution
+  bytes storageHash = 3;
+  bytes codeHash = 4;
```
dwightguth: per our discussion on slack, we will still need a boolean to tell us if the account has empty code, but otherwise we do not need this hash.

src/main/protobuf/extvm/msg.proto
rtkaczyk: Empty code serves as such indication. Especially in our implementation it's straightforward, because if the code is empty we simply don't save it, which works both for creating new accounts and modifying existing ones.

I understand this might be a bit obscure. Do you think we should add a flag to make it more explicit?

dwightguth: That sounds like it should be fine, but I think you should add a comment to the protobuf about it.

src/main/protobuf/extvm/msg.proto
```diff
@@ -0,0 +1,141 @@
...
+message GetCode {
+  bytes address = 1;
+}
+
+message Code {
+  bytes code = 1;
+}
+
+message GetBlockhash {
+  int32 offset = 1;
```
dwightguth: this is the 0-256 offset of the blockhash relative to the current block, not the block number of the block whose hash we want, correct?

rtkaczyk: correct

src/main/protobuf/extvm/msg.proto
```diff
@@ -0,0 +1,141 @@
...
+
+message Blockhash {
+  bytes hash = 1;
+}
+
+message CallResult {
+  bytes returnData = 1;
+  bytes gasRemaining = 2;
+  bytes gasRefund = 3;
+  bool error = 4;
```
dwightguth: one thing I realized when working on the integration today is that in IELE, an account call returns an integer exit code, rather than simply success/failure. So you may wish this to be a bytes.

PR #394, opened 2018-01-30
[EC-414] Feature/fast syc update targetblock

KonradStaniec:
Gist of pr: 
Changing fast sync download to (simillar was proposed by Radek in #380) :

1.  download state after the blocks
2.  upon reaching the target block + safe margin (called X), determine a new updated target
3. repeat above until, upon reaching the updated target block, we are in fact 500 blocks behind (+- some margin, say 10) the top of the chain
4. now downloading of the state is the same with regard to pruning in the source nodes, yet the state root hash has been validated

In case of failures, there is parameter which sets maximum number of those failures (not getting enough fresh new target, or target failed to validate) and after those, i declare fastsync failure similar to #380.

Other things:
~- As far of testing: I still need to test switch to regular sync on network (i am on 700k block),for now i have tested switch starting syncing to fake test block with some low number (30k) and everythink went fine.~
- Most of the changes are in tests (needed to refactor to write more tests)
- Fixed a bug which we have in determining next block to validate in (https://github.com/input-output-hk/mantis/blob/phase/release1_1/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala#L196)
at the end of sync random number can be greater than targetblock number and we could  fail to validate  several blocks.

Change is pretty extensive, so any discussion appreciated about possible failure modes or some edge case i may omitted.


KonradStaniec:
fix to our bug stated in the entry: https://github.com/input-output-hk/mantis/pull/394/files#diff-3766981368048abce715f8396a3aed8cR757.
I am also open to creating separate pr jus to fix this wdyt ?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -149,87 +169,158 @@ class FastSync(
...
+          context become this.receive
+          processSyncing()
+        } else {
+          syncState = syncState.copy(targetBlockUpdateFailures = syncState.targetBlockUpdateFailures + 1)
+          scheduler.scheduleOnce(syncRetryInterval, self, WaitForDownloadsToFinish(state))
+        }
+
+      case PersistSyncState => persistSyncState()
+
+      case WaitForDownloadsToFinish(sState) => updateTargetBlock(sState)
```
LukasGasior1: isn't it already handled in `handleCommonMessages`?

LukasGasior1: also naming seems to be a little misleading, you receive something that sounds like a command to wait: `WaitForDownloadsToFinish` but the action is updating target block. Why not just call this message `UpdateTargetBlock(state)`?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -149,87 +169,158 @@ class FastSync(
...
+        } else {
+          log.info("Asking for new target block")
+          val targetBlockSelector =
+            context.actorOf(FastSyncTargetBlockSelector.props(etcPeerManager, peerEventBus, syncConfig, scheduler))
+          targetBlockSelector ! FastSyncTargetBlockSelector.ChooseTargetBlock
+          context become waitingForTargetBlockUpdate(state)
+        }
+      } else {
+        log.warning(s"Sync failure! Number of targetBlock Failures reached maximum.")
+        sys.exit(1)
```
LukasGasior1: could we start downloading the state for the best block we currently have, instead of exiting?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -149,87 +169,158 @@ class FastSync(
...
+      syncState = syncState.updateNextBlockToValidate(header, K, X)
+    }
+
+    private def processHeader(header: BlockHeader, peer: Peer) : Either[HeaderProcessingResult , (BlockHeader, BigInt)] = for {
+      validatedHeader  <- validateHeader(header, peer)
+      parentDifficulty <- getDifficulty(header)
+    } yield (validatedHeader, parentDifficulty)
+
+    private def getDifficulty(header: BlockHeader) = {
+      blockchain.getTotalDifficultyByHash(header.parentHash).toRight(HeadersProcessingFinished)
```
LukasGasior1: I'm not quite sure I get this. Why does it return `HeadersProcessingFinished` if there's no difficulty for given header?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -343,7 +434,7 @@ class FastSync(
         val evm = account.map(_.codeHash)
         val storage = account.map(_.storageRoot)
 
-        blockchain.saveNode(ByteString(n.hash), n.toBytes, initialSyncState.targetBlock.number)
+        blockchain.saveNode(ByteString(n.hash), n.toBytes, syncState.targetBlock.number)
```
LukasGasior1: hehe this one could lead to tricky errors if you forgot to update it :smile:

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -111,14 +124,21 @@ class FastSync(
...
-        removeRequestHandler(sender())
-        handleBlockHeaders(peer, blockHeaders)
+        log.info("*** Received {} block headers in {} ms ***", blockHeaders.size, timeTaken)
+        requestedHeaders.get(peer).foreach{ requestedNum =>
+          removeRequestHandler(sender())
+          requestedHeaders -= peer
+          if (requestedNum == blockHeaders.size)
+            handleBlockHeaders(peer, blockHeaders)
+          else
+            blacklist(peer.id, blacklistDuration, "wrong number of headers in response")
```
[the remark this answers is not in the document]
KonradStaniec: yep, you are right, according to https://github.com/ethereum/wiki/wiki/Ethereum-Wire-Protocol#new-model-syncing-pv62, the peer can send us less than requested. I will change the conditions to blacklist only when we receive more.

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
KonradStaniec: it is just the way to stop processing headers as soon as we encounter one without a parent. 
In earlier code version it `insertHeader` returned `false` and stopped processing restof headers. 
I will make it more transparent what is happening

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
KonradStaniec: heh, this one has bitten me at the beginning of work at this pr :P

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
KonradStaniec: hmm, we definitely could but I am not sure it would help here, as  we only validate every n-th block, and then every block in interval  (pivot -  X, pivot + X), so to be sure that current best block is safe we would need to discard the last X blocks and then download and validate all blocks to the point from `currentBest - X`  to `currentBest + X.`
This procedure can also fail, so we would be back to declaring a failure at some point in time.

KonradStaniec:
I definitely agree with splitting it up (in separate pr as this one already has grown too much), I will make youtrack task for it.

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -499,24 +591,18 @@ class FastSync(
         val now = Instant.now()
         val peers = unassignedPeers
           .filter(p => peerRequestsTime.get(p).forall(d => d.plusMillis(fastSyncThrottle.toMillis).isBefore(now)))
-        val blockChainPeers = blockChainOnlyPeers.toSet
-        (peers -- blockChainPeers)
-          .take(maxConcurrentRequests - assignedHandlers.size)
-          .toSeq.sortBy(_.ref.toString())
-          .foreach(assignWork)
         peers
```
ntallar: Minor detail: With this change the `blockChainOnlyPeers` was left unused, should we remove it?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -80,15 +89,19 @@ class FastSync(
...
         startWithState(initialSyncState)
       }
   }
 
+  // scalastyle:off number.of.methods
   private class SyncingHandler(initialSyncState: SyncState) {
 
     private val BlockHeadersHandlerName = "block-headers-request-handler"
 
+    private var requestedHeaders: Map[Peer, BigInt] = Map.empty
```
ntallar: Should we remove a peer from this map if a request failed (in the `handleRequestFailure` method)? As this map is checked when assigning work (in `assignBlockchainWork`) it seems it could be the case that headers are no longer requested if the map is not cleared.

KonradStaniec: great catch 👍

PR #400, opened 2018-02-02
update readme

KonradStaniec:


src/test/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManagerSpec.scala
```diff
@@ -44,7 +44,7 @@ class PeerDiscoveryManagerSpec extends FlatSpec with Matchers with MockFactory w
 
     val expectedFindNodeResponse = SendMessage(FindNode(ByteString(nodeStatus.nodeId), expectedTime), remoteUdpAddress)
 
-    Thread.sleep(500)
+    Thread.sleep(1500)
```
AlanVerbner: Why this change was introduced in a `README` update PR?

build.sbt
```diff
@@ -2,7 +2,7 @@ enablePlugins(JDKPackagerPlugin, JavaAppPackaging, SolidityPlugin)
 
 val commonSettings = Seq(
   name := "mantis",
-  version := "1.0-cli-rc1",
+  version := "1.0-cli",
```
AlanVerbner: Do we need to update somewhere in `Daedalus` where Mantis is launched as well?

src/test/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManagerSpec.scala
mcsherrylabs: Is this still necessary?

build.sbt
KonradStaniec: Actually this branch is totally separated from branch integrating with `Daedalus`  so no changes necessary

src/test/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManagerSpec.scala
KonradStaniec: Cli release is a totally different branch from daeadalus release branch. This timeout increase was introduced already in `phase/daeadlus` release and in `phase/release1_1` so to avoid spurious tests fails it made sense to introduce it here.

PR #401, opened 2018-02-02
remove redundant fields from protobuf messages

rtkaczyk:
Tests passed locally (still need to address the CI issue).
TODOs to be addressed in follow-up PR.

### DIFF
--- a/src/main/protobuf/extvm/msg.proto
+++ b/src/main/protobuf/extvm/msg.proto
@@ -1,6 +1,17 @@
 syntax = "proto3";
 package io.iohk.ethereum.extvm;
 
+message Hello {
+  string version = 1;
+
+  // A blockchain related configuration for the VM for the entire session.
+  // Currently only Mantis-EVM is supported. Other VMs may ingore this
+  // or may required different configurations.
+  oneof config {
+    EthereumConfig ethereumConfig = 2;
+  }
+}
+
 message CallContext {
   bytes ownerAddr = 1;
   bytes callerAddr = 2;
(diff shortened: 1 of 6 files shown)


src/main/protobuf/extvm/msg.proto
```diff
@@ -102,20 +88,24 @@ message Blockhash {
...
-  bool error = 4;
-  repeated ModifiedAccount modifiedAccounts = 5;
-  repeated bytes deletedAccounts = 6;
-  repeated bytes touchedAccounts = 7;
-  repeated LogEntry logs = 8;
+  // required for IELE VM
+  bytes returnCode = 2;
+  bytes gasRemaining = 3;
+  bytes gasRefund = 4;
+  bool error = 5;
```
rtkaczyk: We could extract this from `returnCode` but this would require different logic per VM. We only need a simple `true` or `false` answer so retaining this field is probably simpler.

src/main/protobuf/extvm/msg.proto
```diff
@@ -1,6 +1,17 @@
 syntax = "proto3";
 package io.iohk.ethereum.extvm;
 
+message Hello {
+  string version = 1;
+
+  // A blockchain related configuration for the VM for the entire session.
+  // Currently only Mantis-EVM is supported. Other VMs may ingore this
+  // or may required different configurations.
+  oneof config {
```
dwightguth: can we have a separate message containing the oneof that is shared in common between Hello and CallContext? something like:
```
message Hello {
  string version = 1;
  BlockchainConfig config = 2;
}

message CallContext {
  ...
  BlockchainConfig config = 11;
}

message BlockchainConfig {
  oneof config {
    EthereumConfig ethereumConfig = 1;
  }
}
```

There would seem to be a bug in the code that generates ocaml code from a .proto file, and it incorrectly generates the fields of oneof fields as required instead of optional, which means that we cannot support creating an instance of CallContext in OCAML without first creating an EthereumConfig, which is somewhat undesirable for testing purposes.

Another alternative would be to add a second case to both oneofs with a IeleConfig that is just an empty message; this could also work. I'm comfortable with whichever you'd prefer, I just don't want us to have to create an entire EthereumConfig in order to test the IELE side of the connection because of this bug, and trying to get the bug fixed upstream will probably take quite some time.

rtkaczyk: I would go with the empty `IeleConfig` - it fits what I had in mind here, and we may still end up needing it non-empty.

rtkaczyk: @dwightguth The way Scala code generator works for `oneof` messages is that it always creates a class `Empty`. So for this field we have 2 possibilities: it will either be a `Hello.Config.EthereumConfig` or a `Hello.Config.Empty` - which is how we achieve optionality. Are you sure there isn't something like that in OCaml?

dwightguth: I'm sure, yes. The protobuf docs say that oneof is an optional field, so the ocaml behavior is definitely a bug, but unless you want me to work on getting it fixed upstream, we need to work around it.

rtkaczyk: I've added empty `IeleConfig` as another option. 
You said you cannot create a `CallContext`, but does the bug also mean you decode a `CallContext` if sent to you with empty `config` field?

LukasGasior1 (approved):

src/main/protobuf/extvm/msg.proto
dwightguth: I'm not sure what would happen if you tried to decode it, but my guess is it would either decode incorrectly or throw some kind of exception.

PR #405, opened 2018-02-08
[EC-445] Modify storage to work on ByteStrings

LukasGasior1:


src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -235,10 +236,11 @@ class BlockchainImpl(
     }
 
   override def getAccountStorageAt(rootHash: ByteString, position: BigInt): ByteString = {
-    storageMpt(
+    val value = storageMpt(
       rootHash,
       nodesKeyValueStorageFor(None)
-    ).get(UInt256(position)).getOrElse(UInt256(0)).bytes
+    ).get(position).getOrElse(BigInt(0))
+    ByteString(BigIntegers.asUnsignedByteArray(value.bigInteger))
```
rtkaczyk: Hmm, is this different from `ByteString(value.toByteArray)` ?

src/main/scala/io/iohk/ethereum/domain/package.scala
```diff
@@ -2,20 +2,29 @@ package io.iohk.ethereum
...
+    override def toBytes(input: BigInt): Array[Byte] =
+      ByteUtils.padLeft(ByteString(BigIntegers.asUnsignedByteArray(input.bigInteger)), 32).toArray[Byte]
   }
 
-  val rlpUInt256Serializer = new ByteArraySerializable[UInt256] {
-    override def fromBytes(bytes: Array[Byte]): UInt256 = ByteString(bytes).toUInt256
-    override def toBytes(input: UInt256): Array[Byte] = input.toBytes
+  val rlpBigIntSerializer = new ByteArraySerializable[BigInt] {
+    override def fromBytes(bytes: Array[Byte]): BigInt = {
+      BigIntegers.fromUnsignedByteArray(rlp.decode[ByteString](bytes).toArray[Byte])
```
rtkaczyk: We already have an RLP codec for BigInts (RLPImplicits.scala)

src/test/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxySpec.scala
```diff
@@ -8,6 +8,7 @@ import io.iohk.ethereum.vm.{EvmConfig, Generators}
 import org.scalatest.{FlatSpec, Matchers}
 import org.spongycastle.util.encoders.Hex
 
+// scalastyle:off
```
rtkaczyk: Still needed?

src/test/scala/io/iohk/ethereum/vm/CreateOpcodeSpec.scala
```diff
@@ -92,7 +92,7 @@ class CreateOpcodeSpec extends WordSpec with Matchers {
 
         newAccount.balance shouldEqual fxt.endowment
         result.world.getCode(fxt.newAddr) shouldEqual fxt.contractCode.code
-        result.world.getStorage(fxt.newAddr).load(0) shouldEqual 42
+        result.world.getStorage(fxt.newAddr).load(UInt256(0)) shouldEqual UInt256(42).toBigInt
```
rtkaczyk: Looks like an artifact of the previous refactoring: `.load(0) shouldEqual BigInt(42)`

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
LukasGasior1: seems it's not different

src/test/scala/io/iohk/ethereum/ledger/InMemoryWorldStateProxySpec.scala
LukasGasior1: removed

src/main/scala/io/iohk/ethereum/domain/package.scala
```diff
@@ -2,20 +2,29 @@ package io.iohk.ethereum
...
+import io.iohk.ethereum.utils.ByteUtils
+import org.spongycastle.util.BigIntegers
 
 package object domain {
 
-  val byteArrayUInt256Serializer = new ByteArrayEncoder[UInt256] {
-    override def toBytes(input: UInt256): Array[Byte] = input.bytes.toArray[Byte]
+  val byteArrayBigIntSerializer = new ByteArrayEncoder[BigInt] {
+    override def toBytes(input: BigInt): Array[Byte] =
+      ByteUtils.padLeft(ByteString(BigIntegers.asUnsignedByteArray(input.bigInteger)), 32).toArray[Byte]
```
rtkaczyk: I don't think this will work correctly:
```
scala> BigIntegers.fromUnsignedByteArray(BigIntegers.asUnsignedByteArray(BigInt(-1).bigInteger))
res3: java.math.BigInteger = 255
```

Also, why padding?

src/main/scala/io/iohk/ethereum/domain/package.scala
rtkaczyk: ...and it works for nonnegative values only. I see no point, I think it can be changed

LukasGasior1: I've been debugging ETS for the past hour and I was pretty sure using exisitng BigInt encoder was the cause for failures. Let me double check that

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
rtkaczyk: I think it is, compare:

```
scala> BigIntegers.fromUnsignedByteArray(BigIntegers.asUnsignedByteArray(BigInt(-1).bigInteger))
res3: java.math.BigInteger = 255

scala> BigInt(BigInt(-1).toByteArray)
res4: scala.math.BigInt = -1
```

I think the latter method should be used in here and in `RLPImplicits`

src/main/scala/io/iohk/ethereum/domain/package.scala
LukasGasior1: and for that conversion - I believe you only need to convert keys one way, to byte array

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
LukasGasior1: `ByteString(BigIntegers.asUnsignedByteArray(value.bigInteger))` and `ByteString(value.toByteArray)` seem to work the same way, it's the other way conversion that behaves differently

LukasGasior1: `BigIntegers.fromUnsignedByteArray` vs BigInt apply

src/main/scala/io/iohk/ethereum/domain/package.scala
rtkaczyk: > padding is needed for Ethereum support

What exactly do you mean?

LukasGasior1: Ethereum expects serialized keys to be 32-bytes long (in previous implementation see: `input.bytes.toArray[Byte]` - `UInt256.bytes` adds the padding

rtkaczyk: Wait, exactly 32 bytes or at least?

LukasGasior1: For EVM it's exactly 32-bytes (bounded by usage of uint256). For other VM's it can be shorter or longer, BUT, if we want to keep a single implementation, it's the only place we can add this padding for EVM support

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
LukasGasior1: Changed to `ByteString(value.toByteArray)`

src/main/scala/io/iohk/ethereum/domain/package.scala
LukasGasior1: OK, it seems the failures were because of missing padding in key serializer. I've changed it back to use existing BigInt encoder here

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
LukasGasior1: > I think the latter method should be used in here and in RLPImplicits
we can't just use BigInt apply in RLPImplicits, we need to decode the bytes from rlp format

LukasGasior1:
Added `ethCompatibleStorage` to `BlockchainConfig`. TODO: unit tests for non-Ethereum storage.

rtkaczyk:
LGTM, pending conflict resolution

LukasGasior1:
@rtkaczyk merged

PR #406, opened 2018-02-15
Fix performance bottle neck in storages

KonradStaniec:
1. Replace builidng list of nodes to delete in pruning
from append to seq to prepend to list.
2. During serialization of nodes use the fact that
value is already array[Byte] and key is bytestring

As for potential improvements, during sync from master locally:

#### Syncing to 200k blocks
Baseline - 55min
Improved - 45min

#### 30min of syncing from some later block
Baseline:
```
22:06:15 [i.i.e.blockchain.sync.RegularSync] - Block: 756826. Peers: 1 (0 blacklisted)
22:16:20 [i.i.e.blockchain.sync.RegularSync] - Block: 762330. Peers: 1 (0 blacklisted)
10min - 5504
22:36:17 [i.i.e.blockchain.sync.RegularSync] - Block: 776538. Peers: 1 (0 blacklisted)
30 min - 19712 blocks
```

Improved:
```
11:30:18 [i.i.e.blockchain.sync.RegularSync] - Block: 756826. Peers: 1 (0 blacklisted)
11:40:19 [i.i.e.blockchain.sync.RegularSync] - Block: 762714. Peers: 1 (0 blacklisted)
10min - 5888
12:00:19 [i.i.e.blockchain.sync.RegularSync] - Block: 778266. Peers: 1 (0blacklisted)
30min - 21440
```
Diffrence is 1729 blocks in same unit of time, which is ~9% improvement

### DIFF
--- a/src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
+++ b/src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
@@ -24,6 +24,17 @@ trait DataSource {
     */
   def get(namespace: Namespace, key: Key): Option[Value]
 
+
+  /**
+    * This function obtains the associated value to a key, if there exists one. It assumes that
+    * caller already properly serialized key. Useful when caller knows some pattern in data to
+    * avoid generic serialization.
+    *
+    * @param key
+    * @return the value associated with the passed key.
+    */
+  def getOptimized(key: Array[Byte]): Option[Array[Byte]]
+
   /**
     * This function updates the DataSource by deleting, updating and inserting new (key-value) pairs.
     *
@@ -34,6 +45,19 @@ trait DataSource {
     * @return the new DataSource after the removals and insertions were done.
     */
   def update(namespace: Namespace, toRemove: Seq[Key], toUpsert: Seq[(Key, Value)]): DataSource
+
+
+  /**
+    * This function updates the DataSource by deleting, updating and inserting new (key-value) pairs.
+    * It assumes that caller already properly serialized key and value.
+    * Useful when caller knows some pattern in data to avoid generic serialization.
+    *
+    * @param toRemove which includes all the keys to be removed from the DataSource.
+    * @param toUpsert which includes all the (key-value) pairs to be inserted into the DataSource.
+    *                 If a key is already in the DataSource its value will be updated.
+    * @return the new DataSource after the removals and insertions were done.
+    */
+  def updateOptimized(toRemove: Seq[Array[Byte]], toUpsert: Seq[(Array[Byte], Array[Byte])]): DataSource
 
   /**
     * This function updates the DataSource by deleting all the (key-value) pairs in it.
--- a/src/main/scala/io/iohk/ethereum/db/storage/NodeStorage.scala
+++ b/src/main/scala/io/iohk/ethereum/db/storage/NodeStorage.scala
@@ -9,12 +9,33 @@ import io.iohk.ethereum.db.storage.NodeStorage._
   *   Key: hash of the RLP encoded node
   *   Value: the RLP encoded node
   */
-class NodeStorage(val dataSource: DataSource) extends KeyValueStorage[NodeHash, NodeEncoded, NodeStorage] {
+class NodeStorage(val dataSource: DataSource) {
 
   val namespace: IndexedSeq[Byte] = Namespaces.NodeNamespace
-  def keySerializer: NodeHash => IndexedSeq[Byte] = _.toIndexedSeq
-  def valueSerializer: NodeEncoded => IndexedSeq[Byte] = _.toIndexedSeq
-  def valueDeserializer: IndexedSeq[Byte] => NodeEncoded = _.toArray
+  private val specialNameSpace = namespace.head
+
+  def specialSerializer(nodeHash: NodeHash): Array[Byte] = {
+    (specialNameSpace +: nodeHash).toArray
+  }
+
+  def get(key: NodeHash): Option[NodeEncoded] = dataSource.getOptimized(specialSerializer(key))
+
+  /**
+    * This function updates the KeyValueStorage by deleting, updating and inserting new (key-value) pairs
+    * in the current namespace.
+    *
+    * @param toRemove which includes all the keys to be removed from the KeyValueStorage.
+    * @param toUpsert which includes all the (key-value) pairs to be inserted into the KeyValueStorage.
+    *                 If a key is already in the DataSource its value will be updated.
+    * @return the new KeyValueStorage after the removals and insertions were done.
+    */
+  def update(toRemove: Seq[NodeHash], toUpsert: Seq[(NodeHash, NodeEncoded)]): NodeStorage = {
+    val newDataSource = dataSource.updateOptimized(
+      toRemove = toRemove.map(specialSerializer),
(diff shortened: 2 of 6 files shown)


src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -166,15 +166,21 @@ object ReferenceCountNodeStorage extends PruneSupport with Logger {
...
+    snapshotKeys.foreach {hash =>
+      for {
+        snapshot <- nodeStorage.get(hash).map(snapshotFromBytes)
         node <- nodeStorage.get(snapshot.nodeKey).map(storedNodeFromBytes)
         if node.references == 0 && node.lastUsedByBlock <= blockNumber
-      } yield snapshot.nodeKey
-      nodesToRemove :+ toRemove
-    }.flatten
+      } yield {
+        nodesToRemove = snapshot.nodeKey :: nodesToRemove
```
AlanVerbner: 👍

src/main/scala/io/iohk/ethereum/db/storage/NodeStorage.scala
```diff
@@ -9,12 +9,33 @@ import io.iohk.ethereum.db.storage.NodeStorage._
...
-class NodeStorage(val dataSource: DataSource) extends KeyValueStorage[NodeHash, NodeEncoded, NodeStorage] {
+class NodeStorage(val dataSource: DataSource) {
 
   val namespace: IndexedSeq[Byte] = Namespaces.NodeNamespace
-  def keySerializer: NodeHash => IndexedSeq[Byte] = _.toIndexedSeq
-  def valueSerializer: NodeEncoded => IndexedSeq[Byte] = _.toIndexedSeq
-  def valueDeserializer: IndexedSeq[Byte] => NodeEncoded = _.toArray
+  private val specialNameSpace = namespace.head
+
+  def specialSerializer(nodeHash: NodeHash): Array[Byte] = {
```
AlanVerbner: I'm not sure if I understand the naming here. Why is it _special_ ?

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
```diff
@@ -24,6 +24,17 @@ trait DataSource {
...
+
+  /**
+    * This function obtains the associated value to a key, if there exists one. It assumes that
+    * caller already properly serialized key. Useful when caller knows some pattern in data to
+    * avoid generic serialization.
+    *
+    * @param key
+    * @return the value associated with the passed key.
+    */
+  def getOptimized(key: Array[Byte]): Option[Array[Byte]]
```
AlanVerbner: I'm not quite sure I follow where the improvement was made here :thinking:

src/main/scala/io/iohk/ethereum/db/storage/NodeStorage.scala
KonradStaniec: hmm naming probably could be better :) but it is special in that sense that it differs from all our aother serializes which converts `K` to `IndexedSeq[Byte]`. This one just prepend namespace byte to bytestring and gives us Array[Byte], shortcuting this our normal serialization `K -> IndexedSeq[Byte]-> Array[Byte]`

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
KonradStaniec: Normal data source `get` has as an argument `IndexedSeq[Byte]`, and prepends `namespace` byte to it and converts it `Array[Byte]`, also it returns `IndexedSeq[Byte]` which need further deserialization. This one assumes that all of this was already done and takes `Array[Byte]`. Also it returns `Array[Byte]` to

Out node storage is most busy of all storages and this constant conversions just started showing on profiling.
This is especially true for value, which serializtion route is `Array[Byte] -> IndexedSeq[Byte] -> Array[Byte]`

AlanVerbner: What do you think about moving that logic to the `DataSource` as a `get` with a different signature?

KonradStaniec: it is defined in `DataSource`, I named it differently hoping that it will be little explanation why we have two get with different signatures. You think just renaming it would be more clear?

AlanVerbner: I think of _namespaces_ as some kind of column families and it's related to the DataSource. That's why i'm not quite sure it's ok to bypass that at storage level.

KonradStaniec: I agree that it is not pretty and breaks our boundaries between `storages` and `dataSource` that why I named methods in data source differently and added docs to them.

 `    * This function obtains the associated value to a key, if there exists one. It assumes that
    * caller already properly serialized key. Useful when caller knows some pattern in data to
    * avoid generic serialization.`

Maybe some caution in those docs about this namespace and boundry breaking, would be helpful? i mean in other storages this change wont make much sense as they not used as much as `nodeStorage` especially when

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -166,15 +166,21 @@ object ReferenceCountNodeStorage extends PruneSupport with Logger {
     * @param nodeStorage
     * @return
     */
-  private def getNodesToBeRemovedInPruning(blockNumber: BigInt, snapshotKeys: Seq[NodeHash], nodeStorage: NodeStorage): Seq[NodeHash] =
-    snapshotKeys.foldLeft(Seq.empty[Option[NodeHash]]) { (nodesToRemove, snapshotKey) =>
-      val toRemove = for {
-        snapshot <- nodeStorage.get(snapshotKey).map(snapshotFromBytes)
+  private def getNodesToBeRemovedInPruning(blockNumber: BigInt, snapshotKeys: Seq[NodeHash], nodeStorage: NodeStorage): Seq[NodeHash] = {
```
ntallar: Do you think maybe a similar improvement in [`update`](https://github.com/input-output-hk/mantis/blob/phase/release1_1/src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala#L52) also could improve the performance?

KonradStaniec: probably yes, but these are nodes from one transaction (there is relativly few of them) so i never saw it bottlenecked in that place, so for now i would leave it at leat until further works of improving pruning performance

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
AlanVerbner: What about this? To avoid turning this into an endless discussion, lets move forward as it it and let's add a task in youtrack to do a second analysis round about this and maybe do a refactor if necessary.

KonradStaniec: I have created task in yt to not get it lost

PR #407, opened 2018-02-16
[EC-466] Pluggable consensus

loverdos:
This PR includes the initial design for pluggable consensus and everything that has been demonstrated in demo0, demo1 milestones. 

A subsequent PR will include the updated pluggable consensus design.

All workstreams have been squashed to one commit. For the historical development, please have a look at one of 
* [`consensus_as_a_service_EC466_historical`](https://github.com/input-output-hk/mantis/tree/feature/consensus_as_a_service_EC466_historical) in the main repo.
* [`consensus_as_a_service_EC466_historical`](https://github.com/loverdos/mantis-fork/tree/feature/consensus_as_a_service_EC466_historical) in my fork.

### DIFF
--- a/src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
+++ b/src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
@@ -1,15 +1,16 @@
 package io.iohk.ethereum.ets.blockchain
 
 import io.iohk.ethereum.blockchain.sync.EphemBlockchainTestSetup
+import io.iohk.ethereum.consensus.{ConsensusBuilder, ConsensusConfigBuilder}
 import io.iohk.ethereum.domain.Block.BlockDec
 import io.iohk.ethereum.domain.{Account, Address, Block, UInt256}
 import io.iohk.ethereum.ets.common.AccountState
 import io.iohk.ethereum.ledger.Ledger.VMImpl
 import io.iohk.ethereum.ledger._
 import io.iohk.ethereum.network.p2p.messages.PV62.BlockBody
-import io.iohk.ethereum.nodebuilder.{ActorSystemBuilder, BlockchainConfigBuilder, SyncConfigBuilder, ValidatorsBuilder}
+import io.iohk.ethereum.nodebuilder.{ActorSystemBuilder, BlockchainConfigBuilder, ShutdownHookBuilder, SyncConfigBuilder, ValidatorsBuilder}
 import io.iohk.ethereum.utils.BigIntExtensionMethods._
-import io.iohk.ethereum.utils.BlockchainConfig
+import io.iohk.ethereum.utils.{BlockchainConfig, Logger}
 import org.spongycastle.util.encoders.Hex
 
 import scala.util.{Failure, Success, Try}
@@ -19,7 +20,11 @@ abstract class ScenarioSetup(vm: VMImpl, scenario: BlockchainScenario)
   with ValidatorsBuilder
   with SyncConfigBuilder
   with BlockchainConfigBuilder
-  with ActorSystemBuilder {
+  with ActorSystemBuilder
+  with ConsensusBuilder
+  with ConsensusConfigBuilder
+  with ShutdownHookBuilder
+  with Logger {
 
   val emptyWorld = blockchain.getWorldStateProxy(-1, UInt256.Zero, None)
 
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -198,26 +198,49 @@ mantis {
     transaction-timeout = 2.minutes
   }
 
-  mining {
-    # Maximum number of ommers kept in the pool
-    ommers-pool-size = 30
+  consensus {
+    # FIXME this was moved here from the `mining` section
+    active-timeout = 5 seconds
 
+    # New timeout to replace `ommer-pool-query-timeout` for uses outside ethash mining
+    get-transaction-from-pool-timeout = 5.seconds
+
+    # Moved from the `mining` section
+    # Miner's coinbase address
+    coinbase = "0011223344556677889900112233445566778899"
+
+    # Moved from the `mining` section
+    # Extra data to add to mined blocks
+    header-extra-data = "mantis"
+
+    #####################
+    # Moved from the `mining` section because this is needed by BlockGenerator, which we use in other consensus
+    # protocols as well.
+    #####################
     # This determines how many parallel eth_getWork request we can handle, by storing the prepared blocks in a cache,
     # until a corresponding eth_submitWork request is received
     block-cashe-size = 30
 
-    # Miner's coinbase address
-    coinbase = "0011223344556677889900112233445566778899"
+    # See io.iohk.ethereum.consensus.Protocol for the available protocols.
+    # Declaring the protocol here means that a more protocol-specific configuration
+    # is pulled from the corresponding consensus implementation.
+    # For example, in case of ethash, a section named `mining` is used.
+    protocol = ethash
 
-    active-timeout = 5 seconds
+    # If true then the consensus protocol uses this node for mining.
+    # In the case of ethash PoW, this means mining new blocks, as specified by Ethereum.
+    # In the general case, the semantics are due to the specific consensus implementation.
+    mining-enabled = false
+  }
+
+  # This is the section dedicated to Ethash mining.
+  # The consensus protocol is selected by setting `mantis.consensus.protocol = ethash`
+  # but for historical reasons we keep the name `mining` here (instead of `ethash`).
+  mining {
+    # Maximum number of ommers kept in the pool
+    ommers-pool-size = 30
 
     ommer-pool-query-timeout = 5.seconds
-
-    # Extra data to add to mined blocks
-    header-extra-data = "mantis"
-
-    # If true this node will mine new blocks
-    mining-enabled = false
 
     ethash-dir = ${user.home}"/.ethash"
 
--- a/src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
@@ -1,0 +1,41 @@
+package io.iohk.ethereum.consensus
+
+import io.iohk.ethereum.domain.Block
+import io.iohk.ethereum.ledger.BlockExecutionError.ValidationBeforeExecError
+import io.iohk.ethereum.ledger.BlockExecutionSuccess
+import io.iohk.ethereum.nodebuilder.Node
+import io.iohk.ethereum.validators.BlockHeaderValidator
+
+/**
+ * Abstraction for a consensus protocol.
+ *
+ * @see [[io.iohk.ethereum.consensus.Protocol Protocol]]
+ */
+// FIXME Lot's of stuff to do...
+trait Consensus {
+  type Config <: AnyRef /*Product*/
+
+  def protocol: Protocol
+
+  def config: FullConsensusConfig[Config]
+
+  /** Returns `true` if this is the standard Ethereum PoW consensus protocol (`ethash`). */
+  final def isEthash: Boolean = protocol.isEthash
+
+  /**
+   * Starts the consensus protocol on the current `node`.
+   */
+  def startProtocol(node: Node): Unit
+
+  def stopProtocol(): Unit
+
+  /**
+   * Provides the [[io.iohk.ethereum.validators.BlockHeaderValidator BlockHeaderValidator]] that is specific
+   * to this consensus protocol.
+   */
+  // FIXME Probably include the whole of [[io.iohk.ethereum.validators.Validators]].
+  def blockHeaderValidator: BlockHeaderValidator
+
+  // Ledger uses this in importBlock
+  def validateBlockBeforeExecution(block: Block): Either[ValidationBeforeExecError, BlockExecutionSuccess] = ???
+}
--- a/src/main/scala/io/iohk/ethereum/consensus/ConsensusConfig.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/ConsensusConfig.scala
@@ -1,0 +1,104 @@
+package io.iohk.ethereum.consensus
+
+import akka.util.ByteString
+import com.typesafe.config.{Config ⇒ TypesafeConfig}
+import io.iohk.ethereum.domain.Address
+import io.iohk.ethereum.nodebuilder.ShutdownHookBuilder
+import io.iohk.ethereum.utils.Logger
+import io.iohk.ethereum.validators.BlockHeaderValidatorImpl
+
+import scala.concurrent.duration.{FiniteDuration, _}
+
+/**
+ * Provides generic consensus configuration. Each consensus protocol implementation
+ * will use its own specific configuration as well.
+ *
+ * @param protocol Designates the consensus protocol.
+ * @param activeTimeout
+ * @param miningEnabled Provides support for generalized "mining". The exact semantics are up to the
+ *                      specific consensus protocol implementation.
+ */
+final case class ConsensusConfig(
+  protocol: Protocol,
+
+  coinbase: Address,
+
+  // NOTE Moved from [[io.iohk.ethereum.consensus.ethash.MiningConfig MiningConfig]]
+  activeTimeout: FiniteDuration,
+
+  // NOTE Moved from [[io.iohk.ethereum.consensus.ethash.MiningConfig MiningConfig]]
+  headerExtraData: ByteString, // only used in BlockGenerator
+
+  // NOTE Moved from [[io.iohk.ethereum.consensus.ethash.MiningConfig MiningConfig]]
+  blockCacheSize: Int, // only used in BlockGenerator
+
+  getTransactionFromPoolTimeout: FiniteDuration,
+
+  miningEnabled: Boolean
+)
+
+object ConsensusConfig extends Logger {
+  object Keys {
+    final val Consensus = "consensus"
+    final val Protocol = "protocol"
+    final val Coinbase = "coinbase"
+    final val ActiveTimeout = "active-timeout"
+    final val HeaderExtraData = "header-extra-data"
+    final val BlockCacheSize = "block-cashe-size"
+    final val MiningEnabled = "mining-enabled"
+    final val GetTransactionFromPoolTimeout = "get-transaction-from-pool-timeout"
+  }
+
+
+  final val AllowedProtocols = Set(
+    Protocol.Names.Ethash,
+    Protocol.Names.Demo0,
+    Protocol.Names.AtomixRaft
+  )
+
+  final val AllowedProtocolsError = (s: String) ⇒ Keys.Consensus +
+    " was '" + s + "'" +
+    " but it should be one of " +
+    AllowedProtocols.map("'" + _ + "'").mkString(",")
+
+  private def readProtocol(consensusConfig: TypesafeConfig): Protocol = {
+    val protocol = consensusConfig.getString(Keys.Protocol)
+
+    // If the consensus protocol is not a known one, then it is a fatal error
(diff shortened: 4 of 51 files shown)


src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
```diff
@@ -19,7 +20,11 @@ abstract class ScenarioSetup(vm: VMImpl, scenario: BlockchainScenario)
   with ValidatorsBuilder
   with SyncConfigBuilder
   with BlockchainConfigBuilder
-  with ActorSystemBuilder {
+  with ActorSystemBuilder
+  with ConsensusBuilder
+  with ConsensusConfigBuilder
+  with ShutdownHookBuilder
+  with Logger {
```
rtkaczyk: Why were all of these added?

src/main/resources/application.conf
```diff
@@ -198,26 +198,49 @@ mantis {
...
+    # FIXME this was moved here from the `mining` section
+    active-timeout = 5 seconds
 
-    # This determines how many parallel eth_getWork request we can handle, by storing the prepared blocks in a cache,
-    # until a corresponding eth_submitWork request is received
-    block-cashe-size = 30
+    # New timeout to replace `ommer-pool-query-timeout` for uses outside ethash mining
+    get-transaction-from-pool-timeout = 5.seconds
 
+    # Moved from the `mining` section
```
rtkaczyk: Those comments serve as documentation, not a changelog. Do you plan on keeping the comments with historical info or are they just temporary?

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
```diff
@@ -0,0 +1,41 @@
...
+// FIXME Lot's of stuff to do...
+trait Consensus {
+  type Config <: AnyRef /*Product*/
+
+  def protocol: Protocol
+
+  def config: FullConsensusConfig[Config]
+
+  /** Returns `true` if this is the standard Ethereum PoW consensus protocol (`ethash`). */
+  final def isEthash: Boolean = protocol.isEthash
```
rtkaczyk: I didn't find any usages for it, which is good because I don't think such methods belong in object oriented design.

src/main/scala/io/iohk/ethereum/consensus/FullConsensusConfig.scala
```diff
@@ -0,0 +1,31 @@
...
+  )
+
+  /**
+   * There are APIs that expect that the standard Ethash consensus is running and so depend
+   * on either its configuration or general PoW semantics.
+   * This is a method that can handle such cases via a respective if/then/else construct:
+   * if we run under [[io.iohk.ethereum.consensus.Ethash Ethash]] the `_if` function is called,
+   * otherwise the `_else` value is computed.
+   */
+  final def ifEthash[A](_if: MiningConfig ⇒ A)(_else: ⇒ A): A =
```
rtkaczyk: Why not just implement the method `isEthash` (`is` not `if`) in `EthService`?

src/main/scala/io/iohk/ethereum/consensus/FullConsensusConfig.scala
```diff
@@ -0,0 +1,31 @@
...
+import io.iohk.ethereum.consensus.ethash.MiningConfig
+
+case class FullConsensusConfig[C <: AnyRef /*Product*/](
+  generic: ConsensusConfig,
+  specific: C
+) {
+  // Sanity check
+  require(
+    generic.protocol.isEthash == ifEthash(_ ⇒ true)(false),
+    "generic.protocol.isEthash == ifEthash(_ ⇒ true)(false)"
```
rtkaczyk: It took me a while to understand what this check does. It would be helpful if the message was something like "the specific protocol settings must match the generic ones".
As noted in other comments however, I'm not so fond if `isEthash` and `ifEthash` being implemented the consensus classes themselves.

src/main/scala/io/iohk/ethereum/consensus/ConsensusConfig.scala
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Provides generic consensus configuration. Each consensus protocol implementation
+ * will use its own specific configuration as well.
+ *
+ * @param protocol Designates the consensus protocol.
+ * @param activeTimeout
+ * @param miningEnabled Provides support for generalized "mining". The exact semantics are up to the
+ *                      specific consensus protocol implementation.
+ */
+final case class ConsensusConfig(
```
rtkaczyk: No change required, just a question: at some point we considered adding `final` to all our case classes, but decided it's not worth the hassle. Is there any other argument for this than `one shouldn't extend case classes`? What's your general position on this?

src/main/resources/application.conf
loverdos: Just temporary, when the final places are decided these comments will be removed. We do not even have a `mining` section anymore, it is called `ethash` now, following up on a comment by @LukasGasior1 in #395

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
loverdos: Here is the story of `isEthash` and `ifEthash`. Some code in EthService would need to run conditionally only if we are running under Ethash consensus. Also there would be a sanity check in a test case that Mantis includes at least Ethash as consensus. Thus the introduction of `isEthash`. Of course, `ifEthash` subsumes this functionality and provides an even more general one, whether more exact details are provided as a parameter. This means we can retire `isEthash`. And regarding `ifEthash`, essentially a polymorphic `fold` over ethash and non-ethash consensus, I was thinking of moving it to a utility class (an `implicit class extends AnyVal`), so as the Consensus interface is clean of such functionality. I suggest we revisit these design decisions in the next PR.

src/main/scala/io/iohk/ethereum/consensus/FullConsensusConfig.scala
loverdos: I need to revisit this whole requirement anyway, so thanks for the comment; the message indeed could be more useful in another form.

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
loverdos: Let me recheck this. In the next PR, I have a new trait that just mixes in everything you need to get a `consensus` so as to not write all these `with` every time but I am not sure it applies here.

src/main/scala/io/iohk/ethereum/consensus/FullConsensusConfig.scala
LukasGasior1: :+1: If this check (ethash vs non-ethash) is just for RPC it should live there.

src/main/scala/io/iohk/ethereum/consensus/atomixraft/Miner.scala
```diff
@@ -0,0 +1,162 @@
...
+import scala.concurrent.duration._
+import scala.util.{Failure, Success}
+
+class Miner(
+  blockchain: Blockchain,
+  blockGenerator: BlockGenerator,
+  pendingTransactionsManager: ActorRef,
+  syncController: ActorRef,
+  config: AtomixRaftConfig,
+  ethService: EthService,
```
LukasGasior1: it doesn't seem to be needed here

src/main/scala/io/iohk/ethereum/consensus/demo/Demo.scala
```diff
@@ -0,0 +1,188 @@
...
+import akka.util.ByteString
+import io.iohk.ethereum.crypto.{kec256, kec512}
+import io.iohk.ethereum.utils.ByteUtils._
+import org.spongycastle.util.BigIntegers
+import org.spongycastle.util.encoders.Hex
+import java.lang.Integer.remainderUnsigned
+
+import scala.annotation.tailrec
+
+private[demo] object Demo {
```
LukasGasior1: can we get rid of it yet?

LukasGasior1 (approved):
I've added some comments, not any blockers though.

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
loverdos: It's complicated. `Validators` are now (as in the next PR) the responsibility of `Consensus`, so the source of truth for them is consensus, they are not a standalone entity. They do not actually use any other consensus functionality but they are **part** of it. In fact `ValidatorsBuilder` [reduces to](https://github.com/input-output-hk/mantis/blob/4b4c76d5cffe4f28ef696bff4ec9fb2e3fbc3405/src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala#L352..L356):

```scala
trait ValidatorsBuilder {
  self: ConsensusBuilder =>

  lazy val validators = consensus.validators
}
```

Note: the way we stack and mix traits, and use them especially in the tests, is a source of complexity that makes it very difficult to cope with all these changes. A tiny refactor somewhere (e.g. in `Ledger`) looks like an amplification attack to the whole codebase :)

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
rtkaczyk: What do you propose we do to remember we need to revisit? A `FIXME` comment, or a YT ticket?

src/main/scala/io/iohk/ethereum/consensus/ethash/Miner.scala
```diff
@@ -24,14 +25,14 @@ import scala.util.{Failure, Random, Success, Try}
...
-    ethService: EthService)
-  extends Actor with ActorLogging {
+  blockchain: Blockchain,
+  blockGenerator: BlockGenerator,
+  ommersPool: ActorRef,
+  pendingTransactionsManager: ActorRef,
+  syncController: ActorRef,
+  config: FullConsensusConfig[MiningConfig],
+  ethService: EthService
+) extends Actor with ActorLogging {
```
rtkaczyk: Not a big deal, but the [style guide](https://docs.scala-lang.org/style/declarations.html) recommends params to be indented with 4 spaces.

loverdos: :smile:  That style guide is inconsistent (4 vs 2 spaces) and not even Lightbend uses it, e.g. check [akka](https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/routing/Balancing.scala#L68..L72). I usually indent with 2 spaces, and Twitter does [the same](http://twitter.github.io/effectivescala/#Formatting-Whitespace).

loverdos: But of course, I am happy to adhere to any convention we have, the only reason I followed these patterns is out of habbit and because that I was not aware of any local conventions.

src/main/scala/io/iohk/ethereum/consensus/FullConsensusConfig.scala
loverdos: This check is currently implemented via a reflective pattern match that locally reveals the actual config implementation. I would like to keep it near where config implementations are (definition site) rather than near the use site.

In an alternative implementation strategy, e.g. using polymorphism so that only the ethash-related implementation knows how to use the `_if` branch (because it's the only one with a `MiningConfig` instance), the above remark is strengthened even more.

Also `isEthash` alone is not enough, we need the exact `MiningConfig` instance. My [other comment](https://github.com/input-output-hk/mantis/pull/407#discussion_r171520270) is also relevant.

src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftConfig.scala
```diff
@@ -0,0 +1,71 @@
+package io.iohk.ethereum.consensus
+package atomixraft
```
rtkaczyk: WDYT about moving this to `io.iohk.ethereum.utils`? I'm not saying it's best, but it would be consistent.

Also, the way we did `*Config` classes was just about holding the bare values: strings, numbers, durations. This holds object like `AtomixNode` or `File` and has logic to parse them.

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
loverdos: I am looking at it almost every day while working on the next PR, so it will not be forgotten. Most probably the resolution is going to look like this: Since we absolutely need this polymorphic decision making the `ifEthash` is going to be implemented somewhere, and `isEthash` is not fundamental: it is a utility easily implemented on top of `ifEthash`. Most probably, I will get `ifEthash` out of `Consensus` as an extention method (via the `implicit class extends Any` mechanism).

src/main/scala/io/iohk/ethereum/consensus/atomixraft/Miner.scala
```diff
@@ -0,0 +1,162 @@
...
+import io.iohk.ethereum.nodebuilder.Node
+import io.iohk.ethereum.transactions.PendingTransactionsManager
+import io.iohk.ethereum.transactions.PendingTransactionsManager.PendingTransactionsResponse
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import scala.util.{Failure, Success}
+
+class Miner(
```
rtkaczyk: The word "miner" is specific to PoW. I believe "forger" is the most prevalent in PoS and everything else that doesn't depend on _work_.

src/main/scala/io/iohk/ethereum/consensus/atomixraft/Miner.scala
loverdos: Indeed it is [gone](https://github.com/input-output-hk/mantis/blob/feature/consensus_as_a_service_EC458/src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftMiner.scala#L21..L26) now.

src/main/scala/io/iohk/ethereum/consensus/ethash/Miner.scala
rtkaczyk: I know it's not widely adhered to, but where did you find inconsistency within it?

Anyway it's just a proposal. Feel free to disregard it. I personally like the 4 space rule - it makes it clearer where parameters end. I'm always trying to promote conforming to this guide, but our codebase is already inconsistent in this and other regards.

src/main/resources/application.conf
```diff
@@ -198,26 +198,49 @@ mantis {
...
+
+    # See io.iohk.ethereum.consensus.Protocol for the available protocols.
+    # Declaring the protocol here means that a more protocol-specific configuration
+    # is pulled from the corresponding consensus implementation.
+    # For example, in case of ethash, a section named `mining` is used.
+    protocol = ethash
+
+    # If true then the consensus protocol uses this node for mining.
+    # In the case of ethash PoW, this means mining new blocks, as specified by Ethereum.
+    # In the general case, the semantics are due to the specific consensus implementation.
```
rtkaczyk: In the ultimate version I'd like to see something like: 

```
consensus {
  protocol = xyz

  xyz {
    mining-enabled = false
  }
}
```

Even if it means repeating some settings in the protocol specific sections.

src/main/resources/application.conf
```diff
@@ -198,26 +198,49 @@ mantis {
...
     mining-enabled = false
+  }
```
rtkaczyk: Even if they are not used, protocol specific settings (Raft) should be added here, because this file serves also as a developer's reference - so we know what settings are possible. This can be added in the next iteration though.

src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftConsensus.scala
```diff
@@ -0,0 +1,144 @@
...
+        .withDirectory(dataDir)
+        .build
+
+      val server = RaftServer.builder(localNode.id)
+        .withThreadModel(ThreadModel.SHARED_THREAD_POOL)
+        .withClusterService(clusterService)
+        .withProtocol(protocol)
+        .withStorage(raftStorage)
+        /*.withElectionTimeout(Duration.ofMillis(3000))
+        .withHeartbeatInterval(Duration.ofMillis(300))*/
```
rtkaczyk: Remove or add comment

src/main/scala/io/iohk/ethereum/consensus/atomixraft/Miner.scala
```diff
@@ -0,0 +1,162 @@
...
+      log.info("***** Mined block " + block.header.number)
+      syncController ! RegularSync.MinedBlock(block)
+      self ! StartMining
+    }
+    else {
+      lostLeadership()
+    }
+  }
+
+  //noinspection ScalaStyle
```
rtkaczyk: Is it the same as `scalastyle:off`. Why is it needed?

src/main/scala/io/iohk/ethereum/consensus/atomixraft/MiningConfig.scala
```diff
@@ -0,0 +1,5 @@
+package io.iohk.ethereum.consensus.atomixraft
+
+trait MiningConfig {
```
rtkaczyk: Can we remove this?

src/main/scala/io/iohk/ethereum/consensus/ethash/validators/BlockHeaderValidatorImpl.scala
```diff
@@ -0,0 +1,82 @@
...
+import akka.util.ByteString
+import io.iohk.ethereum.crypto
+import io.iohk.ethereum.domain.BlockHeader
+import io.iohk.ethereum.utils.BlockchainConfig
+import io.iohk.ethereum.validators.BlockHeaderError.HeaderPoWError
+import io.iohk.ethereum.validators.BlockHeaderValidatorImpl.PowCacheData
+import io.iohk.ethereum.validators.{BlockHeaderError, BlockHeaderValid, BlockHeaderValidator}
+
+// NOTE Copied parts from [[io.iohk.ethereum.validators.BlockHeaderValidatorImpl]]
+class BlockHeaderValidatorImpl(blockchainConfig: BlockchainConfig) extends BlockHeaderValidator {
```
rtkaczyk: The validators will be refactored in the next iteration, right? Will we still need 2 complementary implementations of `BlockHeaderValidator`? (I understand we can have even more, each specific to a given protocol, I'm not sure though about having 2 different ones for a single protocol)

src/test/scala/io/iohk/ethereum/validators/BlockHeaderValidatorSpec.scala
```diff
@@ -144,201 +147,201 @@ class BlockHeaderValidatorSpec extends FlatSpec with Matchers with PropertyCheck
...
-  }
+  }*/
 
-  it should "validate correctly a block whose parent is in storage" in new EphemBlockchainTestSetup {
-    blockchain.save(validBlockParent)
-    blockHeaderValidator.validate(validBlockHeader, blockchain) match {
-      case Right(_)  => succeed
-      case _ => fail
-    }
+it should "validate correctly a block whose parent is in storage" in new EphemBlockchainTestSetup {
```
rtkaczyk: Previous indentation was correct - starting here, and continuing through the remainder of this file

rtkaczyk (commented):
LGTM. All my comments are either of minor significance or things that could be addressed in the next iterations. Approval pending merge/rebase.

loverdos:
* [x] Rebase on `phase/iele_testnet`
* [x] All checks pass

We are good to go @rtkaczyk

rtkaczyk (approved):

PR #408, opened 2018-02-16
[EC-447] redefine VM entry point

rtkaczyk:
The VM has been refactored so that it better reflects _Θ_ and _Λ_ functions, as defined in the Yellow Paper. Consequently VM now handles value transfer message call and account creation for contract creation.

TODO:
- [x]  unit tests for new functionality moved to `VM` class (`ContractCreationSpec`)


src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainTestConfig.scala
```diff
@@ -182,7 +183,11 @@ class HomesteadToDaoAt5 extends BlockchainTestConfig {
 }
 
 class Eip158Config extends BlockchainTestConfig {
+  override val frontierBlockNumber = -1
```
LukasGasior1: why -1 and not just 0?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -673,82 +673,56 @@ abstract class CreateOp extends OpCode(0xf0, 3, 1, _.G_create) {
...
-
-      val maxCodeSizeExceeded = state.config.maxCodeSize.exists(codeSizeLimit => contractCode.size > codeSizeLimit)
-      val enoughGasForDeposit = totalGasRequired <= startGas
-
-      val creationFailed = maxCodeSizeExceeded || result.error.isDefined ||
-        !enoughGasForDeposit && state.config.exceptionalFailedCodeDeposit
-
-      if (creationFailed) {
+    //FIXME: to avoid calculating this twice, we could adjust state.gas prior to execution in OpCode#execute
+    //not sure how this would affect other opcodes [EC-243]
```
LukasGasior1: do we still need this comment?

src/main/scala/io/iohk/ethereum/vm/ProgramState.scala
```diff
@@ -4,56 +4,63 @@ import akka.util.ByteString
...
-  def apply[W <: WorldStateProxy[W, S], S <: Storage[S]](context: ProgramContext[W, S]): ProgramState[W, S] =
+  def apply[W <: WorldStateProxy[W, S], S <: Storage[S]](
+      vm: VM[W, S],
+      context: ProgramContext[W, S],
+      env: ExecEnv): ProgramState[W, S] = {
+
     ProgramState(
-      context = context,
-      gas = context.startGas,
+      vm = vm,
```
LukasGasior1: why do we need vm in ProgramState?

src/main/scala/io/iohk/ethereum/vm/VM.scala
```diff
@@ -1,36 +1,88 @@
...
+      import context._
+      import org.spongycastle.util.encoders.Hex
+      log.trace(s"caller:  $callerAddr | recipient: $recipientAddr | gasPrice: $gasPrice | value: $value | inputData: ${Hex.toHexString(inputData.toArray)}")
     }
+
+    if (context.recipientAddr.isEmpty)
+      create(context)._1
+
+    else
+      call(context, context.recipientAddr.get)
```
LukasGasior1: why not just match on recipient addr?

src/main/scala/io/iohk/ethereum/vm/VM.scala
```diff
@@ -1,36 +1,88 @@
...
-        finalState.returnData,
-        finalState.gas,
-        finalState.world,
-        finalState.addressesToDelete,
-        finalState.logs,
-        finalState.internalTxs,
-        finalState.gasRefund,
-        finalState.error)
+  def run(context: ProgramContext[W, S]): ProgramResult[W, S] = {
+    {
```
LukasGasior1: still needed?

src/main/scala/io/iohk/ethereum/vm/VM.scala
```diff
@@ -1,36 +1,88 @@
...
+        val initialState: PS = ProgramState(this, context1, env)
+        exec(initialState).toResult
+      }
+    }
+
+  /**
+    * Contract creation - Λ function in YP
+    */
+  private[vm] def create(context: PC): (PR, Address) =
+  if (!isValidCall(context))
```
LukasGasior1: formatting

src/main/scala/io/iohk/ethereum/vm/ProgramState.scala
rtkaczyk: Because `VM` is now type-parameterised (it's just to avoid type parameterising every single method in it).

src/main/scala/io/iohk/ethereum/vm/VM.scala
rtkaczyk: doh 👍.. it's a leftover from something I tried earlier

src/main/scala/io/iohk/ethereum/vm/VM.scala
rtkaczyk: Well, doesn't hurt. It's on `trace`

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
rtkaczyk: Well, it still applies, but I doubt we'll fix it anytime soon

src/main/scala/io/iohk/ethereum/vm/VM.scala
LukasGasior1: I meant just the curly brace

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainTestConfig.scala
rtkaczyk: I guess they could be `0`, but then we'd either have to bump the relevant fork numbers to `1` or modify `EvmConfig.forBlock` method (which doesn't handle ties very well). But I'd also argue that `-1` is more explicitly stating that a given fork is not occurring in the tested chain.

src/ets/scala/io/iohk/ethereum/ets/vm/VMSuite.scala
```diff
@@ -13,6 +13,15 @@ import org.scalatest._
...
 class VMSuite extends FreeSpec with Matchers with Logger {
 
+  val vm = new TestVM {
+//    //TODO: instead of this and saving empty caller account in ScenarioBuilder, consider changing context.endowment to
+//    //TODO: option and don't make transfer in VM#call if it's None
+//    override def isValidCall(context: PC): Boolean =
+//      context.callDepth == 0 ||
+//      context.endowment <= context.world.getBalance(context.callerAddr) &&
+//      context.callDepth <= EvmConfig.MaxCallDepth
+  }
```
rtkaczyk: Forgot to remove this

src/main/scala/io/iohk/ethereum/vm/VM.scala
rtkaczyk: Ahh, I added the curly brace because of the `import context._` to avoid polluting the remaining scope. Especially now with the match statement (`case Some(recipientAddr)`) there would be shadowing, which is something I try to avoid.

src/main/scala/io/iohk/ethereum/nodebuilder/VmSetup.scala
```diff
@@ -3,31 +3,62 @@ package io.iohk.ethereum.nodebuilder
...
+
+  /**
+    * Runs a standard VM binary that takes $port and $host as input arguments
+    */
+  private def startStandardVmProcess(externalConfig: ExternalConfig): Unit = {
+    import externalConfig._
+    require(executablePath.isDefined, s"VM type '$vmType' requires the path to binary to be provided")
+    // TODO: we also need host parameter in iele node
+    new ProcessBuilder(executablePath.get, port.toString, host)
+      .redirectOutput(Redirect.INHERIT)
```
dwightguth: I'm reasonably confident this should also call redirectError.

PR #409, opened 2018-02-20
Add possibility to configure VM

LukasGasior1:


rtkaczyk:
Code looks good but I'm not sure committing the binary. I was thinking we would include path to the executable in the config.

Also path could be optional, and if not provided we don't run it. That would be useful if:
- the VM is running on another machine
- the executable requires some input arguments that we don't support yet

WDYT?

src/main/resources/application.conf
```diff
@@ -455,10 +455,19 @@ mantis {
     filter-manager-query-timeout = 3.minutes
   }
 
-  extvm {
-    host = "127.0.0.1"
-    port = 8888
+  vm {
+    # local | external
+    mode = "local"
```
rtkaczyk: Maybe `s/local/internal/g` as "local" could be interpreted as running on the same machine?

LukasGasior1:
I was going to say that doing this executable config doesn't make much sense because of input args (port for IELE or logger conf for Mantis, for example) but it reminded me of something else: we should probably also include Mantis VM binary in dist. And if we do that the same way we ship our main binary, we wouldn't have to provide logger conf as a parameter anymore. Then, if we standarize VM input args, say "./binary host port", it actually makes sense to use that exec path (it would assume standard args format, and if you need something else you can just run it yourself and not rely on Mantis config).

TLDR: I'll add Mantis VM binary to dist and executable path config

LukasGasior1:
@rtkaczyk I've pushed the changes

rtkaczyk:
Just a note: I don't think we need to require from the VMs that input arguments follow any given standard. We could easily accept other VMs, by creating different process builders as was done so far. Of course that requires code changes for any new non-standard VM to be supported, so this is good for now 👍

PR #410, opened 2018-02-21
[CGKIELE-53] Pluggable Consensus II

loverdos:
This PR continues the work for Pluggable Consensus in Mantis, providing the necessary architectural elements and modifying the core of Mantis in order to accomodate them. It can be seen as the conclusion of the work started in the [previous PR](https://github.com/input-output-hk/mantis/pull/407). 

You can find more technical information in the accompanying [Design Doc for Pluggable Consensus](https://github.com/input-output-hk/mantis/wiki/Pluggable-Consensus).

It all starts with [`Consensus`](https://github.com/input-output-hk/mantis/blob/feature/consensus_as_a_service_EC458/src/main/scala/io/iohk/ethereum/consensus/Consensus.scala).


loverdos:
It seems that an empty commit is necessary in order to avoid those OOM errors during compilation.

loverdos:
Cherry picked @rtkaczyk's fix for the ETS.

src/main/resources/application.conf
```diff
@@ -247,6 +242,32 @@ mantis {
...
+
+    # All the other nodes in the cluster, in the same format as with local-node
+    bootstrap-nodes = [
+      "ID2:IP:PORT",
+      "ID3:IP:PORT",
+      "ID4:IP:PORT"
+    ]
+
+    # Transient data directory
+    data-dir = "atomix-raft-data"
```
LukasGasior1: how about `${mantis.datadir}/"atomix-raft-data"`?

src/main/resources/application.conf
```diff
@@ -247,6 +242,32 @@ mantis {
     mine-rounds = 100000
   }
 
+  # This is the section dedicated to mining in the context of `atomix-raft` consensus.
```
LukasGasior1: It's not only about mining :)

src/main/resources/application.conf
```diff
@@ -199,32 +199,28 @@ mantis {
...
   consensus {
-    # FIXME this was moved here from the `mining` section
     active-timeout = 5 seconds
 
     # New timeout to replace `ommer-pool-query-timeout` for uses outside ethash mining
     get-transaction-from-pool-timeout = 5.seconds
 
-    # Moved from the `mining` section
     # Miner's coinbase address
+    # Also used in non-Ethash consensus.
```
LukasGasior1: I can imagine consensus algorithms that do not require it. What do you think about putting it in consensus-specific config? Like `ethash.coinbase` or `atomix-raft.coinbase`

src/main/resources/application.conf
```diff
@@ -199,32 +199,28 @@ mantis {
   }
 
   consensus {
-    # FIXME this was moved here from the `mining` section
     active-timeout = 5 seconds
```
LukasGasior1: in the current form the meaning of this config is really confusing: `consensus.active-timeout` - sounds like there's a timeout for consensus or something. As it's only used for hashrate reporting in one of RPC apis, maybe `rpc` config is a good place to put it.

src/main/resources/application.conf
```diff
@@ -199,32 +199,28 @@ mantis {
...
 
     # New timeout to replace `ommer-pool-query-timeout` for uses outside ethash mining
     get-transaction-from-pool-timeout = 5.seconds
```
LukasGasior1: same as above, this has nothing to do with consensus. How about moving it to `txPool`?

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
```diff
@@ -1,41 +1,187 @@
...
+   * to this consensus protocol implementation.
+   */
   type Config <: AnyRef /*Product*/
 
   def protocol: Protocol
 
   def config: FullConsensusConfig[Config]
 
-  /** Returns `true` if this is the standard Ethereum PoW consensus protocol (`ethash`). */
-  final def isEthash: Boolean = protocol.isEthash
```
LukasGasior1: :+1:

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
```diff
@@ -1,41 +1,187 @@
...
+  /**
+   * This is the VM used to prepare and generate blocks.
```
LukasGasior1: VM does not prepare nor generate blocks :)

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
```diff
@@ -1,41 +1,187 @@
...
   def startProtocol(node: Node): Unit
 
+  /**
+   * Stops the consensus protocol on the current node.
+   * This is called internally when the node terminates.
+   */
   def stopProtocol(): Unit
 
+  // Note Ledger uses this in importBlock
+  def validateBlockBeforeExecution(
```
LukasGasior1: maybe we can include this method in `Validators`?

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
```diff
@@ -1,41 +1,187 @@
...
+    vm = vm,
+    signedTxValidator = validators.signedTransactionValidator,
+    blockchain = blockchain,
+    blockchainConfig = blockchainConfig
+  )
+
+  /**
+   * This is used by the [[io.iohk.ethereum.consensus.Consensus#blockGenerator blockGenerator]].
+   */
+  def blockPreparator: BlockPreparator = this._blockPreparator
```
LukasGasior1: why the extra val? can't it just be `override val blockPreparator = new BlockPreparetor(..)`?

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
```diff
@@ -1,41 +1,187 @@
...
+
+  /**
+   * This is the VM used to prepare and generate blocks.
+   */
+  def vm: VMImpl = theVm
```
LukasGasior1: again, can't we just make `override val vm: VMImpl,` on class parameter list?

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
```diff
@@ -1,41 +1,187 @@
...
+
+  def config: FullConsensusConfig[C] = fullConsensusConfig
```
LukasGasior1: same here

src/main/scala/io/iohk/ethereum/consensus/Protocol.scala
```diff
@@ -13,49 +13,38 @@ sealed trait Protocol {
   def name: String
 
   /** Returns `true` if this is the standard Ethereum PoW consensus protocol (`ethash`). */
-  final def isEthash: Boolean =
-    this match {
-      case Ethash ⇒ true
-      case _ ⇒ false
-    }
+  def isEthash: Boolean
```
LukasGasior1: :(

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGenerator.scala
```diff
@@ -0,0 +1,62 @@
...
+ *      [[io.iohk.ethereum.ledger.BlockPreparator BlockPreparator]]
+ *
+ */
+trait BlockGenerator {
+  /**
+   * The type of consensus-specific data used in the block generation process.
+   * For example, under [[io.iohk.ethereum.consensus.ethash.EthashConsensus EthashConsensus]],
+   * this represents the [[io.iohk.ethereum.network.p2p.messages.PV62.BlockBody#uncleNodesList ommers]].
+   */
+  type X
```
LukasGasior1: I don't quite like this idea. Can't we do something like this instead:
`def generateBlock[Data <: BlockGenerationData](..., data: Data)`
and for ethash, for example: `case class EthashBlockGenerationData(ommers: Ommers): extends BlockGenerationData` or `RaftBlockGenerationData(raft specific stuff for generating blocks)`

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGenerator.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  /** An empty `X` */
+  def emptyX: X
+
+  // FIXME This is currently used only for Ethash
```
LukasGasior1: why don't we put it in ethash specific generator then?

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorImpl.scala
```diff
@@ -0,0 +1,184 @@
...
+      .scanLeft(BigInt(0), None: Option[SignedTransaction]) { case ((accumulatedGas, _), stx) => (accumulatedGas + stx.tx.gasLimit, Some(stx)) }
+      .collect { case (gas, Some(stx)) => (gas, stx) }
+      .takeWhile { case (gas, _) => gas <= blockGasLimit }
+      .map { case (_, stx) => stx }
+
+    transactionsForBlock
+  }
+
+  //returns maximal limit to be able to include as many transactions as possible
+  protected def calculateGasLimit(parentGas: BigInt): BigInt = {
```
LukasGasior1: I think it's ethash-specific, isn't it?

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
rtkaczyk: It is _used_ in the process

src/main/resources/application.conf
rtkaczyk: 🤔 I agree, but then again if a consensus protocol doesn't specify block rewards then this setting is just ignored. So I think it's fine to leave it here.

src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftMiner.scala
```diff
@@ -18,26 +18,26 @@ import scala.concurrent.Future
 import scala.concurrent.duration._
 import scala.util.{Failure, Success}
 
-class Miner(
+class AtomixRaftMiner(
```
rtkaczyk: As noted on the [previous PR](https://github.com/input-output-hk/mantis/pull/407#discussion_r171842862) _mining_ is strictly related to PoW, whereas _forging_ (_forge_, _forger_) is a more general term. Can we replace the references to _mining_ within the context of Raft? If, through abstraction, _forging_ leaks to Ethash, that's fine too.

rtkaczyk:
Looks like we're missing changes in `universal` config files. The new configuration may deserve a separate file: `consensus.conf`

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorImpl.scala
```diff
@@ -0,0 +1,184 @@
...
+import scala.collection.immutable
+
+/**
+ * This is a skeleton for a generic [[io.iohk.ethereum.consensus.blocks.BlockGenerator BlockGenerator]].
+ *
+ * @param blockchain
+ * @param blockchainConfig
+ * @param _blockTimestampProvider
+ */
+abstract class BlockGeneratorImpl(
```
LukasGasior1: It seems this class is a collection of utility functions used by specific block generators. I don't like two things about it:
1. the name, it implies it's a standalone implementation, while it isn't
2. I can smell Java-style in it :smile: how about something along these lines:
```
trait BlockGeneratorUtils {
   def blockchain: Blockchain
   ...
   utilityFunctions()
}

class EthashBlockGenerator extends ... with BlockGeneratorUtils
```

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGenerator.scala
```diff
@@ -0,0 +1,62 @@
...
+  def getPrepared(powHeaderHash: ByteString): Option[PendingBlock]
+
+  /**
+   * This function returns the block currently being mined block with highest timestamp
+   */
+  def getPendingBlock: Option[PendingBlock]
+
+  def getPendingBlockAndState: Option[PendingBlockAndState]
+
+  def blockTimestampProvider: BlockTimestampProvider
```
LukasGasior1: why does it need to be a part of BlockGenerator interface? It's not used anywhere other than specific implementations

src/main/scala/io/iohk/ethereum/consensus/ConsensusBuilder.scala
```diff
@@ -1,58 +1,65 @@
...
-    val consensus = new AtomixRaftConsensus(blockchainConfig, fullConfig)
+  protected def buildAtomixRaftConsensus(): atomixraft.AtomixRaftConsensus = {
+    val specificConfig = atomixraft.AtomixRaftConfig(mantisConfig)
+    val fullConfig = newConfig(specificConfig)
+    val consensus = AtomixRaftConsensus(vm, blockchain, blockchainConfig, fullConfig)
     consensus
   }
 
-  private def loadConsensus(): Consensus = {
+  protected def buildConsensus(): TestConsensus = {
```
LukasGasior1: I have mixed feelings about `TestConsensus` leaking to main app interfaces. What do you think about using `Consensus` here and casting it to `TestConsensus` in tests?

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
```diff
@@ -1,41 +1,187 @@
...
+  /**
+   * Returns the [[io.iohk.ethereum.consensus.blocks.BlockGenerator BlockGenerator]]
+   * this consensus protocol uses.
+   */
+  def blockGenerator: BlockGenerator
 
   /**
    * Starts the consensus protocol on the current `node`.
    */
   def startProtocol(node: Node): Unit
```
LukasGasior1: I noticed specific implementations miss `override` keyword for these methods

src/main/scala/io/iohk/ethereum/consensus/Protocol.scala
LukasGasior1: this is not used anywhere, is it? as far as I see ifEthash uses a match now

src/main/scala/io/iohk/ethereum/consensus/ethash/validators/EthashValidators.scala
```diff
@@ -0,0 +1,8 @@
+package io.iohk.ethereum.consensus
+package ethash.validators
+
+import io.iohk.ethereum.consensus.validators.Validators
```
LukasGasior1: we still have `BlockHeaderValidator` in `io.iohk.ethereum.validators` which doesn't seem to be needed anymore. Can we delete it (the whole package)?

src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
```diff
@@ -0,0 +1,307 @@
...
+ *
+ * With its introduction, we:
+ *
+ *   1. Avoid a direct dependency of [[io.iohk.ethereum.consensus.Consensus Consensus]] on
+ *      [[io.iohk.ethereum.ledger.Ledger Ledger]].
+ *   2. Extract a substantial chunk of functionality outside [[io.iohk.ethereum.ledger.Ledger Ledger]],
+ *      in an attempt to modularize it.
+ *
+ */
+class BlockPreparator(
```
rtkaczyk: I was wondering what's the difference between block preparation and block generation. I can see `BlockPreparator` handles block execution so why not `BlockExecutor`? 

But from the doc I understood that it was meant to take that responsibility from `Ledger`. In the `Ledger` however, methods like `executeTransactions` still exist, so I'm a bit confused. Why the duplication?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -446,15 +447,17 @@ class EthService(
     }
   }
 
+  //noinspection ScalaStyle
```
rtkaczyk: Is it the same as scalastyle:off. Why is it needed?

build.sbt
```diff
@@ -4,7 +4,7 @@ val commonSettings = Seq(
   name := "mantis",
   version := "1.0-daedalus-rc1",
   scalaVersion := "2.12.1",
-  testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-l", "MinerSpec") // miner tests disabled by default
+  testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-l", "EthashMinerSpec") // miner tests disabled by default
```
rtkaczyk: I remember, Christos, you had a problem a with loading the DAG on CircleCI. Was it because this test was not disabled after renaming? I wonder what's the point of keeping it if it's never run...

src/main/resources/application.conf
loverdos: This was in the `mining` section, so according to the original idea it seems to belong to the current `ethash` section. I must have moved it from `mining` to the generic `consensus` and forgot to re-place it under `ethash` while refactoring. 

Given the above, can you both confirm that a better place indeed is `rpc`?

src/main/resources/application.conf
loverdos: Why do you say so? It is a timeout specified and needed by a client of the pool, not the pool itself. It is currently used in both the ethereum miner and the raft "miner". So it seems to be precisely about consensus, a desing that pre-existed, I did not introduce it.

Also, in the pre-pluggable-consensus setup, the ethereum `Miner` used `ommer-pool-query-timeout` for the same reason inside the `mining` section (so in a consensus-related section). I just introduced another timeout for better reflecting the intent, since `ommer-pool-query-timeout` was properly used elsewhere (but again in the `Miner`).

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
loverdos: Yes, I think it is the right thing to do. I started moving this stuff outside consensus but there was too much to refactor for this round so I eventually [gave up](https://github.com/input-output-hk/mantis/pull/410/commits/d8bf2f32ec411dff0eaede837ced5f5c99ac6ecd#diff-8f88229ce5e1d25dc74496902accebacL13-L19).
How about moving both this method and also `validateBlockAfterExecution` in `Validators`?

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
loverdos: I'm thinking about changing the wording to `VM used while preparing and generating blocks` in order to reduce confusion.

src/main/scala/io/iohk/ethereum/consensus/Protocol.scala
loverdos: [Gone](https://github.com/input-output-hk/mantis/pull/410/commits/ee65603bd983cc41373b32c4442cdf9b1bed4968)!

src/main/resources/application.conf
LukasGasior1: I'd still go for `rpc`

src/main/resources/application.conf
LukasGasior1: To be honest it's really just a boilerplate needed because we use the ask pattern. I, personally, would just make it a constant in code. My point was, it has nothing to do with consensus because it's strictly an implementation detail.

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGenerator.scala
loverdos: Why don't you like it? 
We naturally express the idea that the shape of "ommers" (or whatever concept applies) depends on the `BlockGenerator`, so its type must be a property of it.

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGenerator.scala
loverdos: Too much refactoring :) 
I can prepare a patch and even save us from one more [`FIXME`](https://github.com/input-output-hk/mantis/blob/feature/consensus_as_a_service_EC458/src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala#L500-L513).

The good news is that all the affected tests already expect the *correct* `BlockGenerator` subclass, that is `EthashBlockGenerator`. For example:

* [in EthServiceSpec](https://github.com/input-output-hk/mantis/blob/feature/consensus_as_a_service_EC458/src/test/scala/io/iohk/ethereum/jsonrpc/EthServiceSpec.scala#L396) we use this [type](https://github.com/input-output-hk/mantis/blob/feature/consensus_as_a_service_EC458/src/test/scala/io/iohk/ethereum/jsonrpc/EthServiceSpec.scala#L860).

src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftMiner.scala
loverdos: Yes, as I have explained, "miner" and "mining" is used very loosely/liberally in our non-PoW use-cases. This was in the absense of a better name and since naming is hard I did not want to spend more cycles on it :)

Happy to discuss and find a more appropriate terminology though. Forging sounds cool. Is it used anywhere else or are we the first ones to do it?

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGenerator.scala
loverdos: Indeed. I [moved it](https://github.com/input-output-hk/mantis/pull/410/commits/e218b514624e7bf256a40776e837e1c971146dd9) in `TestBlockGenerator`. 

Note: although technically not used from a `TestBlockGenerator` either, we do have a method there that provides a new `BlockTimestampProvider` that is used from a test. So it is OK, since we introduce the concept of this provider there.

src/main/scala/io/iohk/ethereum/consensus/ConsensusBuilder.scala
loverdos: Leaking is not good and this is a compromise trying to a void the cast. But I think there is a better way: provide a `TestConsensusBuilder` that has the correct subtype and which is only used in the tests. I'll try it and if it does not complicate the codebase, I'll commit a patch. Does that sound OK?

src/main/scala/io/iohk/ethereum/consensus/ethash/validators/EthashValidators.scala
loverdos: [Yes](https://github.com/input-output-hk/mantis/pull/410/commits/c9d0062c3f0c25451c8fb9776b88e6c9015d83a7)!

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
loverdos: I guess the `scalastyle:off` at the top only apply to the two specific styles mentioned but I do not remeber why I introcuced this, so it is [gone now](https://github.com/input-output-hk/mantis/pull/410/commits/63ad75053be3d4047e966b071915f0bb445f4279)!

build.sbt
loverdos: Indeed that is the case. Should have changed it here as well but forgot to.

You make a good point but I think it is for discussion in another venue ;)

src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftMiner.scala
rtkaczyk: Some examples:
https://cardanodocs.com/glossary/#follow-the-satoshi
https://medium.com/@karthik.seshu/cryptocurrency-proof-of-work-vs-proof-of-stake-e1eee1420b10
http://www.cryptomorrow.com/2017/10/23/staking-wallet/
(just text-search for "forge")

Now, I don't think this term has been officially dubbed, but I've seen it used many times in the context. Another word that appears is _minting_, but I think it's less common.

loverdos: I thought about minting as well but yes, it does not seem to be widely used.

On second thought, `forge` may lead to [negative connotations](https://www.merriam-webster.com/dictionary/forger), so I am not sure it is a good idea anymore.

rtkaczyk: I think that's a secondary meaning, so I wouldn't worry about that. But it might be best to ask a native speaker, @mcsherrylabs ?

src/main/resources/application.conf
loverdos: @rtkaczyk ?

src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
loverdos: I named it `BlockPreparator` because of its `prepareBlock` method. This is a dependency that `Consensus` would have on `Ledger`, and the one I wanted to avoid. So I started by taking this out of `Ledger` and so I created  a class named `BlockPreparator` because it had this important `prepareBlock` method. The remaining functionality that you see in `BlockPreparator` is the transitive dependencies needed in order to fully implement `prepareBlock`. These were taken from `Ledger` as well. But I did not remove the respective methods from `Ledger`, I only delegated to `BlockPreparator`. The reason for that is because these methods were already used in tests and I did not want to be carried away to a more massive refactoring.

HTH

src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftMiner.scala
mcsherrylabs: Wow. Much intricacy. `Minting` almost works, but is associated specifically with coin. I like `forge` and for a native speaker it's very clear which meaning is being used. Of course not all readers will be native speakers. It's a good point that `AtomixRaftForger` or just seeing `forger` in the code might cause the reader to pause and scratch their head. :thinking: ..... not going to overthink this. I think `forge` is the best candidate. The root cause of the confusion is the English language and that's out of scope.

build.sbt
mcsherrylabs: (@loverdos  You can bring that up on Tuesday?)

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGenerator.scala
loverdos: OK, I did prepare the [patch](https://github.com/input-output-hk/mantis/pull/410/commits/375064acb3dc4d89441dde359941dea1f88bb87d).

src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftMiner.scala
loverdos: `RaftForger` [it is](https://github.com/input-output-hk/mantis/pull/410/commits/8c5433f2145e75a6c95ac9c9f7199bd67f6b7aac)!

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorImpl.scala
loverdos: As the scaladoc clarifies, `This is a skeleton for a ... BlockGenerator`.

src/main/scala/io/iohk/ethereum/consensus/ConsensusBuilder.scala
loverdos: OK, [implemented](https://github.com/input-output-hk/mantis/pull/410/commits/adcdccb373deb14533176fcd5cde327cdc112aea).

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
loverdos: If we were missing `override` we would not go past the compiler.

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
loverdos: [Changed](https://github.com/input-output-hk/mantis/pull/410/commits/3a855d3dd91e8d728692544a09d46fdf45a7d85b)

src/main/resources/application.conf
rtkaczyk: Yes, `rpc`

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
loverdos: [Moved](https://github.com/input-output-hk/mantis/pull/410/commits/f07dabba8eca7eba850f1caf1035788b1b952111)

src/main/resources/application.conf
loverdos: [Moved](https://github.com/input-output-hk/mantis/pull/410/commits/3282958dbbdf135454ce0805bb69aed710e6afc1).

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
LukasGasior1: Well, it compiles without it, but I usually consider it a good habit to still add the modifier to make it immediately visible which methods are overriden

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorImpl.scala
LukasGasior1: Ok, lets leave it. I may be biased here as I usually prefer "composition-style inheritance" (with traits) rather than a abstract base implementation classes.

src/main/resources/application.conf
loverdos: [Moved](https://github.com/input-output-hk/mantis/pull/410/commits/410386a3cd1878595e4ed95ce245d0a30c18a482) `get-transaction-from-pool-timeout` to `txPool`.

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorImpl.scala
LukasGasior1: True, it's used there. It just seems like a leaky abstraction - we have _generic_ skeleton for BlockGenerator, which contains ethash (subclass) specific stuff.

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
loverdos: There is no extra `val`, just one: `_blockPreparator`.

LukasGasior1: And that's the one I'm referring to :)
Why not:
```
override val blockPreparator: BlockPreparator = new BlockPreparator(
    vm = vm,
    signedTxValidator = validators.signedTransactionValidator,
    blockchain = blockchain,
    blockchainConfig = blockchainConfig
  )
```
instead of
```
  protected val _blockPreparator = new BlockPreparator(
    vm = vm,
    signedTxValidator = validators.signedTransactionValidator,
    blockchain = blockchain,
    blockchainConfig = blockchainConfig
  )
  def blockPreparator: BlockPreparator = this._blockPreparator
```

LukasGasior1: Ok, nvm, I see why :) This class is meant to be inherited, we want to keep it a def on this level.

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorImpl.scala
rtkaczyk: https://en.wikipedia.org/wiki/Composition_over_inheritance
(I wouldn't call mixing traits "composition")

src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
rtkaczyk: I see, maybe we should a task to refactor/clean up those tests. For example the method `LedgerImpl#executePreparedTransactions` is only used in tests, so the avoidance of refactoring the tests is it's only reason to exist.

As for the name, it seems that `BlockPreparator` is currently used for 2 things: block preparation and block execution (of already prepared blocks, like those received from peers). Because preparation includes and depends on execution, I would say it's more appropriate to name it `BlockExecutor` or `TxExecutor`. It's more OK if `*Executor` has an extra functionality that is preparation, rather `Preparator` be used mainly for something other than preparation. Note that `LedgerImp#prepareBlock` is also only used in tests, so it probably  should be removed, and the tests refactored.

rtkaczyk: I mean, no code changes required here, but tell me what do you think about refactoring the tests as a follow-up. Is it worth the effort?

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
rtkaczyk: We should have a style convention for things like that. If you ask me, I tend to avoid `override` if it's not needed cause it is an eyesore. The points made above are valid though.

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGenerator.scala
```diff
@@ -0,0 +1,58 @@
...
+   * This function returns the block currently being mined block with highest timestamp
+   */
+  def getPendingBlock: Option[PendingBlock]
+
+  def getPendingBlockAndState: Option[PendingBlockAndState]
+
+  /**
+   * Generates the next block.
+   *
+   * @param ommers We call it `ommers`, since this is the original use-case in Ethash but in general it can be anything
```
rtkaczyk: Maybe we should call it `x` and skip this comment? 😉 
To me the explanation on the type declaration is sufficient.
(Btw. I'm okay with it being a path-dependent type)

src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
loverdos: I have no doubt we must attempt to refactor the tests in a a post-PR follow-up action.

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorImpl.scala
rtkaczyk: I would agree with Lukasz about the name, though. `*Impl` is a widely recognised convention and may be confusing here. How about `BlockGeneratorSkeleton` or `BlockGeneratorPartialImpl`?

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorImpl.scala
```diff
@@ -0,0 +1,174 @@
...
+import io.iohk.ethereum.domain._
+import io.iohk.ethereum.ledger.Ledger.{BlockPreparationResult, BlockResult}
+import io.iohk.ethereum.ledger.{BlockPreparator, BloomFilter}
+import io.iohk.ethereum.mpt.{ByteArraySerializable, MerklePatriciaTrie}
+import io.iohk.ethereum.network.p2p.messages.PV62.BlockBody
+import io.iohk.ethereum.network.p2p.messages.PV62.BlockHeaderImplicits._
+import io.iohk.ethereum.utils.BlockchainConfig
+import io.iohk.ethereum.utils.ByteUtils.or
+
+import scala.collection.immutable
```
rtkaczyk: Is this on purpose? It's not used consistently here.

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorImpl.scala
loverdos: My personal preference for many years has been `*Skeleton` for such partial implementations, so I am wondering why I did not follow it here in the first place. I think I've lost touch with the `*Impl` convention that you mention, to me it sounds like _implementation detail_ in general, whether partial or not. I tend to prepend `Std` for _standard_ implementations.

So, I can rename it to `BlockGeneratorSkeleton`.

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorImpl.scala
```diff
@@ -0,0 +1,174 @@
...
+      BlockHeader(
+        parentHash = parent.header.hash,
+        ommersHash = ByteString(kec256(ommers.toBytes: Array[Byte])),
+        beneficiary = beneficiary.bytes,
+        stateRoot = ByteString.empty,
+        //we are not able to calculate transactionsRoot here because we do not know if they will fail
+        transactionsRoot = ByteString.empty,
+        receiptsRoot = ByteString.empty,
+        logsBloom = ByteString.empty,
+        difficulty = difficulty.calculateDifficulty(blockNumber, blockTimestamp, parent.header),
```
rtkaczyk: I don't think difficulty makes any sense outside of PoW context. Definitely not for now, but maybe we should follow up on this and avoid this calculation for RAFT?

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
rtkaczyk: Well, I don't get it. `ConsensusImpl` already inherits those defs from `Consensus`, what's the point of implementing them here rather than:

```
class AtomixRaftConsensus private(
  val vm: VMImpl,
  val blockchain: BlockchainImpl,
  val blockchainConfig: BlockchainConfig,
  val fullConsensusConfig: FullConsensusConfig[AtomixRaftConfig],
  val validators: Validators,
  val blockGenerator: AtomixRaftBlockGenerator
) extends ConsensusImpl[AtomixRaftConfig] ...
```

Also, as Lukasz noted above, why the pair:

```
protected val _blockPreparator = ...
def blockPreparator: BlockPreparator = ...
```

instead of one:

```
val blockPreparator: BlockPreparator = ...
```

These are not big issues, but it seems they unnecessarily increase the volume of code. Is there a purpose for these constructions that I'm missing?

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorImpl.scala
rtkaczyk: I'm not sure if `gasLimit` is ethash- or Pow-specific. At any rate you might argue than _in general_ consensuses have a gas limit, and specific one may choose to ignore it (in practice setting it to a huge number)

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorImpl.scala
loverdos: [Renamed](https://github.com/input-output-hk/mantis/pull/410/commits/f84265952159ab3aac2df5ded097234f31a4dc1a)

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorImpl.scala
loverdos: The original

```scala
  private def prepareTransactions(transactions: Seq[SignedTransaction], blockGasLimit: BigInt) = { ...}
```
lacked a return type and when I had the IDE fill it in it was `List[SignedTransaction]`, so I guess that was how `immutable.Seq` appeared in the scene: Not too specific (a `List`) but definetly reflecting that fact that we are dealing with an `immutable.Seq`.

src/main/resources/application.conf
loverdos: [Here](https://github.com/input-output-hk/mantis/pull/410/commits/fdbbf31ddc732603a60d91c16424b191b2bb950a)

rtkaczyk:
🎉

PR #411, opened 2018-02-22
[EC-469] Caching Nodes

KonradStaniec:
Pr implements caching layer for nodes. 

Current persising route for nodes goes as: `ReferenceCountinStorage/ArchieStorage -> NodeStorge -> DB`, where `ReferenceCountingStorage/ArchiveStorage` is responsible for pruning, and `NodeStorage` for serialization and talking to DB.
This pr changes  adds to this route `CachedNodeStorage` which is responsible for talking with cache:
`ReferenceCountinStorage/ArchiveStorage -> CachedNodeStorage -> NodeStorge -> DB`

Final update:
Cache is a simple mutable map, with a synchronized update and get methods. It checks if its ready for cleanup after each saved blocks. Cache has 2 persist conditions `maxsize` which is currently set to `400000`nodes which equal roughly to `250mb`, and `maxtime` which ensures that we lose at most `maxtime` of work when quit (currently set to 5min).
Of course, to this change have a maximal effect we need to reduce kept history to do as much as possible work in memory. For now, i settled to 64.

These parameters need some tuning,  for now, most of persists happens because of the time limit.

Also, I still need to add docs, but I want to wait for pr comments to see what will be main questions and doubts.

Performance gains on 1 to 1000000 regular sync from a master node ( data for pruning turned on): 
`Baseline  - 10h25min`
`Improved - 6h43min`


loverdos:
I am not sure of the actual requirements for this cache. Would something like [caffeine](https://github.com/ben-manes/caffeine) be appropriate here, instead of rolling our own?

KonradStaniec:
I actually explored this possibility but with https://github.com/blemale/scaffeine ( it is more pleasant to work with) and to be truthful I haven't used even half of the features it has and need to implement my own like flush all data to disk when full ( maybe I missed something in the api), so I as simple as it gets.

As for requirements:
- Atomic `update` in respect to `get`.
- clear and persist all cache data to disk when full or not persisted long enough to not lose a lot of work

Basically, it works as follows:
After each transaction, we save all nodes to cache.
After each block, there is pruning action which clear junk nodes (their reference count is equal to 0) from the cache.
After Pruning, Check if it is the time to save it to disk.
If it time to persist, persist and save new best block number

So for each block, we add some nodes and deletes some junk nodes, and cache grows only with useful data which we persist when there is a good moment for it.  
The main speedup comes from the fact that pruning now takes all data from memory, not from disk.

loverdos:
@KonradStaniec Thanks for the detailed explanation!

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -175,10 +175,10 @@ class FastSync(
     private def discardLastBlocks(startBlock: BigInt, blocksToDiscard: Int): Unit = {
       (startBlock to ((startBlock - blocksToDiscard) max 1) by -1).foreach { n =>
         blockchain.getBlockHeaderByNumber(n).foreach { headerToRemove =>
-          blockchain.removeBlock(headerToRemove.hash, saveParentAsBestBlock = false)
+          blockchain.removeBlock(headerToRemove.hash, withState = false)
         }
       }
-      blockchain.saveBestBlockNumber((startBlock - blocksToDiscard - 1) max 0)
+      appStateStorage.putBestBlockNumber((startBlock - blocksToDiscard - 1) max 0)
```
AlanVerbner: I'm not quite sure about this change. Shouldn't we change `io.iohk.ethereum.domain.BlockchainImpl#bestKnownBlock` as well?

src/main/scala/io/iohk/ethereum/db/storage/CachedKeyValueStorage.scala
```diff
@@ -0,0 +1,39 @@
...
+    if (inMemory)
+      cache.update(toRemove, toUpsert)
+    else
+      storage.update(toRemove, toUpsert)
+
+    apply(cache, storage)
+  }
+
+  def persist(): Boolean = {
+    if (cache.shouldPersist) {
```
AlanVerbner: Here the cache is dumped to disk and everything is cleared. Have we considered doing something like a _sliding window_ where we only dump old stuff that has survived? I'm just curious about it, I don't think that should be done within this task.

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
KonradStaniec: I have actually totally gone around cache in case of fast sync, any node save just write it to storage, so best block is also written to storage when neccesssery.

src/main/scala/io/iohk/ethereum/db/storage/CachedKeyValueStorage.scala
KonradStaniec: I was considering it but taking into account the timeline for 1.1 i came to the conclusion I won't make it in time and it is better to have this simple flush all semantics and implement sliding window maybe for 1.2 .

src/main/scala/io/iohk/ethereum/db/storage/CachedKeyValueStorage.scala
```diff
@@ -0,0 +1,39 @@
...
+import io.iohk.ethereum.db.cache.Cache
+import scala.collection._
+
+trait CachedKeyValueStorage[K, V, T <: CachedKeyValueStorage[K, V, T]] extends SimpleMap[K, V, T]{
+  type I <: KeyValueStorage[K, V, I]
+  protected val storage: I
+  protected val cache: Cache[K, V]
+  def apply(cache: Cache[K,V] , storage: I): T
+
+  def get(key: K): Option[V] = cache.get(key) orElse storage.get(key)
```
AlanVerbner: Do you think it could improve performance if we store in the cache everything we read? If so, maybe we can leave a comment here for a follow up task

KonradStaniec: hmm I tested this idea by creating an implementation of `Cache[K,V]` which cached all data in `Scaffeine` cache, and there weren't big performance gains. ( although maybe `todo` is worth adding as this was quick check so i could missed something out)

src/main/scala/io/iohk/ethereum/db/storage/CachedKeyValueStorage.scala
AlanVerbner: Cool. Should we add a task marked for future releases?

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -219,8 +228,13 @@ class BlockchainImpl(
 
   override def getTotalDifficultyByHash(blockhash: ByteString): Option[BigInt] = totalDifficultyStorage.get(blockhash)
 
-  override def getBestBlockNumber(): BigInt =
-    appStateStorage.getBestBlockNumber()
+  override def getBestBlockNumber(): BigInt = {
+    val bestBlockNum = appStateStorage.getBestBlockNumber()
+    if (bestKnownBlock.get() > bestBlockNum)
```
AlanVerbner: is there any case where this is not true?

AlanVerbner:
@KonradStaniec Do you have any size metrics where this version is compared to the previous one? I thin we can use it as a quick check in order to verify everything is deleted as supposed to

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
KonradStaniec: It is not true just after mantis start (then `bestKnownBlock` is initialized to 0) so always best-saved block number will be returned. The same case is just after fast sync finish. (fast sync do not update `bestKnownBlock` but saves all data db  and update best block number in db)

KonradStaniec:
I have left it for the night with current settings and 1milion block dir size is `2.38 GB ` , the baseline size was `3.76 GB` the difference comes from the fact i have reduced kept history in this pr to do as much work in memory.

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
ntallar: Could this initialization in 0 cause an issue when [it's value is saved to the best block number in `appStateStorage`](https://github.com/input-output-hk/mantis/pull/411/files#diff-abb3f698c5418831dcb5f183fbf61937R383)?

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -260,13 +274,15 @@ class BlockchainImpl(
 
   override def save(hash: ByteString, evmCode: ByteString): Unit = evmCodeStorage.put(hash, evmCode)
 
-  override def saveBestBlockNumber(number: BigInt): Unit =
-    appStateStorage.putBestBlockNumber(number)
+  override def saveBestBlockNumber(number: BigInt): Unit = {
```
ntallar: Minor detail: Maybe it would be better rename this method to `saveBestKnownBlock` to distinguish it from `AppStateStorage.saveBestBlockNumber`?

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
KonradStaniec: hmm it is a though question, i do not see code path that could lead to it, but theoretically it is possible.
As I think about solution to handle it more defensively, maybe it would be better to call `getBestBlockNumber()`, instead of this `bestKnownBlock.get()`. That way we would dont risk this 0 block save, and all `bestKnownBlock` hanlding would be in `getBestBlockNumber()`. Wdyt?

ntallar: Hmm that would work. Maybe we could also initialize the `bestKnownBlock` with `appStateStorage.getBestBlockNumber()`, I'd prefer that as it keeps the value `bestKnownBlock` more consistent with what it should be on initialization, wdyt?

KonradStaniec: It sounds good but has this one caveat that we have a lot of `mock[BlockchainImpl]` in tests, and they would just throw `nullpointers` on missing `appStateStorage` during initialization. Maybe you know some work around it?

ntallar: That's a good point, there's none that I know of.
We should go with your solution then :+1:

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -332,19 +347,49 @@ class BlockchainImpl(
...
+    } else{
+      PruningMode.prune(pruningMode, blockNumber, cachedNodeStorage, inMemory = true)
+    }
+  }
+
+  def rollbackStateChangesMadeByBlock(blockNumber: BigInt): Unit ={
+    val currentBestSavedBlock = appStateStorage.getBestBlockNumber()
+
+    if (blockNumber <= currentBestSavedBlock)
+      PruningMode.rollback(pruningMode, blockNumber, cachedNodeStorage, inMemory = false)
```
ntallar: Should we clear the cache in this case?
If a node was deleted at a block `x` (it's new reference count is in cache) and we rollback to the block `x-1`, wouldn't we still have in cache the reference count after the delete? If that's so this would be eventually be persisted. WDYT?

ntallar: I think we are thinking of the same scenario, but just to double check, it would be the case that a rollback is done to block `best - 1` but the state of `best` is left in memory, so if persisted the state would become `best` (while the best block is now `best - 1`), right?

KonradStaniec: yes, we would save this stale state to the disk, although I dont think it should generate more problems than just junk on disk. ( best - 1 state is still intact on disk) 
So when later when executing `newBest`  it would have all needed nodes and its snapshots would just overwrite old ones.

ntallar: :thinking: I'm not sure if that's the case, @AlanVerbner , won't persisting that state possibly wrongly increase or decrease the reference of a node?

AlanVerbner: Saving junk nodes into DB is _not an issue_ The problem is when you mess around with snapshots as them are used to `undo` changes.
Let's say that you rollback between `[N, N+10]`. That means that we will replace by data in the snapshots for that range. If we rollback again between `[N+3, M], M > N + 10` and we have invalid data (snapshots that are no longer valid) we will undo with wrong data.

I need to re-read this specific case but it seems there might be some invalid snapshots in the cache, am I right @KonradStaniec @ntallar?

KonradStaniec: @AlanVerbner if I understand correctly `rollback` do not generate snapshots, only all state changes to from given block? If that's the case there won't be stale snapshots, only duplicated state of the best block (from the diagram) which can be further persisted on the wrong branch.

AlanVerbner: `rollback` does not generate snapshots. It uses them to replace data that's no longer valid.

src/main/resources/application.conf
```diff
@@ -334,7 +335,7 @@ mantis {
 
   sync {
     # Whether to enable fast-sync
-    do-fast-sync = true
+    do-fast-sync = false
```
ntallar: Why was this flag changed?

src/test/resources/rnd.sh
```diff
@@ -0,0 +1,31 @@
+# Simple script to randomly kill running mantis instance
```
ntallar: :+1:

src/main/resources/application.conf
```diff
@@ -66,14 +66,15 @@ mantis {
...
-        "enode://3f5f14647126dc39323447e22cb548369e1377a4e789ad48a6dc6680df1a39a28c46d36e79165fa155fdfaa67337d9703e1f029d5a788306fdef2030669a3bba@50.64.94.28:30303",
-        "enode://4fca8ecaf9bd12b805b2b99d2ed6a28e62223707d16c53dd746e2a70f067308b0d8d9a769412f9b323b3410083dfef3eeadd139fd6946535692c1b31467f1080@159.203.78.75:30303",
-        "enode://57be9e74b22ff3ea1bd3fedeb2ced310387dd176866793e273c7712305d8e4677f8913c86f93dfa8810e1cdb4177e5f87112db8748199a7771baf8dced63688b@104.41.188.223:30303",
-        "enode://94072bbbf1d3e5648afc573bbaf79b14a26ac48380f635fde32782631329263fe7a347251079f9abd3a2678f5bc5e672f8e6aff93a27f0f8f4e0e4f961dac68d@1.226.84.230:50303",
-        "enode://98b863da48ab8bef2339a823d552f3619fd8e892425ae40c6812c6f7e4a0afb4f9591b012183e89a63bb01c5085d0e96aa5f0812652335fb0ac946d6aaf15881@118.178.57.121:30303",
-        "enode://fba5a07e283d517a2680bcfc7aeb498ac2d246d756556a2ebd5edeb39496491c47a6d27e27f82833b7d7d12defc8de994de04bb58beb72472649f9a323006820@41.135.121.6:30303"
+        "enode://7ff0a073bf2ee9e94fae86de914eca0c0266c0b7051f4592ff0e9820706a085919ea667459897cde4efe3f0465f608cf402bde861983ce6ae3f9185df371aea7@138.197.135.19:30303",
```
ntallar: Should we also make this changes in `src/universal/conf/network.conf`?

ntallar (commented):
Apart from some minor comments, LGTM!

src/main/resources/application.conf
KonradStaniec: as spoken, it was wrong change in config

src/main/resources/application.conf
KonradStaniec: as spoken, it was wrong change in config

KonradStaniec:
It was mistake on on my side (changes to `application.conf` were from other branch i have locally), i reverted it to commit `3465b89` (baseline)

src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
```diff
@@ -125,7 +125,7 @@ object DumpChainApp extends App with NodeKeyBuilder with SecureRandomBuilder wit
 
     override def saveNode(nodeHash: NodeHash, nodeEncoded: NodeEncoded, blockNumber: BigInt): Unit = ???
 
-    override def removeBlock(hash: ByteString, saveParentAsBestBlock: Boolean, rollbackState: Boolean = true): Unit = ???
+    override def removeBlock(hash: ByteString, ithState: Boolean = true): Unit = ???
```
AlanVerbner: small typo here?

AlanVerbner (approved):
lgtm

ntallar (approved):
LGTM!

PR #413, opened 2018-02-23
[EC-482] Additional fields in syncing api (live after death)

KonradStaniec:
## Description

Adding more info to `eth_syncing`.

- original target: `phase/release1_1`


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -325,6 +325,10 @@ class FastSync(
         .addPendingNodes(hashesToRequest.flatten)
         .copy(downloadedNodesCount = syncState.downloadedNodesCount + nodeData.values.size)
 
+
+      appStateStorage.putPulledStateNodes(syncState.downloadedNodesCount)
```
LukasGasior1: I wonder, do we need to save those to leveldb? We already do that as a part of sync state (persistSyncState). Maybe we can load it from there on startup and keep as a var somewhere? Or maybe EthService could just ask FastSync for `SyncState`?

KonradStaniec: good call, it is just case of using existing `FastSyncStorage`  to get currently persisted state.

LukasGasior1: this should work too although keep in mind this state is only persisted every 1 minute or so (I think)

KonradStaniec: yep every minute, but I think we will be good with it, we just want to give the user some feedback that something is happening

mirkoAlic:
After a little discussion with @ntallar, we conclude is better to not add BREAKS_API changes like this one, given some tools could be parsing the `eth_syncing` response.

PR #414, opened 2018-02-23
Update external VM config

LukasGasior1:


LukasGasior1:
@rtkaczyk I think we're good with current config? If so, I'll close this PR

rtkaczyk:
yes, I think we're good

PR #417, opened 2018-03-05
[EC-484] Implement ecip1041

KonradStaniec:
Docs: https://github.com/ethereumproject/ECIPs/blob/master/ECIPs/ECIP-1041.md
Afaik it only affects classic and morden.

### DIFF
--- a/src/universal/conf/ethereumHF.conf
+++ b/src/universal/conf/ethereumHF.conf
@@ -26,6 +26,8 @@ mantis {
     # Setting both values to zero disables difficulty bomb delay
     difficulty-bomb-pause-block-number = "0"
     difficulty-bomb-continue-block-number = "0"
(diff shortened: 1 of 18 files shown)


src/universal/conf/ethereumHF.conf
```diff
@@ -26,6 +26,8 @@ mantis {
     # Setting both values to zero disables difficulty bomb delay
     difficulty-bomb-pause-block-number = "0"
     difficulty-bomb-continue-block-number = "0"
+    # Disabled for ethereumHF
+    difficulty-bomb-removal-block-number = "1000000000000000000"
```
AlanVerbner: Minor comment (no changes required). I still prefer an `Option` here, but I remember it was already discussed for `eip161-block-number`

PR #418, opened 2018-03-07
[EC-472 ]Feature/update bouncy

KonradStaniec:
Updating bouncy castle from 1.56 to 1.59

As I measured old bouncy castle keccak hash functions were pretty slow,  that probably why we used `org.consensusresearch" %% "scrypto` in few places (they way notably faster). Release notes of new bouncy castle version state that there was some improvement in that case.
This pr gives us:
-We are consolidating  the hashing into one library (this was one of security audit suggestions) 
-Performance wise we have a little bump:
Sync 10 minutes from master node:
`Baseline - 70696 blocks`
`Improved - 74304 blocks` 
It is a minor improvement (5%) in the least interesting part of sync ( at the beginning when io does not hit us)
but it gives us the possibility to further optimize `intsToBytes` conversion which is currently main bottleneck in ethash validation.


build.sbt
```diff
@@ -72,12 +71,10 @@ val verifyDeps = Seq(
   "org.scala-lang.modules" % "scala-xml" sha1 "e22de3366a698a9f744106fb6dda4335838cf6a7",
   "de.heikoseeberger" % "akka-http-json4s" sha1 "575ba7e823282bf64ebcb4c04e33d9a6c24b5c31",
   "io.suzaku" % "boopickle" sha1 "85281da7833655a0fb7c91274c7a97cbe61b990c",
-  "org.consensusresearch" % "scrypto" sha1 "20177c8d44689b7d3d398ab1f142daf42b8a978c",
   "com.chuusai" % "shapeless" sha1 "27e115ffed7917b456e54891de67173f4a68d5f1",
   "org.typelevel" % "macro-compat" sha1 "ed809d26ef4237d7c079ae6cf7ebd0dfa7986adf",
   "com.google.guava" % "guava" sha1 "6ce200f6b23222af3d8abb6b6459e6c44f4bb0e9",
-  "org.whispersystems" % "curve25519-java" sha1 "09091eb56d696d0d0d70d00b84e6037dcd3d98b6",
```
AlanVerbner: good catch

PR #421, opened 2018-03-12
[EC-423] Bug/hande tcp udp ports

KonradStaniec:
Now each peer can have udp and tcp ports specified separatly.

The logic of discovery needed to be changed, as the only way to get udp and tcp ports separately is from `neighbours` message.

Also, there are some minor changes that should help with bug that some times our discovery is stuck at some nodes count and wont progress because it missed one or more good peers. (it can still happen but it is rarer from my several runs)

### DIFF
--- a/src/main/scala/io/iohk/ethereum/network/discovery/Node.scala
+++ b/src/main/scala/io/iohk/ethereum/network/discovery/Node.scala
@@ -53,11 +53,6 @@ object NodeParser extends Logger {
 
   type Error = String
 
-  private def combineValidations[A,B](obj: B, eithers: Either[A,B]*): Either[Set[A], B] = {
-    val errors = eithers.collect { case Left(e) => e }
-    if (errors.isEmpty) Right(obj) else Left(errors.toSet)
-  }
-
   private def validateTcpAddress(uri: URI): Either[Error, URI] = {
     Try(InetAddress.getByName(uri.getHost) -> uri.getPort) match {
       case Success(tcpAddress) if tcpAddress._2 != -1 => Right(uri)
@@ -93,6 +88,8 @@ object NodeParser extends Logger {
   }
 
   private def validateNodeUri(node: String): Either[Set[Error], URI] = {
+    import io.iohk.ethereum.utils.ValidationUtils._
+
     val uri = validateUri(node)
     uri match {
       case Left(error) => Left(Set(error))
--- a/src/main/scala/io/iohk/ethereum/network/discovery/Node.scala
+++ b/src/main/scala/io/iohk/ethereum/network/discovery/Node.scala
@@ -9,18 +9,40 @@ import org.spongycastle.util.encoders.Hex
 
 import scala.util.{Failure, Success, Try}
 
-case class Node(id: ByteString, addr: InetSocketAddress) {
+case class Node(id: ByteString, addr: InetAddress, tcpPort: Int, udpPort: Int) {
+
+  lazy val udpSocketAddress = new InetSocketAddress(addr, udpPort)
+  lazy val tcpSocketAddress = new InetSocketAddress(addr, tcpPort)
+
   def toUri: URI = {
-    val host = network.getHostName(addr.getAddress)
-    val port = addr.getPort
-    new URI(s"enode://${Hex.toHexString(id.toArray[Byte])}@$host:$port")
+    val host = network.getHostName(addr)
+    new URI(s"enode://${Hex.toHexString(id.toArray[Byte])}@$host:$tcpPort?discport=$udpPort")
   }
 }
 
 object Node {
+
+  // If there is no udp port specified or it is malformed use tcp as default
+  private def getUdpPort(uri: URI, default: Int): Int = {
+    Option(uri.getQuery).fold(default) { query =>
+      Try {
+        val params = query.split("=")
+        if (params(0) == "discport")
+          params(1).toInt
+        else
+          default
+      } match {
+        case Success(udpPort) => udpPort
+        case Failure(_) => default
+      }
+    }
+  }
+
   def fromUri(uri: URI): Node = {
     val nodeId = ByteString(Hex.decode(uri.getUserInfo))
-    Node(nodeId, new InetSocketAddress(uri.getHost, uri.getPort))
+    val address = InetAddress.getByName(uri.getHost)
+    val tcpPort = uri.getPort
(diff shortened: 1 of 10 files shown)


src/main/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManager.scala
```diff
@@ -42,13 +45,21 @@ class PeerDiscoveryManager(
...
+    // At the beginning ping all bootstrap nodes
+    if (firstScan) {
+      bootStrapNodesInfo.foreach{ nodeInfo =>
+        sendPing(Endpoint.makeEndpoint(nodeInfo.node.udpSocketAddress, nodeInfo.node.tcpPort), nodeInfo.node.udpSocketAddress, nodeInfo)
+      }
+
+      firstScan = false
+    }
+
+    // Ping a random sample from currently pinged nodes without the answer
```
AlanVerbner: There was a comment on the previous version about 

> One optimisation worth considering is pinging them only every n-th scan

The goal of this is the same, right?

KonradStaniec: Yes, it is one of its goals but this approach is more robust imo.
At the beginning of syncing, the sample will be composed only from our bootstrap nodes so unless we would get more new node we will constantly re-ping bootstraps.
And later, It can help us with the case that perfectly good peer pong will be missed because of udp having no guarantees of delivery. (we will constantly re-ping a sample from unanswered ping)
This approach is not perfect but should help us with cases, that because of missed some valid pongs we will be stuck. (now it can happen from time to time and one solution is to restart mantis with hope that this time `pongs` will get through)

src/main/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManager.scala
```diff
@@ -54,15 +73,17 @@ class PeerDiscoveryManager(
...
-        nodesInfo += newNodeInfo.node.id -> newNodeInfo
-        sendMessage(FindNode(ByteString(nodeStatusHolder().nodeId), expirationTimestamp), from)
-      } else {
-        val (earliestNode, _) = nodesInfo.minBy { case (_, node) => node.addTimestamp }
-        nodesInfo -= earliestNode
-        nodesInfo += newNodeInfo.node.id -> newNodeInfo
+      pingedNodes.get(packet.nodeId).foreach { newNodeInfo =>
+        val nodeInfoUpdatedTime = newNodeInfo.copy(addTimestamp = clock.millis())
+        pingedNodes -= newNodeInfo.node.id
+        if (nodesInfo.size < discoveryConfig.nodesLimit) {
```
AlanVerbner: (Minor) This logic seems to be quite similar to `updatePingedNodes` .. maybe we can create a function to share the common stuff (check the size, remove the oldest one, etc)

src/main/scala/io/iohk/ethereum/network/discovery/Node.scala
```diff
@@ -9,65 +9,112 @@ import org.spongycastle.util.encoders.Hex
...
-    */
-  def parseNode(node: String): Either[Set[Throwable], Node] = {
 
-    val maybeUri = Try(new URI(node))
+  type Error = String
 
-    val maybeScheme = maybeUri.map(_.getScheme).flatMap { scheme =>
-      if (scheme == NodeScheme) Success(scheme)
-      else Failure(new Exception(s"Invalid node scheme $scheme, it should be $NodeScheme"))
+  private def combineValidations[A,B](obj: B, eithers: Either[A,B]*): Either[Set[A], B] = {
```
AlanVerbner: Maybe this could be moved to `package io.iohk.ethereum.utils` package, wdyt?

src/main/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManager.scala
```diff
@@ -54,15 +73,17 @@ class PeerDiscoveryManager(
...
+          nodesInfo += newNodeInfo.node.id -> nodeInfoUpdatedTime
+          sendMessage(FindNode(ByteString(nodeStatusHolder().nodeId), expirationTimestamp), from)
```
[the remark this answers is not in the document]
KonradStaniec: I was thinking about it also, there is the second person who thinks similarly I think I will go that way 👍

KonradStaniec:
I have Intruduced all comments, and also changed `pong` validation instead getting stored ping by `nodeId`  I now use `ping` mdc, it was sugested in Eclipse paper and it is more robust, also nodeId can change ( few of our bootstraps nodes has new ids, with old approach we were rejecting pongs from them)

src/main/scala/io/iohk/ethereum/utils/ValidationUtils.scala
```diff
@@ -0,0 +1,16 @@
+package io.iohk.ethereum.utils
+
+object ValidationUtils {
```
AlanVerbner: :+1:

src/main/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManager.scala
```diff
@@ -158,7 +165,12 @@ object PeerDiscoveryManager {
 
   }
 
-  case class DiscoveryNodeInfo(node: Node, addTimestamp: Long)
+
+  sealed abstract class TimedInfo {
+    def addTimestamp: Long
+  }
+  case class DiscoveryNodeInfo(node: Node, addTimestamp: Long) extends TimedInfo
+  case class PingInfo(nodeinfo: DiscoveryNodeInfo, addTimestamp: Long) extends TimedInfo
```
ntallar: Why do we need to store 2 `addTimestamp`s for a node (the one in `PingInfo` and `DiscoveryNodeInfo` in it)? Are we still using the `nodeInfo.addTimestamp`?

KonradStaniec: we are using `nodeInfo.addTimestamp` to sort nodes which responded with pong, and we update this timestamp after their response. Also we scan only most fresh nodes

`pingInfo.addTimestamp` is used to record time of pings for which there can be no answer, and is used to remove oldest unanswered ping.

KonradStaniec:
agreed with @AlanVerbner  and @ntallar  to look for root cause of connectivity issues in other pr as it may not be related to this pr (checked it out on other branches)

PR #422, opened 2018-03-13
fix BlockchainTests OOM

rtkaczyk:


src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
```diff
@@ -1,35 +1,43 @@
...
-import io.iohk.ethereum.utils.{BlockchainConfig, Logger}
+import io.iohk.ethereum.utils.BlockchainConfig
+import io.iohk.ethereum.validators._
 import org.spongycastle.util.encoders.Hex
 
 import scala.util.{Failure, Success, Try}
 
+object ScenarioSetup {
+  // FIXME: this is a temporary (hopefully) fix for BlockchainSuite. It avoids using `ValidatorsBuilder` which causes
+  // the tests to run into OOM errors (for reason currently unknown)
```
mcsherrylabs: This is worth getting to the bottom of. It's been floating around for a while. It's trying to tell us something (!!) Potentially something to do with lazy init.... :thinking:

rtkaczyk: > It's been floating around for a while.

I am not aware of any previous incidents when `ValidatorsBuilder` was causing an OOM. This was introduced in #407.
Or do you mean OOM errors in general?

mcsherrylabs: I mean OOM in general. I *think* I've noticed an uptick in occurrences. Although I've no hard facts to back that up. I could be wrong?

loverdos:
s/Run/Ran/

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
rtkaczyk: Well, this will always be a memory hungry beast. And the reasons for going OOM vary case by case, I think.

PR #425, opened 2018-03-19
Use an exact amount of heap memory with SBT

loverdos:
We also have SBT print the exact process arguments by using the `-v` parameter.

These are mostly for when running under CircleCI. We have noticed random OOM errors during compilation (= compiler crashes) due to not having enough memory under CircleCI. This PR ensures that we specify exactly our requirements.

The amount of heap memory chosen, `1024`, has never failed me locally and seems reasonable, so it was the first candidate.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,8 @@
 version: 2
 jobs:
   build:
+    environment:
+      - SBT: "sbt -v -mem 1024"
     docker:
       - image: rtkaczyk/mantis-circleci:v4
     steps:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -17,24 +17,24 @@ jobs:
 
       - run:
           name: scalastyle
-          command: sbt scalastyle test:scalastyle
+          command: sbt -v -mem 1024 scalastyle test:scalastyle
```
rtkaczyk: We could probably avoid repetition by utilising the `environment` section:

```
jobs:
  build:
    environment:
      - SBT: "sbt -v -mem 1024"
    steps:
      - run:
          name: scalastyle
          command: $SBT scalastyle test:scalastyle
```

Also note that `-mem` does not necessarily force the amount:

```
$ SBT="sbt -v -mem 1024"
$ $SBT scalastyle
[process_args] java_version = '1.8.0_112'
# Executing command line:
java
-XX:ReservedCodeCacheSize=128m
-Xmx4096m
-Xms512m
-Xss64m
-XX:MaxMetaspaceSize=1024m
-jar
/usr/local/Cellar/sbt/0.13.12/libexec/sbt-launch.jar
scalastyle
```

rtkaczyk: With regard to memory it seems the priority is:
1. `JAVA_OPTS` var
2. `sbtopts` file
3. `mem` switch

That's good, it means we can still override this memory setting within Circle's web UI

loverdos: I implemented the repetition avoidance suggestion.

rtkaczyk (approved):

LukasGasior1 (approved):

PR #426, opened 2018-03-19
Add testmode

LukasGasior1:
- added JSON RPC via Unix Domain Socket
- added testmode
- 1831 out of 1837 Solidity test cases passing (6 failing)

TODO:
- unit tests (at least for IPC, test_ endpoints are kind of covered by those Solidity tests anyway)
- fix remaining 6 Solidity tests (possibly another PR) // edit: it seems, at lest some of these tests, are failing because of Mantis not properly handling "pending" block parameter in RPC api, see task: EC-510 for more details


src/main/scala/io/iohk/ethereum/jsonrpc/IeleJsonMethodsImplicits.scala
```diff
@@ -0,0 +1,104 @@
...
+    def toEitherOpt[A, B](opt: Option[Either[A, B]]): Either[A, Option[B]] =
+      opt.map(_.right.map(Some.apply)).getOrElse(Right(None))
+
+    for {
+      from <- toEitherOpt((obj \ "from").extractOpt[String].map(extractBytes))
+      to <- toEitherOpt((obj \ "to").extractOpt[String].map(extractBytes))
+      gas <- optionalQuantity(obj \ "gas")
+      gasPrice <- optionalQuantity(obj \ "gasPrice")
+      value <- optionalQuantity(obj \ "value")
+      functionName = (obj \ "functionName").extractOpt[String]
```
dwightguth: we've previously used the name "function" for this field instead of "functionName". If you don't have a strong opinion, it may be simpler to do the same thing here.

src/main/scala/io/iohk/ethereum/jsonrpc/TestJsonMethodsImplicits.scala
```diff
@@ -0,0 +1,110 @@
...
+        ByteString(Hex.decode(key.replace("0x", ""))) -> AccountConfig(None, wei)
+      }
+      Right(accounts)
+    }
+
+    private def extractBlockchainParams(blockchainParamsJson: JValue): Either[JsonRpcError, BlockchainParams] = {
+      for {
+        eIP150ForkBlock <- extractQuantity(blockchainParamsJson \ "EIP150ForkBlock")
+        eIP158ForkBlock <- extractQuantity(blockchainParamsJson \ "EIP150ForkBlock")
+        accountStartNonce <- extractQuantity(blockchainParamsJson \ "EIP150ForkBlock")
```
dwightguth: this code looks a little suspicious to me. Is this a copy/paste error or was it deliberate?

LukasGasior1: lol! I totally overlooked that.

src/universal/conf/testmode.conf
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  pruning.mode = "archive"
+
+  blockchain = {
+    eip160-block-number = "0"
+  }
+
+  network.rpc {
+    apis = "eth,web3,net,personal,daedalus,test"
```
dwightguth: this should include iele if we want to use testmode for the solidity-to-iele test suite, right?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -544,13 +563,29 @@ class EthService(
 
   def call(req: CallRequest): ServiceResponse[CallResponse] = {
     Future {
-      doCall(req)(ledger.simulateTransaction).map(r => CallResponse(r.vmReturnData))
+      doCall(req)(ledgerHolder().simulateTransaction).map(r => CallResponse(r.vmReturnData))
+    }
+  }
+
+  def ieleCall(req: IeleCallRequest): ServiceResponse[CallResponse] = {
```
dwightguth: iele transactions return a list of values rather than a single value, so I believe you want to make a IeleCallResponse class and perform rlp decoding in this method.

LukasGasior1: fixed

src/main/scala/io/iohk/ethereum/extvm/VMClient.scala
```diff
@@ -14,13 +14,17 @@ import scala.annotation.tailrec
...
+    externalVmConfig: VmConfig.ExternalConfig,
     messageHandler: MessageHandler,
     testMode: Boolean)
   extends Logger {
 
   def sendHello(version: String, blockchainConfig: BlockchainConfig): Unit = {
     val config = BlockchainConfigForEvm(blockchainConfig)
-    val configMsg = msg.Hello.Config.EthereumConfig(buildEthereumConfigMsg(BlockchainConfigForEvm(blockchainConfig)))
+    val configMsg = externalVmConfig.vmType match {
+      case VmConfig.ExternalConfig.VmTypeIele => msg.Hello.Config.IeleConfig(buildIeleConfigMsg())
```
dwightguth: based on our discussion on slack, weren't we planning for our near-term solution to be to make IeleConfig also be passed on every CallContext?

rtkaczyk:
Could you add some info on how to setup and run solidity tests?

LukasGasior1:
@rtkaczyk 
- first, you need to build solidity from sources: https://solidity.readthedocs.io/en/latest/installing-solidity.html#building-from-source
- then you start Mantis in testmode (uncomment include testmode.conf)
- and run the tests: ` ETH_TEST_IPC="~/.mantis-testmode/mantis.ipc" ./test/soltest -- --no-smt`

by the way, DO NOT REVIEW :smile: I'm resolving conflicts and doing a small refactor on this occasion

src/main/scala/io/iohk/ethereum/jsonrpc/TransactionRequest.scala
```diff
@@ -20,8 +20,19 @@ case class TransactionRequest(
       nonce = nonce.getOrElse(defaultNonce),
       gasPrice = gasPrice.getOrElse(defaultGasPrice),
       gasLimit = gasLimit.getOrElse(defaultGasLimit),
-      receivingAddress = to,
+      receivingAddress = if (to.contains(Address(0))) None else to,
```
LukasGasior1: It seems this is the default behavior in cpp-ethereum (and what solidity tests expect). To me it (kind of) makes sense, but is a bit counterintuitive for caller (what if you really want to send tx to address 0x0?). WDYT? Other solution is to use this logic only in testmode.

src/main/resources/application.conf
```diff
@@ -144,50 +147,60 @@ mantis {
...
-      certificate-keystore-path = null
+      http {
+        # JSON-RPC mode
+        # Available modes are: http, https
+        # Choosing https requires creating a certificate and setting up 'certificate-keystore-path' and
+        # 'certificate-password-file'
+        # See: https://github.com/input-output-hk/mantis/wiki/Creating-self-signed-certificate-for-using-JSON-RPC-with-HTTPS
+        mode = "http"
+
+        # Whether to enable JSON-RPC endpoint
```
loverdos: Should this be
> Whether to enable JSON-RPC HTTP(S) endpoint
?

Also, minor, we could use the shorter `HTTP(S)` instead of  `HTTP/HTTPS`

loverdos:
There is a compilation error:
```
[error] /mantis/src/snappy/scala/io/iohk/ethereum/snappy/Prerequisites.scala:53: not found: type ValidatorsBuilder
[error]   private val components = new StdTestConsensusBuilder with ValidatorsBuilder with SyncConfigBuilder {
[error]                                                             ^
[error] one error found
```

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
```diff
@@ -353,12 +382,6 @@ trait OmmersPoolBuilder {
   lazy val ommersPool: ActorRef = system.actorOf(OmmersPool.props(blockchain, ommersPoolSize))
 }
 
-trait ValidatorsBuilder {
```
loverdos: This deletion causes a compilation error in snappy. Check `io.iohk.ethereum.snappy.Prerequisites.components`.

src/main/scala/io/iohk/ethereum/jsonrpc/TransactionRequest.scala
rtkaczyk: Well, you know... https://gastracker.io/addr/0x0000000000000000000000000000000000000000

This is of course a result of many mistakes, but maybe, for reasons beyond my imagination, someone has interest in sending Ether there. I'd say if it's not too much trouble go with testmode only, if not, don't worry about it.

Btw, the expression could be simplified as:
```
to.filter(_ != Address(0))
```

src/main/resources/application.conf
```diff
@@ -17,6 +17,9 @@ mantis {
   # timeout for shutting down the ActorSystem
   shutdown-timeout = "15.seconds"
 
+  # Whether to run Mantis in test mode (enables test_ RPC endpoints; similar to --test flag in cpp-ethereum)
```
rtkaczyk: Does it change other functionality? If not, could be moved into `rpc` section.

LukasGasior1: It does. If running in testmode we use different Ledger/Consensus implementations.

rtkaczyk: In that case maybe we should document how the functionality differs in test mode? Or is that already documented somewhere else?

src/main/scala/io/iohk/ethereum/testmode/TestLedgerWrapper.scala
```diff
@@ -0,0 +1,19 @@
...
+import io.iohk.ethereum.utils.Config.SyncConfig
+
+/**
+  * Provides a ledger instance with modifiable blockchain config (used in testmode).
+  */
+class TestLedgerWrapper(
+    blockchain: BlockchainImpl,
+    syncConfig: SyncConfig,
+    consensus: Consensus,
+    var blockchainConfig: BlockchainConfig) { // var as it's modifiable by test_ RPC endpoints
```
rtkaczyk: Couldn't we just create a new instance of this class?

src/main/scala/io/iohk/ethereum/testmode/TestmodeConsensus.scala
```diff
@@ -0,0 +1,69 @@
...
+import io.iohk.ethereum.network.p2p.messages.PV62
+import io.iohk.ethereum.nodebuilder._
+import io.iohk.ethereum.utils.BlockchainConfig
+
+class TestmodeConsensus(
+    override val vm: VMImpl,
+    blockchain: BlockchainImpl,
+    blockchainConfig: BlockchainConfig,
+    consensusConfig: ConsensusConfig,
+    var blockTimestamp: Long = 0) // var, because it can be modified by test_ RPC endpoints
```
rtkaczyk: Similar as above, couldn't we create a new instance?

rtkaczyk:
> unit tests (at least for IPC, test_ endpoints are kind of covered by those Solidity tests anyway)

I take it this was done?

src/main/scala/io/iohk/ethereum/jsonrpc/server/ipc/JsonRpcIpcServer.scala
```diff
@@ -0,0 +1,119 @@
...
+    log.info(s"Starting IPC server: ${config.socketFile}")
+
+    removeSocketFile()
+
+    serverSocket = new UnixDomainServerSocket(config.socketFile)
+    new Thread {
+      override def run(): Unit = {
+        while (!serverSocket.isClosed) {
+          val clientSocket = serverSocket.accept()
+          new ClientThread(jsonRpcController, clientSocket).start()
```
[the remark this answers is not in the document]
LukasGasior1: It's one thread per connection; it stops when connection closes

src/main/scala/io/iohk/ethereum/testmode/TestLedgerWrapper.scala
LukasGasior1: Not really. It's a bit tricky, because Ledger is being passed as a parameter to various other components on initialization (see NodeBuilder.scala). I wanted to create new instances, but I didn't find an elegant way to update those classes with new ledger (without introducing a var elsewhere).

rtkaczyk:
@LukasGasior1 I'm unable to run the tests. Trying the command you provided:

```
$ ETH_TEST_IPC="~/.mantis-testmode/mantis.ipc" ./test/soltest -- --no-smt
bash: ./test/soltest: No such file or directory
```

So I tried:
```
$ find . -name soltest
./build/test/soltest

$ ETH_TEST_IPC="~/.mantis-testmode/mantis.ipc" ./build/test/soltest -- --no-smt
Test setup error: std::exception: No test path specified. The --testpath argument is required.
```

Any hints?

src/main/scala/io/iohk/ethereum/jsonrpc/server/ipc/JsonRpcIpcServer.scala
```diff
@@ -0,0 +1,119 @@
...
+        }
+      }
+    }.start()
+  }
+
+  def close(): Unit = {
+    serverSocket.close()
+    removeSocketFile()
```
loverdos: If `close()` throws an `Exception`, the file will not be removed.

src/main/scala/io/iohk/ethereum/jsonrpc/server/ipc/JsonRpcIpcServer.scala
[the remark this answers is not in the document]
loverdos: Why not use a thread pool?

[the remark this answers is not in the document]
LukasGasior1: good idea

loverdos:
@LukasGasior1 Could you `git cherry-pick ` 0b033e29bb28261d8b8cf70f90199b83c708ad8d ? It will make testing a bit easier.

src/main/scala/io/iohk/ethereum/jsonrpc/server/ipc/JsonRpcIpcServer.scala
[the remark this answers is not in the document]
LukasGasior1: @loverdos after taking a second thought I'm not sure if it's the best idea. What are the benefits of using a thread pool here?

rtkaczyk:
```
$ ETH_TEST_IPC="$HOME/.mantis-testmode/mantis.ipc" ./build/test/soltest -- --no-smt
Running 1837 test cases...
/home/radek/iele/solidity/test/libsolidity/SolidityEndToEndTest.cpp(1502): error in "blockchain": Invalid encoded data
   Result                                                           Expectation
   ..............................................................1c ..............................................................1c
   ........................1212121212121212121212121212121212121212 ........................1212121212121212121212121212121212121212
 X ...............................................................6 ...............................................................7

/home/radek/iele/solidity/test/libsolidity/SolidityEndToEndTest.cpp(1552): error in "now": Invalid encoded data
   Result                                                           Expectation
   ...............................................................1 ...............................................................1
 X .............................................................22c .............................................................22d

/home/radek/iele/solidity/test/libsolidity/SolidityEndToEndTest.cpp(10784): error in "snark": check callContractFunction("pair()") == encodeArgs(true) failed
/home/radek/iele/solidity/test/libsolidity/SolidityEndToEndTest.cpp(10785): error in "snark": check callContractFunction("verifyTx()") == encodeArgs(true) failed
/home/radek/iele/solidity/test/contracts/Wallet.cpp(574): error in "multisig_value_transfer": check balanceAt(destination) == 100 failed [0 != 100]
/home/radek/iele/solidity/test/contracts/Wallet.cpp(644): error in "revoke_transaction": check balanceAt(destination) == 100 failed [0 != 100]

*** 6 failures detected in test suite "SolidityTests"
```

Looking good, but do we have a plan to fix the remaining tests?

LukasGasior1:
@rtkaczyk  I've spent some time trying to figure them out. I think they may be related to the bug with "pending" as block param. To be honest, I wouldn't like to spend more time investigating it. Especially that this task was originally a minor thing.

src/main/scala/io/iohk/ethereum/jsonrpc/server/ipc/JsonRpcIpcServer.scala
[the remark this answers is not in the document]
loverdos: Avoid a new thread for each new connection?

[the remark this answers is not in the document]
LukasGasior1: You mean sharing one thread for multiple connections? I think it doesn't make sense with current approach (we're using blocking socket reading)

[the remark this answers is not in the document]
loverdos: I am mostly concerned with preventing an unbounded number of threads and/or outstanding requests and I was thinking about using an appropriately configured `Executor`.

loverdos:
@LukasGasior1 
So, trying to understand what is going on: We introduce a new service, `TestService`, which we interface to the outer world via the existing `JsonRpcController`. The `TestService` initially starts with the original Mantis configuration but it will deviate as soon as some calls are made to:

```scala
  private def handleTestRequest(testService: TestService): PartialFunction[JsonRpcRequest, Future[JsonRpcResponse]] = ...
```

The crucial observation is that `TestService` and the related _testmode_ facilities live in parallel to the rest of Mantis. If this is correct, then we do not have to worry about introducing inconsistencies to the rest of Mantis.

Having said that, using `var`s probably exposes the state a bit too eagerly. We could model state changes by respective `updateXXX` methods, e.g. `updateBlockChainConfig`. This resembles _firing an event_ and gives the receiver the advantage of internally doing whatever else is necessary. This way we can fight possible inconsistencies _inside_ the whole testmode. 

So, for example, when we update `blockchainConfig` https://github.com/input-output-hk/mantis/blob/493da554d09857816f037fe27b43c2b2c169ef79/src/main/scala/io/iohk/ethereum/testmode/TestLedgerWrapper.scala#L16 in `TestService`https://github.com/input-output-hk/mantis/blob/322183788c0d2c450316d42f93e9776b7718271e/src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala#L103 what does it mean to just set a new value? Does it have to propagate further?

I am not saying there are inconsistencies with the current state change approach but I can identify a potential source.

LukasGasior1:
@loverdos 

The main problem here is that, when we update the blockchain config in testmode, it should also be updated globally (wherever Ledger or related classes are used - miner etc). I solved that by introducing `TestLedgerWrapper` (notice the `ledger` implementation there is a `def`) and `TestLedgerBuilder.TestLedgerProxy` which delegates to the ledger in this mutable wrapper. What I wanted to avoid here is to couple non-test logic (sync, block execution etc) with testmode functionality (note, that with this solution all remaining classes don't even know they're using test/mutable ledger).
> what does it mean to just set a new value? Does it have to propagate further?

yes, subsequent references to ledger (from any place) will use this new confg - thanks do this wrapper mechanism.
This approach may not be perfect, but it in my opinion it has two strong pros:
- existing code stays untouched and is not aware of testmode existence
- this mechanism is _only_ used in testmode, it does not affect a "normally" running node in _any_ way, see:
```
class StdNode extends BaseNode with StdLedgerBuilder with StdConsensusBuilder
class TestNode extends BaseNode with TestLedgerBuilder with TestmodeConsensusBuilder with TestServiceBuilder
```

rtkaczyk:
@LukasGasior1  can this be merged now?

PR #432, opened 2018-03-30
[cgkiele-138,139] Polymorphic difficulty calculation & validation

loverdos:
Some extra features are also present:
* Bump Scala version to `2.12.5`.
* Bump `sbt-verify` version to `0.4.1`.
* Have `sbt-verify` issue warnings instead of errors. This is a temporary solution to some inconsistent behaviour we have observed, so that the developer experience is a bit better.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,7 @@ enablePlugins(JDKPackagerPlugin, JavaAppPackaging, SolidityPlugin)
 val commonSettings = Seq(
   name := "mantis",
   version := "1.0-daedalus-rc1",
-  scalaVersion := "2.12.1",
+  scalaVersion := "2.12.5",
   testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-l", "EthashMinerSpec") // miner tests disabled by default
 )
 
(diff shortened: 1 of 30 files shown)


build.sbt
```diff
@@ -54,7 +54,7 @@ val dep = {
 }
 
 val verifyDeps = Seq(
-  "org.scala-lang" % "scala-library" sha1 "dd235d04037dc6f4b6090257872dd35359a563ce",
+  "org.scala-lang" % "scala-library" sha1 "4861b00d921952f2dc9a024198bc72590fd7dc5e",
```
rtkaczyk: So, what was the problem/solution?

loverdos: `sbt-verify`: locally the last build step ("additional compilation & dist") fails, regardless of Scala version. No solution for that but I think we should decouple the scala version upgrade from whatever `sbt-verify` does. The above changes are just to take into account the modified `2.12.5` scala jars.

rtkaczyk: Yes, it fails locally for me too :(

> decouple the scala version upgrade from whatever sbt-verify does

How can we do that? Unless you mean removing `sbt-verify` (or treating the failed checks as warnings) which I would be happy with 😁

project/plugins.sbt
```diff
@@ -6,5 +6,6 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-native-packager" % "1.0.6")
 addSbtPlugin("com.thoughtworks.sbt-api-mappings" % "sbt-api-mappings" % "latest.release")
 addSbtPlugin("uk.co.josephearl" % "sbt-verify" % "0.3.0-SNAPSHOT")
 addSbtPlugin("com.thesamet" % "sbt-protoc" % "0.99.12")
+addSbtPlugin("com.timushev.sbt" % "sbt-updates" % "0.3.4")
```
rtkaczyk: Interesting, do you intend it for local use only, or should we include it in the build pipeline?

rtkaczyk:
As long as we intend on retaining `sbt-verify` checks could we also use this PR to update its version to `0.4.1`? It's kind of a bug right now...

project/plugins.sbt
loverdos: I propose we should (but not necessarily with changes in this PR). It gives you a very good idea of how old your libraries are.

loverdos:
The build is looking good so far, I'll remove the <kbd>`do-not-merge`</kbd> label when it finishes succesfully.

PR #433, opened 2018-04-03
[EC-509] Bug/fastsync snapshots

KonradStaniec:
Problem: During fast sync we are saving nodes + nodes snapshots under target block key. Later during regular sync at block `target block number+history` we need to prune those snapshots ( we do not prune any nodes, as every node saved during fast sync has a number of references > 0 ). Currently, there are around 24 million nodes downloaded during fast sync which translates roughly to 48 million random reads during pruning of target block ( read snapshot + read node). Also we are doing those reads on already pretty big db (23gb), so their speed departs from speed mentioned in benchmarks (leveldb read speed deteriorates with db size)

Solution: Do not save snapshots for target block nodes as they are not needed. It is extremely (impossible) unlikely we will ever rollback state to the target block, as currently in fast sync blocks are validated up to `targetblock+24`, so during regular sync first 24 blocks should execute without any problems.

Testing: Already fast synced to top of the chain and have been running regular sync for over 72h without problems


src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
```diff
@@ -94,7 +94,7 @@ class GenesisDataLoader(
     val initalRootHash = MerklePatriciaTrie.EmptyRootHash
 
     val stateMptRootHash = genesisData.alloc.zipWithIndex.foldLeft(initalRootHash) { case (rootHash, (((address, AllocAccount(balance)), idx))) =>
-      val ephemNodeStorage =  PruningMode.nodesKeyValueStorage(pruning.ArchivePruning, nodeStorage)(Some(idx - genesisData.alloc.size))
+      val ephemNodeStorage = PruningMode.nodesKeyValueStorage(pruning.ArchivePruning, nodeStorage, withSnapshotsSave = true)(Some(idx - genesisData.alloc.size))
```
AlanVerbner: shouldn't this one be false as the one that  @ntallar mentions in https://github.com/input-output-hk/mantis/pull/433/files#r178815627

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -166,7 +166,7 @@ trait Blockchain {
 
   def saveBestKnownBlock(number: BigInt): Unit
 
-  def saveNode(nodeHash: NodeHash, nodeEncoded: NodeEncoded, blockNumber: BigInt): Unit
+  def saveNode(nodeHash: NodeHash, nodeEncoded: NodeEncoded, blockNumber: BigInt, withSnapshotSave: Boolean): Unit
```
AlanVerbner: Although this solution will work, I'm not quite sure if this continues polluting upper layers with pruning internal detail. I'm wondering if we could configure FastSync with a storage that won't save snapshots and regular sync with one that will instead. Do you think it's too much? Maybe we work on a refactor on the next version

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
```diff
@@ -116,7 +116,7 @@ class GenesisDataLoader(
           " Use different directory for running private blockchains."))
       case None =>
         ephemDataSource.getAll(nodeStorage.namespace)
-          .foreach { case (key, value) => blockchain.saveNode(ByteString(key.toArray[Byte]), value.toArray[Byte], 0) }
+          .foreach { case (key, value) => blockchain.saveNode(ByteString(key.toArray[Byte]), value.toArray[Byte], 0, withSnapshotSave = true) }
```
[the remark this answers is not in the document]
KonradStaniec: good point, we probably do not need snapshots to revert past genesis block 👍

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
KonradStaniec: good point, we probably do not need snapshots to revert past genesis block 👍

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
KonradStaniec: I don't think it is too much, it would be beneficial to refactor it the way you say it. Especially that we already have tasks to refactor both fast and regular sync. I will leave task also for this refactoring.
During implementing this solution I was thinking about it but decided that last moment refactors are not best idea here.

KonradStaniec:
Also to not multiply numbers of pr-s unnecessary I have bumped `sbt-verify` plugin version in this pr as there is already correct docker image `rtkaczyk/mantis-circleci:v6`

PR #435, opened 2018-04-04
[cgkiele-143] testnet provisioning scripts & conf

loverdos:


loverdos:
This work is now tracked elsewhere.

PR #436, opened 2018-04-09
[CGKIELE-107] mallet

rtkaczyk:
The minimalist CLI wallet for KEVM/IELE testnet.

See [the list of missing features](https://iohk.myjetbrains.com/youtrack/issue/CGKIELE-107#comment=93-19538)

Additional improvements:
- a fix for `sbt-verify`: to regenerate checksums simply type `verifyGenerate`. This will create a file `verify.sbt` in the root dir - no need for manual copying of the output


build.sbt
```diff
@@ -143,7 +148,14 @@ val verifyDeps = Seq(
...
   "org.objenesis" % "objenesis" sha1 "272bab9a4e5994757044d1fc43ce480c8cb907a4",
-  "org.hamcrest" % "hamcrest-all" sha1 "63a21ebc981131004ad02e0434e799fd7f3a8d5a"
+  "org.hamcrest" % "hamcrest-all" sha1 "63a21ebc981131004ad02e0434e799fd7f3a8d5a",
+
+  // mallet
+  "com.github.scopt" % "scopt" sha1 "e078455e1a65597146f8608dab3247bf1eb92e6e",
+  "org.jline" % "jline" sha1 "dfb4e9e15e981634155ce063fa697b2b8964d507",
+  "org.scala-lang.modules" % "scala-parser-combinators" sha1 "3c1c5475ece77c41e18dd971f8f818c091e4961c",
+  // FIXME: WTF?
+  "com.google.j2objc" % "j2objc-annotations" sha1 "976d8d30bebc251db406f2bdb3eb01962b5685b3"
```
loverdos: I did a bit of searching. This is a dependency pulled in via guava `v22` (check image below) and is discussed [here](https://github.com/google/guava/issues/2824).


selection_108

rtkaczyk: What uses guava?

loverdos: Using `sbt` with `addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.0")`:
```
$ sbt whatDependsOn com.google.guava guava 22.0
[info] com.google.guava:guava:22.0
[info]   +-io.atomix:atomix-utils:2.1.0-beta1
[info]   | +-io.atomix:atomix-cluster:2.1.0-beta1
[info]   | [...]   
[info]   +-org.consensusresearch:scrypto_2.12:1.2.0-RC3 [S]
[info]   | [...]
[info]   +-org.scorexfoundation:iodb_2.12:0.3.0 [S]
[info]     
```

src/main/scala/io/iohk/ethereum/consensus/ethash/EthashMiner.scala
```diff
@@ -93,7 +93,7 @@ class EthashMiner(
         val startTime = System.currentTimeMillis()
         val mineResult = mine(headerHash, block.header.difficulty.toLong, dagSize, dag, miningConfig.mineRounds)
         val time = System.currentTimeMillis() - startTime
-        val hashRate = (mineResult.triedHashes * 1000) / time
+        val hashRate = if (time > 0) (mineResult.triedHashes * 1000) / time else Long.MaxValue
```
loverdos: I was wondering if we could use a "better" (=monotonic) clock, e.g. `System.nanoTime()` whose purpose is to measure elapsed time. There is no strong guarantee that this is indeed going to be monotonic because it ultimately depends on the OS facilities but this is generally not a problem in Linux nowadays. 

Another possible route is to not update the hash rate if we cannot reliably compute it.

loverdos: Note: guava has a `com.google.common.base.StopWatch` class for this purpose.

src/main/scala/io/iohk/ethereum/mallet/common/Util.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  def exceptionToString(ex: Throwable): String = {
+    val sw = new StringWriter()
+    sw.append(ex.getMessage + "\n")
+    ex.printStackTrace(new PrintWriter(sw))
+    sw.toString
+  }
+
+  implicit class OptionOps[T](opt: Option[T]) {
+    def toEither[U](left: U): Either[U, T] =
```
loverdos: There is `scala.Option.toRight`, so we can avoid introducing a new construct.

src/main/scala/io/iohk/ethereum/consensus/ethash/EthashMiner.scala
rtkaczyk: I'd rather not spend too much time on this, this is not really a big problem anyway. I can add a FIXME though

src/main/scala/io/iohk/ethereum/mallet/common/Util.scala
rtkaczyk: lol, how did I miss that? :)

build.sbt
rtkaczyk: @loverdos I removed the `j2objc-annotations` artifact from my local cache. After rebuilding, the checksum changed and it passed the verification. See also the updated PR description

build.sbt
```diff
@@ -161,15 +73,16 @@ val root = project.in(file("."))
...
+      verifyOutputFile in verifyGenerate := baseDirectory.value / "verify.sbt",
       verifyOptions in verify := VerifyOptions(
         includeBin = true,
         includeScala = true,
         includeDependency = true,
         excludedJars = Nil,
-        warnOnUnverifiedFiles = true,
-        warnOnUnusedVerifications = true
+        warnOnUnverifiedFiles = false,
+        warnOnUnusedVerifications = false
```
LukasGasior1: why this change? it seems keeping it true is more secure

src/main/scala/io/iohk/ethereum/mallet/interpreter/AST.scala
```diff
@@ -0,0 +1,39 @@
...
+  /**
+    * Base trait for literal types of arguments from user input.
+    * Note: these are different and require mapping to command parameter types
+    */
+  sealed trait Literal {
+    def input: String
+  }
+
+  case class Quoted(input: String) extends Literal {
+    def unqoute: String = unquote(input)
```
LukasGasior1: typo: unqoute -> unquote

src/universal/bin/mantis
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+./bin/mantis-app -Dconfig.file=./conf/mantis.conf -Dlogback.configurationFile=./conf/logback.xml -J-Xss10M
```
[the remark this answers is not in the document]
rtkaczyk: Good point, I was actually going to explore another way with those scripts, thanks for reminding me.

src/main/scala/io/iohk/ethereum/mallet/interpreter/Parameter.scala
```diff
@@ -0,0 +1,107 @@
...
+      case q: Quoted => ByteString(q.unqoute)
+    }
+  }
+
+  case object Addr20 extends ParamType {
+    type T = Address
+
+    protected val convert = {
+      case h: Hex if h.bytes.nonEmpty && h.bytes.length <= 20 =>
+        Address(h.bytes)
```
LukasGasior1: can we also support quoted addresses?

build.sbt
rtkaczyk: `false` means error instead of warning. I actually didn't plan on changing this, but it happened during some fixes I applied wrt `sbt-verify`. 

I'm not having any problems with `sbt-verify` now, so I wouldn't mind staying with `false`. WDYT?

src/main/scala/io/iohk/ethereum/mallet/interpreter/Parameter.scala
rtkaczyk: Why?

build.sbt
LukasGasior1: Makes sense, I didn't know it means an error :)

src/main/scala/io/iohk/ethereum/mallet/interpreter/Parameter.scala
LukasGasior1: My first intuition was to try with quotes. Also geth console uses quoted addresses (it doesn't even accept unquoted ones)

verify.sbt
```diff
@@ -0,0 +1,100 @@
...
+  "io.netty" % "netty-transport-native-unix-common" sha1 "fee46a4e2b4f2f096532ca443f4f63e08b205318",
+  "io.atomix" % "atomix-raft" sha1 "e45af08b70220dd010d6f193e3d38147b4be35fe",
+  "io.atomix" % "atomix-primitive" sha1 "73d9bc7c859856832178afd271f1a573e6a7c7e6",
+  "io.atomix" % "atomix-storage" sha1 "136f0b221acbc2680f099b8ff3a34f8cc1592fe7",
+  "io.atomix" % "atomix-primary-backup" sha1 "1c895965e3e67a152ffbccb4283b6cee91b4ea61",
+  "io.netty" % "netty-tcnative-boringssl-static" sha1 "ff5f2d6db5aaa1b4df1b381382cd6581844aad9d",
+  "com.github.scopt" % "scopt" sha1 "e078455e1a65597146f8608dab3247bf1eb92e6e"
+)
+
+verifyOptions in verify := VerifyOptions(
```
LukasGasior1: is this needed here? how does it relate to the options in build.sbt?

rtkaczyk: This file is auto-generated, see the PR description

LukasGasior1:
One thing I noticed: we probably don't want to store the `importPrivateKey` command in the history

rtkaczyk:
> importPrivateKey command in the history

That's a really good point!

src/main/scala/io/iohk/ethereum/mallet/interpreter/Parameter.scala
rtkaczyk: I guess it would be easy to do, but for now I wouldn't go further than adding a `TODO` comment to consider this.
Firstly, I'd rather have the syntax be strict, and secondly there's supporting account aliases (which would be quoted) on the list missing features.

LukasGasior1:
Another one: `-d` flag does not work:
```
ubuntu@ip-10-0-1-82:~/mantis-faucet/mantis-1.0-daedalus-rc1$ ./bin/mallet -d "/home/ubuntu/keystore" "http://10.0.1.82:8546"                                                                                                               
[residual] arg = '/home/ubuntu/keystore'                                                                                                                                                                                                  
[residual] arg = 'http://10.0.1.82:8546'
# Executing command line:
java
-Dconfig.file=./conf/mantis.conf                                                                                                                                                                                                          
-Dlogback.configurationFile=./conf/logback.xml                                                                                                                                                                                            
-Xss10M
-Dconfig.file=./conf/mallet.conf                                                                                                                                                                                                          
-cp                                                                                                                                                                                                                                       

-jar                                                                                                                                                                                                                                      
/home/ubuntu/mantis-faucet/mantis-1.0-daedalus-rc1/lib/mantis.mantis-1.0-daedalus-rc1-launcher.jar                                                                                                                                        
mallet                                                                                                                                                                                                                                    
/home/ubuntu/keystore
http://10.0.1.82:8546

Error: node URI scheme must be explicit and either 'http' or 'https'                                                                                                                                                                      
Error: Unknown argument 'http://10.0.1.82:8546'                                                                                                                                                                                           
Try --help for more information.                               
```

full `--data-dir` works just fine

rtkaczyk:
@LukasGasior1 `-d` was conflicting with the launcher's script debug option, but adding `--` fixed the issue.

Also on the subject, note there are no longer separate `mantis` and `mantis-app` scripts. Turns out that adding `-Dconfig.file=...` in the mallet script overrides the defaults from `application.ini` (which was the original problem that the hack tried to solve)

In other news:
- `importPrivateKey` is no longer saved in history
- absence of a receipt is correctly handled
- receipts are displayed as formatted JSON

src/main/scala/io/iohk/ethereum/mallet/main/OptionParser.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  private implicit val addrRead: scopt.Read[Address] =
+    scopt.Read.reads { str =>
+      val bytes = StringUtil.hexToBytes(str)
+      require(bytes.nonEmpty && bytes.length <= Address.Length, "invalid address length")
+      Address(bytes)
+    }
+
+  def apply(args: Seq[String]): Option[ClOptions] = {
+    val parser = new scopt.OptionParser[ClOptions]("scopt") {
```
LukasGasior1: super minor: the string argument here is a `programName`, guess it should be `Constants.AppName` rather than `"scopt"`

rtkaczyk: Not that minor, good catch

src/main/scala/io/iohk/ethereum/mallet/service/RpcClient.scala
```diff
@@ -0,0 +1,97 @@
...
+
+import scala.concurrent.duration._
+import scala.concurrent.{Await, Future}
+import scala.util.{Failure, Success, Try}
+
+object RpcClient {
+  // TODO: CL option to enable akka logging
+  private val akkaConfig = ConfigFactory.load("mallet")
+
+  implicit val actorSystem = ActorSystem("mallet_rpc", akkaConfig)
```
LukasGasior1: why not instantiate it in Mallet app and pass to RpcClient as a parameter?

src/main/scala/io/iohk/ethereum/mallet/main/ConstPasswordReader.scala
```diff
@@ -0,0 +1,15 @@
+package io.iohk.ethereum.mallet.main
+
+import io.iohk.ethereum.mallet.service.PasswordReader
+
+/**
+  * This implementation of [[PasswordReader]] is in non-interactive mode when the password is
```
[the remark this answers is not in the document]
rtkaczyk: Isn't it a bug? The `PasswordReader` is imported. See [this scaladoc](https://github.com/input-output-hk/mantis/blob/phase/iele_testnet/src/main/scala/io/iohk/ethereum/domain/Blockchain.scala#L28)

build.sbt
```diff
@@ -186,7 +96,6 @@ scalacOptions := Seq(
   "-unchecked",
   "-deprecation",
   "-feature",
-  "-Xfatal-warnings",
   "-Xlint:unsound-match",
```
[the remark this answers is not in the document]
rtkaczyk: I can re-enable it. I always disable it when working on something complicated, because of unused imports. I'm adding and removing code back and forth, and going to imports for each such change kills my concentrations and infuriates me.

src/main/scala/io/iohk/ethereum/mallet/service/RpcClient.scala
rtkaczyk: I meant to avoid `ActorSystem` leaking out of `RpcClient` cause it's an internal detail. Obviously this is not the case now, because I forgot to add `shutdown()` method to `RpcClient` - then the actor system could be private to `RpcClient`. What do you think about  such approach?

LukasGasior1: To me, actor system is not really internal to rpc client. Even though for mallet it's only used here, it has an impact on the whole app (for example, keeps it alive until it's shutdown), that's why I'd opt for putting in on app-level and pass (perhaps implicitly) to rpc client. It would also make RpcClient's dependencies more clear as well as make it reusable.

LukasGasior1: also, is there any reason why we're using the `actorSystem.dispatcher` instead of global EC or, even better, implicit ec passed as a constructor param?

rtkaczyk: What's wrong with using `dispatcher`?

LukasGasior1: since you're using blocking operation (await), doesn't that mean that one of the threads that akka uses internally will be blocked? I know, in this case it doesn't matter since RpcClient is the only component using actors, but for the reasons mentioned above (reusability) I think it would be better to let the user of this class choose which EC he prefers (or, alternatively use global ec that won't block system dispatcher's thread).

src/main/scala/io/iohk/ethereum/mallet/main/ConstPasswordReader.scala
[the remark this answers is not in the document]
rtkaczyk: @loverdos wdyt?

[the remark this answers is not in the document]
loverdos: It works only if the referenced thing is in the same package, not if it is imported. I do not know if this is intended behaviour or a bug though. So I just use the FQN on these occasions.

src/main/resources/mallet.conf
```diff
@@ -0,0 +1,28 @@
+akka {
+  # to enable logging use: ["akka.event.slf4j.Slf4jLogger"]
+  loggers = []
+
+  loglevel = OFF
+
+  ssl-config {
+    trustManager = {
+      stores = [
+        # TODO: move to Wiki maybe?
```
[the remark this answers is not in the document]
rtkaczyk: As discussed, it will be done in #445

build.sbt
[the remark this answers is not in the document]
loverdos: There is a keystroke in IntelliJ that takes care of unused imports :)

src/main/scala/io/iohk/ethereum/mallet/service/RpcClient.scala
```diff
@@ -0,0 +1,108 @@
...
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.util.{Failure, Success, Try}
+
+object RpcClient {
+
+  /**
+    * This factory method is defining an ActorSystem, ActorMaterializer and ExecutionContext for
+    * the [[RpcClient]]. To customize these dependencies use [[RpcClient]]'s constructor
+    */
+  def apply(node: Uri): RpcClient = {
```
rtkaczyk: @LukasGasior1 hopefully this should address your remarks about reusability, while limiting the scope of dependencies in mallet.

LukasGasior1: ok, let's leave it that way

PR #437, opened 2018-04-12
[EC-522] Make fastSync more resisilent to failures

KonradStaniec:
Handling of two cases:
1. Handle the case when we are constantly receiving 0 headers from one peer. As we are choosing our blockheader peer as first free peer, it could happen that we will ask headers from a peer which constantly send us 0, it can greatly impair our syncing.
2. Handle case when peer will send us header from the future e.g our best block is `10` and we requested blocks from `11` to `20`, and peer sends us block/blocks which start from block `5000`.  Of course, this block will fail validation, but it will change our best block to `5000 - N - 1`. Currently `N` is equal to `2048` so our best block number would be `2951` which is obviously incorrect.
We are always asking our peers for headers from `bestBlock + 1`, so if we receive some headers from beyond that we can assume something is wrong: maybe malicious peer, maybe some error on another peer side, or we have some hidden bug in our code (i could not find any suspicious)

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
@@ -132,10 +132,10 @@ class FastSync(
         requestedHeaders.get(peer).foreach{ requestedNum =>
           removeRequestHandler(sender())
           requestedHeaders -= peer
-          if (blockHeaders.size <= requestedNum)
+          if (blockHeaders.nonEmpty && blockHeaders.size <= requestedNum && blockHeaders.head.number == syncState.bestBlockHeaderNumber + 1)
             handleBlockHeaders(peer, blockHeaders)
           else
-            blacklist(peer.id, blacklistDuration, "wrong number of headers in response")
+            blacklist(peer.id, blacklistDuration, "wrong blockheaders response (empty or not chain forming)")
         }
 
       case ResponseReceived(peer, BlockBodies(blockBodies), timeTaken) =>
--- a/src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
+++ b/src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
@@ -18,7 +18,7 @@ import io.iohk.ethereum.network.p2p.messages.PV62.{BlockBody, _}
 import io.iohk.ethereum.network.p2p.messages.PV63.{GetNodeData, GetReceipts, NodeData, Receipts}
 import io.iohk.ethereum.network.{EtcPeerManagerActor, Peer}
 import io.iohk.ethereum.utils.Config.SyncConfig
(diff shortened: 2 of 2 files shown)


src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
```diff
@@ -151,19 +151,54 @@ class SyncControllerSpec extends FlatSpec with Matchers with BeforeAndAfter with
...
+    val targetNumber = 200000
+    val bestNumber = targetNumber - 10
+    val startState = defaultState.copy(
+      targetBlock = baseBlockHeader.copy(number = targetNumber ),
+      bestBlockHeaderNumber = bestNumber,
+      safeDownloadTarget = targetNumber + 10
+    )
+
+    startWithState(startState)
+    Thread.sleep(1.seconds.toMillis)
```
vargasmontero: Why do we sleep 1s here and 2s below?

vargasmontero:
I'm testing the code changes with a new sync. It's half way through. Should be ready for tomorrow.

jtownson:
I tested this code by running 
* fast sync from blocks 5 million until switching to regular sync and then to completion of regular sync (~5,700,000 blocks).
* fast sync from scratch for a few thousand blocks (in case the addition of `bestBlock + 1` caused some form of out-by-one-error).

No problems encountered.

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
KonradStaniec: Actually when I used 1s,  fastsync was not quick enough (at least one time) to ask for the headers before I sent this future block. (it was probably one-time flicker but I tend to be pretty conservative in case of this `sleeps`, especially as they tend to fail especially on circleci)

KonradStaniec:
just 2 cents from my testing, over the weekend, synced to top 2 times. No problems encountered.

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -132,10 +132,10 @@ class FastSync(
         requestedHeaders.get(peer).foreach{ requestedNum =>
           removeRequestHandler(sender())
           requestedHeaders -= peer
-          if (blockHeaders.size <= requestedNum)
+          if (blockHeaders.nonEmpty && blockHeaders.size <= requestedNum && blockHeaders.head.number == syncState.bestBlockHeaderNumber + 1)
```
ntallar: Minor detail: What do you think of calling `checkHeadersChain` here instead of in `handleBlockHeaders`, so that all the validations are done in a single place? (or the other way around)

KonradStaniec: Actually it is a good idea, but I would abstain from it for now as there is separate refactor ticket (https://iohk.myjetbrains.com/youtrack/issue/EC-524) and I would like to minimise code changes for now

PR #439, opened 2018-04-16
Add faucet

LukasGasior1:
Depends on #436


rtkaczyk:
Other than the dependency, is it ready to review and test? Could you show some example requests?

LukasGasior1:
@rtkaczyk not really yet. 1) it depends on mallet PR 2) I'm adding one more functionality (limiting the number of requests)

LukasGasior1:
the request is just a simple POST to http://host:port/faucet?address=0x...

rtkaczyk:
In that case shouldn't it be a GET? We could then simply paste the URL to the browser. Also, we make a request to _get_ some funds ;)

LukasGasior1:
But you _post_ the request to get the funds, you get them as a result. Same way you _post_ `/ticket` to _get_ (i.e create) a new ticket. To me post seems more appropriate here

LukasGasior1:
I think you're confusing two meanings of _get_, you get the funds - yes, but you don't _get_ anything in http response

rtkaczyk:
I was more concerned with practicality. If we do a GET then we can just say "here's your faucet link, click it". But it's just a suggestion, technically you are correct, POST is the appropriate method for this kind of interaction.

src/main/scala/io/iohk/ethereum/faucet/Faucet.scala
```diff
@@ -0,0 +1,35 @@
...
+import io.iohk.ethereum.mallet.service.RpcClient
+import io.iohk.ethereum.utils.{Config, Logger}
+import scala.concurrent.ExecutionContext.Implicits.global
+
+import scala.util.{Failure, Success}
+
+object Faucet extends Logger {
+
+  def main(args: Array[String]): Unit = {
+    val config = FaucetConfig(Config.config)
```
rtkaczyk: This makes faucet config dependent on mantis config. 
I suggest `FaucetConfig(ConfigFactory.load())` and `-Dconfig.file=./conf/faucet.conf` in `universal/bin/faucet-server`. Then we don't have to modify `mantis.conf` when running.

src/main/resources/application.conf
```diff
@@ -548,6 +548,44 @@ mantis {
     }
   }
 
+  faucet {
```
rtkaczyk: I suggest moving this outside of mantis section (as in `universal/conf/faucet.conf`) and into a separate file

src/main/resources/application.conf
```diff
@@ -548,6 +548,44 @@ mantis {
...
+    # Wallet address used to send transactions from
+    wallet-address = "0x00"
+
+    # Password to unlock faucet wallet
+    wallet-password = ""
+
+    # Path to directory where wallet key is stored
+    keystore-dir = "~/.mantis-faucet-keystore"
```
rtkaczyk: `~` won't work here

src/main/scala/io/iohk/ethereum/faucet/FaucetApi.scala
```diff
@@ -0,0 +1,83 @@
...
+      latestRequestTimestamps.put(clientAddr, timeMillis)
+
+      val res = for {
+        nonce <- rpcClient.getNonce(wallet.address)
+        txId <- rpcClient.sendTransaction(prepareTx(Address(targetAddress), nonce))
+      } yield txId
+
+      res match {
+        case Right(txId) =>
+          complete(StatusCodes.OK, s"0x${Hex.toHexString(txId.toArray[Byte])}")
```
rtkaczyk: Let's  add logging for the successful case, too. It will be interesting to see how often this is used (include IP and ETH addresses and transfer value)

src/main/scala/io/iohk/ethereum/faucet/FaucetApi.scala
```diff
@@ -0,0 +1,83 @@
...
+import io.iohk.ethereum.rlp
+import io.iohk.ethereum.network.p2p.messages.CommonMessages.SignedTransactions.SignedTransactionEnc
+import org.spongycastle.util.encoders.Hex
+
+class FaucetApi(
+    rpcClient: RpcClient,
+    keyStore: KeyStore,
+    config: FaucetConfig,
+    clock: Clock = Clock.systemUTC())
+  extends Json4sSupport
```
rtkaczyk: Is it used?

src/main/scala/io/iohk/ethereum/faucet/Faucet.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.util.{Failure, Success}
+
+object Faucet extends Logger {
+
+  def main(args: Array[String]): Unit = {
+    val config = FaucetConfig(ConfigFactory.load())
+
+    val rpcClient = new RpcClient(config.rpcAddress)
```
rtkaczyk: Does not compile (system, materializer, ec)

LukasGasior1:
@rtkaczyk logging added

PR #441, opened 2018-04-16
Add universal configuration for a custom private testnet

loverdos:


rtkaczyk:
Note: this should be merged after #442 and include the following settings:
```
atomix-raft.block-forging-delay = 15.seconds
sync.blacklist-duration = 0
sync.broadcast-new-block-hashes = false
```

src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftConfig.scala
```diff
@@ -45,8 +45,10 @@ object AtomixRaftConfig extends Logger {
       case _ ⇒ throw new IllegalArgumentException("parts.length != 1, 2, 3")
     }
 
+  //noinspection ScalaStyle
   def parseNode(id_host_port: String): AtomixNode = {
     val parts = id_host_port.split(":")
+    log.info(s"atomix-raft: id_host_port = ${id_host_port} => parts = ${parts.toList}")
```
rtkaczyk: Minor: not sure if it's the best place for such logging, as a reader might expect this to be a pure function and want to use it in different context.

src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftConfig.scala
```diff
@@ -45,8 +45,10 @@ object AtomixRaftConfig extends Logger {
       case _ ⇒ throw new IllegalArgumentException("parts.length != 1, 2, 3")
     }
 
+  //noinspection ScalaStyle
```
rtkaczyk: Still needed?

src/universal/conf/private-testnet.conf
```diff
@@ -0,0 +1,74 @@
...
+    # Custom genesis JSON file
+    # null value indicates using default genesis definition that matches the main network
+    custom-genesis-file = "conf/private-genesis.json"
+  }
+
+  network {
+    discovery {
+      # Listening interface for discovery protocol
+      interface = ${mantis.network.server-address.interface}
+    }
```
rtkaczyk: I think we should disable discovery and override bootstrap nodes

src/universal/conf/private-testnet.conf
```diff
@@ -0,0 +1,74 @@
...
+      # Choosing https requires creating a certificate and setting up 'certificate-keystore-path' and
+      # 'certificate-password-file'
+      # See: https://github.com/input-output-hk/mantis/wiki/Creating-self-signed-certificate-for-using-JSON-RPC-with-HTTPS
+      mode = "http"
+
+      # Listening address of JSON-RPC HTTP/HTTPS endpoint
+      interface = ${mantis.network.server-address.interface}
+
+      # Domains allowed to query RPC endpoint. Use "*" to enable requests from any domain.
+      cors-allowed-origins = []
```
rtkaczyk: I wonder if we should set this to `"*"`

loverdos: Yes, I changed this to match exactly what we do now in the testnet, which is actually the primary motivation for this PR.

src/universal/conf/private-testnet.conf
```diff
@@ -0,0 +1,88 @@
...
+
+    # Custom genesis JSON file
+    # null value indicates using default genesis definition that matches the main network
+    custom-genesis-file = "conf/private-genesis.json"
+  }
+
+  network {
+    discovery {
+      # Turn discovery of/off
+      # discovery-enabled = true
```
rtkaczyk: So, don't we want that disabled?

src/universal/conf/private-testnet.conf
```diff
@@ -0,0 +1,88 @@
...
+      # Listening interface for discovery protocol
+      interface = ${mantis.network.server-address.interface}
+
+      # Listening port for discovery protocol
+      # port = 30303
+
+      # Set of initial nodes
+      bootstrap-nodes = [
+        "enode://18a551bee469c2e02de660ab01dede06503c986f6b8520cb5a65ad122df88b17b285e3fef09a40a0d44f99e014f8616cf1ebc2e094f96c6e09e2f390f5d34857@127.0.0.1:30303",
+      ]
```
rtkaczyk: Minor, as it will have to be modified anyway, but wouldn't we prefer empty list or even:

```
bootstrap-nodes = [
  # fill in with node URI for the private network, e.g.:
  # "enode://18a551bee469c2e02de660ab01dede06503c986f6b8520cb5a65ad122df88b17b285e3fef09a40a0d44f99e014f8616cf1ebc2e094f96c6e09e2f390f5d34857@127.0.0.1:30303"
]
```

PR #442, opened 2018-04-17
[CGKIELE-154] stabilize sync for iele testnet

rtkaczyk:


src/main/scala/io/iohk/ethereum/blockchain/sync/BlockBroadcast.scala
```diff
@@ -25,7 +26,11 @@ class BlockBroadcast(val etcPeerManager: ActorRef) {
       case (peer, peerInfo) if shouldSendNewBlock(newBlock, peerInfo) => peer }.toSet
 
     broadcastNewBlock(newBlock, peersWithoutBlock)
-    broadcastNewBlockHash(newBlock, peersWithoutBlock)
+
+    if (syncConfig.broadcastNewBlockHashes) {
+      // the usefulness of this message is debatable, especially in private networks
```
loverdos: Since we may need to make a decision in the future, can we start the comment as `// NOTE` or `// Note`, so that it is easily `grep`-able?

LukasGasior1: Isn't it used to inform other peers about the new blocks we have? Why do you think it's not needed?

LukasGasior1: (I think what Christos means is adding a "Note" prefix to comments)

loverdos: Yes, adding a Note prefix to comments. But no stopper here.

KonradStaniec: @rtkaczyk `newBlocks` message is only sent to a subset of peers (if I remember correctly `sqrt(n)` peers, where `n ` is the number of current peers), `newBlockHashes` is sent to all peers. 
And main reason for both of them to exist is that `newBlockHashes` is much more lightweight  and not load peer network that much (`newBlockHashes`  is only a hash, and then need to download block directly, `newBlocks` is the whole new block )

rtkaczyk: I've added the "NOTE" just in case, but I still don't think there's anything to decide here

rtkaczyk: The main problems with `NewBlockHashes`: 
- it's sent to the same peers that receive `NewBlock`, which is redundant
- it's handled differently than `NewBlock`, where orphaned block is ignored
- `NewBlock` should be completely sufficient, because nodes that received the message _should_ forward it to the ones that didn't

PR #451, opened 2018-05-10
[CGKIELE-187] node private key generation tool

rtkaczyk:
ECDSA keygen tool - as per request from DevOps


loverdos:
Just saw there is a merge conflict ...

src/main/resources/application.conf
```diff
@@ -9,7 +9,7 @@ mantis {
   datadir = ${user.home}"/.mantis"
 
   # The unencrypted private key of this node
-  node-key-file = ${mantis.datadir}"/nodeId.keys"
+  node-key-file = ${mantis.datadir}"/node.key"
```
LukasGasior1: it may be worth to update this in .gitignore as well

rtkaczyk: I don't see why we'd need that file it `.gitignore`. Must be a very old entry. Removed

PR #461, opened 2018-05-18
[CKGIELE-189] synchronise extvm / constant block gas limit

rtkaczyk:
- entry point to extvm is `synchronized` - this fixes `eth_call`, `eth_estimateGas` and normal transactions while on `ethash`
- configurable constant block gas limit
- `BlockchainConfig` as case class

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -383,6 +383,10 @@ mantis {
     # if true, account storage will use Ethereum-specific format for storing keys/value in MPT (32 byte)
     # if false, generic storage for arbitrary length integers will be used
     eth-compatible-storage = true
+
+    # true: gas limit is constant for all blocks in the blockchain and equal to genesis block gas limit
+    # false: the default Ethereum formula will be applied (based on genesis block)
+    constant-block-gas-limit = false
   }
 
   sync {
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -384,9 +384,9 @@ mantis {
     # if false, generic storage for arbitrary length integers will be used
     eth-compatible-storage = true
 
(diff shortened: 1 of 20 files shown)


src/main/resources/application.conf
```diff
@@ -383,6 +383,10 @@ mantis {
     # if true, account storage will use Ethereum-specific format for storing keys/value in MPT (32 byte)
     # if false, generic storage for arbitrary length integers will be used
     eth-compatible-storage = true
+
+    # true: gas limit is constant for all blocks in the blockchain and equal to genesis block gas limit
+    # false: the default Ethereum formula will be applied (based on genesis block)
+    constant-block-gas-limit = false
```
rtkaczyk: After some thought, I think it will be better if we specify the number here, rather than basing on genesis block gas limit (`null` for std Ethereum formula). That way we can adjust the limit without discarding the whole blockchain.

src/main/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorSkeleton.scala
```diff
@@ -143,12 +144,15 @@ abstract class BlockGeneratorSkeleton(
...
 
   //returns maximal limit to be able to include as many transactions as possible
-  protected def calculateGasLimit(parentGas: BigInt): BigInt = {
-    val GasLimitBoundDivisor: Int = 1024
-
-    val gasLimitDifference = parentGas / GasLimitBoundDivisor
-    parentGas + gasLimitDifference - 1
-  }
+  protected def calculateGasLimit(parentGas: BigInt): BigInt =
+    blockchainConfig.constantBlockGasLimit match {
```
LukasGasior1: minor: you can use `.getOrElse` here

PR #462, opened 2018-05-21
modify sbt verify version to 0.4.1

Lbqds:
* fix sbt.ResolveException because of sbt-verify verison changed to 0.4.1 [here](https://github.com/input-output-hk/sbt-verify/pull/3)
* remove dup scalacheck dependency


mirkoAlic:
[TOO OLD]

PR #466, opened 2018-05-24
CGKIELE-168: Add more metrics

loverdos:
Also:
 - Fix a bug with null `environment` and `deployment`
 - Require non-empty values for the above when `enabled = true`
 - Improve documentation in configuration


src/main/scala/io/iohk/ethereum/metrics/MetricsConfig.scala
```diff
@@ -38,6 +38,16 @@ object MetricsConfig {
     val environment = config.getString(Keys.Environment)
     val deployment = config.getString(Keys.Deployment)
 
+    if(enabled) {
```
rtkaczyk: But if it's disable, then not specifying `environment` or `deployment` will still throw an exception, right?

PR #472, opened 2018-06-01
Simplifying the SBT invocations in the nix derivations.

krisajenkins:


jobsets/mantis.nix
```diff
@@ -7,13 +7,17 @@ stdenv.mkDerivation {
...
   buildInputs = [ scala sbt sbtVerify unzip ];
 
+  configurePhase = ''
+    export HOME="$NIX_BUILD_TOP"
+  '';
+
   buildPhase = ''
-    cp -r ${sbtVerify}/.ivy .
-    cp -r ${sbtVerify}/.sbt .
+    cp -r ${sbtVerify}/.ivy $HOME
```
loverdos: It should be `.ivy2`

krisajenkins: Ah, thanks @loverdos!

Actually, now I look at this, it can come out entirely. It was a caching-convenience that I needed when building on my laptop. But now it's on the server the speed doesn't matter as much as the recipe being clear. I'll take it out. :-)

PR #476, opened 2018-06-07
Use `iele` in the version and remove `daedalus`

loverdos:
Ref CGKIELE-303


loverdos:
Need to consider many more places, e.g. https://github.com/input-output-hk/mantis/blob/973e0c8088ed692776ef92e11b008477b8390b7e/jobsets/mantis.nix#L23

loverdos:
Need to re-consider how we do this.

PR #485, opened 2018-06-12
Update libraries

mmrozek:
Update libraries to newest versions.
I had to use AtomicReferance instead of akka agents because agents are already deprecated and will be removed from akka shortly.
There are also some cosmetic changes caused by updating akka testkit.


src/main/scala/io/iohk/ethereum/faucet/FaucetApi.scala
```diff
@@ -31,9 +31,9 @@ class FaucetApi(
     case Left(err) => throw new RuntimeException(s"Cannot unlock wallet for use in faucet (${config.walletAddress}), because of $err")
   }
 
-  private val corsSettings = CorsSettings.defaultSettings.copy(
-    allowGenericHttpRequests = true,
-    allowedOrigins = config.corsAllowedOrigins)
+  private val corsSettings = CorsSettings.defaultSettings.
+    withAllowGenericHttpRequests(true).
```
agnikowa: weir formatting, I think it should start with a dot not end with it

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -29,10 +29,9 @@ trait JsonRpcHttpServer extends Json4sSupport {
 
   def corsAllowedOrigins: HttpOriginRange
 
-  val corsSettings = CorsSettings.defaultSettings.copy(
-    allowGenericHttpRequests = true,
-    allowedOrigins = corsAllowedOrigins
-  )
+  val corsSettings = CorsSettings.defaultSettings.
+    withAllowGenericHttpRequests(true).
```
agnikowa: also formatting think here

src/main/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManager.scala
```diff
@@ -106,7 +104,7 @@ class PeerDiscoveryManager(
     }
   }
 
-  private def getTcpPort: Int = nodeStatusHolder().serverStatus match {
+  private def getTcpPort: Int = nodeStatusHolder.get().serverStatus match {
```
agnikowa: I don't know much about `AtomicReference ` but I'm curious if this `get()` returned in all calls in this class is the same or not? If first maybe we should think about extracting it to value to get rig of all this gets. If later skip this comment

mmrozek: AtomicReference works as every Atomic* class in java. Under the hood there is a synchronisation on the reference. We shouldn't extract the value because we will lose atomicity.

KonradStaniec:
generally pr looks good, but: 
1.  Circle Ci not works, probably because we have switched off building from forks. I do not know if i have privilages to turn it on, and also https://circleci.com/docs/1.0/fork-pr-builds/ mentions that it could by some security implications so i am relcutant to mingle with it. Wdyt @mcsherrylabs ?
2. Your commits are unverified, so maybe you did not configure it with github somehow ?

mmrozek:
I add PGP key to my github account so now commits should be verified

mmrozek:
@KonradStaniec please check one more time. CI checks have passed.

mmrozek:
I move the code from fork to local branch because of issues with ci.

PR #486, opened 2018-06-12
Update receipt to include status and returnData fields

LukasGasior1:


src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
```diff
@@ -254,14 +255,26 @@ class BlockPreparator(
...
           case Right(_) =>
-            val TxResult(newWorld, gasUsed, logs, _, _) = executeTransaction(stx, blockHeader, worldForTx)
+            val TxResult(newWorld, gasUsed, logs, rd, vmError) = executeTransaction(stx, blockHeader, worldForTx)
+
+            val (status, returnData) =
+              if (blockchainConfig.ethCompatibilityMode) (None, None)
+              else (
+                Some(vmError match {
+                  case Some(WithReturnCode(returnCode)) => returnCode
+                  case Some(OutOfGas) => ByteString(0x05)
```
loverdos: Can we introduce a few constants with meaningful names, instead of using these magic numbers `0x05` etc?

src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
```diff
@@ -254,14 +255,26 @@ class BlockPreparator(
...
+                  case Some(_) => ByteString(0x04)
```
loverdos: Same as above

src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
```diff
@@ -254,14 +255,26 @@ class BlockPreparator(
...
+                  case None => ByteString(0x00)
```
loverdos: and here ...

LukasGasior1:
> As as comment in the PR description (or in the relevant ticket) we need to know how the PR affects the deployment (configuration changes, can it be a rolling deployment?, other areas that we need to be careful about).

Good point. What has changed in this regard is the config: `eth-compatible-storage` it has been replaced with more generic `eth-compatibility-mode`, which for IELE testnet should be set to false.

PR #490, opened 2018-06-14
Feature/eip 198

KonradStaniec:
One of new precompiled contracts from Byzantium fork

Spec: 
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-198.md

Left to do:
Currently protobuf dont support `byzatiumBlockNumber` field so it is hardcoded in necessery places. After inclusion to protobuf it will need to be changed 

ets test are located in `ets/BlockchainTests/GeneralStateTests/stPreCompiledContracts` dir , to run it you need to enable vm in local mode, also currently you need to turn off receipt validation after transaction execution otherwise tests will fail becouse current off lack of https://github.com/ethereum/EIPs/blob/master/EIPS/eip-658.md. Also two tests are failing becouse of lack https://github.com/ethereum/EIPs/blob/master/EIPS/eip-214.md. 
Otherwise 809 out of 811 tests are passing.


src/main/scala/io/iohk/ethereum/vm/PrecompiledContracts.scala
```diff
@@ -93,23 +103,118 @@ object PrecompiledContracts {
...
+          safeAdd(totalLengthBytes, baseLength),
+          safeAdd(safeAdd(totalLengthBytes, baseLength), expLength))
+
+      val multComplexity = getMultComplexity(math.max(baseLength, modLength))
+
+      val adjExpLen = adjExpLength(expBytes, expLength)
+
+      val gasCost = multComplexity * math.max(adjExpLen , 1) / GQUADDIVISOR
+
+      gasCost
```
agnikowa: maybe consider removing this assignment and just have last line as `multiComplexity * math.max(adjExpLen , 1) / GQUADDIVISOR` and function renamed ad `gasCost` or something

src/main/scala/io/iohk/ethereum/vm/PrecompiledContracts.scala
```diff
@@ -93,23 +103,118 @@ object PrecompiledContracts {
...
+
+    private def adjExpLength(expBytes: ByteString, expLength: Int): Long = {
+      val expHead =
+        if (expLength <= lengthBytes)
+          expBytes.padTo(expLength, 0.toByte)
+        else
+          expBytes.take(lengthBytes).padTo(lengthBytes, 0.toByte)
+
+
+      val msb = math.max(ByteUtils.toBigInt(expHead).bitLength - 1, 0)
```
agnikowa: maybe use some more specified naming instead of `msb`

src/main/scala/io/iohk/ethereum/vm/PrecompiledContracts.scala
```diff
@@ -93,23 +103,118 @@ object PrecompiledContracts {
...
+
+    private def getMultComplexity(x: BigInt): BigInt = {
+      val x2 = x * x
+
+      if (x <= 64)
+        x2
+      else if (x <= 1024)
+        x2 / 4 + 96 * x - 3072
+      else
+        x2 / 16 + 480 * x - 199680
```
agnikowa: maybe leave a comment why this specific numbers are used

KonradStaniec: hmm i will leave a link to the spec on top as all is defined there

src/test/scala/io/iohk/ethereum/vm/PrecompiledContractsSpec.scala
```diff
@@ -128,4 +131,33 @@ class PrecompiledContractsSpec extends FunSuite with Matchers with PropertyCheck
     }
   }
 
+  test("MODEXP") {
```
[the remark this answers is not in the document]
KonradStaniec: I have added comment, data are copied form geth , and i assume geth have copied them from ets tests to have some subset run without any other noise of trasaction execution

PR #491, opened 2018-06-18
Feature/eip 649

agnikowa:
needed changes described in: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-649.md


src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/EthashDifficultyCalculator.scala
```diff
@@ -5,27 +5,37 @@ import io.iohk.ethereum.domain.BlockHeader
...
       } else {
         val timestampDiff = blockTimestamp - parentHeader.unixTimestamp
         math.max(1 - timestampDiff / 10, FrontierTimestampDiffLimit)
       }
 
     val extraDifficulty: BigInt =
-      if (blockNumber < difficultyBombRemovalBlockNumber) {
-        val difficultyBombExponent = calculateBombExponent(blockNumber)
+      if (fakeBlockNumber < difficultyBombRemovalBlockNumber) {
+        val difficultyBombExponent = calculateBombExponent(fakeBlockNumber)
```
KonradStaniec: Spec is not clear here by sayin `simply replace the use of block.number, as used in the exponential ice age component`, but i think fakeBlock number should be used only here (as paramater to `calculateBombExponent`). 
example: https://github.com/ethereum/ethereumj/pull/927/files#diff-e6100521843ae5e9fb95013c150f2129R72

src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
```diff
@@ -43,17 +43,32 @@ class BlockPreparator(
...
 
+    val blockNumber = block.header.number
+    // condition to calculate block rewards https://github.com/ethereum/EIPs/blob/master/EIPS/eip-649.md
+    val shouldReduceBlockReward = blockNumber >= blockchainConfig.byzantiumBlockNumber
+
     val minerAddress = Address(block.header.beneficiary)
     val minerAccount = getAccountToPay(minerAddress, worldStateProxy)
-    val minerReward = blockRewardCalculator.calcBlockMinerReward(block.header.number, block.body.uncleNodesList.size)
+    val minerReward = if (shouldReduceBlockReward) {
+      BigInt(10).pow(18) * 3
```
KonradStaniec: I think all changes should be localized to class `BlockRewardCalculator`.
Specs are sayin that: `This is the existing pre-Metropolis formula for uncle rewards, simply adjusted with new_block_reward` , so i think all our current formulas for rewards in https://github.com/input-output-hk/mantis/blob/feature/byzantium_fork/src/main/scala/io/iohk/ethereum/ledger/BlockRewardCalculator.scala are correct, and only change we would need to do there is to change value https://github.com/input-output-hk/mantis/blob/feature/byzantium_fork/src/main/scala/io/iohk/ethereum/ledger/BlockRewardCalculator.scala#L19 to `3,000,000,000,000,000,000` if we are past byzantium block. (it can be done here, or mabe to config passed to calculator can be updated if we are past byzantium block)

src/main/resources/application.conf
```diff
@@ -337,6 +337,10 @@ mantis {
     # Doc: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-161.md
     eip161-block-number = "1000000000000000000"
 
+    # EIP-609 fork block number (ETH Only)
+    # Doc: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-609.md
+    byzantium-block-number = "1000000000000000000"
```
KonradStaniec: It is already in config ( https://github.com/input-output-hk/mantis/pull/491/files#diff-5eab1f3fccda52c11d4e4dbdd0959523R363)

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/EthashDifficultyCalculator.scala
```diff
@@ -5,15 +5,21 @@ import io.iohk.ethereum.domain.BlockHeader
...
+    homesteadBlockNumber,
+    difficultyBombRemovalBlockNumber,
+    byzantiumBlockNumber
+  }
 
   val DifficultyBoundDivision: Int = 2048
   val FrontierTimestampDiffLimit: Int = -99
   val ExpDifficultyPeriod: Int = 100000
   val MinimumDifficulty: BigInt = 131072
+  val RelaxDifficulty: BigInt = BigInt(10).pow(6) * 3
```
KonradStaniec: I think it would be better to write it as number, otherwise scalastyle can complain on magic numbers ( here it does not, but in `BlockRewardCalculator` it is)

src/main/scala/io/iohk/ethereum/ledger/BlockRewardCalculator.scala
```diff
@@ -39,45 +42,53 @@ class BlockRewardCalculator(config: MonetaryPolicyConfig) {
 
 
   def calcBlockMinerReward(blockNumber: BigInt, ommersCount: Int): BigInt = {
-    val era = eraNumber(blockNumber)
-
-    val baseReward = calcMinerBaseReward(era)
-    val ommersReward = calcMinerRewardPerOmmer(era) * ommersCount
-    baseReward + ommersReward
+    if (blockNumber >= byzantiumBlockNumber) newRewardAfterByzantium else {
```
KonradStaniec: Here you don't need this `if`, as miner reward is always composed from `baseReward` and `ommersReward`  even after byzantium. (only code in `else` case is needed)

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/EthashDifficultyCalculator.scala
agnikowa: done

src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
agnikowa: I tried this, but when I applied it, there were more errors than in the real solution, so I just moved all the calculations to the calculator class

src/main/scala/io/iohk/ethereum/ledger/BlockRewardCalculator.scala
```diff
@@ -18,6 +18,9 @@ class BlockRewardCalculator(config: MonetaryPolicyConfig) {
   /** Base block reward in the first era */
   val firstEraBlockReward: BigInt = config.firstEraBlockReward
 
+  /** Block reward for miner after Byzantium Fork */
+  val newRewardAfterByzantium: BigInt = BigInt(10).pow(18) * 3
```
KonradStaniec: I think it would be better to write it as number, otherwise scalastyle can complain on magic numbers. Or maybe define it in configuration is some smart way.
One i can think of, is definig new parameter in `MonetaryPolicyConfig` called like 
` first-era-reduced-block-reward` and in `ethereumHf.conf` setting it to correct reduced value, but in `application.conf`  setting it to current value as there is no byzantium fork on etc chain.
What do you think about that ?  It my be overkill, but rewards are highly consensus related thing, and i am a ltittle uneasy hardcoding them like that

agnikowa: I've actually tried this solution but then when executing tests the program couldn't see it in some weird way, so I just put it in here, but I could try to investigate further

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/EthashDifficultyCalculator.scala
agnikowa: i've used it to not make a mistake with zeros but I'll do as you proposed

src/main/resources/application.conf
agnikowa: good catch, missed in merge

src/test/scala/io/iohk/ethereum/ledger/BlockRewardSpec.scala
```diff
@@ -62,25 +93,47 @@ class BlockRewardSpec extends FlatSpec with Matchers with MockFactory {
...
-          uncleNodesList = ommerMiners.map(a => Fixtures.Blocks.Genesis.header.copy(beneficiary = a.bytes, number = 5))
+          uncleNodesList = ommerMiners.map{ address =>
+            Fixtures.Blocks.Genesis.header.copy(beneficiary = address.bytes, number = 5)
+          }
+        )
+      )
+    }
+
+    def sampleBlockAfterByzantium(minerAddress: Address, ommerMiners: Seq[Address] = Nil): Block = {
+      val baseBlockNumber = BigInt(10).pow(18) * 3 + 1
```
KonradStaniec: It is pretty big number for block number (now tests fail because of conversion `toInt` in reward calculator).
I think it would be the best to: 
Update `byzantium-block-number` in `mantis/src/test/resources/application.conf` to its real value (4375000)
And set this baseBlockNumber to`blockchainConfig.byzantiumBlockNumber`

src/test/scala/io/iohk/ethereum/ledger/BlockRewardSpec.scala
```diff
@@ -51,6 +51,37 @@ class BlockRewardSpec extends FlatSpec with Matchers with MockFactory {
...
+    val afterRewardWorldState: InMemoryWorldStateProxy = ledger.payBlockReward(block, worldState)
+
+    val beforeExecutionBalance1: BigInt = worldState.getGuaranteedAccount(minerAddress).balance
+    val beforeExecutionBalance2: BigInt = worldState.getGuaranteedAccount(ommer1Address).balance
+    val beforeExecutionBalance3: BigInt = worldState.getGuaranteedAccount(ommer2Address).balance
+
+    afterRewardWorldState.getGuaranteedAccount(minerAddress).balance shouldEqual (beforeExecutionBalance1 + afterByzantiumNewBlockReward)
+
+    // new_uncle_reward = (8 - block.number - uncle.number) * new_block_reward / 8
+    val wei: BigInt = BigInt(10).pow(18) * 3
```
KonradStaniec: you can set it to `blockchainConfig.monetaryPolicyConfig.firstEraReducedBlockReward`

src/test/scala/io/iohk/ethereum/ledger/BlockRewardSpec.scala
```diff
@@ -51,6 +51,37 @@ class BlockRewardSpec extends FlatSpec with Matchers with MockFactory {
...
+    val ommer1Address = Address(block.body.uncleNodesList.head.beneficiary)
+    val ommer2Address = Address(block.body.uncleNodesList(1).beneficiary)
+
+    val afterRewardWorldState: InMemoryWorldStateProxy = ledger.payBlockReward(block, worldState)
+
+    val beforeExecutionBalance1: BigInt = worldState.getGuaranteedAccount(minerAddress).balance
+    val beforeExecutionBalance2: BigInt = worldState.getGuaranteedAccount(ommer1Address).balance
+    val beforeExecutionBalance3: BigInt = worldState.getGuaranteedAccount(ommer2Address).balance
+
+    afterRewardWorldState.getGuaranteedAccount(minerAddress).balance shouldEqual (beforeExecutionBalance1 + afterByzantiumNewBlockReward)
```
KonradStaniec: here balance should equal `beforeExecutionBalance1 + afterByzantiumNewBlockReward + nephewReward` 

where `nephewReward = (blockchainConfig.monetaryPolicyConfig.firstEraReducedBlockReward / 2) * 32`

agnikowa: you meant: `val nephewRewards: BigInt = newBlockReward / 32 * 2`

src/main/scala/io/iohk/ethereum/ledger/BlockRewardCalculator.scala
agnikowa: I've work it out, so this reduce block reward is now present in `MonetaryPolicyConfig` as you suggested

src/test/scala/io/iohk/ethereum/ledger/BlockRewardSpec.scala
KonradStaniec: exactly

PR #493, opened 2018-06-18
Feature/eip 140

mmrozek:
EIP-140 - REVERT opcode


src/main/scala/io/iohk/ethereum/vm/VM.scala
```diff
@@ -123,6 +123,8 @@ class VM[W <: WorldStateProxy[W, S], S <: Storage[S]] extends Logger {
     } else if (codeStoreOutOfGas && !config.exceptionalFailedCodeDeposit) {
       // Code storage causes out-of-gas with exceptionalFailedCodeDeposit disabled
       result
+    } else if(result.error.contains(RevertOccurs)) {
```
KonradStaniec: I am not sure this is good place for this condition. Main reason is that if `Revert` would occur, but the conditions in 
https://github.com/input-output-hk/mantis/pull/493/files#diff-5b7d8817413ce869c362b5398e9a65f1L120 would be fullfilled we would return `result.copy(error = Some(OutOfGas), gasRemaining = 0)` so that is definitly not our intention.
Example of revert handling in that place in geth: https://github.com/ethereum/go-ethereum/blob/1da33028ce88c4365d99471977098f4911fd38fa/core/vm/evm.go#L383

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -776,7 +779,7 @@ abstract class CallOp(code: Int, delta: Int, alpha: Int) extends OpCode(code, de
         val stack2 = stack1.push(UInt256.Zero)
         val mem2 = mem1.expand(outOffset, outSize)
         val world1 = state.world.combineTouchedAccounts(result.world)
-        val gasAdjustment = if (error == InvalidCall) -startGas else BigInt(0)
+        val gasAdjustment = if (error == InvalidCall) -startGas else if (error == RevertOccurs) -result.gasRemaining else BigInt(0)
```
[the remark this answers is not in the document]
mmrozek: Please look at: https://github.com/input-output-hk/mantis/pull/500/files#diff-cfbbd0cee300114eb59c735b4e84e997R120.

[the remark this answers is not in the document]
KonradStaniec: You missed second part of my comment : 
`Also in case of REVERT for create op we must handle this return data somehow, the specification states that In case REVERT is used in the context of a CREATE or CREATE2 call, no code is deployed, 0 is put on the stack and the error message is available in the returndata buffer. So in https://github.com/input-output-hk/mantis/pull/493/files#diff-2573bce385123e76cd2b2495911d3b7aL703, there should be something along this lines: val returnData = if (err == RevertOccurs) result.returnData else ByteString.empty and then in https://github.com/input-output-hk/mantis/pull/493/files#diff-2573bce385123e76cd2b2495911d3b7aL706 `

Also i am also little uneasy with handling the memory the same way as in no error case ,it would be safer just extract logic of storing in memory to some other function, and use it only in case of revert and for other errors leave the old way of handling ( just expand memory)

src/main/scala/io/iohk/ethereum/vm/VM.scala
KonradStaniec: case of do not have enough gas is coveret here : https://github.com/input-output-hk/mantis/blob/284578afeffaeefc0fd7368976d51008ff786bcf/src/main/scala/io/iohk/ethereum/vm/OpCode.scala#L196. 

In this place we have create already returned and i think w do not want to change it to `OutOfGas`  by accident (later we will need to extract return data from revert and pass it to return buffer)

src/test/scala/io/iohk/ethereum/vm/CreateOpcodeSpec.scala
```diff
@@ -233,7 +243,26 @@ class CreateOpcodeSpec extends WordSpec with Matchers {
     "refund the correct amount of gas" in {
       result.stateOut.gasRefund shouldBe result.stateOut.config.feeSchedule.R_selfdestruct
     }
+  }
+
+  "initialization includes REVERT opcode" should {
```
KonradStaniec: there should also be the test case that `error message is available in the returndata buffer.` (one of my earlier comments)

src/main/scala/io/iohk/ethereum/vm/VM.scala
mmrozek: Still, we have to check if reversion occurs, because we don't want to deploy the code and don't want to spend gas for code deposit. I think `saveNewContract` is the best place to do that. I could move this logic one level up but it won't change anything.

src/test/scala/io/iohk/ethereum/vm/CreateOpcodeSpec.scala
mmrozek: But we don't have at the moment returndata buffer functionality. It will be in the next pr.

KonradStaniec: oo my bad, i did not raealize that this new opcodes https://github.com/ethereum/EIPs/blob/master/EIPS/eip-211.md are called return buffer

src/main/scala/io/iohk/ethereum/vm/VM.scala
KonradStaniec: I agree that this method is good place to handle possible errors, but i want to avoid risk of changing the returned error from `Revert` to `outOfGas`. I mean if the executions of contract code ended with `Revert` it needs to be propagatet further even if `maxCodeSizeExceeded || (codeStoreOutOfGas && config.exceptionalFailedCodeDeposit` is true

PR #497, opened 2018-06-21
Feature/eip 214

mmrozek:


src/main/scala/io/iohk/ethereum/vm/ProgramState.scala
```diff
@@ -26,14 +27,15 @@ object ProgramState {
   * @param gas current gas for the execution
   * @param world world state
   * @param addressesToDelete list of addresses of accounts scheduled to be deleted
```
agnikowa: consider formatting also ones above

src/test/scala/io/iohk/ethereum/vm/StaticCallOpcodeSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+  val startState = MockWorldState(touchedAccounts = Set.empty, noEmptyAccountsCond = true)
+
+  val fxt = new CallOpFixture(config, startState)
+
+  def stateWithProgram(code: Assembly): MockWorldState =
+    fxt.worldWithoutExtAccount.saveAccount(fxt.extAddr, fxt.accountWithCode(code.code)).saveCode(fxt.extAddr, code.code)
+
+  "STATICCALL" should {
+
+    "calling a program that executes a state-changing opcodes" should {
```
KonradStaniec: I think testing of `Log` programs should be moved to other test, as they do not modify `staticcall.world` directly. They just append logs in program state, and world is modify after whole transactions.
The assertion in this test would be probably smoething like `staticcall.stateOut.logs.size shouldEqual 0` and 
`call.stateOut.logs.size shoud be > 0`

PR #500, opened 2018-06-25
Feature/eip 211

mmrozek:


src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -417,6 +417,28 @@ case object EXTCODECOPY extends OpCode(0x3c, 4, 0, _.G_extcode) {
   }
 }
 
+case object RETURNDATASIZE extends ConstOp(0x3d)(_.returnData.size)
+
+case object RETURNDATACOPY extends OpCode(0x3e, 3, 0, _.G_verylow) {
+  protected def exec[W <: WorldStateProxy[W, S], S <: Storage[S]](state: ProgramState[W, S]): ProgramState[W, S] = {
+    val (Seq(memOffset, dataOffset, size), stack1) = state.stack.pop(3)
+    if(dataOffset + size > state.returnData.size){
+      state.withStack(stack1).withError(ReturnDataOverflow)
```
KonradStaniec: specification says that `if start + length overflows or results in a value larger than RETURNDATASIZE, the current call stops in an out-of-gas condition`  and when we end execution with `OutOfGas` we are setting remaining gas to 0. I am wondering if we should do the same here ?

KonradStaniec (commented):
as this opcodes changes semantics of create and call operations, it would be nice to have some tests with connection to them

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
mmrozek: Yes, but we do the same thing here. In `ProgramState.toResult` method we have `if (error.exists(_.useWholeGas)) 0 else gas`. `ReturnDataOverflow` are using whole gas

agnikowa (approved):

KonradStaniec (approved):

PR #502, opened 2018-06-25
Feature/eip 197

KonradStaniec:
Adding support for precompiled contract for optimal ate pairing on bn128 curve.
All tests from ets tests passed.
Spec: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-197.md
Review and merge after: #490 and #499


src/main/scala/io/iohk/ethereum/crypto/zksnark/FieldElement.scala
```diff
@@ -0,0 +1,575 @@
...
+  val B_Fp2 = mulByConst(NON_RESIDUE.inversed(), Fp.B_Fp)
+}
+
+case class Fp6(a: Fp2, b: Fp2, c: Fp2) extends FieldElement
+
+object Fp6 {
+
+  // Alg.12 in 'High-Speed Software Implementation of the Optimal Ate Pairing over Barreto–Naehrig Curves'
+  def mulByNonResidue(a: Fp6): Fp6 = {
+    val ra = Fp2.NON_RESIDUE * a.c
```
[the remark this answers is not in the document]
KonradStaniec: it is good that way, in paper mentioned in comment there is (alg.12): 
`A = a0 + a1v + a2v^2` and `C = A · γ, C ∈ Fp6` , where `C = c0 + c1v + c2v^2`

and operation:
`c0 ← a2 · ξ` // our  new first component (`ra`)  is equals to third component `a.c` times NON_RESIDUE

and result is:
`return C ← c0 + a0v + a1v^2` // return new element of Fp6 where first component is `c0` which we calculated earlier, second component is `a0` which is first term of a (`a.a`), third component is` a1` which is second term of a(`a.b`)

src/main/scala/io/iohk/ethereum/crypto/zksnark/BN128.scala
```diff
@@ -172,4 +184,57 @@ object BN128 {
...
+  case class BN128G1(p: Point[Fp])
+  object BN128G1 {
+
+    /**
+      * Constructs valid element of subgroup `G1`
+      * To be valid element of subgroup, elements needs to be valid point (have valid coordinates in Fp_2 and to be on curve
+      * Bn128 in Fp
+      * @return [[scala.None]] if element is invald group element, [[io.iohk.ethereum.crypto.zksnark.BN128.BN128G1]]
+      */
+    def apply(xx: ByteString, yy: ByteString): Option[BN128G1] = {
```
mmrozek: I am not sure if apply returning an Option is the best pattern. Maybe you could use some meaningful name for this method.

src/main/scala/io/iohk/ethereum/crypto/zksnark/PairingCheck.scala
```diff
@@ -0,0 +1,169 @@
...
+    * Ate pairing algorithms
+    *
+    * @param pairs Sequence of valid pairs of group elements (a1, b1, a2, b2, ..., ak, bk) from (G_1 x G_2)^k^
+    * @return `true` if `log_P1(a1) * log_P2(b1) + ... + log_P1(ak) * log_P2(bk) = 0` else `false`
+    */
+  def pairingCheck(pairs: Seq[G1G2Pair]): Boolean = {
+
+    var product = FiniteField[Fp12].one
+
+    pairs.foreach{ pair =>
```
mmrozek: Maybe it will be nicer: `pairs.map(p=> millerLoop(p.a, p.b).filter(_ != FiniteField.one).fold(FiniteField.one)(_ *_)`

src/main/scala/io/iohk/ethereum/vm/PrecompiledContracts.scala
```diff
@@ -19,6 +20,7 @@ object PrecompiledContracts {
   val ModExpAddr = Address(5)
   val Bn128AddAddr = Address(6)
   val Bn128MulAddr = Address(7)
+  val bn128PairingAddr = Address(8)
```
mmrozek: It should starts from capital letter

src/main/scala/io/iohk/ethereum/vm/PrecompiledContracts.scala
```diff
@@ -30,7 +32,8 @@ object PrecompiledContracts {
   val byzantiumContracts = contracts ++ Map(
     ModExpAddr -> ModExp,
     Bn128AddAddr -> Bn128Add,
-    Bn128MulAddr -> Bn128Mul
+    Bn128MulAddr -> Bn128Mul,
+    bn128PairingAddr -> Bn128Pairing
```
mmrozek: ditto

src/main/scala/io/iohk/ethereum/crypto/zksnark/PairingCheck.scala
KonradStaniec: I do not like how your approach creates two unnecessary collections (after map and filter), but definitely i can improve my approach by using fold instead of `var` and `foreach`

src/main/scala/io/iohk/ethereum/crypto/zksnark/BN128.scala
KonradStaniec: it is fairly standard smart constructor pattern. Why do you think it is not the best?

src/main/scala/io/iohk/ethereum/vm/PrecompiledContracts.scala
```diff
@@ -297,4 +301,60 @@ object PrecompiledContracts {
...
+    }
+
+    // Method which stops reading another points if one of earlier ones failed (had invalid coordinates, or was not on
+    // BN128 curve
+    private def getPairs(bytes: Iterator[ByteString]): Option[Seq[G1G2Pair]] = {
+      var accum = List.empty[G1G2Pair]
+      while (bytes.hasNext) {
+        getPair(bytes.next) match {
+          case Some(part) => accum = part :: accum
+          case None => return None // scalastyle:ignore
```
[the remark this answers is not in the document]
KonradStaniec: yep i need to agree  it is ugly :)it is not easy to create function to stop processing collection with first failed element fully functional way, it could probably be done with some lazy collections but i afraid intent would be lost there, and here intent is pretty clear

src/main/scala/io/iohk/ethereum/crypto/zksnark/PairingCheck.scala
mmrozek: Maybe you could use views or sth like that

src/main/scala/io/iohk/ethereum/crypto/zksnark/BN128.scala
mmrozek: I am not sure, maybe it is just my personal preference;) When I have a case class I expect that apply returns me an instance of this class.

mmrozek (approved):

agnikowa (approved):

PR #509, opened 2018-06-29
Fix overflow possibility

KonradStaniec:
In case when `dataOffset` is equals for example `-1 mod (2 ^ 256)` and size is equal to `1`, `(dataOffset + size) mod (2 ^ 256) ` will be qual to `0` and there will be possibility to read byte from empty byte array.


mmrozek (approved):
lgtm

agnikowa (approved):

PR #510, opened 2018-06-29
Riemann poc

shmish111:
* events are batched up to be more efficient
* batches are sent once the batch is full and periodically
* events are also buffered
* backpressure causes events to overflow into stdout (as JSON)
* connection failure causes overflow to stdout JSON
* if you do not add riemann configuration then this simply
  provides a JSON stdout logger

With this you should avoid calls to `log.debug` `log.info` etc and send Riemann events instead. Logging should be reserved for detailed debugging/tracing only.

If this is merged then pretty much every `log.{info,debug,error}` can be migrated to a riemann service without really affecting the functionality.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -28,17 +28,17 @@ mantis {
   # Uncomment this to specify, otherwise use the default implementation
   # secure-random-algo = "NativePRNG"
 
-  network {
+  # riemann {
+  #   host = "localhost"
+  #   port = 5555
+  #   buffer-size = 1000
+  #   batch-size = 10
+  #   auto-flush-ms = 1000
+  #   # optional hostname to send to Riemann server, if not provided the client will use InetAddress.getLocalHost().getHostName()
+  #   # host-name = "my-node"
+  # }
 
-    # riemann {
-    #   host = "localhost"
-    #   port = 5555
-    #   buffer-size = 1000
-    #   batch-size = 10
-    #   auto-flush-ms = 1000
-    #   # optional hostname to send to Riemann server, if not provided the client will use InetAddress.getLocalHost().getHostName()
-    #   # host-name = "my-node"
-    # }
+  network {
 
     # Ethereum protocol version
     protocol-version = 63
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -1,4 +1,5 @@
 mantis {
+
   # Identifier used when connecting to other clients
   client-id = "mantis"
 
@@ -28,6 +29,17 @@ mantis {
   # secure-random-algo = "NativePRNG"
 
   network {
+
+    # riemann {
+    #   host = "localhost"
+    #   port = 5555
+    #   buffer-size = 1000
+    #   batch-size = 10
+    #   auto-flush-ms = 1000
+    #   # optional hostname to send to Riemann server, if not provided the client will use InetAddress.getLocalHost().getHostName()
+    #   # host-name = "my-node"
+    # }
+
     # Ethereum protocol version
     protocol-version = 63
 
--- a/src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
@@ -14,7 +14,6 @@ import io.iohk.ethereum.domain._
 import io.iohk.ethereum.mpt.MerklePatriciaTrie
 import io.iohk.ethereum.network.p2p.messages.PV62.BlockBody
 import io.iohk.ethereum.rlp.RLPImplicits._
-import io.iohk.ethereum.utils.ToRiemann._
 import io.iohk.ethereum.utils.Riemann
 import org.json4s.{CustomSerializer, DefaultFormats, Formats, JString, JValue}
 import org.spongycastle.util.encoders.Hex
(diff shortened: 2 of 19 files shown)


src/main/resources/application.conf
```diff
@@ -28,6 +29,17 @@ mantis {
   # secure-random-algo = "NativePRNG"
 
   network {
+
+    # riemann {
```
LukasGasior1: I'm not sure if `network` is the best section to put it. So far it was only intended for eth network stuff. Maybe we can put riemann config in separate section (like we did with `metrics`)?

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
```diff
@@ -47,18 +48,18 @@ class GenesisDataLoader(
           Try(Source.fromResource(customGenesisFile))
         } match {
           case Success(customGenesis) =>
-            log.info(s"Using custom genesis data from: $customGenesisFile")
+            Riemann.ok("genesis loading").attribute("file", customGenesisFile).send
```
LukasGasior1: Does it log as well?

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
```diff
@@ -103,7 +104,7 @@ class GenesisDataLoader(
 
     val header: BlockHeader = prepareHeader(genesisData, stateMptRootHash)
 
-    log.debug(s"prepared genesis header: $header")
+    header.toRiemann.send
```
LukasGasior1: is this a blocking call?

src/main/scala/io/iohk/ethereum/utils/Riemann.scala
```diff
@@ -0,0 +1,264 @@
...
+              new RiemannStdoutClient()
+          }
+        }
+        case None => {
+          log.debug("create new stdout riemann client")
+          new RiemannStdoutClient()
+        }
+      }
+    }
+    c.connect()
```
LukasGasior1: it seems it's called twice for RiemannBatchClient (btw. is this needed for StdoutClient?)

src/main/scala/io/iohk/ethereum/utils/Riemann.scala
```diff
@@ -0,0 +1,264 @@
...
+      .state("error")
+      .tag("exception")
+      .tag(t.getClass().getSimpleName())
+      .description(desc.toString())
+  }
+  def critical(service: String): EventDSL = defaultEvent.state("critical").service(service)
+
+}
+
+object Riemann extends Riemann with Logger {}
```
LukasGasior1: trait Riemann already extends Logger ;)

src/main/scala/io/iohk/ethereum/utils/Riemann.scala
```diff
@@ -0,0 +1,264 @@
...
+        sendBatch
+        log.trace("sent batch of Riemann events")
+      } else {
+        log.trace("wait for next batch of Riemann events")
+        Thread.sleep(5000)
+      }
+    }
+  }
+
+  val flusher = Future {
```
LukasGasior1: why is this needed? as far as I can see you flush after each batch is sent

src/main/scala/io/iohk/ethereum/utils/Riemann.scala
```diff
@@ -0,0 +1,264 @@
...
+
+  val f = Future {
+    while (true) {
+      if (queue.size() > 0) {
+        log.trace("sending batch of Riemann events")
+        sendBatch
+        log.trace("sent batch of Riemann events")
+      } else {
+        log.trace("wait for next batch of Riemann events")
+        Thread.sleep(5000)
```
LukasGasior1: This will happen a lot. Right after sending the last batch the whole thread will be blocked for 5 seconds - it doesn't sound too good. Can we use something non-blocking here? Perhaps `ScheduledThreadPoolExecutor.schedule` or an Actor?

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
LukasGasior1: (A: it prints to stdout if sending to Riemann server fails)

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
LukasGasior1: (A: no)

src/main/resources/application.conf
shmish111: moved

src/main/scala/io/iohk/ethereum/utils/Riemann.scala
shmish111: its idempotent but fixed anyway since its ugly and bad, thanks

src/main/scala/io/iohk/ethereum/utils/Riemann.scala
shmish111: done

src/main/scala/io/iohk/ethereum/utils/Riemann.scala
shmish111: fixed, see below

shmish111: you are correct, removed

src/main/scala/io/iohk/ethereum/utils/Riemann.scala
shmish111: changed: we now have a `Sender` that takes care of emptying the queue and then re-schedules a `new Sender`. This starts at the same time as `client.connect()` and is closed and re-created appropriately.

PR #512, opened 2018-07-02
Feature/enable transition tests

KonradStaniec:
Enable last portion of ets tests: transition from eip161(158) to byzantium

PR #513, opened 2018-07-02
Introduce transaction outcome data type for recipts

agnikowa:
should be merged after: #511 and #512


src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -167,10 +167,15 @@ object PV63 {
...
-    implicit class ReceiptEnc(msg: Receipt) extends RLPSerializable {
+    implicit class ReceiptEnc(receipt: Receipt) extends RLPSerializable {
       override def toRLPEncodable: RLPEncodeable = {
-        import msg._
-        RLPList(postTransactionStateHash, cumulativeGasUsed, logsBloomFilter, RLPList(logs.map(_.toRLPEncodable): _*))
+        import receipt._
+        val stateHash = postTransactionStateHash match {
+          case HashOutcome(hash) => hash.toArray[Byte]
+          case SuccessOutcome => Array(1.toByte)
+          case _ => Array(0.toByte)
```
mmrozek: Shouldn't it be an empty array?

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -190,7 +195,12 @@ object PV63 {
     implicit class ReceiptRLPEncodableDec(val rlpEncodeable: RLPEncodeable) extends AnyVal {
       def toReceipt: Receipt = rlpEncodeable match {
         case RLPList(postTransactionStateHash, cumulativeGasUsed, logsBloomFilter, logs: RLPList) =>
-          Receipt(postTransactionStateHash, cumulativeGasUsed, logsBloomFilter, logs.items.map(_.toTxLogEntry))
+          val stateHash = postTransactionStateHash match {
+            case RLPValue(bytes) if bytes.length > 1 => HashOutcome(ByteString(bytes))
+            case RLPValue(bytes) if bytes.length == 1 && bytes.head == 1 => SuccessOutcome
+            case _ => FailureOutcome
```
mmrozek: I prefer to write explicit all meaningful cases. In this situation value different than RLPValue is treated as a FailureOutcome.

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
agnikowa: you're right, changed.

agnikowa:
there are 3 tests for it in `LedgerSpec` class https://github.com/input-output-hk/mantis/pull/513/files#diff-e44b517ae530cddefaa864eb1c98cda5R690 or do you mean something more?

mmrozek:
Sorry, I missed that. It should be enough.

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV63.scala
```diff
@@ -167,10 +167,15 @@ object PV63 {
   object ReceiptImplicits {
     import TxLogEntryImplicits._
 
-    implicit class ReceiptEnc(msg: Receipt) extends RLPSerializable {
+    implicit class ReceiptEnc(receipt: Receipt) extends RLPSerializable {
       override def toRLPEncodable: RLPEncodeable = {
-        import msg._
-        RLPList(postTransactionStateHash, cumulativeGasUsed, logsBloomFilter, RLPList(logs.map(_.toRLPEncodable): _*))
+        import receipt._
+        val stateHash = postTransactionStateHash match {
```
KonradStaniec: here you can add type to stateHash like : `val stateHash: RLPEncodeable`, and then in cases you do not need to serialize it by hand i.e you can just write:

```
case HashOutcome(hash) => hash
case SuccessOutcome => 1.toByte      
case _ => 0.toByte
```
And proper serialization implicitis will kick in.

PR #518, opened 2018-07-03
Promote logs of business impact from DEBUG

loverdos:
to (usually) INFO. The idea is to avoid the verbosity of DEBUG
and yet be able to understand what is going on, especially
regarding the blockchain progress (or lack thereof). The
usefulness and effectiveness of this patch has been verified
while debugging a real case in production.

Also:

- augment `HeaderGasLimitError` with context to aid debugging
- provide two sub-cases for `InvalidBranch`, thus making the
  precise error clearer
- augment `Block`'s `idTag` with extra info to aid debugging
- use `block.idTag` as the default way to identify a block
  when logging

Ref CGKIELE-397

### DIFF
--- a/src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
@@ -71,9 +71,10 @@ case class BlockHeader(
 
   lazy val hashAsHexString: String = Hex.toHexString(hash.toArray)
 
-  def idTag: String =
-    s"$number: $hashAsHexString"
-
+  def idTag: String = {
+    val numberHex = number.toString(16)
+    s"$number (= 0x$numberHex : 0x$hashAsHexString)"
+  }
 }
 
 object BlockHeader {
--- a/src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
@@ -114,10 +114,10 @@ case class SignedTransaction (
 
   override def toString: String = {
     s"""SignedTransaction {
-         |tx: $tx
-         |signature: $signature
-         |sender: ${Hex.toHexString(senderAddress.bytes.toArray)}
-         |}""".stripMargin
+       |tx: $tx
+       |signature: $signature
+       |sender: ${Hex.toHexString(senderAddress.bytes.toArray)}
+       |}""".stripMargin
   }
 
   def isChainSpecific: Boolean =
--- a/src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
+++ b/src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
@@ -12,7 +12,6 @@ import io.iohk.ethereum.utils.Config.SyncConfig
 import io.iohk.ethereum.utils._
 import io.iohk.ethereum.utils.events._
 import io.iohk.ethereum.vm._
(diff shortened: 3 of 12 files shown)


src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
```diff
@@ -51,8 +51,10 @@ case class BlockHeader(
 
   lazy val hashAsHexString: String = Hex.toHexString(hash.toArray)
 
-  def idTag: String =
-    s"$number: $hashAsHexString"
+  def idTag: String = {
+    val numberHex = number.toString(16)
```
rtkaczyk: Why this change? Most explorers, including our own, show block in decimal.

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -113,11 +114,11 @@ case class SignedTransaction (
...
-    s"""SignedTransaction {
-         |tx: $tx
-         |signature: $signature
-         |sender: ${Hex.toHexString(senderAddress.bytes.toArray)}
-         |}""".stripMargin
+    MoreObjects.toStringHelper(this)
+      .add("tx", tx)
+      .add("signature", signature)
+      .add("senderAddress", Hex.toHexString(senderAddress.bytes.toArray))
+      .toString
```
rtkaczyk: What will this look like when printed? I was also looking into better formatting here, but was hoping for a more automatic solution, here are some ideas: https://stackoverflow.com/questions/15718506/scala-how-to-print-case-classes-like-pretty-printed-tree

Also, do we add the same for `Block(Body/Header)`, `Transaction`, etc.?

shmish111:
I would suggest that if we are interested in these logs then they should be Riemann events rather than logs

loverdos:
@shmish111 Indeed these are events that can help us understand what the application is doing. That they are in the form of logging is a matter of convention and what we had in place. I would like to find all those important events but this can be approached on-demand. Also this PR includes some model changes regarding error data.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -470,12 +479,20 @@ class LedgerImpl(
...
       block = block,
       getBlockHeaderByHash = getHeaderFromChainOrQueue,
       getNBlocksBack = getNBlocksBackFromChainOrQueue
     )
+
+    result match {
+      case Left(ValidationBeforeExecError(reason)) ⇒
+        log.error(s"[$context/validateBeforeExecution] Block: ${block.idTag}. Reason: ${reason}")
+      case _ ⇒
+    }
```
rtkaczyk: Minor, could be more concise: 
```
result.left.foreach { e => logger.error(...) }
```

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -484,13 +501,37 @@ class LedgerImpl(
...
       stateRootHash = stateRootHash,
       receipts = receipts,
       gasUsed = gasUsed
     )
+
+    result match {
+      case Left(error) ⇒
+        log.error(s"[$context/validateAfterExecution] Block: ${block.idTag}. Reason: ${error.reason}")
+      case _ ⇒
+    }
```
rtkaczyk: Minor, conciseness.

PR #519, opened 2018-07-04
 EC-513 Refactor code to avoid leaking internal implementation of pru…

mmrozek:
…ning

### DIFF
--- a/src/it/scala/io/iohk/ethereum/txExecTest/util/FixtureProvider.scala
+++ b/src/it/scala/io/iohk/ethereum/txExecTest/util/FixtureProvider.scala
@@ -62,11 +62,11 @@ object FixtureProvider {
 
       def traverse(nodeHash: ByteString): Unit = fixtures.stateMpt.get(nodeHash).orElse(fixtures.contractMpts.get(nodeHash)) match {
         case Some(m: BranchNode) =>
-          blockchain.nodesKeyValueStorageFor(Some(block.header.number), storages.nodeStorage, withSnapshotsSave = true).update(Nil, Seq(ByteString(m.hash) -> m.toBytes))
+          blockchain.nodesKeyValueStorageFor(Some(block.header.number), storages.nodeStorage).update(Nil, Seq(ByteString(m.hash) -> m.toBytes))
           m.children.collect { case Some(Left(hash)) => hash}.foreach(e => traverse(e))
 
         case Some(m: ExtensionNode) =>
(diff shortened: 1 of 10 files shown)


src/main/scala/io/iohk/ethereum/utils/Config.scala
```diff
@@ -403,6 +401,7 @@ object PruningConfig {
     val pruningMode: PruningMode = pruningConfig.getString("mode") match {
       case "basic" => BasicPruning(pruningConfig.getInt("history"))
       case "archive" => ArchivePruning
+      case "fastsync" => FastSyncPruning(pruningConfig.getInt("history"))
```
KonradStaniec: after FastSync finishes its run, it is switching to  RegularSync and there we need snapshots if we want to prune, and if you read this from config we will always be stuck with fast sync pruning even during regular sync.

src/it/scala/io/iohk/ethereum/txExecTest/util/FixtureProvider.scala
```diff
@@ -62,19 +62,19 @@ object FixtureProvider {
...
+          blockchain.nodesKeyValueStorageFor(Some(block.header.number), storages.nodeStorage).update(Nil, Seq(ByteString(m.hash) -> m.toBytes))
           m.next match {
             case Left(hash) if hash.nonEmpty => traverse(hash)
             case _ =>
           }
 
         case Some(m: LeafNode) =>
           import AccountImplicits._
-          blockchain.nodesKeyValueStorageFor(Some(block.header.number), storages.nodeStorage, withSnapshotsSave = true).update(Nil, Seq(ByteString(m.hash) -> m.toBytes))
+          blockchain.nodesKeyValueStorageFor(Some(block.header.number), storages.nodeStorage).update(Nil, Seq(ByteString(m.hash) -> m.toBytes))
```
agnikowa: maybe you could extract this to some `def` because this combination is used 3 times in here

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -451,11 +451,11 @@ class FastSync(
 
       case n: BranchNode =>
         val hashes = n.children.collect { case Some(Left(childHash)) => childHash }
-        blockchain.saveNode(ByteString(n.hash), n.toBytes, syncState.targetBlock.number, withSnapshotSave = false)
+        blockchain.saveNode(ByteString(n.hash), n.toBytes, syncState.targetBlock.number)
```
agnikowa: same here with this `ByteString(n.hash), n.toBytes, syncState.targetBlock.number`

PR #520, opened 2018-07-05
[EC-331] improve fix for optimising BlockchainSuite

KonradStaniec:


src/main/scala/io/iohk/ethereum/consensus/ethash/validators/EthashBlockHeaderValidator.scala
```diff
@@ -37,22 +37,21 @@ class EthashBlockHeaderValidator(blockchainConfig: BlockchainConfig) extends Blo
...
-
-            val keys = powCaches.keySet().asScala
-            val keysToRemove = keys.toSeq.sorted.take(keys.size - MaxPowCaches + 1)
-            keysToRemove.foreach(powCaches.remove)
-
-            powCaches.put(epoch, data)
-
-            data
-        }
+      Option(powCaches.get(epoch)) match {
```
agnikowa: are you sure that this `powCaches.get(epoch)` wouldn't return null in some point? if so maybe consider using 
```
Try(powCaches.get(epoch)).toOption
```
so we would omit NullPointerException

mmrozek: I think, it is ok. `Option(null) == None`

PR #522, opened 2018-07-05
Fix vm config executable path handling

LukasGasior1:


src/main/resources/application.conf
```diff
@@ -554,7 +554,6 @@ mantis {
       # - kevm: runs a binary provided at `executable-path` with `port` and `host` as arguments (`./executable-path $port $host`)
       # - mantis: if `executable-path` is provided, it will run the binary with `port` and `host` as arguments
       #           otherwise mantis VM will be run in the same process, but acting as an external VM (listening at `host` and `port`)
-      # - none: doesn't run anything, expect the VM to be started by other means
```
rtkaczyk: We should update _all_ comments in this section

LukasGasior1: I think the remaining comments are still valid.

rtkaczyk: The comments suggest that setting type to `kevm` or `iele` runs the binary. It's not clear that it's actually the `exectuable-path` that determines whether the binary is run or not

LukasGasior1:
I've added the additional `run-vm` flag. Now the logic is:
```
if run-vm is false: don't run anything
if run-vm is true:
  if vm is iele|kevm: executable-path is required and the binary is run
  if vm is mantis: executable-path is optional. it will run binary if provided, and if not, it will run the vm server in the same process
```
I think that makes sense now. Of course there's still vm-mode internal which does not need/run any vm server at all.

@rtkaczyk

PR #523, opened 2018-07-06
Introduce endpoint for peer info

agnikowa:
- enable debug api in JsonRPCController
- create debug service and decode/encode methods
- add debug_listPeersInfo method to JsonRpcRequest

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/DebugService.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/DebugService.scala
@@ -1,0 +1,51 @@
+package io.iohk.ethereum.jsonrpc
+
+import akka.actor.ActorRef
+import akka.pattern._
+import akka.util.Timeout
+import io.iohk.ethereum.jsonrpc.DebugService.{ ListPeersInfoRequest, ListPeersInfoResponse }
+import io.iohk.ethereum.network.EtcPeerManagerActor.{ PeerInfo, PeerInfoResponse }
+import io.iohk.ethereum.network.PeerManagerActor.Peers
+import io.iohk.ethereum.network.{ EtcPeerManagerActor, Peer, PeerActor, PeerId, PeerManagerActor }
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+object DebugService {
+
+  case class ListPeersInfoRequest()
+  case class ListPeersInfoResponse(peers: List[PeerInfo])
+
+}
+
+class DebugService(peerManager: ActorRef, etcPeerManager: ActorRef) {
+
+  def listPeersInfo(getPeersInfoRequest: ListPeersInfoRequest): ServiceResponse[ListPeersInfoResponse] = {
+    val result = for {
+      ids <- getPeerIds
+      peers <- Future.traverse(ids)(getPeerInfo)
+    } yield ListPeersInfoResponse(peers)
+
+    result.map(Right(_))
+  }
+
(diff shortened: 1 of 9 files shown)


src/main/scala/io/iohk/ethereum/jsonrpc/DebugService.scala
```diff
@@ -0,0 +1,51 @@
...
+      .recover{ case _ => Peers(Map.empty[Peer, PeerActor.Status]) }
+      .map(_.peers.keySet.map(_.id).toList)
+  }
+
+  private def getPeerInfo(peer: PeerId): Future[PeerInfo] = {
+    implicit val timeout: Timeout = Timeout(5.seconds)
+
+    (etcPeerManager ? EtcPeerManagerActor.PeerInfoRequest(peer))
+      .mapTo[PeerInfoResponse]
+      .recover{ case _ => PeerInfoResponse(None) }
```
mmrozek: You are trying to recover but below you have a `collect` method and ignore this value.

agnikowa: so you suggest to remove this `recover`?

mmrozek: Yes. Probably it could be one-liner: `.mapTo[PeerInfoResponse].collect{ case PeerInfoResponse(Some(info)) => info }

src/main/scala/io/iohk/ethereum/jsonrpc/DebugService.scala
```diff
@@ -0,0 +1,49 @@
...
+      .recover { case _ => Peers(Map.empty[Peer, PeerActor.Status]) }
+      .map(_.peers.keySet.map(_.id).toList)
+  }
+
+  private def getPeerInfo(peer: PeerId): Future[PeerInfo] = {
+    implicit val timeout: Timeout = Timeout(5.seconds)
+
+    (etcPeerManager ? EtcPeerManagerActor.PeerInfoRequest(peer))
+      .mapTo[PeerInfoResponse]
+      .collect { case PeerInfoResponse(Some(info)) => info }
```
KonradStaniec: Unfortunatly  when i tested it by running `curl --data '{"method":"debug_listPeersInfo","params":[],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8546` it sometimes failed with  such stack trace:
`java.util.NoSuchElementException: Future.collect partial function is not defined at: PeerInfoResponse(None)
	at scala.concurrent.Future.$anonfun$collect$2(Future.scala:370)
	at io.iohk.ethereum.jsonrpc.DebugService$$anonfun$getPeerInfo$1.applyOrElse(DebugService.scala:49)
	at io.iohk.ethereum.jsonrpc.DebugService$$anonfun$getPeerInfo$1.applyOrElse(DebugService.scala:49)
	at scala.concurrent.Future.$anonfun$collect$1(Future.scala:370)
	at scala.util.Success.$anonfun$map$1(Try.scala:251)
	at scala.util.Success.map(Try.scala:209)
	at scala.concurrent.Future.$anonfun$map$1(Future.scala:288)`


I think easiest fix would by to do:
`.collect { case PeerInfoResponse(info) => info }` and change function signature to `Future[Option[PeerInfo]]`
then you just need to `flatten` resulting `List[Option[PeerInfo]]`.

It would be good to create `DebugServiceSpec` with both dependecies as test probes to simulate this failure.

agnikowa: done

PR #525, opened 2018-07-09
Add IELE tx validation to EthService

LukasGasior1:


rtkaczyk:
:+1: but shouldn't it be actually/also done in the `Ledger`. Even though that's not an issue for the testnet we basically make Mantis susceptible to crashing be receiving a message over eth protocol.

rtkaczyk:
Because testing depends on this, I suggest we merge this and add task for validation in `Ledger`

LukasGasior1:
I've added a task for it (https://iohk.myjetbrains.com/youtrack/issue/CGKIELE-440)

PR #526, opened 2018-07-10
EC-544 Improve pending transaction manager performance

mmrozek:


src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
```diff
@@ -39,21 +40,28 @@ object PendingTransactionsManager {
...
+    .maximumSize(txPoolConfig.txPoolSize)
+    .removalListener(
+      (notification: RemovalNotification[ByteString, PendingTransaction]) => if (notification.getCause == RemovalCause.EXPIRED) {
+        knownTransactions = knownTransactions.filterNot(_._1 == notification.getKey)
+      }
+    ).build()
+
   /**
     * stores transactions timeouts by tx hash
     */
```
KonradStaniec: Timeouts map is now unused

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
```diff
@@ -2,26 +2,27 @@ package io.iohk.ethereum.transactions
 
 import akka.actor.{Actor, ActorRef, Cancellable, Props}
 import akka.util.{ByteString, Timeout}
+import com.google.common.cache.{Cache, CacheBuilder, RemovalCause, RemovalNotification}
```
KonradStaniec: It seems we do not have this cache in `build.sbt` , so it seems is it some transitive dependency. Maybe it would be worth adding it, to not depend on transitive dependencies ?

mmrozek: It is from `com.twitter.util-collection`. I could add it explicitly but in my opinion, it is unnecessary.

PR #528, opened 2018-07-11
Feature/ec 547 Remove sender field form SignedTransaction

KonradStaniec:
This pr remove sender field from SignedTransaction, main reason is that currently this field would need  be caluclated during deserialization of signedTransactions. On etc network it did not pose much problem, but on eth network when each block can have much more transactions and getting sender is pretty costly operation, it lead to situation than deserializing block packet during fast sync could last 60s or more. It is pretty not acceptable as it leads to unwarranted timeouts and general fast sync slowness.

Now sender is calculated for transaction when necessery, it can lead to some rpc methods being slower it is worth considering some cache for caluclated senders.


src/main/scala/io/iohk/ethereum/keystore/Wallet.scala
```diff
@@ -9,6 +9,6 @@ import org.bouncycastle.crypto.AsymmetricCipherKeyPair
 case class Wallet(address: Address, prvKey: ByteString) {
   lazy val keyPair: AsymmetricCipherKeyPair = keyPairFromPrvKey(prvKey.toArray)
 
-  def signTx(tx: Transaction, chainId: Option[Byte]): SignedTransaction =
+  def signTx(tx: Transaction, chainId: Option[Byte]): (SignedTransaction, Address) =
```
mmrozek: Why are you returning a tuple if you defined SignedTransactionWithSender case class? If you use this class it will be more readable - you won't need to use _._x syntax.

KonradStaniec: ahh good point, i defined this case class in somewhat different context and later forgot that i have here ready. I will change it.

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -109,14 +107,13 @@ object SignedTransaction {
 
 case class SignedTransaction (
   tx: Transaction,
-  signature: ECDSASignature,
-  senderAddress: Address) {
+  signature: ECDSASignature) {
```
rtkaczyk: Why not just convert it to a `lazy val`?

KonradStaniec: tbh i was thinking about it, but:
1. Sender field not make sense in all transactions , it would need to be `Option[Sender]` as in wrongly signed transactions we would not be able recover sender, or in case of leaving it as `Sender` I would need to track when exceptions would need to be cought. 
2. I got more help from compiler when removing this field, when converted to lazy val it just compiles more or less as it was, after removing that field i needed to expicilty call `getSender` in each place deciding what to do.

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
```diff
@@ -55,15 +53,15 @@ object SignedTransaction {
...
     val bytesToSign: Array[Byte] = if (v == ECDSASignature.negativePointSign || v == ECDSASignature.positivePointSign) {
-      generalTransactionBytes(tx)
+      generalTransactionBytes(tx.tx)
     } else {
-      chainSpecificTransactionBytes(tx, chainId)
+      chainSpecificTransactionBytes(tx.tx, chainId)
     }
 
-    val recoveredPublicKey: Option[Array[Byte]] = signature.publicKey(bytesToSign, Some(chainId))
+    val recoveredPublicKey: Option[Array[Byte]] = tx.signature.publicKey(bytesToSign, Some(chainId))
```
agnikowa: you could replace this `tx.signature` with `v` extracted before, but consider change of the name for something more descriptive

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -559,8 +560,12 @@ class EthService(
 
     Try(req.data.toArray.toSignedTransaction) match {
       case Success(signedTransaction) =>
-        pendingTransactionsManager ! PendingTransactionsManager.AddOrOverrideTransaction(signedTransaction)
-        Future.successful(Right(SendRawTransactionResponse(signedTransaction.hash)))
+        if (SignedTransaction.getSender(signedTransaction).isDefined){
+          pendingTransactionsManager ! PendingTransactionsManager.AddOrOverrideTransaction(signedTransaction)
+          Future.successful(Right(SendRawTransactionResponse(signedTransaction.hash)))
+        }else{
```
agnikowa: weird formatting

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -778,7 +783,7 @@ class EthService(
 
       val tx = Transaction(0, req.tx.gasPrice, gasLimit, toAddress, req.tx.value, req.tx.data)
       val fakeSignature = ECDSASignature(0, 0, 0.toByte)
-      SignedTransaction(tx, fakeSignature, fromAddress)
+      SignedTransactionWithSender(SignedTransaction(tx, fakeSignature), fromAddress)
```
agnikowa: maybe you should consider adding object for `SignedTransactionWithSender ` with apply method something like that:
```
def apply(transaction: Transaction, signature: ECDSASignature, sender: Address): SignedTransactionWithSender = {
    SignedTransactionWithSender(SignedTransaction(transaction, signature), sender)
  }
```
to get rid of always adding `SignedTransaction `

src/main/scala/io/iohk/ethereum/jsonrpc/FilterManager.scala
```diff
@@ -226,7 +226,10 @@ class FilterManager(
       .flatMap { case PendingTransactionsManager.PendingTransactionsResponse(pendingTransactions) =>
         keyStore.listAccounts() match {
           case Right(accounts) =>
-            Future.successful(pendingTransactions.filter(pt => accounts.contains(pt.stx.senderAddress)))
+            Future.successful(pendingTransactions.filter{pt =>
```
agnikowa: weird formatting

src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
```diff
@@ -251,16 +254,25 @@ class BlockPreparator(
...
-            (Account.empty(blockchainConfig.accountStartNonce), world.saveAccount(stx.senderAddress, Account.empty(blockchainConfig.accountStartNonce)))
-          )
         val upfrontCost = calculateUpfrontCost(stx.tx)
-        val validatedStx = signedTxValidator.validate(stx, senderAccount, blockHeader, upfrontCost, acumGas)
+        val senderAddress = SignedTransaction.getSender(stx)
+
+        val accountDataOpt = senderAddress.map {address =>
+          world.getAccount(address).map(a => (a, address))
+            .getOrElse(
+              (Account.empty(blockchainConfig.accountStartNonce), address)
```
agnikowa: weird formatting

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
```diff
@@ -88,8 +89,10 @@ class PendingTransactionsManager(txPoolConfig: TxPoolConfig, peerManager: ActorR
 
     case AddOrOverrideTransaction(newStx) =>
       pendingTransactions.cleanUp()
+      // Only validated tranactions are added this way, it is safe to call get
+      val newStxSender = SignedTransaction.getSender(newStx).get
       val obsoleteTxs = pendingTransactions.asMap().asScala.filter(
-        ptx => ptx._2.stx.senderAddress == newStx.senderAddress && ptx._2.stx.tx.nonce == newStx.tx.nonce
+        ptx => SignedTransaction.getSender(ptx._2.stx).contains(newStxSender) && ptx._2.stx.tx.nonce == newStx.tx.nonce
```
agnikowa: maybe consider to make method out of this `SignedTransaction.getSender(ptx._2.stx).contains(newStxSender)` because you're using this combination twice already (`PersonalService`)

src/test/scala/io/iohk/ethereum/jsonrpc/FilterManagerSpec.scala
```diff
@@ -295,20 +296,21 @@ class FilterManagerSpec extends FlatSpec with Matchers with ScalaFutures with No
...
 
+    val tx = Transaction(
+      nonce = 0,
+      gasPrice = 123,
+      gasLimit = 123,
+      receivingAddress = Address("0x1234"),
+      value = 0,
+      payload = ByteString())
+
+
```
agnikowa: remove additional enter, also 344-345

src/test/scala/io/iohk/ethereum/jsonrpc/FilterManagerSpec.scala
```diff
@@ -332,20 +334,21 @@ class FilterManagerSpec extends FlatSpec with Matchers with ScalaFutures with No
...
+    val stx = SignedTransaction.sign(tx, keyPair, None)
     val pendingTxs = Seq(
```
agnikowa: i think it now fits one line, same as 310

src/test/scala/io/iohk/ethereum/ledger/LedgerSpec.scala
```diff
@@ -197,7 +197,7 @@ class LedgerSpec extends FlatSpec with PropertyChecks with Matchers with MockFac
         val BlockResult(resultingWorldState, resultingGasUsed, resultingReceipts) = txsExecResult.right.get
 
         //Check valid world
-        val minerPaymentForTxs = UInt256(stx.tx.gasLimit * stx.tx.gasPrice)
+        val minerPaymentForTxs = UInt256(stx.tx.tx.gasLimit * stx.tx.tx.gasPrice)
```
agnikowa: i know that this is not a refactor, but maybe consider extracting `stx.tx.tx` to val

src/test/scala/io/iohk/ethereum/ledger/LedgerSpec.scala
```diff
@@ -242,7 +242,7 @@ class LedgerSpec extends FlatSpec with PropertyChecks with Matchers with MockFac
     val BlockResult(resultingWorldState, resultingGasUsed, resultingReceipts) = txsExecResult.right.get
 
     //Check valid world
-    val minerPaymentForTxs = UInt256(validStxSignedByOrigin.tx.gasLimit * validStxSignedByOrigin.tx.gasPrice)
+    val minerPaymentForTxs = UInt256(validStxSignedByOrigin.tx.tx.gasLimit * validStxSignedByOrigin.tx.tx.gasPrice)
```
agnikowa: same here with `validStxSignedByOrigin.tx.tx` to val

src/test/scala/io/iohk/ethereum/ledger/LedgerSpec.scala
```diff
@@ -444,14 +444,14 @@ class LedgerSpec extends FlatSpec with PropertyChecks with Matchers with MockFac
...
-      resultingGasUsed shouldBe stx1.tx.gasLimit + stx2.tx.gasLimit
+      resultingGasUsed shouldBe stx1.tx.tx.gasLimit + stx2.tx.tx.gasLimit
 
       //Check valid receipts
       resultingReceipts.size shouldBe 2
       val Seq(receipt1, receipt2) = resultingReceipts
 
       //Check receipt1
-      val minerPaymentForTx1 = UInt256(stx1.tx.gasLimit * stx1.tx.gasPrice)
+      val minerPaymentForTx1 = UInt256(stx1.tx.tx.gasLimit * stx1.tx.tx.gasPrice)
```
agnikowa: same with `stx1.tx.tx` and tx2.tx.tx, it could be more readable as values

src/test/scala/io/iohk/ethereum/transactions/PendingTransactionsManagerSpec.scala
```diff
@@ -74,34 +77,35 @@ class PendingTransactionsManagerSpec extends FlatSpec with Matchers with ScalaFu
...
     val pendingTxs = (pendingTransactionsManager ? GetPendingTransactions).mapTo[PendingTransactionsResponse].futureValue
     pendingTxs.pendingTransactions.size shouldBe 6
-    pendingTxs.pendingTransactions.map(_.stx).toSet shouldBe (msg2.txs.take(2) ++ msg1.txs.takeRight(4)).toSet
+    pendingTxs.pendingTransactions.map(_.stx).toSet shouldBe (tx2.take(2) ++ tx1.takeRight(4)).toSet
   }
 
   it should "not add pending transaction again when it was removed while waiting for peers" in new TestSetup {
-    val msg1 = SignedTransactions(Seq(newStx(1)))
-    pendingTransactionsManager ! MessageFromPeer(msg1, peer1.id)
+    val msg1 = Seq(newStx(1)).toSet
```
agnikowa: maybe use just `Set(newStx(1))`

src/main/scala/io/iohk/ethereum/domain/SignedTransaction.scala
KonradStaniec: not sure what you mean, `v` is only a part of signature, and to calculate public key i need whole signature.

agnikowa: yup, my mistake, ignore this comment

PR #530, opened 2018-07-12
Selective forging

loverdos:
`mining-enabled` has been silently ignored by the code in the case
of `atomix-raft` consensus. The code just assumes a value of
`true`. The assumption is correct regarding the current design,
since all nodes participate in the Raft consensus, but we need
more operational flexibility. In particular, we may need to
temporarily disable forging until a (recovering) node completely
syncs the blockchain and re-enable it afterwards, thus resuming
normal operation.

The current patch inspects the value of `mining-enabled` and acts
accordingly.

Ref CGKIELE-318

### DIFF
--- a/src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftConsensus.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftConsensus.scala
@@ -38,7 +38,7 @@ class AtomixRaftConsensus private(
 
   type Config = AtomixRaftConfig
 
-  private[this] final val miner = new AtomixRaftForgerRef
+  private[this] final val forgerRef = new AtomixRaftForgerRef
 
   private[this] final val raftServer = new RaftServerRef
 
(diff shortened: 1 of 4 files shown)


src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftConsensus.scala
```diff
@@ -53,33 +53,60 @@ class AtomixRaftConsensus private(
...
+    log.info(s"***** Role changed to $role, ${ConsensusConfig.Keys.MiningEnabled}=${config.miningEnabled}")
+
     if(role == RaftServer.Role.LEADER) {
-      onLeader()
+      if(config.miningEnabled) {
+        onLeaderWithMiningEnabled()
+      }
+      else {
+        onLeaderWithMiningDisabled()
+      }
```
shmish111: `if(role == RaftServer.Role.LEADER && config.miningEnabled)`

src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftConsensus.scala
```diff
@@ -53,33 +53,60 @@ class AtomixRaftConsensus private(
...
+    log.warn(s"***** Elected ${RaftServer.Role.LEADER} but ${ConsensusConfig.Keys.MiningEnabled}=${config.miningEnabled}")
+
+    raftServer.run { server ⇒
+      val cluster = server.cluster()
+      val leader = cluster.getLeader
+
+      // We have just become the leader but what is the probability of having a re-election and a new leader,
+      // thus kicking out a different leader?
+      log.info(s"***** Cluster leader is: ${leader}, demoting")
+      try leader.demote().join()
```
shmish111: Is there a race condition whereby the node is a leader for a short while and might mine? I presume not but just checking.

src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftConsensus.scala
shmish111: gotcha

src/main/scala/io/iohk/ethereum/consensus/atomixraft/AtomixRaftConsensus.scala
shmish111: Yeah, its the second part that I was worried _might_ not be implemented, but it is!

loverdos:
Thanks @rtkaczyk

PR #531, opened 2018-07-12
Export build-related info

loverdos:
From a running node, it is now possible to know exactly which
commit the code is based on, along with some other metadata.

This _buildinfo_ is exported via the logging subsystem when the
node boots, and it is also available under the `/buildinfo`
HTTP endpoint (using the HTTP mode of JSON/RPC).

Ref CGKIELE-463

### DIFF
--- a/src/main/scala/io/iohk/ethereum/nodebuilder/StdNode.scala
+++ b/src/main/scala/io/iohk/ethereum/nodebuilder/StdNode.scala
@@ -1,12 +1,13 @@
 package io.iohk.ethereum.nodebuilder
 
 import io.iohk.ethereum.blockchain.sync.SyncController
+import io.iohk.ethereum.buildinfo.MantisBuildInfo
 import io.iohk.ethereum.consensus.StdConsensusBuilder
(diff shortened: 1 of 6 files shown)


src/main/scala/io/iohk/ethereum/nodebuilder/StdNode.scala
```diff
@@ -61,7 +64,17 @@ abstract class BaseNode extends Node {
     nodeMetrics.Start.trigger()
   }
 
+  private[this] def logBuildInfo(): Unit = {
+    val map = MantisBuildInfo.toMap
+
+    val json = Serialization.writePretty(map)(DefaultFormats)
```
shmish111: these 2 lines are duplicated above, could they be a function in `io.iohk.ethereum.buildinfo.MantisBuildInfo` instead?

loverdos: I have a better function that uses this `Serialization` and the `DefaultFormats` and turns any (?) object into json, which is generally a needed fucntionality. I will refactor to at least take your comment into consideration.

PR #532, opened 2018-07-12
Add iele tx validation to tx validator

LukasGasior1:


src/test/scala/io/iohk/ethereum/consensus/validators/SignedTransactionValidatorSpec.scala
```diff
@@ -18,7 +19,7 @@ class SignedTransactionValidatorSpec extends FlatSpec with Matchers {
 
   val blockchainConfig = BlockchainConfig(Config.config)
 
-  val signedTransactionValidator = new StdSignedTransactionValidator(blockchainConfig)
+  val signedTransactionValidator = new StdSignedTransactionValidator(blockchainConfig, VmConfig(VmMode.Internal, None))
```
rtkaczyk: Could we add tests for IELE? Positive and negative cases

LukasGasior1: @rtkaczyk done

PR #534, opened 2018-07-13
Feature/ec 543 jni leveldb

agnikowa:
- add rocksdb as new data source 
- add config for which db to use 

Issue description under: [EC-543](https://iohk.myjetbrains.com/youtrack/issue/EC-543)


build.sbt
```diff
@@ -26,8 +26,9 @@ val dep = {
     "de.heikoseeberger" %% "akka-http-json4s" % "1.11.0",
     "com.typesafe.akka" %% "akka-http-testkit" % akkaHttpVersion % "it,test",
     "io.suzaku" %% "boopickle" % "1.2.6",
-    "org.iq80.leveldb" % "leveldb" % "0.12",
-    "org.iq80.leveldb" % "leveldb-api" % "0.12",
+    "org.iq80.leveldb" % "leveldb" % "0.7",
```
KonradStaniec: hmm i think they latest one is version `0.12` which is our own fork of leveldb with some fixes - https://github.com/input-output-hk/leveldb/releases. I know numeration does not make sense :P

src/main/resources/application.conf
```diff
@@ -550,7 +550,36 @@ mantis {
...
+
+      # This ensures that only 32 open files can be accessed at once
+      max-open-files = 32
+
+      # Force checksum verification of all data that is read from the file system on behalf of a particular read
+      verify-checksums = true
     }
+
+    # Define which database to use [iodb, leveldb, rocksdb]
+    data-source = "rocksdb"
```
KonradStaniec: I like the idea of supporting choosing db but it maybe overkill (expecially as iodb is no longer maintained and used). Actually i think it may be better to support only rocksdb, and fallback to java port of leveldb only went platform does not support rocks db.  wdyt ?

src/main/scala/io/iohk/ethereum/db/dataSource/LevelDBDataSource.scala
```diff
@@ -81,7 +73,21 @@ class LevelDBDataSource(
...
+      import levelDbConfig._
+
+      val options = new Options()
+        .createIfMissing(createIfMissing)
+        .paranoidChecks(paranoidChecks) // raise an error as soon as it detects an internal corruption
+        .verifyChecksums(verifyChecksums) // force checksum verification of all data that is read from the file system on behalf of a particular read
+
+
+      val factory = if (native) {
+        org.fusesource.leveldbjni.JniDBFactory.factory
```
KonradStaniec: If we would like to support LevelDb jni that way we woulnd need similar `ReadWriteLock` as in rocks db to avoid corruption on close or open. This is one reasons to maybe consider supporting only rocks db and java leveldb as fallback.

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -0,0 +1,175 @@
+package io.iohk.ethereum.db.dataSource
+
+import java.util.concurrent.locks.ReentrantReadWriteLock
+
+import org.rocksdb._
+
+class RocksDbDataSource(private var db: RocksDB, private val rocksDbConfig: RocksDbConfig) extends DataSource {
```
KonradStaniec: ethereumj and rocks docs mentions need to load  rocks db library with `RocksDB.loadLibrary();`  - https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/datasource/rocksdb/RocksDbDataSource.java#L69 
Dont we need something like this ?

src/test/scala/io/iohk/ethereum/db/dataSource/LevelDBDataSourceTest.scala
```diff
@@ -1,15 +1,22 @@
 package io.iohk.ethereum.db.dataSource
 
+import java.nio.file.Files
+
 import org.scalatest.FlatSpec
 
 class LevelDBDataSourceTest extends FlatSpec with DataSourceTestBehavior {
```
KonradStaniec: this test is now failing:
`Error during tests:
[error] 	io.iohk.ethereum.db.dataSource.LevelDBDataSourceTest`

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -0,0 +1,191 @@
...
+    * This function obtains the associated value to a key, if there exists one.
+    *
+    * @param namespace which will be searched for the key.
+    * @param key       the key retrieve the value.
+    * @return the value associated with the passed key.
+    */
+  override def get(namespace: Namespace, key: Key): Option[Value] = {
+    RocksDbDataSource.dbLock.readLock().lock()
+    try {
+      val readOptions = new ReadOptions().setVerifyChecksums(rocksDbConfig.verifyChecksums)
```
KonradStaniec: Maybe it is worth passing one ` ReadOptions` object to this `RocksDbDataSource`, it seems costly to create it before each read  ? (it can be created in `apply` or ` createDb` functions)

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -0,0 +1,191 @@
...
+    * @param toRemove  which includes all the keys to be removed from the DataSource.
+    * @param toUpsert  which includes all the (key-value) pairs to be inserted into the DataSource.
+    *                  If a key is already in the DataSource its value will be updated.
+    * @return the new DataSource after the removals and insertions were done.
+    */
+  override def update(namespace: Namespace, toRemove: Seq[Key], toUpsert: Seq[(Key, Value)]): DataSource = {
+    RocksDbDataSource.dbLock.readLock().lock()
+    try {
+      val batch = new WriteBatch()
+      val writeOptions = new WriteOptions().setSync(rocksDbConfig.synchronousWrites)
```
KonradStaniec: All rockDb objects are backed by c++ , and you need to call `close` after using them (https://github.com/facebook/rocksdb/wiki/RocksJava-Basics#memory-management). So even `WriteOptions ` object need to be closed, thats way ethereumj  uses `try-with-resources` for all alocated objects (https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/datasource/rocksdb/RocksDbDataSource.java#L265) as all rocks db object implement `AutoClosableInterface`. 
Unfortunatly scala does not support this idiom, although it is possible to implement it yourself (check out https://medium.com/@dkomanov/scala-try-with-resources-735baad0fd7d). 
Maybe it is worth implementing such util and using it here ?

src/main/resources/application.conf
```diff
@@ -550,7 +550,36 @@ mantis {
 
       # Force checksum verification of all data that is read from the file system on behalf of a particular read
       verify-checksums = true
+
+      # This ensures that only 32 open files can be accessed at once
+      max-open-files = 32
```
KonradStaniec: Why `max-open-files = 32` for leveldb ? Up until now we were using default (`1000`) and everything was ok with it.

src/main/resources/application.conf
```diff
@@ -550,7 +550,36 @@ mantis {
...
+      max-threads = 1
+
+      # This ensures that only 32 open files can be accessed at once
+      max-open-files = 32
+
+      # Force checksum verification of all data that is read from the file system on behalf of a particular read
+      verify-checksums = true
+
+      # Can be turned on to make the write operation not return until the data being written has been pushed all the way to persistent storage
+      synchronous-writes = false
```
KonradStaniec: Why exposing that option ? i mean it is probably performance killer, and we would not enable it ever.

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -0,0 +1,248 @@
...
+    * @param namespace which will be searched for the key.
+    * @param key       the key retrieve the value.
+    * @return the value associated with the passed key.
+    */
+  override def get(namespace: Namespace, key: Key): Option[Value] = {
+    RocksDbDataSource.dbLock.readLock().lock()
+    try {
+      Option(db.get(readOptions, (namespace ++ key).toArray))
+    } catch {
+      case e: Exception =>
```
KonradStaniec: (comment in this line but applies to all exceptions)
I am not sure that swallowing  exceptions this way is good idea e.g it can lead to case during execution of transaction that mantis will think some account does not exist and create new one, or some update will be lost etc.
I think you should catch `case NonFatal(ex)`  (a if it its fatal error we do not do anything ), and log error, and just re throw it as `RuntimeException`, similary to ethereumj `https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/datasource/rocksdb/RocksDbDataSource.java#L280`

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -0,0 +1,248 @@
...
+
+  }
+
+  def apply(rocksDbConfig: RocksDbConfig): RocksDbDataSource = {
+    val (db, readOptions) = createDB(rocksDbConfig)
+    new RocksDbDataSource(db, rocksDbConfig, readOptions)
+  }
+
+  // try-with-resources, source: https://github.com/dkomanov/stuff/blob/master/src/com/komanov/io/package.scala
+  def withResources[R <: AutoCloseable, T](r: => R)(f: R => T): T = {
```
KonradStaniec: I think it can be moved to new file in `utils` package as it can be used in other contexts other than db.

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -0,0 +1,248 @@
...
+  private def createDB(rocksDbConfig: RocksDbConfig): (RocksDB, ReadOptions) = {
+    import rocksDbConfig._
+
+    RocksDB.loadLibrary()
+
+    RocksDbDataSource.dbLock.writeLock().lock()
+
+    val readOptions = new ReadOptions().setVerifyChecksums(rocksDbConfig.verifyChecksums)
+    try {
+      val options = new Options()
```
KonradStaniec: I think we can add here this other table options https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/datasource/rocksdb/RocksDbDataSource.java#L122, as they are really improving performnace of rocks db.

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
agnikowa: good point

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
agnikowa: what do you think about putting all that into config of rocksdb?

KonradStaniec: i think that maybe blockcache size should be configurable, as it will be beneficial to user who has bigger memory supply. But as for other options i think they can be hardocoded, as it is hard to imagine that user would want to to go docs and read what bloom filter does or any of the other options.

build.sbt
```diff
@@ -26,8 +27,9 @@ val dep = {
     "de.heikoseeberger" %% "akka-http-json4s" % "1.11.0",
     "com.typesafe.akka" %% "akka-http-testkit" % akkaHttpVersion % "it,test",
     "io.suzaku" %% "boopickle" % "1.2.6",
-    "org.iq80.leveldb" % "leveldb" % "0.12",
-    "org.iq80.leveldb" % "leveldb-api" % "0.12",
+    "org.iq80.leveldb" % "leveldb-api" % levelDb,
+    "org.iq80.leveldb" % "leveldb" % levelDb,
+    "org.rocksdb" % "rocksdbjni" % "5.14.2",
```
mmrozek: It will be more readable if you extract rocksdb version to variable.

src/main/scala/io/iohk/ethereum/db/dataSource/LevelDBDataSource.scala
```diff
@@ -44,17 +37,25 @@ class LevelDBDataSource(
     */
   override def update(namespace: Namespace, toRemove: Seq[Key], toUpsert: Seq[(Key, Value)]): DataSource = {
     val batch = db.createWriteBatch()
-    toRemove.foreach { key => batch.delete((namespace ++ key).toArray) }
-    toUpsert.foreach { item => batch.put((namespace ++ item._1).toArray, item._2.toArray) }
-    db.write(batch, new WriteOptions())
+
```
mmrozek: Did you investigate merge operator from rocksdb ([merge-operator](https://github.com/facebook/rocksdb/wiki/Merge-Operator))? I wonder if it could improve performance.

KonradStaniec: i remember investigating it some day as it would greatly improve our reference counting but unfortunatly it is not part of java api yet https://github.com/facebook/rocksdb/issues/2282

PR #537, opened 2018-07-13
this adds duplicate options

shmish111:


rtkaczyk:
A bit more context please?

deepfire:
@rtkaczyk, https://iohk.myjetbrains.com/youtrack/issue/CGKIELE-427

loverdos:
How does it affect `sbt dist`? There is a procedure that generates the zip and then a specific structure is assumed for someone who just wants to run Mantis this way ...

rtkaczyk:
This makes it harder to run a local node built with `sbt dist`. I will have to recreate this file each time I build a new dist.

shmish111:
@rtkaczyk could the file be somewhere else as an example config, rather than it being included in the dist? Then you wouldn't need to re-create the file, it would be in the repo.

rtkaczyk:
@shmish111  for my local testing it wouldn't be a problem to simply rename this file. But we have to keep in mind that this code is not 100% about IELE/KEVM testnets. It may be merged back to mainline Mantis (as was done before), or who knows what else it will be used for. So the problem I'm having with this is:
- it has to be well documented, everyone should still know how to run Mantis
- an instruction like: "before running rename XXX to YYY" has that hacky smell 😉 

Instead I'd rather focus on the other source of duplication (which is what actually?). Could we do one of:
- drop the other method of passing JVM args and template this file as any other config file
- have a script remove this file when setting up a deployment
?

PR #539, opened 2018-07-18
some riemann fixes including time which should be in seconds and more…

shmish111:
… structured service names


src/main/scala/io/iohk/ethereum/utils/Riemann.scala
```diff
@@ -65,7 +65,7 @@ trait Riemann extends Logger {
   def defaultEvent: EventDSL = {
     val event = new EventDSL(riemannClient)
     val microsenconds =
```
LukasGasior1: s/microseconds/seconds

PR #542, opened 2018-07-19
Fix status field in tx receipt

LukasGasior1:


src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -340,7 +340,11 @@ class EthService(
             data = txLog.data,
             topics = txLog.logTopics)
         },
-        status = receipt.status,
+        status = receipt.statusCode.map {
+          case s if s == ByteString(BlockPreparator.StatusCodeSuccess) => true
+          case _ => false
+        },
```
rtkaczyk: It's been since I've written any Scala 😁 , but I think this can be simplified to `status = receipt.statusCode.map(_ == ByteString(BlockPreparator.StatusCodeSuccess))`. Minor

rtkaczyk: Also, don't we need `status` for EVM?

LukasGasior1: Simplified. For EVM it's optional, but when it's there it has to be a boolean.

PR #543, opened 2018-07-20
add 2 health checks to Riemann

shmish111:
This adds some new healthchecks that regularly send events to Riemann.


src/main/scala/io/iohk/ethereum/nodebuilder/StdNode.scala
```diff
@@ -100,6 +122,8 @@ abstract class BaseNode extends Node {
     tryAndLogFailure(() => consensus.stopProtocol())
     tryAndLogFailure(() => Await.ready(system.terminate, shutdownTimeoutDuration))
     tryAndLogFailure(() => storagesInstance.dataSources.closeAll())
+    tryAndLogFailure(() => sendExecutor.shutdown())
+    tryAndLogFailure(() => stopJsonRpcHttpServer())
```
LukasGasior1: can we add jsonRpcContrller.shutdown() here instead of adding it to `close` method in various server implementations?

LukasGasior1: I'm not sure if we should introduce such coupling. I mean, currently controller is independent of "server". Neither http server not ipc server is the "owner" of controller - it's a separate component, and that's why I think they shouldn't close it.

PR #544, opened 2018-07-22
Fix sbt-verify version in plugins.sbt

artem0:
At this moment, mantis refers to the outdated version of `sbt-verify` plugin - `0.3.0-SNAPSHOT` while the [latest version](https://github.com/input-output-hk/sbt-verify/blob/master/build.sbt#L7) is `0.4.1`, it breaks down the build, because `sbt` tries to fetch not existing plugin's version from local sbt folder after publishing `sbt-verify`.


mirkoAlic:
[TOO OLD]

PR #547, opened 2018-07-26
Refactor/ec 316 peer manager actor state mutation

agnikowa:


src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -16,230 +16,293 @@ import io.iohk.ethereum.network.discovery.PeerDiscoveryManager
...
+      terminatedPeers.foreach{ peerId =>
         peerEventBus ! Publish(PeerDisconnected(peerId))
-        managerState.removeTerminatedPeer(peerId)
       }
+      context unwatch ref
+      context become listen(pendingPeers -- terminatedPeers, peers -- terminatedPeers)
 
-    case IncomingConnectionHandshakeSuccess(peerId, peer) =>
-      if (managerState.incomingPeers.size >= peerConfiguration.maxIncomingPeers) {
+    case IncomingConnectionHandshakeSuccess(peer) =>
```
KonradStaniec: I think we can get rid of that case if we are already doing refactoring e.g;
1. We can subscribe to `peerEventBus ! Subscribe(SubscriptionClassifier.PeerHandshaked)`  during action creation
2. And then cover case `case PeerEvent.PeerHandshakeSuccessful(peer, _) if peer.incomingConnection` 
and then we cen get rid of 
3. https://github.com/input-output-hk/mantis/pull/547/files#diff-d88dc0e9076bb9b1c58875187ebcd4acR230

I think that solution will be clearer, and use our existing pub-sub infrastrucure. Wdyt ?

agnikowa: check now if you had something like this in mind

PR #548, opened 2018-07-27
Refactor Ledger class

agnikowa:
- create BlockImport for all actions connected with importing new block
- create BlockExecution for all actions connected with block transactions
- create BranchResolution for all actions connected with resolving branch
- adjust test class to new classes

description under ticket: [EC-313](https://iohk.myjetbrains.com/youtrack/issue/EC-313)


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -2,78 +2,71 @@ package io.iohk.ethereum.ledger
...
-   *                         eg. in the importBlock method)
-   */
+  /** Executes a block
+    *
+    * @param alreadyValidated should we skip pre-execution validation (if the block has already been validated,
+    *                         eg. in the importBlock method)
+    */
   def executeBlock(block: Block, alreadyValidated: Boolean = false): Either[BlockExecutionError, Seq[Receipt]]
 
   def simulateTransaction(stx: SignedTransactionWithSender, blockHeader: BlockHeader, world: Option[InMemoryWorldStateProxy]): TxResult
```
KonradStaniec: food for thought: Maybe it would be good to move `simulateTransaction` and `binarySearchGasEstimation` to separate class ? They are highly connected to each other, but not have much in common with other methods. As i see Ledger public interface it should be highly related to importing block to blockchain and providing info about the process, so from current interface the most releveant methods are: 
`checkBlockStatus`
`importBlock`
`resolveBranch`

src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
```diff
@@ -0,0 +1,260 @@
...
+        BlockImportFailed(error.toString)
+
+      case Some(_) =>
+        topBlocks.drop(importedBlocks.length).headOption.foreach { failedBlock =>
+          blockQueue.removeSubtree(failedBlock.header.hash)
+        }
+        BlockImportedToTop(importedBlocks)
+    }
+
+    importedBlocks.foreach { b =>
```
mmrozek: This loop should be inside `debug` and lazy evaluated.

src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
```diff
@@ -0,0 +1,260 @@
...
+    * @return the list of removed blocks along with receipts and total difficulties
+    */
+  private def removeBlocksUntil(parent: ByteString, fromNumber: BigInt): List[BlockData] = {
+    blockchain.getBlockByNumber(fromNumber) match {
+      case Some(block) if block.header.hash == parent =>
+        Nil
+
+      case Some(block) =>
+        val receipts = blockchain.getReceiptsByHash(block.header.hash).get
+        val td = blockchain.getTotalDifficultyByHash(block.header.hash).get
```
mmrozek: I am not sure we should use `get`

src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
agnikowa: you mean something like this: 
```
log.debug{
        lazy val result = importedBlocks.map { b =>
          val header = b.block.header
        s"Imported new block (${header.number}: ${Hex.toHexString(header.hash.toArray)}) to the top of chain \n"
        }
        result.toString
      }
```

src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
agnikowa: I'll use for {} yield instead

src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
mmrozek: It still won't be lazy (I think).
```
log.debug("{}", {
 lazy val result = importedBlocks.map { b =>
 val header = b.block.header
 s"Imported new block (${header.number}: ${Hex.toHexString(header.hash.toArray)}) to the 
 top of chain \n"
 }
 result.toString
})
```

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -424,19 +112,19 @@ class LedgerImpl(
       UnknownBlock
   }
 
-  def executeBlock(block: Block, alreadyValidated: Boolean = false): Either[BlockExecutionError, Seq[Receipt]] = {
+  override def executeBlock(block: Block, alreadyValidated: Boolean = false): Either[BlockExecutionError, Seq[Receipt]] = {
```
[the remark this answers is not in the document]
KonradStaniec: yep thats what i have in mind.

src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
```diff
@@ -0,0 +1,264 @@
...
+      blockQueue.enqueueBlock(block, currentBestBlock.header.number) match {
+        case Some(Leaf(leafHash, leafTd)) if isBetterBranch(block, currentBestBlock, leafTd, currentTd) =>
+          log.debug("Found a better chain, about to reorganise")
+          reorganiseChain(leafHash, leafTd)
+
+        case _ =>
+          BlockEnqueued
+      }
+    })
+  }(validationContext)
```
KonradStaniec: In old version `reorganise` was taking `blockexecutioncontext` as implicit, and wa using it. I would like to keep it that way. I mean the callet would suprised that now all his execution is happening on validation context.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -497,80 +154,72 @@ class LedgerImpl(
...
-        }
-      case None => worldState
+  override def resolveBranch(headers: Seq[BlockHeader]): BranchResolutionResult = {
+    if (!branchResolution.areHeadersFormChain(headers) || headers.last.number < blockchain.getBestBlockNumber()) {
+      InvalidBranch
+    } else {
+      // Dealing with a situation when genesis block is included in the received headers,
+      // which may happen in the early block of private networks
+      val result = for {
+        genesisHeader      <- blockchain.getBlockHeaderByNumber(0)
```
KonradStaniec: actually blockchain has a method called `genesisHeader`. 

Also i am wondering why not to move this method to`BranchResolution`, and let ledeger only delegete to it in `resolveBranch` ?  As now situation is kinda weird, we have bunch of methods related to resolution in one module, and main resolution method in other module using those methods.

agnikowa:
also I've run ETS tests and the results are the same as in `phase/release2_0` branch:

```
bcBlockGasLimitTest - Tests passed: 30
bcExploitTest - Tests passed: 20
bcForgedTest - Tests passed: 143
bcForkStressTest - Tests passed: 10
bcGasPricerTest - Tests passed: 15
bcInvalidHeaderTest - Tests passed: 110
bcMultiChainTest - Tests passed: 50
bcRandomBlockhashTest - Tests passed: 480
bcStateTests - Tests ignored: 1, passed: 44
bcTotalDifficultyTest - Tests passed: 55
bcUncleHeaderValidity - Tests passed: 140
bcUncleTest - Tests passed: 115
bcValidBlockTest - Tests failed: 1, passed: 104 (callRevert_Byzantium) Block has invalid gas used, expected 21006 but got 50000
bcWalletTest - Tests passed: 25
GeneralStateTests - Tests failed: 1925, passed: 7695, ignored: 60
TransitionTests - Tests passed: 13```

File: src/main/scala/io/iohk/ethereum/ledger/BranchResolution.scala [hunk 7]
```diff
@@ -0,0 +1,77 @@
...
+      // Dealing with a situation when genesis block is included in the received headers,
+      // which may happen in the early block of private networks
+      val result = for {
+        genesisHeader      <- blockchain.getBlockHeaderByNumber(0)
+        givenHeadOfHeaders <- headers.headOption
+        isGenesisNumber     = givenHeadOfHeaders.number == genesisHeader.number
+        isGenesisHash       = givenHeadOfHeaders.hash == genesisHeader.hash
+        reachedGenesis      = isGenesisHash && isGenesisNumber
+        parentIsKnown       = blockchain.getBlockHeaderByHash(givenHeadOfHeaders.parentHash).isDefined
+      } yield parentIsKnown || reachedGenesis
```
mmrozek: If `reachedGenesis` is true you unnecessary call the `blockchain.getBlockHeaderByHash`. You could write this code in different order.

src/main/scala/io/iohk/ethereum/ledger/BlockExecution.scala
```diff
@@ -0,0 +1,175 @@
...
+    */
+  def executeBlocks(blocks: List[Block], parentTd: BigInt): (List[BlockData], Option[BlockExecutionError]) = {
+    @tailrec
+    def go(
+      executedBlocks: List[BlockData],
+      remainingBlocks: List[Block],
+      parentTd: BigInt,
+      error: Option[BlockExecutionError]
+    ):(List[BlockData], Option[BlockExecutionError]) ={
+      if (remainingBlocks.isEmpty) {
```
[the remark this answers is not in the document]
agnikowa: To be honest this snippet of code comes from PR that was approved not so long ago, so I didn't change it, but I'll investigate and try to create proper tests for this use case.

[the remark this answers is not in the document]
agnikowa: I've added proper test in `BlockExecutionSpec` class and it looks that it works ok for both cases: 
- two blocks with txs (that first one has invalid tx)
- two blocks with txs (that last one has invalid tx)

PR #554, opened 2018-07-31
Replace some logging with Riemann events

shmish111:
There are now quite a few Riemann events in place of logging, this means after this PR you really would be better off running a simple Riemann server (`docker run --rm -p 5555:5555 -p 5556:5556 riemannio/riemann`) and using `rcl` with `jq`. 

Installing `rcl` requires either nix (it's built in our Hydra) or stack at the moment as I don't have any need to distribute it anywhere else.

This is basically structured logging from a developer point of view and my intention would be to remove all `log.*` calls except for a few odd `log.debug` calls perhaps and ofc the logs that other libraries produce such as akka and netty.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,28 @@ After installing the `sbt-verify` library to your local repository checkout this
 
 This creates a distribution zip.
 
+### Logs
+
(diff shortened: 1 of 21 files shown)


loverdos:
My "comments" are in https://github.com/input-output-hk/mantis/pull/570. If there are no counter-comments, feel free to merge #570 here and this one back to `phase/iele_testnet` @shmish111. We can always get it in better shape in another round.

PR #555, opened 2018-08-01
Refactor/ec 320 regularsync

agnikowa:
Transform all vars from `RegularSync` class into params

PR #564, opened 2018-08-08
Refactor/fast sync

agnikowa:


agnikowa:
duplicated, already under #555

PR #565, opened 2018-08-08
NotifyPeers refactor to avoid sending to separate msg for each peer

mmrozek:


agnikowa (approved):
lgtm

src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala
```diff
@@ -99,19 +100,22 @@ class PendingTransactionsManager(txPoolConfig: TxPoolConfig, peerManager: ActorR
...
-        .filter(stx => pendingTransactions.asMap().containsKey(stx.hash)) // signed transactions that are still pending
-        .filterNot(isTxKnown(_, peer.id)) // and not known by peer
-
-      if (txsToNotify.nonEmpty) {
-        etcPeerManager ! EtcPeerManagerActor.SendMessage(SignedTransactions(txsToNotify), peer.id)
-        txsToNotify.foreach(setTxKnown(_, peer.id))
+      peers.foreach{
+        peer => val txsToNotify = signedTransactions
+          .filter(stx => pendingTransactions.asMap().containsKey(stx.hash)) // signed transactions that are still pending
+          .filterNot(isTxKnown(_, peer.id)) // and not known by peer
```
KonradStaniec: as we are doing improvments here, maybe we would move `isTxKnown(_, peer.id)` to filter above to have something like `.filter(stx => pendingTransactions.asMap().containsKey(stx.hash) && !isTxKnown(_, peer.id))` to avoid creating one unecessery collection ?

KonradStaniec: nvm :D

KonradStaniec (approved):

PR #571, opened 2018-08-13
jsonrpc: Make request body strict

loverdos:
This avoids observed non-deterministic behavior that bubbles up as
request parsing errors (probably caused by consuming a stream
more than once).

Ref CGKIELE-570


src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -45,24 +51,57 @@ trait JsonRpcHttpServer extends Json4sSupport {
...
+            mainRoute
+        }
+      }
+    }
+
+  val route: Route =
+    cors(corsSettings) {
+      handleRejections(myRejectionHandler) {
+        handleExceptions(myExceptionHandler) {
+          limitsRoute
```
rtkaczyk: This probably comes with a cost for some edgy edge cases... could we do `if (maxContentLength > 0) limitsRoute else mainRoute` ?

PR #575, opened 2018-08-16
Feature/ec 576 bitwise shifting instructions in evm

agnikowa:
Specification under [link](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-145.md)


src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -313,6 +313,40 @@ case object NOT extends UnaryOp(0x19, _.G_verylow)(~_) with ConstGas
 
 case object BYTE extends BinaryOp(0x1a, _.G_verylow)((a, b) => b getByte a) with ConstGas
 
+// logical shift left
```
mmrozek: Shift left and right are the same apart of shifting direction. It will be nicer if you extract the code to common ShiftingOp

PR #577, opened 2018-08-16
[CGKIELE-591] log  json rpc requests

rtkaczyk:


src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -102,12 +102,39 @@ trait JsonRpcHttpServer extends Json4sSupport {
...
-  private def handleRequest(request: JsonRpcRequest) = {
-    complete(jsonRpcController.handleRequest(request))
+  private def handleRequest(request: JsonRpcRequest) = extractClientIP { ip =>
+    complete {
+      val requestId = request.id.flatMap(_.extractOpt[String]).getOrElse("n/a")
+
+      log.info(s"Processing new JSON RPC request [$requestId] from [$ip]:\n ${serialization.write(request)}")
+
+      val responseF = jsonRpcController.handleRequest(request)
+      responseF.onComplete {
```
loverdos: Minor, we could use `andThen` to both signify that we have an effect and reuse the returned value, which is the original future.

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -102,12 +102,39 @@ trait JsonRpcHttpServer extends Json4sSupport {
     complete(httpResponseF)
   }
 
-  private def handleRequest(request: JsonRpcRequest) = {
-    complete(jsonRpcController.handleRequest(request))
+  private def handleRequest(request: JsonRpcRequest) = extractClientIP { ip =>
+    complete {
+      val requestId = request.id.flatMap(_.extractOpt[String]).getOrElse("n/a")
+
+      log.info(s"Processing new JSON RPC request [$requestId] from [$ip]:\n ${serialization.write(request)}")
```
loverdos: At `INFO` debug level, we should probably log only the success or error situtation, so the initial log statement can be either discarded or changed to `log.debug`.
Something like:

```scala
// either debug or do not log this at all.
log.debug(s"Processing new JSON RPC request [$requestId] from [$ip]:\n ${serialization.write(request)}")
[...]
case Success(_) => log.info(s"Request [$requestId] successful:\n${serialization.write(request)}")
case Failure(ex) => log.error(s"Request [$requestId] failed::\n${serialization.write(request)}", ex)
```

rtkaczyk: I would say that logging the request body is of greater importance (business impact) than logging the success. The point of this logging is that when someone comes to us with a complaint that something doesn't work (either from Mallet, or interacting with RPC directly) then we can see that request and try to reproduce the problem.

loverdos: Of course, that's why in the snippet I suggested logging the body in both success and failure. But no real objection here. Logging the actual request is needed anyway.

rtkaczyk: Ah, I see, didn't notice the body there. But I still think it should be logged first in case something goes terribly wrong and we don't even get that far.

PR #578, opened 2018-08-20
Use rocks db native column families

KonradStaniec:
Pr:
1. Add usage of native rocks db column families, which improve mantis performance.
Some measurments (syncing 1h from master node)
 `Baseline - 250305 blocks synced`
`columnfamilies- 375553 blocks synced`
2. Add some missing `close` calls to rocks db objects,  which lack of could result in potential memory leaks

Change tested by syncing to the top of the chain via fast sync and then regular syncing for around 1day.


mmrozek (approved):
lgtm

agnikowa (approved):

PR #579, opened 2018-08-22
Add minimal benchmarking programs

KonradStaniec:
Simple programs to generate benchmarking data, and to benchmark run over this data. Just minimal implementation to get started, if there will be need it can be extended to some options, stats, or maybe even regression test. 

- Original target: `phase/release2_0`

### DIFF
--- a/src/benchmark/scala/io/iohk/ethereum/bench/Benchmark.scala
+++ b/src/benchmark/scala/io/iohk/ethereum/bench/Benchmark.scala
@@ -1,0 +1,162 @@
+package io.iohk.ethereum.bench
+
+//import io.iohk.ethereum.domain.BlockchainImpl
+import java.io.File
+
+import scala.concurrent.duration.Duration
+import java.util.concurrent.Executors
+
+import akka.actor.{Actor, ActorRef, ActorSystem}
+import io.iohk.ethereum.bench.BlockProducer.{GetBlocks, NewBlocks}
+import io.iohk.ethereum.consensus.{ConsensusConfigBuilder, StdConsensusBuilder}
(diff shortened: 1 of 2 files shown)


src/benchmark/scala/io/iohk/ethereum/bench/Benchmark.scala
```diff
@@ -0,0 +1,97 @@
...
+  val context = ExecutionContext.fromExecutor(exe)
+
+  val source = Source.fromFile("blocks.txt")
+
+  genesisDataLoader.loadGenesisData()
+
+  val lines = source.getLines()
+  val startTime = System.nanoTime()
+  lines.foreach { line =>
+    val block = Hex.decode(line).toBlock
```
kapke: I think that `Future.traverse` would be a good fit here: https://www.scala-lang.org/api/current/scala/concurrent/Future$.html#traverse[A,B,M[X]%3C:TraversableOnce[X]](in:M[A])(fn:A=%3Escala.concurrent.Future[B])(implicitcbf:scala.collection.generic.CanBuildFrom[M[A],B,M[B]],implicitexecutor:scala.concurrent.ExecutionContext):scala.concurrent.Future[M[B]]

Also there's one thing that concerns me a bit - In regular sync imports are called from Akka actor which can also impact its performance to some degree (implicit switches of threads between messages). I don't know if this is something that's relevant in context of this benchmark though. Maybe it would be a good way for future to set up `Importer` actor here with fake fetcher, etc.?

mirkoAlic:
At this point i'm not sure if we want to include this. But if at some point we need or we plan to add proper benchmarking, we could base on this.

PR #582, opened 2018-08-30
Handle newHeads websocket subscriptions

LukasGasior1:


src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -566,6 +569,11 @@ class RegularSync(
   private def notDownloading(): Boolean =
     headersQueue.isEmpty && waitingForActor.isEmpty && !resolvingBranches
 
+  private def publishEvents(newBlocks: Seq[Block]): Unit = {
```
rtkaczyk: Don't we need this in `handleBlocks` as well?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
```diff
@@ -566,6 +569,11 @@ class RegularSync(
...
+    newBlocks.foreach { newBlock =>
+      context.system.eventStream.publish(NewHead(newBlock))
```
rtkaczyk: Do we need to publish it for all blocks and not just the latest one?

rtkaczyk:
Could you provide more information about this (links)? How do I test it using web3?

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
LukasGasior1: For all blocks

src/main/scala/io/iohk/ethereum/blockchain/sync/RegularSync.scala
LukasGasior1: true, we do

LukasGasior1:
@rtkaczyk https://web3js.readthedocs.io/en/1.0/web3-eth-subscribe.html#subscribe-newblockheaders this uses the `newHeads` subscription

rtkaczyk:
Is it really called `newHeads`? The doc only mentions `newBlockHeaders`

LukasGasior1:
newHeads is the underlying json rpc call (https://github.com/ethereum/go-ethereum/wiki/RPC-PUB-SUB#newheads, newBlockHeaders is web3 naming

PR #586, opened 2018-09-10
Feature/ec 587 more efficient serializer for block headers storage

agnikowa:


KonradStaniec:
It looks good but it would be nice to know at least some rudimentary performance numbers, like 2-3h sync from master node.  ( It applies to other pr-s as well)

agnikowa:
I've done performance tests:
- baseline: 127221.1 blocks/10mins
- this branch: 133504.7 blocks/10mins
which gives ~5% speed up

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
```diff
@@ -13,18 +16,63 @@ import io.iohk.ethereum.network.p2p.messages.PV62.BlockHeaderImplicits._
...
+  override def valueDeserializer: IndexedSeq[Byte] => BlockHeader =
+    bytes => Unpickle[BlockHeader].fromBytes(ByteBuffer.wrap(bytes.toArray[Byte]))
 
   override protected def apply(dataSource: DataSource): BlockHeadersStorage = new BlockHeadersStorage(dataSource)
 
 }
 
 object BlockHeadersStorage {
   type BlockHeaderHash = ByteString
+  type BlockHeaderBody = (
```
mmrozek: Maybe you could add comments what the parameters mean.

PR #587, opened 2018-09-10
Feature/ec 588 more efficient serializer for receipt storage

agnikowa:


agnikowa:
I've done performance tests:
- baseline: 127221.1 blocks/10mins
- this branch: 133488.7 blocks/10mins
which gives ~5% speed up

PR #588, opened 2018-09-11
EIP-1234 Constantinople Difficulty Bomb Delay and Block Reward Adjust…

mmrozek:
…ment

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -403,6 +403,9 @@ mantis {
       # Reduced block reward after Byzantium fork
       first-era-reduced-block-reward = "3000000000000000000"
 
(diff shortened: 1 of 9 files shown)


src/main/resources/application.conf
```diff
@@ -403,6 +403,9 @@ mantis {
       # Reduced block reward after Byzantium fork
       first-era-reduced-block-reward = "3000000000000000000"
 
+      # Reduced block reward after Constantinople fork
+      first-era-constantinople-reduced-block-reward = "2000000000000000000"
```
KonradStaniec: You need to apply similar change in file `ethereumHF.conf` which ovverrides default values for eth chain

PR #589, opened 2018-09-13
Optimize reference counting performance

KonradStaniec:
Pr introduces concept of deathrow - structure which hold keys of all nodes which references droped to 0 during block execution. 
During pruning phase then, it is not needed to get all snapshots from db making many db calls, but one call for deathrow will be enough. Optimisation help especially at higher blocks.


src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -28,18 +28,22 @@ import io.iohk.ethereum.utils.Logger
...
   * and number of snapshots) and therefore, fast access to each snapshot individually.
   */
-class ReferenceCountNodeStorage(nodeStorage: NodesStorage, blockNumber: Option[BigInt] = None)
+class ReferenceCountNodeStorage(nodeStorage: NodesStorage,
+                                blockNumber: Option[BigInt] = None)
   extends NodesKeyValueStorage {
 
   import ReferenceCountNodeStorage._
 
+  private val deathRowKey = drRowKey(blockNumber.get)
```
agnikowa: I don't think this is safe, to call get here

KonradStaniec: yep, you are right my bad. it should have been in `update` method.

src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -98,6 +115,16 @@ class ReferenceCountNodeStorage(nodeStorage: NodesStorage, blockNumber: Option[B
 
 object ReferenceCountNodeStorage extends PruneSupport with Logger {
 
+  val nodeKeyLength = 32
+
+  def drRowKey(bn: BigInt): ByteString = {
+    ByteString("dr".getBytes()) ++ ByteString(bn.toByteArray)
```
mmrozek: Maybe deathRows could be stored in separate column family (separate cache, etc)?

KonradStaniec: it is possibility, but i do not think it is worth it, there is only as many deathrows as history blocks in pruning, which currently is 64, and separate column for only 64 key value pairs sound like overkill.

PR #591, opened 2018-09-20
Improve Mpt trie performance

KonradStaniec:
1st batch of improvements to mpt trie, changes:
Reduce unnecessery allocations of wrappers (`Option`, `Either`).
Use Arrays and while loops where possible.

changes to `MptNode` encoding enables further refactorings and possible performance gains but i did not want to make this pr to big. (extending `MptNode` required some additional exception throws in pattern matching clauses to satisfy compiler)

Performance gains (30min sync ):
Baseline - 288577blocks
Improved - 301057blocks
~5%

### DIFF
--- a/src/it/scala/io/iohk/ethereum/txExecTest/util/FixtureProvider.scala
+++ b/src/it/scala/io/iohk/ethereum/txExecTest/util/FixtureProvider.scala
@@ -84,8 +84,8 @@ object FixtureProvider {
             }
           }
 
-        case None =>
         case _ =>
+
       }
 
       traverse(block.header.stateRoot)
--- a/src/it/scala/io/iohk/ethereum/txExecTest/util/FixtureProvider.scala
+++ b/src/it/scala/io/iohk/ethereum/txExecTest/util/FixtureProvider.scala
@@ -14,7 +14,7 @@ import MptNodeEncoders._
 import ReceiptImplicits._
 import io.iohk.ethereum.db.cache.MapCaches
 import io.iohk.ethereum.db.storage.pruning.{ArchivePruning, PruningMode}
-import io.iohk.ethereum.mpt.{BranchNode, ExtensionNode, LeafNode, MptNode}
+import io.iohk.ethereum.mpt.{BranchNode, ExtensionNode, HashNode, LeafNode, MptNode}
 import org.bouncycastle.util.encoders.Hex
 
 import scala.io.Source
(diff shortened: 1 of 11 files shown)


src/main/scala/io/iohk/ethereum/mpt/Node.scala
```diff
@@ -25,6 +27,8 @@ sealed abstract class MptNode {
     val encoded = encode
     if (encoded.length < 32) ByteString(encoded) else ByteString(hash)
   }
+
+  def isNull: Boolean
```
agnikowa: probably you could set it as false, and only override it to true in `NullNode`

src/main/scala/io/iohk/ethereum/mpt/Node.scala
```diff
@@ -62,10 +72,50 @@ case class BranchNode(children: Seq[Option[Either[ByteString, MptNode]]], termin
     */
   def updateChild(childIndex: Int, childNode: MptNode): BranchNode = {
     val childCapped = childNode.capped
-    BranchNode(children.updated(childIndex, Some(if (childCapped.length == 32) Left(childCapped) else Right(childNode))), terminator)
+    val updatedChild = if (childCapped.length == 32) HashNode(childCapped) else childNode
```
agnikowa: maybe you should extract this 32 to some value about capped length

agnikowa: also `if (childCapped.length == 32) HashNode(childCapped) else childNode` structure is used 3 times (on 129 and 160 line)

src/it/scala/io/iohk/ethereum/txExecTest/util/FixtureProvider.scala
```diff
@@ -85,6 +85,7 @@ object FixtureProvider {
           }
 
         case None =>
+        case _ =>
```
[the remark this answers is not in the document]
KonradStaniec: Actually this is only used in integration testing so if something will go wrong it will dealt upstream in tests.

src/main/scala/io/iohk/ethereum/mpt/HexPrefix.scala
```diff
@@ -42,25 +42,44 @@ object HexPrefix {
...
+    * Needs to be as fast possible, which requires usage of var's and mutable arrays.
     * @param nibbles byte[]
     * @return array with bytes combining pairs of nibbles
     *
     */
   def nibblesToBytes(nibbles: Array[Byte]): Array[Byte] = {
     require(nibbles.length % 2 == 0)
-    nibbles.grouped(2).map{case Array(n1,n2) => (16*n1 + n2).toByte}.toArray
+    val newArray = new Array[Byte](nibbles.length / 2)
+    var i = 0
```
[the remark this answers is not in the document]
KonradStaniec: acutally in  my experinece foreachs are slower than plain old while loops, it is also visible in other benchmarsks: http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#foreach-performance

PR #592, opened 2018-09-21
Optimise ethash

KonradStaniec:
Optimisation to our ethash alg. After this improvments, bottleneck is localized in calculating `kec512` hash.
Baseline (20Min): `204609 blocks`
Improved (20min):  `225601 blocks`

Roughly 10% improvement.


src/main/scala/io/iohk/ethereum/utils/ByteUtils.scala
```diff
@@ -70,21 +76,93 @@ object ByteUtils {
...
+    val ret = new Array[Byte](ints.length * 4)
+    intsToBytesMut(ints, ret, bigEndian)
+    ret
   }
 
   def getIntFromWord(arr: Array[Byte]): Int = {
     ByteBuffer.wrap(arr, 0, 4).order(ByteOrder.LITTLE_ENDIAN).getInt
   }
 
+  def intsToBytesMut(arr: Array[Int], b: Array[Byte], bigEndian: Boolean) {
```
agnikowa: I would suggest to add comments to all this methods with `var`s and `while`s  or one on top of the file - that it is because of performance, so it wouldn't be changed if someone would think of refactoring

src/main/scala/io/iohk/ethereum/utils/ByteUtils.scala
```diff
@@ -70,21 +85,115 @@ object ByteUtils {
...
+        off += 1
+        b(off) = (ii & 0xFF).toByte
+        off += 1
+
+        i = i + 1
+      }
+    }
+  }
+
+
```
agnikowa: to much enters

src/main/scala/io/iohk/ethereum/utils/ByteUtils.scala
```diff
@@ -70,21 +85,115 @@ object ByteUtils {
...
+    val ret = new Array[Int](bytes.length / 4)
+    bytesToIntsMut(bytes, ret, bigEndian)
+    ret
   }
 
-  def intsToBytes(ints: Array[Int]): Array[Byte] = {
-    val buf = ByteBuffer.allocate(ints.length * 4).order(ByteOrder.LITTLE_ENDIAN)
-    buf.asIntBuffer().put(ints)
-    buf.array()
+  def intsToBytes(ints: Array[Int], bigEndian: Boolean): Array[Byte] = {
```
agnikowa: you have almost everywhere `bigEndian` set to false, maybe you could add setting it to false?

KonradStaniec: to be honest i would rather have the user of this function to make decision about which conversions should be used

PR #596, opened 2018-09-28
Divide RegularSync into Fetcher and Importer actors

kapke:
This PR divides RegularSync actor in a Fetcher and Importer ones.
While most code of Importer is copy&paste of code from current RegularSync responsible for importing, for Fetcher it looks differently.

Whole idea of how it's supposed to work is as follows:
1. Fetcher receives signal to start with number of block it can start with
2. Fetcher fetches headers, it has only headers now
3. Fetcher has headers, it can fetch matching block bodies and new headers in the same time
4. Fetcher received bodies, it takes matching headers and puts whole block in buffer
5. Importer polls from time to time to pick blocks it can import
6. Once Importer receives some blocks, imports them and immediately picks next bunch of blocks.
7. At some point fetcher will have drained buffer and will not receive headers as requested - that means it's on top, so:
  a) it sends proper signal to importer
  b) importer knows now that it can handle mined blocks
  c) fetcher will forward messages with new blocks in the network
  d) if top is lost - proper signal is sent to importer

Of course - there are some additional things to handle:
 - [x] - error propagation to fetcher (peer blacklisting, redownload blocks that were imported with error)
 - [x] - resolving missing state node without need to refetch all blocks
 - [x] - make broadcaster an actor with PeerListSupport so Importer doesn't need to know almost anything about networking
 - [x] - importing new block and mined block differs only in log messages, so simple abstraction for creating them would allow to remove that duplication.
 - [x] - clean up in SyncController
 - [x] - ensure all places in current code using messages from RegularSync are adjusted
 - [x] - write tests
 - [x] - properly handle ommers


src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,270 @@
...
+  sealed trait ImporterMsg
+  case object Start extends ImporterMsg
+  case object OnTop extends ImporterMsg
+  case object NotOnTop extends ImporterMsg
+  case class MinedBlock(block: Block) extends ImporterMsg
+  case class ImportNewBlock(block: Block, peerId: PeerId) extends ImporterMsg
+
+  case class ImporterState(isOnTop: Boolean, importing: Boolean)
+  object ImporterState {
+    val initial: ImporterState = ImporterState(isOnTop = false, importing = false)
```
mmrozek: The same thing. It could be much simpler if you move these methods to case class

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -0,0 +1,430 @@
...
+  val bufferSize = 10000
+
+  scheduler.schedule(1.second, 1.second, self, PrintStatus)
+  peerEventBus ! Subscribe(MessageClassifier(Set(NewBlock.code, NewBlockHashes.code), PeerSelector.AllPeers))
+
+  override def receive: Receive = idle(FetcherState.initial)
+
+  private def idle(state: FetcherState): Receive =
+    handleCommonMessages(state, idle) orElse {
+      case Start(blockNr) => state |> FetcherState.withLastBlock(blockNr) |> fetchBlocks
```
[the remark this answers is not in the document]
kapke: It's usually called "pipeline operator". It's defined in mouse - bunch of helpers to cats. It works like pipe in bash: takes value on the left and apply it to function on the right side. It's really cool in situations where couple of functions needs to be composed ad-hoc to compute some result. It also allows to avoid temporary variables for intermediate steps (which can be very hard to name properly)

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -0,0 +1,447 @@
...
+    case RequestFailed(peer, reason) if state.isStateNodeFetcher(sender()) =>
+      log.debug(s"Failed state node request from peer {} with reason {}", peer, reason)
+      blacklistIfHandshaked(peer, reason)
+      state.stateNodeFetcher.foreach(fetcher => {
+        fetchStateNode(fetcher.hash, fetcher.replyTo, state.withoutHeadersFetcher())
+      })
+  }
+
+  private def handleNewBlockMessages(state: FetcherState): Receive = {
+    case MessageFromPeer(NewBlockHashes(_), _) =>
```
KonradStaniec: Why such handling of new block hashes ? In principle we should request header for that hash, and then body and then do import.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -0,0 +1,447 @@
...
+      blacklist(blockProvider.id, syncConfig.blacklistDuration, reason)
+      fetchBlocks(newState)
+    case FetchBlocks => fetchBlocks(state)
+  }
+
+  private def handleHeadersMessages(state: FetcherState): Receive = {
+    case ResponseReceived(peer, BlockHeaders(headers), _) if state.isHeadersFetcher(sender()) =>
+      log.debug("Fetched headers for blocks {} by {}", headers.map(_.number).mkString(","), sender())
+
+      val newState = state.validatedHeaders(headers) match {
```
KonradStaniec: I do not see anywhere code to resolve branches - in old sync it was done by method `Ledger.resolveBranches`. 
Main idea is that you can have current best imported chain `(blocknumber, blockHash, parentHash)`:
`Genesis -> (1 , a, genesis)-> (2, b, a) -> (3,  c, b) `
but canonical chain now looks like:
`Genesis -> (1 , a, genesis)-> (2, d, a) -> (3,  e, d) `
so when you now receive header `(4, f, e)` ,  it will now fail vaildation , as we do not have its parent in blockchain. Worse, we will receive this header from all peers (as it is now canonical) and we wont be able move forward. 
As you look at old regular sync in that situation, we were asking peer for `n` headers  from current best to best - n , and then we were looking for common parent to determine best chain.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -0,0 +1,447 @@
...
+    if (!state.isFetching && !newState.hasFetchedTopHeader) {
+      scheduler.scheduleOnce(10.seconds, context.self, FetchBlocks)
+    }
+
+    context become started(newState)
+  }
+
+  private def tryFetchHeaders(fetcherState: FetcherState): FetcherState =
+    Some(fetcherState)
+      .filterNot(_.isFetchingHeaders)
```
KonradStaniec: tbh i do not like this filter trains, on each one of them i must consider hard what filter does, does it leave elements which pass the predicate or other way around.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,314 @@
...
+      pendingTransactionsManager ! RemoveTransactions(block.body.transactionList)
+    }
+  }
+
+  private def importWith(state: ImporterState, importFuture: => Future[Behavior]): Unit = {
+    val newState = ImporterState.importingBlocks(state)
+    context become running(newState)
+    importFuture.onComplete {
+      case Failure(ex) => throw ex
+      case Success(behavior) => context become behavior(ImporterState.notImportingBlocks(newState))
```
mmrozek: You shouldn't switch context in `onComplete`. It isn't processed by the same thread.

From Akka documentation:
```
/**
...
   * WARNING: Only valid within the Actor itself, so do not close over it and
   * publish it to other threads!
...
  implicit val context: ActorContext
```

src/main/scala/io/iohk/ethereum/blockchain/sync/PeerListSupport.scala
```diff
@@ -42,4 +42,16 @@ trait PeerListSupport {
     case PeerDisconnected(peerId) if handshakedPeers.exists(_._1.id == peerId) =>
       removePeer(peerId)
   }
+
+  def peerById(peerId: PeerId): Option[Peer] =
+    peersToDownloadFrom find {
```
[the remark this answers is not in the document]
kapke: I just did a copy&pate from older code and I did not know how can it be simplified - that's what I was looking for.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
kapke: Switching it into plain `filter` would be a good fix to you? TBH I had the same problem at the beginning but now it's pretty clear to me. Also it seems nicer to me to write (and read) `filterNot(sth)` instead of `filter(!sth)`

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
kapke: Good catch

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
kapke: I remember there was something weird about handling this message and it appeared to me better to just schedule next fetch instead of handling it directly. I will check it in detail and provide more detailed response.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
KonradStaniec: yep simple filter would be good 👍

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
kapke: The issue is that request for headers still has to be made. So instead of fetching headers for that specific hashes (while it's possible that there are some missing headers in between) is worse option that just doing a request for next bunch of headers and then bodies.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,350 @@
...
+              "block import error {}, current best block is {} and best block with hash is {}",
+              err,
+              blockchain.getBestBlockNumber(),
+              getStartingBlockNumber())
+            val notImportedBlocks = blocks.drop(importedBlocks.size)
+
+            err match {
+              case e: MissingNodeException =>
+                fetcher ! BlockFetcher.FetchStateNode(e.hash)
+                ResolvingMissingNode(NonEmptyList(notImportedBlocks.head, notImportedBlocks.tail))
```
agnikowa: what if `notImportedBlocks` is empty? same in line 152

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,350 @@
...
+        .successful(resolveBranch(blocks))
+        .flatMap({
+          case Right(blocksToImport) => handleBlocksImport(blocksToImport)
+          case Left(resolvingFrom) => Future.successful(ResolvingBranch(resolvingFrom))
+        })
+    })
+
+  private def handleBlocksImport(blocks: List[Block]): Future[NewBehavior] =
+    tryImportBlocks(blocks)
+      .map(value => {
```
agnikowa: why not use only curly brace here?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,350 @@
...
+            tryImportBlocks(restOfBlocks, blocks.head :: importedBlocks)
+
+          case ChainReorganised(_, newBranch, _) =>
+            tryImportBlocks(restOfBlocks, newBranch.reverse ::: importedBlocks)
+
+          case DuplicateBlock | BlockEnqueued =>
+            tryImportBlocks(restOfBlocks, importedBlocks)
+
+          case err @ (UnknownParent | BlockImportFailed(_)) =>
+            log.debug(s"block ${Block.number(blocks.head)} import failed")
```
agnikowa: start the message with capital letter and maybe consider extracting `blocks.head` to some val

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,350 @@
...
+    def doLog(entry: ImportMessages.LogEntry): Unit = log.log(entry._1, entry._2)
+
+    importWith({
+      doLog(importMessages.preImport())
+      ledger
+        .importBlock(block)(context.dispatcher)
+        .tap(importMessages.messageForImportResult _ andThen doLog)
+        .tap {
+          case BlockImportedToTop(importedBlocksData) =>
+            val (blocks, receipts) = importedBlocksData.map(data => (data.block, data.td)).unzip
```
agnikowa: are this receipts? because you're getting `data.td`, so the name is misleading

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,350 @@
...
+      ommersPool ! RemoveOmmers(block.header :: block.body.uncleNodesList.toList)
+      pendingTransactionsManager ! RemoveTransactions(block.body.transactionList)
+    })
+  }
+
+  private def importWith(importFuture: => Future[NewBehavior])(state: ImporterState): Unit = {
+    val newState = ImporterState.importingBlocks(state)
+    context become running(newState)
+    importFuture.onComplete {
+      case Failure(ex) => throw ex
```
agnikowa: this is only handled in one function (importBlock) and in other is omitted (importBlocks), is this on purpose?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,350 @@
...
+    broadcastNewBlocks(newBlocks)
+  }
+
+  private def broadcastNewBlocks(blocks: List[NewBlock]): Unit = broadcaster ! BroadcastBlocks(blocks)
+
+  private def updateTxAndOmmerPools(blocksAdded: Seq[Block], blocksRemoved: Seq[Block]): Unit = {
+    blocksRemoved.headOption.foreach(block => ommersPool ! AddOmmers(block.header))
+    blocksRemoved.foreach(block => pendingTransactionsManager ! AddTransactions(block.body.transactionList.toSet))
+
+    blocksAdded.foreach(block => {
```
agnikowa: why using `(` and `{` instead of one `{`?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,350 @@
...
+        val currentBlock = Block.number(blocks.head)
+        val goingBackTo = currentBlock - syncConfig.branchResolutionRequestSize
+        val msg = s"Unknown branch, going back to block nr $goingBackTo in order to resolve branches"
+
+        log.debug(msg)
+        fetcher ! BlockFetcher.InvalidateBlocksFrom(goingBackTo, msg, shouldBlacklist = false)
+        Left(goingBackTo)
+      case InvalidBranch =>
+        val goingBackTo = Block.number(blocks.head)
+        val msg = s"invalid branch, going back to $goingBackTo"
```
agnikowa: should start with capital letter as in above `UnknownBranch` case

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/NewRegularSync.scala
```diff
@@ -0,0 +1,72 @@
...
+    scheduler: Scheduler
+) extends Actor
+    with ActorLogging {
+  import NewRegularSync._
+
+  val fetcher: ActorRef =
+    context.actorOf(BlockFetcher.props(peersClient, peerEventBus, syncConfig, scheduler), "block-fetcher")
+  val broadcaster: ActorRef = context.actorOf(
+    BlockBroadcasterActor
+      .props(new BlockBroadcast(etcPeerManager, syncConfig), peerEventBus, etcPeerManager, syncConfig, scheduler))
```
agnikowa: why not name `BlockBroadcasterActor` also?

src/main/scala/io/iohk/ethereum/utils/FutureOps.scala
```diff
@@ -0,0 +1,17 @@
...
+import akka.actor.Scheduler
+
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.concurrent.duration.FiniteDuration
+
+object FutureOps {
+  implicit class FutureOps[T](future: Future[T]) {
+
+    def delayedBy(delay: FiniteDuration)(implicit ec: ExecutionContext, scheduler: Scheduler): Future[T] =
+      future.flatMap(value => {
```
agnikowa: maybe this doubled brackets are not needed

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -0,0 +1,477 @@
...
+  }
+
+  private def started(state: FetcherState): Receive =
+    handleCommonMessages(Some(state)) orElse
+      handleCommands(state) orElse
+      handleNewBlockMessages(state) orElse
+      handleHeadersMessages(state) orElse
+      handleBodiesMessages(state) orElse
+      handleStateNodeMessages(state) orElse {
+      case msg => log.debug("Unhandled msg {} from {}", msg, sender())
```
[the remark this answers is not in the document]
kapke: It will work - `log.debug` takes parameters by value, not by name. Other thing is that most probably this case is going to be removed.

src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
```diff
@@ -0,0 +1,151 @@
...
+        syncConfig.peerResponseTimeout,
+        etcPeerManager,
+        peerEventBus,
+        requestMsg = requestMsg,
+        responseMsgCode = responseMsgCode
+      )(classTag, scheduler, toSerializable))
+
+  private def selectPeer(peerSelector: PeerSelector): Option[Peer] =
+    peerSelector match {
+      case BestPeer => bestPeer(peersToDownloadFrom)
```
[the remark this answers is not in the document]
kapke: To make this code easier to update when new case is added - compiler should notify that there's missing case

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -0,0 +1,477 @@
...
+    case BlockImportFailed(blockNr, reason) =>
+      val (peerId, newState) = state.invalidateBlocksFrom(blockNr)
+      peerId.foreach(id => peersClient ! BlacklistPeer(id, reason))
+      fetchBlocks(newState)
+  }
+
+  private def handlePickedBlocks(state: FetcherState)(
+      pickResult: Option[(NonEmptyList[Block], FetcherState)]): FetcherState =
+    pickResult
+      .tap({
```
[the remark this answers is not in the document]
kapke: It seems that they are not

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,350 @@
...
+        case ResolvingMissingNode(blocksToRetry) => resolvingMissingNode(blocksToRetry)
+        case ResolvingBranch(from) => resolvingBranch(from)
+      }
+      context become behavior(newState)
+    case PickBlocks => pickBlocks(state)
+  }
+
+  private def resolvingMissingNode(blocksToRetry: NonEmptyList[Block])(state: ImporterState): Receive = {
+    case BlockFetcher.FetchedStateNode(nodeData) =>
+      val node = nodeData.values.head
```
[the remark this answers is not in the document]
kapke: Yes, it's validated in fetcher before sending it further

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
kapke: It won't happen if there is catched error.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
kapke: It's handled - notice that this function is a generic wrapper for both `importBlocks` and `importBlock`

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
kapke: But how? I need more than one reference to block, so I can't use `_`. I also have to put `{` and `}` due to 2 statements in this block

src/main/scala/io/iohk/ethereum/utils/FutureOps.scala
kapke: AFAIK they are

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/NewRegularSync.scala
```diff
@@ -0,0 +1,72 @@
...
+  val importer: ActorRef =
+    context.actorOf(
+      BlockImporter.props(fetcher, ledger, blockchain, syncConfig, ommersPool, broadcaster, pendingTransactionsManager),
+      "block-importer")
+
+  override def receive: Receive = {
+    case Start => importer ! BlockImporter.Start
+    case MinedBlock(block) => importer ! BlockImporter.MinedBlock(block)
```
KonradStaniec: Shouldn't you update also update `EthashMiner` to send `NewRegularSync.MinedBlock` instead of `RegularSync.MinedBlock` ?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -0,0 +1,477 @@
...
+        case (blocks, newState) =>
+          sender() ! PickedBlocks(blocks)
+          newState.importer ! (if (newState.isOnTop) OnTop else NotOnTop)
+      })
+      .fold(state)(_._2)
+
+  private def fetchBlocks(state: FetcherState): Unit = {
+    log.debug("Fetching blocks")
+    val newState = state |> tryFetchHeaders |> tryFetchBodies
```
KonradStaniec: I have mixed feeling about this construct, it looks like simple function application but actually it makes requests, schedule retries etc.  May export it to separate function and add some scala doc explain what is happening behind scenes ?

src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
```diff
@@ -0,0 +1,151 @@
...
+            val handler =
+              makeRequest(peer, message, responseMsgCode(message), toSerializable)(scheduler, responseClassTag(message))
+            val newRequesters = requesters + (handler -> requester)
+            context become running(newRequesters)
+          case None =>
+            log.debug("No suitable peer found to issue a request")
+            requester ! NoSuitablePeer
+        }
+      case PeerRequestHandler.ResponseReceived(peer, message, _) =>
+        val requestHandler = sender()
```
KonradStaniec: code under `PeerRequestHandler.ResponseReceived` is almost duplication of `PeerRequestHandler.RequestFailed(peer, reason)`, maybe extract it to separate method ?

kapke: Fixed

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/NewRegularSync.scala
kapke: done

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,350 @@
...
+        val goingBackTo = Block.number(blocks.head)
+        val msg = s"invalid branch, going back to $goingBackTo"
+
+        log.debug(msg)
+        fetcher ! BlockFetcher.InvalidateBlocksFrom(goingBackTo, msg)
+        Right(Nil)
+    }
+
+  @tailrec
+  private def getStartingBlockNumber(current: BigInt = blockchain.getBestBlockNumber()): BigInt =
```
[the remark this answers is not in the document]
kapke: It's removed now. I added this to debug that bug with inconsistencies

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,350 @@
...
+    pendingTransactionsManager: ActorRef
+) extends Actor
+    with ActorLogging {
+  import BlockImporter._
+
+  implicit val ec: ExecutionContext = context.dispatcher
+
+  context.setReceiveTimeout(syncConfig.syncRetryInterval)
+
+  val blocksBatchSize = 50
```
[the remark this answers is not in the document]
kapke: It's now

src/main/scala/io/iohk/ethereum/utils/FutureOps.scala
kapke: Or not really :) Fixed

src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
```diff
@@ -0,0 +1,153 @@
...
+    statusSchedule.cancel()
+  }
+
+  def running(requesters: Requesters): Receive =
+    handleBlacklistMessages orElse handlePeerListMessages orElse {
+      case PrintStatus =>
+        log.debug(
+          """PeersClient status:
+          | requests: {}
+          | available peers: {}""".stripMargin,
```
agnikowa: probably you could skip this `stripMargin` and have it in one line

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -0,0 +1,185 @@
...
+    if (nr > knownTop) {
+      withKnownTopAt(nr)
+    } else {
+      this
+    }
+  def withPossibleNewTopAt(nr: Option[BigInt]): BlockFetcherState = nr.map(withPossibleNewTopAt).getOrElse(this)
+
+  def withPeerForBlocks(peerId: PeerId, blocks: Seq[BigInt]): BlockFetcherState =
+    copy(
+      blockProviders = blockProviders ++ blocks.map(block => block -> peerId)
```
agnikowa: probably could fit in one line instead of 3

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,331 @@
...
+  private def handleTopMessages(state: ImporterState, currentBehavior: Behavior): Receive = {
+    case OnTop => context become currentBehavior(ImporterState.onTop(state))
+    case NotOnTop => context become currentBehavior(ImporterState.notOnTop(state))
+  }
+
+  private def running(state: ImporterState): Receive = handleTopMessages(state, running) orElse {
+    case ReceiveTimeout => self ! PickBlocks
+    case BlockFetcher.PickedBlocks(blocks) => importBlocks(blocks)(state)
+    case MinedBlock(block) =>
+      if (!state.importing && state.isOnTop) {
```
agnikowa: probably could be a method in `ImporterState`, because it is also used below

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,331 @@
...
+    log.debug("Starting Regular Sync, current best block is {}", startingBlockNumber)
+    fetcher ! BlockFetcher.Start(self, startingBlockNumber)
+    context become running(ImporterState.initial)
+  }
+
+  private def pickBlocks(state: ImporterState): Unit = {
+    val msg = state.resolvingBranchFrom.fold[BlockFetcher.FetchMsg](
+      BlockFetcher.PickBlocks(syncConfig.blocksBatchSize)
+    )(
+      from => BlockFetcher.StrictPickBlocks(from, startingBlockNumber)
```
agnikowa: maybe it could be in one line (99) instead of 3

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -0,0 +1,331 @@
...
+            tryImportBlocks(restOfBlocks, blocks.head :: importedBlocks)
+
+          case ChainReorganised(_, newBranch, _) =>
+            tryImportBlocks(restOfBlocks, newBranch.reverse ::: importedBlocks)
+
+          case DuplicateBlock | BlockEnqueued =>
+            tryImportBlocks(restOfBlocks, importedBlocks)
+
+          case err @ (UnknownParent | BlockImportFailed(_)) =>
+            log.error(s"block ${Block.number(blocks.head)} import failed")
```
agnikowa: should start with uppercase and probably you could use `{}` like in the rest of the logs here

src/main/scala/io/iohk/ethereum/domain/Block.scala
```diff
@@ -50,4 +51,10 @@ object Block {
   }
 
   def size(block: Block): Long = (block.toBytes: Array[Byte]).length
+
+  def number(block: Block): BigInt = block.header.number
```
agnikowa: maybe put this into case class?

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/NewRegularSyncSpec.scala
```diff
@@ -0,0 +1,633 @@
...
+        peersClient.setAutoPilot(new BranchResolutionAutoPilot(didResponseWithNewBranch = false, testBlocks))
+
+        Await.result(ledger.importBlock(genesis), remainingOrDefault)
+
+        regularSync ! RegularSync.Start
+
+        peerEventBus.expectMsgClass(classOf[Subscribe])
+        peerEventBus.reply(MessageFromPeer(NewBlock(testBlocks.last, Block.number(testBlocks.last)), defaultPeer.id))
+
+        awaitCond(ledger.bestBlock == alternativeBlocks.last, 10 minutes)
```
agnikowa: isn't 10minutes too long?

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/NewRegularSyncSpec.scala
```diff
@@ -0,0 +1,633 @@
...
+
+  abstract class Fixture(_system: ActorSystem)
+      extends TestKitBase
+      with EphemBlockchainTestSetup
+      with TestSyncConfig
+      with SecureRandomBuilder {
+    implicit override lazy val system: ActorSystem = _system
+    override lazy val syncConfig: SyncConfig =
+      defaultSyncConfig.copy(blockHeadersPerRequest = 2, blockBodiesPerRequest = 2)
+    val handshakedPeers: PeersMap = (0 to 5).map(peerId).map(getPeer).map(peer => (peer, getPeerInfo(peer))).toMap
```
agnikowa: probably you could use one map instead of 3

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/NewRegularSyncSpec.scala
```diff
@@ -0,0 +1,633 @@
...
+          MessageFromPeer(NewBlockHashes(List(BlockHash(Block.hash(newBlock), Block.number(newBlock)))), defaultPeer.id)
+
+        peersClient.expectMsgPF() {
+          case PeersClient.Request(GetBlockHeaders(_, _, _, _), _, _) => true
+        }
+      }
+    }
+  }
+
+  abstract class Fixture(_system: ActorSystem)
```
agnikowa: maybe you could extract this as `NewRegularSyncFixture` because this file is already big

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/NewRegularSyncSpec.scala
kapke: It is. I forgot to change that value after debugging.

src/main/resources/application.conf
```diff
@@ -119,7 +119,7 @@ mantis {
       max-mpt-components-per-message = 200
 
       # Maximum number of peers this node can connect to
-      max-outgoing-peers = 40
+      max-outgoing-peers = 200
```
KonradStaniec: Is this intentional ? I think we can increase it but i would be careful as 200 peers could easily clog the system. Maybe set it to `80` ?

PR #597, opened 2018-09-28
Merkele patricia trie refactoring/improvements

KonradStaniec:
Refactoring takes into account trie-ness of mpt trie, and defines generic dispatch method which can traverse trie in depth by separatly defined visitors.
Pros:
- Separation of trie traversing from from operation on nodes
- We do not hash/encode every node but only those changed along the way from root
- We can easliy define and compose different visitors i.e visitor which reads whole trie into memory, which can be used in tests, or visitor which will track parent-children relationschips which can be used in pruning.
Cons:
-Visitor pattern is a little bit verbose


src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainActor.scala
```diff
@@ -111,8 +111,8 @@ class DumpChainActor(peerManager: ActorRef, peerMessageBus: ActorRef, startBlock
       val nodes = NodeData(stateNodes).values.indices.map(i => NodeData(stateNodes).getMptNode(i))
 
       val children = nodes.flatMap {
-        case n: BranchNode => n.children.collect { case HashNode(h) => h }
-        case ExtensionNode(_, HashNode(h), _, _) => Seq(h)
+        case n: BranchNode => n.children.collect { case HashNode(h) => ByteString(h) }
```
mmrozek: When we are dealing with hashes we are doing a lot of boxing/unboxing or making a conversions. Probably is not the issue for now but maybe we should think about using one type everywhere

src/main/scala/io/iohk/ethereum/mpt/Node.scala
```diff
@@ -4,64 +4,78 @@ import java.util
...
+      false
+    } else{
+      val compared = obj.asInstanceOf[MptNode]
+      hash sameElements compared.hash
+    }
   }
 
-  def isNull: Boolean = false
+  override def hashCode(): Int = {
+    17 + util.Arrays.hashCode(hash)
```
mmrozek: Why 17?

src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainActor.scala
KonradStaniec: For now i focused to avoid as much conversions as possible inside mpt trie, and left those outside intact. If benchmarks will show that that they are somehow bottleneck then it can be tackled in other pr.

src/main/scala/io/iohk/ethereum/mpt/Node.scala
KonradStaniec: Just needed some constant to avoid having `MptNode` having same hashcode as it hash ( which is `Array[Byte]`)
I chose 17 as i remeberd it was used in `Effective Java` - `Store some constant nonzero value, say, 17, in an int variable called result.`

src/main/scala/io/iohk/ethereum/mpt/MerklePatriciaTrie.scala
```diff
@@ -42,131 +46,44 @@ object MerklePatriciaTrie {
...
-    case extension: ExtensionNode =>
-      RLPList(HexPrefix.encode(nibbles = extension.sharedKey.toArray[Byte], isLeaf = false), this.nodeEnc.encode(extension.next))
-    case branch: BranchNode =>
-      val encoded = new Array[RLPEncodeable](MerklePatriciaTrie.ListSize)
-      var i = 0
-      while (i < BranchNode.numberOfChildren) {
-        val nodeEncoded = this.nodeEnc.encode(branch.children(i))
-        encoded(i) = nodeEncoded
-        i = i + 1
+  private def getFromHash(nodeId: Array[Byte], source: NodesKeyValueStorage): MptNode =
```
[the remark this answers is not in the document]
KonradStaniec: There are two reasons:
1. If the node requested by mpt trie is not in db ( or some cache), it is means that our trie is inconsistent state and all further operations on it do not make much sense. If we would return `None`  upstream code would not have way of knowing if value was not present in db or the trie was inconsistent.
2. We have special mechanism in regular sync which catches those exceptions, and download nodes on demand from network. (we have it because fastSync or regulars sync is not perfect and sometimes we do not have all nodes needed)

src/main/scala/io/iohk/ethereum/mpt/Node.scala
agnikowa: maybe add this in a comment, so it would be clear in the future

PR #604, opened 2018-10-17
Create separate layer to handle mpt nodes

KonradStaniec:
this pr clears current mess with node handling which was scattered accross many places.
Now there is: 
- `MptStorage` which  responsible for providing nodes to mpt trie, and serlialiation/deserialization of those nodes.
- `StateStorage` which creates and manages different kind of possible storages (archive, reference counting and specials case of reference counting during fast sync and genesis loading)


src/main/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorage.scala
```diff
@@ -28,19 +28,13 @@ import io.iohk.ethereum.utils.Logger
...
-                                blockNumber: Option[BigInt] = None)
-  extends NodesKeyValueStorage {
+class ReferenceCountNodeStorage(nodeStorage: NodesStorage, bn: BigInt) extends NodesKeyValueStorage {
 
   import ReferenceCountNodeStorage._
 
-  override def get(key: ByteString): Option[NodeEncoded] = nodeStorage.get(key).map(storedNodeFromBytes).map(_.nodeEncoded.toArray)
-
-  override def update(toRemove: Seq[NodeHash], toUpsert: Seq[(NodeHash, NodeEncoded)]): NodesKeyValueStorage = {
+  def get(key: ByteString): Option[NodeEncoded] = nodeStorage.get(key).map(storedNodeFromBytes).map(_.nodeEncoded.toArray)
```
mmrozek: It could be in 1 `map`. It always one unboxing less

src/main/scala/io/iohk/ethereum/db/storage/StateStorage.scala
```diff
@@ -0,0 +1,128 @@
...
+                                   private val cachedNodeStorage: CachedNodeStorage,
+                                   private val pruningHistory: BigInt) extends StateStorage {
+  override def forcePersist: Unit = {
+    cachedNodeStorage.forcePersist()
+  }
+
+  override def onBlockSave(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlock: () => Unit): Unit = {
+    val blockToPrune = bn - pruningHistory
+
+    if (blockToPrune <= currentBestSavedBlock){
```
mmrozek: It could be simplified to `ReferenceCountNodeStorage.prune(blockToPrune, cachedNodeStorage, inMemory = blockToPrune > currentBestSavedBlock)`

src/main/scala/io/iohk/ethereum/db/storage/StateStorage.scala
```diff
@@ -0,0 +1,128 @@
...
+      ReferenceCountNodeStorage.prune(blockToPrune, cachedNodeStorage, inMemory = true)
+    }
+
+    if (cachedNodeStorage.persist()) {
+      updateBestBlock()
+    }
+  }
+
+  override def onBlockRollback(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlock: () => Unit): Unit = {
+    if (bn <= currentBestSavedBlock)
```
mmrozek: ditto

src/test/scala/io/iohk/ethereum/db/storage/ReadOnlyNodeStorageSpec.scala
```diff
@@ -2,35 +2,39 @@ package io.iohk.ethereum.db.storage
...
 import akka.util.ByteString
 import io.iohk.ethereum.db.dataSource.EphemDataSource
+import io.iohk.ethereum.mpt.LeafNode
 import org.scalatest.{FlatSpec, Matchers}
 
 class ReadOnlyNodeStorageSpec extends FlatSpec with Matchers {
 
   "ReadOnlyNodeStorage" should "not update dataSource" in new TestSetup {
-    readOnlyNodeStorage.put(ByteString("key1"), ByteString("Value1").toArray)
+    val newLeaf = LeafNode(ByteString(1), ByteString(1))
```
agnikowa: this `newLeaf` probably could be extracted for both this test cases

src/test/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorageSpec.scala
```diff
@@ -28,20 +28,20 @@ class ReferenceCountNodeStorageSpec extends FlatSpec with Matchers {
   }
 
   it should "not remove a key that was inserted after deletion when pruning" in new TestSetup {
-    val storage = new ReferenceCountNodeStorage(nodeStorage, blockNumber = Some(1))
+    val storage = new ReferenceCountNodeStorage(nodeStorage, bn = 1)
```
agnikowa: maybe it could be some `mkNodeStorage` method in `TestSetup` that accept only block number, because nodeStorage is unchanged

src/test/scala/io/iohk/ethereum/db/storage/StateStorageSpec.scala
```diff
@@ -0,0 +1,168 @@
...
+
+class StateStorageSpec extends FlatSpec with Matchers with PropertyChecks with ObjectGenerators {
+
+  "ArchiveStateStorage" should "save node directly to db" in new TestSetup {
+    forAll(keyValueByteStringGen(32)) { keyvals =>
+      keyvals.foreach { case (key, value) =>
+        archiveStateStorage.saveNode(key, value , 10)
+      }
+
+      keyvals.foreach { case (key, value) =>
```
agnikowa: why not put it in one `foreach`?

src/test/scala/io/iohk/ethereum/jsonrpc/EthServiceSpec.scala
```diff
@@ -1019,4 +1021,6 @@ class EthServiceSpec extends FlatSpec with Matchers with ScalaFutures with MockF
       noEmptyAccounts = false, ethCompatibleStorage = true)
   }
 
+  val stateStorage = StateStorage.createTestStateStorage(EphemDataSource())._1
```
agnikowa: it's not used here

src/main/scala/io/iohk/ethereum/db/storage/StateStorage.scala
```diff
@@ -0,0 +1,121 @@
...
+
+class ArchiveStateStorage(private val nodeStorage: NodeStorage,
+                          private val cachedNodeStorage: CachedNodeStorage) extends StateStorage {
+
+  override def forcePersist: Unit = {
+    cachedNodeStorage.forcePersist()
+  }
+
+  override def onBlockSave(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlock: () => Unit): Unit = {
+    if (cachedNodeStorage.persist()) {
```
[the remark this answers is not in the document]
KonradStaniec: actually persist in `CachedKeyValueStorage` has signature `def persist(): Boolean`

src/test/scala/io/iohk/ethereum/db/storage/ReferenceCountNodeStorageSpec.scala
KonradStaniec: we would also need to pass `CachedNodeStorage` of `NodeStorage` so i am not sure if we would gain anything by it.

src/test/scala/io/iohk/ethereum/db/storage/StateStorageSpec.scala
KonradStaniec: I wanted to first save all elements and then check than all of those elements are present. Puting it all in one `foreach` would be saving and checking one by one.  Not a big difference but i would leave it as it is.

PR #605, opened 2018-10-22
Refactor/change index seq to array of byte in data storage

agnikowa:
I've done performance tests:

baseline: 127221.1 blocks/10mins
this branch: 128925.7 blocks/10mins
which gives ~1% speed up

### DIFF
--- a/src/main/scala/io/iohk/ethereum/db/storage/KnownNodesStorage.scala
+++ b/src/main/scala/io/iohk/ethereum/db/storage/KnownNodesStorage.scala
@@ -14,7 +14,7 @@ class KnownNodesStorage(val dataSource: DataSource) extends KeyValueStorage[Stri
   val key = "KnownNodes"
 
   val namespace: IndexedSeq[Byte] = Namespaces.KnownNodesNamespace
-  def keySerializer: String => IndexedSeq[Byte] = _.getBytes
+  def keySerializer: String => Array[Byte] = _.getBytes
   def valueSerializer: Set[String] => IndexedSeq[Byte] = _.mkString(" ").getBytes
   def valueDeserializer: IndexedSeq[Byte] => Set[String] = (valueBytes: IndexedSeq[Byte]) => new String(valueBytes.toArray).split(' ').toSet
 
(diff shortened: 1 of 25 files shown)


mmrozek:
I think, 10 min tests are too short to give us valuable info. If I remember correctly the noticeable impact for performance appeared much later.

src/main/scala/io/iohk/ethereum/db/storage/KnownNodesStorage.scala
```diff
@@ -14,9 +14,9 @@ class KnownNodesStorage(val dataSource: DataSource) extends KeyValueStorage[Stri
   val key = "KnownNodes"
 
   val namespace: IndexedSeq[Byte] = Namespaces.KnownNodesNamespace
-  def keySerializer: String => IndexedSeq[Byte] = _.getBytes
-  def valueSerializer: Set[String] => IndexedSeq[Byte] = _.mkString(" ").getBytes
-  def valueDeserializer: IndexedSeq[Byte] => Set[String] = (valueBytes: IndexedSeq[Byte]) => new String(valueBytes.toArray).split(' ').toSet
+  def keySerializer: String => Array[Byte] = _.getBytes
+  def valueSerializer: Set[String] => Array[Byte] = _.mkString(" ").getBytes
+  def valueDeserializer: Array[Byte] => Set[String] = (valueBytes: Array[Byte]) => new String(valueBytes).split(' ').toSet
```
mmrozek: It could be written better without conversion to String.

agnikowa:
Ok, I'll do 1h tests.

src/main/scala/io/iohk/ethereum/db/storage/KnownNodesStorage.scala
agnikowa: Do you mean something like this: ```valueBytes.toSet.map(_.toString)```?

agnikowa: Do you mean something like this: `valueBytes.toSet.map(_.toString)`?

mmrozek: No. I have in mind something like that:
```
var result = Set.empty[String]
var (x, y) = valueBytes.span(_ != ' ')
result += x.mkString
y = y.drop(1)

while(y.nonEmpty){
  val (b, c) = y.span(_ != ' ')
  result += b.mkString
  y = c.drop(1)
}
```
But I am not sure if it is better solution ;)

agnikowa:
so I've done another batch of tests, this time:
5 times per 1 hour and I've got average for such conditions:

baseline: 632352.6 blocks/1h
this branch: 635615.6 blocks/1h
which gives ~0,5% speed up

mmrozek:
So probably it is not worth to merge. What do you think? It is more readable for you?

agnikowa:
I think you're right. It's not worth to merge. I'm closing it.

PR #606, opened 2018-11-06
Add new field sealEngine in BlockchainScenario to handle small change…

agnikowa:
… in the ETS tests format

according to this [link](https://github.com/ethereum/tests/issues/480) we need to provide new field `sealEngine` that accepts two values (for now) `Ethash` and `NoProof` in our `BlockchainScenario`
I've run ETS tests with results as follows: 
- before: failed: 15780, passed: 525
- after: failed: 4903, passed: 11402

### DIFF
--- a/src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainTestConfig.scala
+++ b/src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainTestConfig.scala
@@ -236,14 +236,14 @@ object Eip158ToByzantiumAt5Config extends BlockchainTestConfig {
 }
 
 object Validators {
-  val frontierValidators = EthashValidators(FrontierConfig)
-  val homesteadValidators = EthashValidators(HomesteadConfig)
-  val eip150Validators = EthashValidators(Eip150Config)
-  val frontierToHomesteadValidators = EthashValidators(FrontierToHomesteadAt5)
-  val homesteadToEipValidators = EthashValidators(HomesteadToEIP150At5)
-  val homesteadToDaoValidators = EthashValidators(HomesteadToDaoAt5)
-  val eip158Validators = EthashValidators(Eip158Config)
-  val byzantiumValidators = EthashValidators(ByzantiumConfig)
-  val constantinopleValidators = EthashValidators(ConstantinopleConfig)
-  val eip158ToByzantiumValidators = EthashValidators(Eip158ToByzantiumAt5Config)
+  def frontierValidators(shouldSkipPoW: Boolean): EthashValidators = EthashValidators(FrontierConfig, shouldSkipPoW)
+  def homesteadValidators(shouldSkipPoW: Boolean): EthashValidators = EthashValidators(HomesteadConfig, shouldSkipPoW)
+  def eip150Validators(shouldSkipPoW: Boolean): EthashValidators = EthashValidators(Eip150Config, shouldSkipPoW)
+  def frontierToHomesteadValidators(shouldSkipPoW: Boolean): EthashValidators = EthashValidators(FrontierToHomesteadAt5, shouldSkipPoW)
+  def homesteadToEipValidators(shouldSkipPoW: Boolean): EthashValidators = EthashValidators(HomesteadToEIP150At5, shouldSkipPoW)
+  def homesteadToDaoValidators(shouldSkipPoW: Boolean): EthashValidators = EthashValidators(HomesteadToDaoAt5, shouldSkipPoW)
+  def eip158Validators(shouldSkipPoW: Boolean): EthashValidators = EthashValidators(Eip158Config, shouldSkipPoW)
+  def byzantiumValidators(shouldSkipPoW: Boolean): EthashValidators = EthashValidators(ByzantiumConfig, shouldSkipPoW)
+  def constantinopleValidators(shouldSkipPoW: Boolean): EthashValidators = EthashValidators(ConstantinopleConfig, shouldSkipPoW)
+  def eip158ToByzantiumValidators(shouldSkipPoW: Boolean): EthashValidators = EthashValidators(Eip158ToByzantiumAt5Config, shouldSkipPoW)
 }
--- a/src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainTestConfig.scala
+++ b/src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainTestConfig.scala
@@ -235,15 +235,32 @@ object Eip158ToByzantiumAt5Config extends BlockchainTestConfig {
     MonetaryPolicyConfig(5000000, 0.2, BigInt("5000000000000000000"), BigInt("3000000000000000000"))
 }
 
-case class Validators(shouldSkipPoW: Boolean) {
-  val frontierValidators = EthashValidators(FrontierConfig, shouldSkipPoW)
-  val homesteadValidators = EthashValidators(HomesteadConfig, shouldSkipPoW)
-  val eip150Validators = EthashValidators(Eip150Config, shouldSkipPoW)
-  val frontierToHomesteadValidators = EthashValidators(FrontierToHomesteadAt5, shouldSkipPoW)
-  val homesteadToEipValidators = EthashValidators(HomesteadToEIP150At5, shouldSkipPoW)
-  val homesteadToDaoValidators = EthashValidators(HomesteadToDaoAt5, shouldSkipPoW)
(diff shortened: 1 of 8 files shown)


src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
```diff
@@ -32,22 +32,29 @@ object ScenarioSetup {
...
     val storagesInstance = new SharedEphemDataSources with Pruning with Storages.DefaultStorages
     BlockchainImpl(storagesInstance.storages)
   }
 }
 
 abstract class ScenarioSetup(_vm: VMImpl, scenario: BlockchainScenario) {
 
-  val (blockchainConfig, validators) = buildBlockchainConfig(scenario.network)
+  // according to: https://github.com/ethereum/tests/issues/480 only "NoProof" value should change our current implementation
+  val shouldSkipPoW: Boolean = scenario.sealEngine match {
```
mmrozek: It could be `scenario.sealEngine.exists(_ == "NoProof")`

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainTestConfig.scala
```diff
@@ -236,14 +236,14 @@ object Eip158ToByzantiumAt5Config extends BlockchainTestConfig {
...
-  val homesteadValidators = EthashValidators(HomesteadConfig)
-  val eip150Validators = EthashValidators(Eip150Config)
-  val frontierToHomesteadValidators = EthashValidators(FrontierToHomesteadAt5)
-  val homesteadToEipValidators = EthashValidators(HomesteadToEIP150At5)
-  val homesteadToDaoValidators = EthashValidators(HomesteadToDaoAt5)
-  val eip158Validators = EthashValidators(Eip158Config)
-  val byzantiumValidators = EthashValidators(ByzantiumConfig)
-  val constantinopleValidators = EthashValidators(ConstantinopleConfig)
-  val eip158ToByzantiumValidators = EthashValidators(Eip158ToByzantiumAt5Config)
+  def frontierValidators(shouldSkipPoW: Boolean): EthashValidators = EthashValidators(FrontierConfig, shouldSkipPoW)
```
mmrozek: Maybe the cleaner solution will be if we make 
```
case class Validators(shouldSkipPoW: Boolean){
...
}
```

src/main/scala/io/iohk/ethereum/consensus/ethash/validators/EthashBlockHeaderValidator.scala
```diff
@@ -22,15 +30,14 @@ class EthashBlockHeaderValidator(blockchainConfig: BlockchainConfig) extends Blo
...
-    validatePoW(blockHeader)
-
-  /**
-   * Validates [[io.iohk.ethereum.domain.BlockHeader.nonce]] and [[io.iohk.ethereum.domain.BlockHeader.mixHash]] are correct
-   * based on validations stated in section 4.4.4 of http://paper.gavwood.com/
-   *
-   * @param blockHeader BlockHeader to validate.
-   * @return BlockHeader if valid, an [[io.iohk.ethereum.consensus.validators.BlockHeaderError.HeaderPoWError]] otherwise
-   */
+    if (shouldSkipPoW) Right(BlockHeaderValid) else validatePoW(blockHeader)
```
KonradStaniec: I am not sure that ets test-s should have such impact on our production code. Have you tried maybe creating sub-class of `EthashBlockHeaderValidator` which always validate  to `Right(BlockHeaderValid)`,  and then passing it to `EthashValidators `  as parametr in `apply` ? 
How much more work would it be  ? ( just asking as maybe it to much hassle to do that, becouse little mess in class dependencies becouse of swapable consensus code)

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainTestConfig.scala
```diff
@@ -235,15 +235,15 @@ object Eip158ToByzantiumAt5Config extends BlockchainTestConfig {
...
-  val homesteadValidators = EthashValidators(HomesteadConfig)
-  val eip150Validators = EthashValidators(Eip150Config)
-  val frontierToHomesteadValidators = EthashValidators(FrontierToHomesteadAt5)
-  val homesteadToEipValidators = EthashValidators(HomesteadToEIP150At5)
-  val homesteadToDaoValidators = EthashValidators(HomesteadToDaoAt5)
-  val eip158Validators = EthashValidators(Eip158Config)
-  val byzantiumValidators = EthashValidators(ByzantiumConfig)
-  val constantinopleValidators = EthashValidators(ConstantinopleConfig)
-  val eip158ToByzantiumValidators = EthashValidators(Eip158ToByzantiumAt5Config)
+case class Validators(shouldSkipPoW: Boolean) {
```
KonradStaniec: As i rembemer i was all created in object, because otherwise pow cache would need to be recreated before each test which leads to really long running times for ets test. What is ets time after this change ? (Maybe it did not changed much because of this new field, which disabled pow for most of tests)

mmrozek: Good question. Still, if the validators will be a `def`s they will have to be recreated.

agnikowa: this branch: 19m 45s 162ms
constantinople branch: 36s 295ms

so I'll change it ti previous one

src/main/scala/io/iohk/ethereum/consensus/ethash/validators/EthashBlockHeaderValidator.scala
agnikowa: I'll check it out

src/ets/scala/io/iohk/ethereum/ets/blockchain/BlockchainTestConfig.scala
mmrozek: To which one? If you change to defs inside an object it won't change anything (I guess) because it will be evaluated each time

agnikowa: with defs it is taking 18m 24s 488ms. I tried also with trait and only `shouldSkipPoW` as a def but it takes: 18m 5s 836ms.

I've done something like @KonradStaniec mention in his previous comment and now tests run in: 9m 1s 652ms while I check

src/ets/scala/io/iohk/ethereum/ets/blockchain/EthashTestBlockHeaderValidator.scala
```diff
@@ -0,0 +1,22 @@
...
+import io.iohk.ethereum.consensus.validators.{ BlockHeaderError, BlockHeaderValid, BlockHeaderValidatorSkeleton }
+import io.iohk.ethereum.domain.BlockHeader
+import io.iohk.ethereum.utils.BlockchainConfig
+
+class EthashTestBlockHeaderValidator(blockchainConfig: BlockchainConfig) extends BlockHeaderValidatorSkeleton(blockchainConfig) {
+  import EthashBlockHeaderValidator._
+
+  // NOTE the below comment is from before PoW decoupling
+  // we need concurrent map since validators can be used from multiple places
+  protected val powCaches: java.util.concurrent.ConcurrentMap[Long, PowCacheData] = new java.util.concurrent.ConcurrentHashMap[Long, PowCacheData]()
```
KonradStaniec: minor: Probably not needed here

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
```diff
@@ -32,22 +32,26 @@ object ScenarioSetup {
...
     val storagesInstance = new SharedEphemDataSources with Pruning with Storages.DefaultStorages
     BlockchainImpl(storagesInstance.storages)
   }
 }
 
 abstract class ScenarioSetup(_vm: VMImpl, scenario: BlockchainScenario) {
 
-  val (blockchainConfig, validators) = buildBlockchainConfig(scenario.network)
+  // according to: https://github.com/ethereum/tests/issues/480 only "NoProof" value should change our current implementation
+  def shouldSkipPoW: Boolean = scenario.sealEngine.contains("NoProof")
```
KonradStaniec: minor: it can be `val`

PR #608, opened 2018-11-09
EC-596 - Improve our handling of sending discovery message

mmrozek:
I changed the SendMessage to SendPacket because I want to avoid second encoding. I thought also about moving the extraction message outside the DiscoveryListener (listener don't need to know anything about message type). What do you think?

### DIFF
--- a/src/test/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManagerSpec.scala
+++ b/src/test/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManagerSpec.scala
@@ -38,6 +38,8 @@ class PeerDiscoveryManagerSpec extends FlatSpec with Matchers with MockFactory w
     val pongDecoded = getPacket(pong)
     val pongMessageReceiced = MessageReceived(pong, remoteUdpAddress, pongDecoded)
 
+    pongDecoded.validated().isDefined should be (true)
+
     val nodeInfo = DiscoveryNodeInfo.fromNode(Node(pongDecoded.nodeId.get, remoteUdpAddress.getAddress, remoteUdpPort, remoteUdpPort))
 
     discoveryPeerManager.underlyingActor.pingedNodes += pingPingPacketDecoded.mdc -> PingInfo(nodeInfo, timestamp)


src/main/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManager.scala
```diff
@@ -95,10 +94,11 @@ class PeerDiscoveryManager(
       case ServerStatus.Listening(address) =>
         val from = Endpoint.makeEndpoint(address, getTcpPort)
         val ping = Ping(ProtocolVersion, from, toEndpoint, expirationTimestamp)
-        getPacketData(ping).foreach{ key =>
+        val packet = encodePacket(ping, nodeStatusHolder.get().key)
+        getPacketData(packet).foreach{ key =>
```
KonradStaniec: Maybe it would be better to not create `Packet` object ? I mean now it is created in :
https://github.com/input-output-hk/mantis/pull/608/files#diff-9fc6e18903892ffc78f87ca4811bb1c5R101 and
https://github.com/input-output-hk/mantis/pull/608/files#diff-9fc6e18903892ffc78f87ca4811bb1c5R117
and it is possibly high cost operation ( each packet has `val nodeId` which is calculated from public key).
And all we need is two `ByteStrings` which we can recover from encoded packet in line `97` like:
`val packetMdc          = packet.take(Packet.MdcLength)` and
`val packetDataHash = crypto.kec256(packet.drop(Packet.DataOffset))`.

Second option is to make `nodeId` a `lazy val`, but then we would need to change `decodePacket` as now it is using `.get` on `nodeId` to signalize error.

mmrozek:
The problem with moving the extraction logic is that now we are sending an extracted message to all subscribers. If we move the logic, all actors will have to do it by itself. We will duplicate the decoding part.

src/main/scala/io/iohk/ethereum/network/discovery/package.scala
```diff
@@ -0,0 +1,83 @@
...
+
+    val forSha = sigBytes ++ Array(msg.packetType) ++ encodedData
+    val mdc = crypto.kec256(forSha)
+
+    ByteString(mdc ++ sigBytes ++ Array(msg.packetType) ++ encodedData)
+  }
+
+  private[discovery] def extractMessage(packet: Packet): Try[Message] = Try {
+    packet.packetType match {
+      case Ping.packetType => rlp.decode[Ping](packet.data.toArray[Byte])
```
agnikowa: maybe this `packet.data.toArray[Byte]` could be extracted to some val?

PR #610, opened 2018-11-15
fix ets tests

KonradStaniec:
Fixing most of failing constantinipole ets tests (most errors were related to new sstore gas calculation)

Only one failing test remain: 
GeneralStateTests/stShift/sar_2^256-1_0_d0g0v

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
@@ -28,7 +28,8 @@ object ProgramContext {
       callDepth = 0,
       world = world,
       initialAddressesToDelete = Set(),
-      evmConfig = evmConfig
+      evmConfig = evmConfig,
+      originalWorld = world
     )
   }
 }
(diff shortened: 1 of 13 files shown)


src/main/scala/io/iohk/ethereum/vm/ProgramContext.scala
```diff
@@ -60,6 +61,8 @@ object ProgramContext {
   * @param initialAddressesToDelete contains initial set of addresses to delete (from lower depth calls)
   * @param evmConfig                evm config
   * @param staticCtx                a flag to indicate static context (EIP-214)
+  * @param originalWorld            state of the world at the beginning og the current transaction, read-only,
```
agnikowa: *of

src/main/scala/io/iohk/ethereum/vm/VM.scala
```diff
@@ -76,13 +76,29 @@ class VM[W <: WorldStateProxy[W, S], S <: Storage[S]] extends Logger {
       // to empty values.
       val conflict = context.world.nonEmptyCodeOrNonceAccount(newAddress)
 
+      /*
```
agnikowa: you should use scaladoc style

PR #613, opened 2018-11-22
Restructure configuration to allow easy network switch

kapke:
All is needed now to run Mantis with different network is to provide system property `mantis.blockchains.network`. AFAIK the simplest way is through JAVA_OPTS env var

Things that still needs to be done:

  - [x] migrate config for private network
  - [x] ensure it's all working properly in `dist` mode
  - [x] ensure testmode is working


src/main/resources/chains/base.conf
```diff
@@ -0,0 +1,144 @@
...
+    block-extra-data-range = 10
+
+    # Address to send funds when draining
+    refund-contract-address = null
+
+    # List of accounts to be drained
+    drain-list = null
+  }
+
+  # Starting nonce a an empty account. Some networks (like Morden) use different values.
```
agnikowa: typo, should be something like `nonce of an empty`

src/test/resources/application.conf
```diff
@@ -16,48 +16,81 @@ mantis {
...
-  network.discovery.bootstrap-nodes = [
-    "enode://375fc4e3712013120f392d909c57aead2461ba919130075e8fcd007e8326971a514741af0debff783ebdcb8c82ab72c8eee9a31adab629232e52d1e8e9b9b4b6@127.0.0.1:30340",
-    "enode://375fc4e3712013120f392d909c57aead2461ba919130075e8fcd007e8326971a514741af0debff783ebdcb8c82ab72c8eee9a31adab629232e52d1e8e9b9b4b7@10.0.0.0:30340"
-  ]
-
   network.rpc.apis = "eth,web3,net,personal,daedalus,debug"
 
-  blockchain {
-    eip155-block-number = "3000000"
+  blockchains {
```
agnikowa: why not put this into `test.conf` file?

kapke: Are you asking why it's not in `src/main/resources/chains/test.conf` or why it's not in `src/test/resources/test.conf`? 
The former - cause it's only for tests and thus I don't think it should present be in app resources
The latter - I don't think it's neccessary.

src/universal/conf/storage.conf
```diff
@@ -34,16 +34,17 @@ mantis {
...
-    # - basic: reference count based pruning
+    # - archive:  No pruning is performed
+    # - basic:    reference count based pruning
+    # - inmemory: reference count inmemory pruning
     #
-    # After changing, please delete previous db before starting the client:
+    # After changing, please delete previous db before starting the client
     #
-    # mode = "basic"
+    # mode = "archive"
```
KonradStaniec: For what chains it will be set up as archive with this new config encoding ?
Should not each of new `eth.conf`, `etc.conf` files have different prunning settings ? 
As for current config the settings where
inmemory - ETH
basic - everything else

kapke: It's handled now by separate launcher script (in `src/universal/bin/mantis-eth`) for ETH which sets these settings. To keep things rather simple and not make revolution in code I wanted to avoid loading different configuration files which override whole app settings.

PR #617, opened 2018-12-07
fix configs

KonradStaniec:
eth and ropsten don not have concept of eras, so 1st era should last as long as possible 50 000 000 block is around 21 years.

### DIFF
--- a/src/main/resources/chains/eth.conf
+++ b/src/main/resources/chains/eth.conf
@@ -22,6 +22,7 @@
   chain-id = "1"
 
   monetary-policy {
+    era-duration = 50000000
     reward-reduction-rate = 0
   }
 
--- a/src/main/resources/chains/eth.conf
+++ b/src/main/resources/chains/eth.conf
@@ -22,7 +22,14 @@
   chain-id = "1"
 
   monetary-policy {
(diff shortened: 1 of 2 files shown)


mmrozek:
If they don't have such concepts maybe a better idea is to keep the configuration file as is and handle this case in `MonetaryPolicyConfig` class (ex. if this parameter doesn't exist, we could set max value)

KonradStaniec:
i do not know if it worth now spliting `MonetaryPolicyConfig` to two or more classes, and then refactoring our reward calcluator to accomodate that. But maybe i will just switch this value to `Integer.MaxValue` in config ? I mean everywhere else we are using this convention.

mmrozek:
I agree with you. I have the same in mind. Just simple checking if `era-duration` exists, if not set to `Integer.MaxValue`. You could also add a comment why. I think it will be clean enough solution.

KonradStaniec:
Yep it would be good, but as currently stands `era-duration` will be always present, as all configs inherit it from base config, solution to that is to remove monetary policy from base, and define separatly in each config, but this is quite repetetive so i am not sure it worth it.

PR #620, opened 2018-12-14
Add cacluclatin senders in background. Fix mem performance

KonradStaniec:
1. Add caluclating senders in background - It was used earlier in regular sync so this is nothing new

2. Fix performance of store operation- 
As it seems `MSTORE` is one of most frequent used opcodes in eth, and our courrent implementation clogs on it when synchronising on top of the chain.
[https://ibb.co/jJ7wvLv](url) - almost 5/6 of vm time spent in store.
This pr introduces optimised implementation.
- Microbenchmarks (which i have written on the side and are not 100% reliable) shows 4x optimization with that implementation
- block processing speed as i am currently running eth increased from `50blocks / 5min` to `50blocks/2min`
- visual vm shows that now only 1/4 vm time is spent in `MSTORE`

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/Memory.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/Memory.scala
@@ -39,8 +39,33 @@ class Memory private(private val underlying: ByteString) {
     val newUnderlying: ByteString =
       if (data.isEmpty)
         underlying
-      else
-        underlying.take(idx).padTo(idx, 0.toByte) ++ data ++ underlying.drop(idx + data.length)
+      else {
+        val currentLength = underlying.length
+        val dataLength = data.length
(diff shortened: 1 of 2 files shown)


src/main/scala/io/iohk/ethereum/vm/Memory.scala
```diff
@@ -39,8 +39,33 @@ class Memory private(private val underlying: ByteString) {
...
+
+        var i = 0
+        while (i < currentLength) {
+          newData(i) = underlying(i)
+          i += 1
+        }
+
+        var u = 0
+        i = idx
+        while (i < newLen && u < dataLength) {
```
mmrozek: I think you don't ned to check `i < newLen`

PR #621, opened 2018-12-14
Mantis launchers readme

kapke:


README.md
```diff
@@ -50,6 +50,10 @@ The bootstrap database files can be downloaded from [here](https://github.com/in
 
 To access the command line version of this release go to [daedalus-cli](https://github.com/input-output-hk/mantis/tree/phase/daedalus-cli)
 
+Depending on network you want to join you can use appropriate launcher, all can be found in `bin` directory:
+  - `mantis-etc` - for joining Ethereum Classic network
```
KonradStaniec: `mantis-etc`  also has enabled pruning by default as default option is defined in `application.conf`, but it is basic version that tries to cleat all junk.
I think i would leave out info about pruning from general readme.

PR #623, opened 2018-12-17
Feature/eth test update

KonradStaniec:
update ets and fix extcodehash opcode


src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -376,7 +376,21 @@ case object EXTCODEHASH extends OpCode(0x3F, 1, 1, _.G_balance) with ConstGas {
   protected def exec[W <: WorldStateProxy[W, S], S <: Storage[S]](state: ProgramState[W, S]): ProgramState[W, S] = {
     val (accountAddress, stack1) = state.stack.pop
     val address = Address(accountAddress)
-    val accountExists = state.world.accountExists(address)
+
+    /**
+      * Specification of EIP1052 - https://eips.ethereum.org/EIPS/eip-1052, says that we should return 0 if
```
agnikowa: probably without this `if` at the end

PR #626, opened 2018-12-19
Rpc tests fixes

kapke:
1. Changed configuration a bit to make it simpler to patch mantis client
1. Added script for patching mantis client with required configuration so manual step other than running this script is not required anymore
1. Changed README a make whole process just bunch of copy&paste commands
1. Updated tests that had too strong checks/assumptions


src/rpcTest/resources/README.md
```diff
@@ -1,30 +1,46 @@
...
-10. turn off mining in `~/mantis/conf/consensus.conf`
-11. run mantis client
-12. go to `mantis` source dir and run `sbt`, next run privet net portion of tests without mining
-    via `rpcTest:testOnly -- -n PrivNetNoMining`
+
+1. Go to `~/mantis_build` directory and run mantis on ETC mainnet with command:
+
+        ./bin/mantis-etc -Dmantis.sync.do-fast-sync=false -Dmantis.network.discovery.discovery-enabled=true -Dmantis.network.rpc.http.mode=http
+        
+1. Ensure it has at least `150000` blocks.
```
mmrozek: Where is point 2? :)

src/rpcTest/resources/README.md
```diff
@@ -1,30 +1,46 @@
...
+1. Go to `mantis` source dir and run 
+
+        sbt "rpcTest:testOnly -- -n MainNet"
```
mmrozek: You could format commands in a different way to improve readability

src/rpcTest/resources/privateNetConfig/conf/rpc-test-private-genesis.json
```diff
@@ -0,0 +1,21 @@
...
+  "difficulty": "0x400",
+  "ommersHash": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
+  "timestamp": "0x00",
+  "coinbase": "0x0000000000000000000000000000000000000000",
+  "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+  "alloc": {
+    "316158e265fa708c623cc3094b2bb5889e0f5ca5": {
+      "balance": "100000000000000000000"
+    },
+	"b9ec69316a8810db91c36de79c4f1e785f2c35fc": {
```
agnikowa: this should be aligned to first element of `alloc`

src/rpcTest/resources/README.md
kapke: Take a look of that file in github or some editor supporting Markdown preview :)

You'll notice that:
  - in markdown each point can have a number `1.` and renderer takes care of assigning proper numbers
  - double indentation causes a command to be nicely shown in highlighted block of code

mmrozek: I agree with you at the second point. But the points number should be corrected. Not everybody will be using Markdown preview :)

PR #642, opened 2020-08-24
[ETCM-23] Add etc fork blocks to config

KonradStaniec:


src/main/resources/application.conf
```diff
@@ -518,7 +504,7 @@ mantis {
       block-cache-size = 33554432
     }
 
-    # Define which database to use [leveldb, rocksdb], iodb is not currently used
+    # Define which database to use [rocksdb], iodb is not currently used
```
mmrozek, line 507: Do we need that anymore?

mmrozek (approved):
LGTM

src/main/resources/application.conf
KonradStaniec: probably not, but I would leave it as it is for now (as it is not critical). We can improve it later, when we will have working sync.

PR #644, opened 2020-08-25
[ETCM-28] petersburg changes

mmrozek:
Changes related to Petersburg fork

### DIFF
--- a/src/main/resources/chains/eth.conf
+++ b/src/main/resources/chains/eth.conf
@@ -37,6 +37,10 @@
   # https://ecips.ethereumclassic.org/ECIPs/ecip-1088
   phoenix-block-number = "1000000000000000000"
 
+  # Petersburg fork block number (ETH only)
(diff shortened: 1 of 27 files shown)


src/main/resources/chains/eth.conf
```diff
@@ -37,6 +37,10 @@
   # https://ecips.ethereumclassic.org/ECIPs/ecip-1088
   phoenix-block-number = "1000000000000000000"
 
+  # Petersburg fork block number (ETH only)
+  # https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1716.md
+  petersburg-block-number = "7280000"
```
KonradStaniec, line 42: hmm mayble as we are adding it, we should also change constantinople fork block number in config to `7280000` ? (https://media.consensys.net/the-constantinople-hard-fork-what-you-need-to-know-d438a91dec3f)

KonradStaniec (approved):

PR #645, opened 2020-08-25
[ETCM-19] chainId opcode

mmrozek:
EIP-1344: ChainID opcode

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
@@ -4,6 +4,7 @@ import akka.util.ByteString
 import io.iohk.ethereum.domain.UInt256
 import io.iohk.ethereum.utils.BlockchainConfig
 import EvmConfig._
+import io.iohk.ethereum
 import io.iohk.ethereum.vm
 
 // scalastyle:off number.of.methods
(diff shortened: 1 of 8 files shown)


KonradStaniec (commented):
I would add at least one unit test maybe in `OpCodeFunSpec` ?

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -105,6 +108,11 @@ object EvmConfig {
     opCodeList = AghartaOpCodes
   )
 
+  val PhoenixConfigBuilder: EvmConfigBuilder = config => PhoenixConfigBuilder(config).copy(
```
KonradStaniec, line 111: Here it shoud be:
` val PhoenixConfigBuilder: EvmConfigBuilder = config => AghartaConfigBuilder(config).copy(`

mmrozek: Yes, my bad

mmrozek:
@KonradStaniec, There is a generic test for all `ConstOp`s

KonradStaniec:
ok, but then from what I see it uses byzantium config https://github.com/input-output-hk/mantis/blob/etcm-19-chainid-opcode/src/test/scala/io/iohk/ethereum/vm/OpCodeFunSpec.scala#L17, to use it we would need change this to phoenix conifg ?

KonradStaniec (approved):

PR #646, opened 2020-08-26
[ETCM-18] EIP-1108: Reduce alt_bn128 precompile gas costs

mmrozek:
https://eips.ethereum.org/EIPS/eip-1108

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/BlockchainConfigForEvm.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/BlockchainConfigForEvm.scala
@@ -2,6 +2,8 @@ package io.iohk.ethereum.vm
 
 import io.iohk.ethereum.domain.UInt256
 import io.iohk.ethereum.utils.BlockchainConfig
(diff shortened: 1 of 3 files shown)


src/main/scala/io/iohk/ethereum/vm/BlockchainConfigForEvm.scala
```diff
@@ -22,9 +24,36 @@ case class BlockchainConfigForEvm(
...
+
+  }
+}
 
 object BlockchainConfigForEvm {
+
+  object EtcForks extends Enumeration {
+    type EtcFork = Value
+    val BeforeAtlantis, Atlantis, Agharta, Phoenix = Value
+    // TODO add Petersburg and Istanbul
```
KonradStaniec, line 48: there is no Petersburg and Istanbul for ETC, or did you mean that add those to config ? If yes then imo it would be better to thiss coment to EthForks

mmrozek: You're right. I will move the comment

KonradStaniec (approved):

PR #650, opened 2020-08-28
Etcm 22 structured definitions for net gas metering

mmrozek:
https://eips.ethereum.org/EIPS/eip-2200

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
@@ -4,7 +4,9 @@ import akka.util.ByteString
 import io.iohk.ethereum.crypto.kec256
 import io.iohk.ethereum.domain.{Account, Address, TxLogEntry, UInt256}
 import io.iohk.ethereum.domain.UInt256._
-import io.iohk.ethereum.vm.BlockchainConfigForEvm.EthForks
+import io.iohk.ethereum.vm.BlockchainConfigForEvm.EtcForks.EtcFork
+import io.iohk.ethereum.vm.BlockchainConfigForEvm.{EtcForks, EthForks}
+import io.iohk.ethereum.vm.BlockchainConfigForEvm.EthForks.EthFork
 
 // scalastyle:off magic.number
 // scalastyle:off number.of.types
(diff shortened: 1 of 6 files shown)


src/test/scala/io/iohk/ethereum/vm/SSTOREOpCodeGasPostConstantinopleSpec.scala
```diff
@@ -10,47 +10,76 @@ import akka.util.ByteString.{empty => bEmpty}
...
+      ("60006000556002600055", 1, 5812, 0),
+      ("60026000556000600055", 1, 5812, 15000),
+      ("60026000556003600055", 1, 5812, 0),
+      ("60026000556001600055", 1, 5812, 4200),
+      ("60026000556002600055", 1, 5812, 0),
+      ("60016000556000600055", 1, 5812, 15000),
+      ("60016000556002600055", 1, 5812, 0),
+      ("60016000556001600055", 1, 1612, 0),
+      ("600160005560006000556001600055", 0, 40818, 19200),
+      ("600060005560016000556000600055", 1, 10818, 19200)
```
KonradStaniec, line 68: I found out that geth has two additional test cases : https://github.com/ethereum/go-ethereum/blob/b9df7ecdc3d3685180ceb29665bab59e9f614da5/core/vm/gas_table_test.go#L76. 
```
	{1, 2306, "0x6001600055", 2306, 0, ErrOutOfGas},                            // 1 -> 1 (2300 sentry + 2xPUSH)
	{1, 2307, "0x6001600055", 806, 0, nil},                                     // 1 -> 1 (2301 sentry + 2xPUSH)
```
 Most probably testing some edge cases. Imo we should try to add them, as all this gass couting makes me nervous,

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -593,52 +595,67 @@ case object MSTORE8 extends OpCode(0x53, 2, 0, _.G_verylow) {
...
-              state.config.feeSchedule.G_sreset - state.config.feeSchedule.G_sload
+    val currentBlockNumber = state.env.blockHeader.number
+    val etcFork = state.config.blockchainConfig.etcForkForBlockNumber(currentBlockNumber)
+    val ethFork = state.config.blockchainConfig.ethForkForBlockNumber(currentBlockNumber)
+
+    val eip2200Enabled = isEip2200Enabled(etcFork, ethFork)
+    val eip1283Enabled = isEip1283Enabled(ethFork)
+
+    if(eip2200Enabled && state.gas <= state.config.feeSchedule.G_callstipend){
+        state.withError(OutOfGas)
```
KonradStaniec, line 606: Usually `OutOfGas` errors zeros caller remaining gas like ` state.copy(gas = 0).withError(OutOfGas)`. Do you think we should try to do the same here ? (maybe it worth checking parity of geth)

src/test/scala/io/iohk/ethereum/vm/SSTOREOpCodeGasPostConstantinopleSpec.scala
mmrozek: Added

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
mmrozek: The tests from previous comment check that

KonradStaniec (approved):

PR #652, opened 2020-09-01
[FIX] fast sync fix

KonradStaniec:


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -315,14 +315,18 @@ class FastSync(
       if (checkHeadersChain(headers)) {
         processHeaders(peer, headers) match {
           case ParentDifficultyNotFound(header) =>
-            log.debug("Parent difficulty not found for block {}, not processing rest of headers", header.number)
-            processSyncing()
+            // We could end in wrong fork and get blocked so we should rewind our state a little
+            // we blacklist peer just in case we got malicious peer which would send us bad blocks, forcing us to rollback
+            // to genesis
+            log.info("Parent difficulty not found for block {}, not processing rest of headers", header.idTag)
```
mirkoAlic, line 321: maybe all this logs that end up in a blacklisting action could be used `.warn` instead of `info`, in that way we could easily search for them. WDYT?

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -315,14 +315,18 @@ class FastSync(
       if (checkHeadersChain(headers)) {
         processHeaders(peer, headers) match {
           case ParentDifficultyNotFound(header) =>
-            log.debug("Parent difficulty not found for block {}, not processing rest of headers", header.number)
-            processSyncing()
+            // We could end in wrong fork and get blocked so we should rewind our state a little
+            // we blacklist peer just in case we got malicious peer which would send us bad blocks, forcing us to rollback
```
mirkoAlic, line 319: sounds hardcore..i mean, we blacklist the Peer. but still we rewind to genesis? =S

KonradStaniec: if we blacklist this peer and rewind state a little, we hopefully will get some blocks from other peer which will steer us to main chain.

mirkoAlic (approved):
LGTM!

PR #653, opened 2020-09-01
[FIX] Change algorithm of target block selection

mmrozek:


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncTargetBlockSelector.scala
```diff
@@ -83,12 +82,22 @@ class FastSyncTargetBlockSelector(
...
-      val (mostUpToDatePeer, mostUpToDateBlockHeader) = receivedHeaders.maxBy(_._2.number)
+
+      val (_, mostUpToDateBlockHeader) = receivedHeaders.maxBy(_._2.number)
       val targetBlock = mostUpToDateBlockHeader.number - syncConfig.targetBlockOffset
-      peerEventBus ! Subscribe(MessageClassifier(Set(BlockHeaders.code), PeerSelector.WithId(mostUpToDatePeer.id)))
-      etcPeerManager ! EtcPeerManagerActor.SendMessage(GetBlockHeaders(Left(targetBlock), 1, 0, reverse = false), mostUpToDatePeer.id)
+
+      val peersToAsk = receivedHeaders.toList.sortBy(-_._2.number)
+        .takeWhile { case (_, header) => header.number > targetBlock }
+        .map(_._1).take(3) // todo ???
```
KonradStaniec, line 91: here we could probably use the `minPeersToChooseTargetBlock`

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncTargetBlockSelector.scala
```diff
@@ -83,12 +82,22 @@ class FastSyncTargetBlockSelector(
...
     if (receivedHeaders.size >= minPeersToChooseTargetBlock) {
-      val (mostUpToDatePeer, mostUpToDateBlockHeader) = receivedHeaders.maxBy(_._2.number)
+
+      val (_, mostUpToDateBlockHeader) = receivedHeaders.maxBy(_._2.number)
       val targetBlock = mostUpToDateBlockHeader.number - syncConfig.targetBlockOffset
-      peerEventBus ! Subscribe(MessageClassifier(Set(BlockHeaders.code), PeerSelector.WithId(mostUpToDatePeer.id)))
-      etcPeerManager ! EtcPeerManagerActor.SendMessage(GetBlockHeaders(Left(targetBlock), 1, 0, reverse = false), mostUpToDatePeer.id)
+
+      val peersToAsk = receivedHeaders.toList.sortBy(-_._2.number)
+        .takeWhile { case (_, header) => header.number > targetBlock }
```
KonradStaniec, line 90: we should probably make sure that we have enough peers, and if not then reschedule asking for new block

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncTargetBlockSelector.scala
```diff
@@ -98,27 +107,41 @@ class FastSyncTargetBlockSelector(
...
-          context become idle
+          blacklist(peerId, blacklistDuration, "Did not respond with target block header, blacklisting")
+          context become waitingForTargetBlock(updatedPeersToAsk, targetBlockNumber, timeout, headers)
       }
 
     case TargetBlockTimeout =>
-      blacklist(peer.id, blacklistDuration, s"did not respond with target block header (timeout), blacklisting and scheduling retry in $startRetryInterval")
+      peersToAsk.foreach{
+        peerId =>
+          blacklist(peerId, blacklistDuration, "Did not respond with target block header (timeout), blacklisting")
```
KonradStaniec, line 141: we should proably have a timeout per peer to blackilst only the one which did not provide us with block.

KonradStaniec (commented):
In general i think we shoud have something like that,and this is good direction. Maybe it is also worth checking how geth is choosing their target block.

Also quick fix for now, maybe to increase `minPeersToChooseTargetBlock` in config it wil minimise the chance of ending in bad fork

src/main/scala/io/iohk/ethereum/blockchain/sync/FastSyncTargetBlockSelector.scala
mmrozek: It is done here because `peersToAsk` is updated after each response and contains only not responding peers

mmrozek:
Close for now. Fast sync requires a much more sophisticated refactor

PR #658, opened 2020-09-07
[Chore] Remove unused iodb

mirkoAlic:


mmrozek:
There still is a dependency in `build.sbt` file - https://github.com/input-output-hk/mantis/blob/develop/build.sbt#L30

PR #660, opened 2020-09-08
[Chore] Project build updates

mirkoAlic:
# Description
    
As part of the catching up, the goal of this PR is to prepare the build for better supporting of the nix build.

 - Update Sbt
 - Update Scala
 - Remove sbt-verify package: In the near future will covered by nix build
 - Update Readme

# Testing

Crucial components should continue working: CircleCi, local build, etc, etc.


build.sbt
```diff
@@ -2,8 +2,8 @@ enablePlugins(JDKPackagerPlugin, JavaAppPackaging, SolidityPlugin)
 
 val commonSettings = Seq(
   name := "mantis",
-  version := "2.0",
-  scalaVersion := "2.12.5",
+  version := "3.0",
+  scalaVersion := "2.12.10",
```
kapke, line 6: Minor: `2.12.12`?

build.sbt
```diff
@@ -78,17 +78,8 @@ val root = project.in(file("."))
...
-        includeBin = true,
-        includeScala = true,
-        includeDependency = true,
-        excludedJars = Nil,
-        warnOnUnverifiedFiles = false,
-        warnOnUnusedVerifications = false
-      ),
       executableScriptName := name.value,
-      dist in Universal := ((dist in Universal) dependsOn verify).value
+      dist in Universal := (dist in Universal).value
```
kapke, line 82: this can be removed now

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version = 0.13.13
\ No newline at end of file
+sbt.version = 1.2.8
```
kapke, line 1: We don't rely on anything fancy here, is 1.3 feasible then?

mirkoAlic: let me try..

mirkoAlic: with 1.3.13:
```
# Unit tests step
[info] Compiling 143 Scala sources to /home/circleci/project/target/scala-2.12/test-classes ...
/usr/share/sbt/bin/sbt-launch-lib.bash: line 58:   789 Killed                  "$@"
```
^ i tried to increase memory but it doesn't worked, so i decided to turn back to 1.2.8

kapke: 👍 
Did you update launcher script as well? It has changed a bit between 1.2.x and 1.3 to accomodate for specifically different approach to memory

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     environment:
-      - SBT: "sbt -v -mem 1024"
+      - SBT: "sbt -v -mem 2048 -J-Xmx4g -J-Xss10M"
```
kapke, line 5: setting `-mem` and `-Xmx` is redundant

src/rpcTest/resources/privateNetConfig/conf/application.ini
```diff
@@ -1 +1 @@
--Dconfig.file=./conf/rpc-test-private.conf -Dlogback.configurationFile=./conf/rpc-test-logback.xml -J-Xss10M
+-Dconfig.file=./conf/rpc-test-private.conf -Dlogback.configurationFile=./conf/rpc-test-logback.xml -J-Xmx4g -J-Xss10M
```
kapke, line 1: MAntis should be more than fine with 2g of memory with disabled mining. Do we really want to hardcode that value here and not on deployment side?

kapke (approved):
LGTM!

PR #661, opened 2020-09-09
[Chore] Add nix

mirkoAlic:
## Description
This enables building Mantis in a sandbox.

## Proposed Solution

Use Sbtix to generate Nix expressions from
sbt configs. The generated Nix expressions are functionally equivalent to the
(already removed) verify.sbt file and must be updated on the same
occasions that verify.sbt would (i.e., whenever dependencies are updated or changed).

# Testing

- to generate nix expressions:

``` shell
nix-shell -p nixUnstable
sbtix-gen-all2
```

- to build:

``` shell
nix-build
```

- to run:

``` shell
result/bin/mantis -Dconfig.file=src/main/resources/application.conf
```


project/manual-repo.nix
```diff
@@ -0,0 +1,26 @@
+# this file must still be generated manually.
```
mirkoAlic, line 1: why we need both: `./manual-repo.nix`, `project/manual-repo.nix`?

nix/pkgs/mantis/unwrapped.nix
```diff
@@ -0,0 +1,51 @@
+# this file originates from SBTix
```
mirkoAlic, line 1: can we improve the comment, i mean, this file is originates from SBTix but as far as i know we also added some stuff.

project/project/repo.nix
```diff
@@ -0,0 +1,1872 @@
+# This file was automatically generated by Sbtix. Do not edit manually.
```
mirkoAlic, line 1: if this is automatically generated, why we need to push it to the repo?

src/universal/bin/mallet
```diff
@@ -1,5 +0,0 @@
-#!/bin/bash
-
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
mirkoAlic, line 3: we discovered why this was causing troubles?

shell.nix
```diff
@@ -0,0 +1,12 @@
+{ sources ? import nix/sources.nix, pkgs ? import ./nix { } }:
+
+if __getEnv "BUILDKITE" == "true" then
```
mirkoAlic, line 3: builkite integration will be included in a future sprint right?

kreisys:
Added Hydra builds. Can we merge this already so that other branches get hydra builds as well?

mirkoAlic:
> Added Hydra builds. Can we merge this already so that other PRs get hydra builds as well?

In order to get this merge you should:
1. Solve conflicts in `build.sbt` (Probably a re-fresh of the `repo.nix` files is needed).
^ I didn't do it, because i'm still struggling nix issues that i don't understand...So is up to you guys.

2. I asked/raise some concerns in the comments, is not a blocker for get this PR merged, but if you could respond them, it would be nice.

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version = 1.3.13
+sbt.version = 1.2.8
```
mirkoAlic, line 1: why? i mean, it was necessary?

mirkoAlic:
This needs a dependencies regeneration again. And i can't do in it (still not working in my machine), so please someone try to.

project/build.properties
kreisys: Me and @sevanspowell were pulling our hair on this all night... briefly, given that:
- Sbtix doesn't work correctly with SBT 1.3 (it creates a `repo.nix` that references local path)
- We don't really know how to fix Sbtix for SBT 1.3
- We don't know how to estimate how long it would take to figure it out
- We have some aggressive time constraints
- SBT 1.3 is at best a nice-to-have (this is an assumption, correct me if I'm wrong)
- Other urgent tasks are blocked by this not being merged

therefore we made an executive decision to go back to `1.2.8` which was already known to work...

kreisys:
I regenerated and now it fails to build.

project/build.properties
mirkoAlic: first of all thanks for the clarification and all the effort you are doing! Yeah, fully agree, it was not a requirement. We can get rid off it, given are current deadlines. (On devops-sync call we also agreed upon that)

project/project/repo.nix
```diff
@@ -5,9 +5,190 @@
       "sbtVersion" = "1.2.8";
     }];
   "repos" = {
+    "nix-local-preloaded" = "";
     "nix-public" = "";
   };
   "artifacts" = {
+    "nix-local-preloaded/file:/Users/kreisys/.sbtix/preloaded/com/eed3si9n/shaded-scalajson_2.12/1.0.0-M4/shaded-scalajson_2.12-1.0.0-M4.jar" = {
```
mirkoAlic, line 12: karma, is that you?

PR #663, opened 2020-09-11
[CHORE] Testnet config

mmrozek:
# Description

Initial config for the private testnet:

1. Node syncs using regular sync
2. Mining is enabled
3. Phoenix fork config will be a base for chain
4. Disabled features:
- pruning
- discovery
- blacklisting

### DIFF
--- a/src/main/resources/chains/private.conf
+++ b/src/main/resources/chains/private.conf
@@ -1,8 +1,20 @@
 {
-  include "base.conf"
-  chain-id = "0x2A"
-  network-id = 42
-  custom-genesis-file = "chains/private-genesis.json"
-  dao = null
-  bootstrap-nodes = []
+    include "base.conf"
+    frontier-block-number = "0"
+    homestead-block-number = "0"
+    eip150-block-number = "0"
+    eip155-block-number = "0"
+    eip160-block-number = "0"
+    difficulty-bomb-pause-block-number = "0"
+    difficulty-bomb-continue-block-number = "0"
+    difficulty-bomb-removal-block-number = "0"
+    atlantis-block-number = "0"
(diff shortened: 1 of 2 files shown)


kreisys (approved):
I cherry-picked the commit on top of `chore/update-sbt-add-nix` and it seems to make my problems go away so :shipit: plz

src/universal/conf/private-testnet.conf
```diff
@@ -12,12 +12,21 @@ mantis {
...
+        mode = "archive"
+    }
   }
 
   blockchains {
     network = "private"
   }
 
+  consensus {
+    coinbase = "0011223344556677889900112233445566778899" # has to be changed for each node
```
mirkoAlic, line 26: why this is part of the default?

src/main/resources/chains/private.conf
```diff
@@ -1,8 +1,20 @@
 {
-  include "base.conf"
-  chain-id = "0x2A"
-  network-id = 42
-  custom-genesis-file = "chains/private-genesis.json"
-  dao = null
-  bootstrap-nodes = []
+    include "base.conf"
+    frontier-block-number = "0"
```
KonradStaniec, line 3: I would double check if we do not have some boolean conflicts when are block numbers are equal to 0 (like in ets test sometimes we have fork block equal to `-1` to keep the ordering between forks)

KonradStaniec (approved):
Small comment to double check, also you need to sign your commit :)

src/universal/conf/private-testnet.conf
mmrozek: I wanted to have anything here(it is the default from application.conf). For each node, we have to obtain a different account

PR #666, opened 2020-09-11
[CHORE] Mordor testnet support

ntallar:
# Description

Configurations were taken from geth
- Bootstrap nodes: https://github.com/etclabscore/core-geth/blob/master/params/bootnodes_mordor.go
- Consensus configuration: https://github.com/etclabscore/core-geth/blob/master/params/config_mordor.go
- Genesis block: https://github.com/etclabscore/core-geth/blob/master/params/genesis_mordor.go

## Differences in config with Geth

- Mantis has disabled EIP161 while Geth doesn't, as it's enabled by activating the Atlantis hard fork

# Testing

Starting up Mantis with Mordor selected and syncing to the top of the chain

I already did that on my machine and successfully synced around 2.2 million blocks


mirkoAlic (approved):
LGTM!

PR #667, opened 2020-09-14
[ETCM-107] Backport logging configuration

kapke:
# Description

Backport Logging configuration to allow:
  - logs in JSON (currently disabled, cc @mirkoAlic)
  - setting logs configuration within application.conf
  - automatic adjustment of logs placement depending on datadir setting (important for Luna)

# Important changes introduced

This PR changes logging configuration in a way that may affect deployments. How can I check whether that's true?

# Testing

1. Check whether logs are in same place as previously
2. Check whether logs follow settings under `mantis.logging`


src/main/resources/logback.xml
```diff
@@ -26,8 +42,16 @@
...
+        <if condition='p("ASJSON").contains("true")'>
+            <then>
+                <appender-ref ref="STASH" />
+            </then>
+            <else>
+                <appender-ref ref="STDOUT" />
+            </else>
+        </if>
         <appender-ref ref="FILE" />
+        <appender-ref ref="METRICS" />
```
mirkoAlic, line 54: should be included now?

mirkoAlic (approved):
(only one minor comment/doubt) LGTM!

src/main/resources/application.conf
```diff
@@ -536,6 +536,17 @@ mantis {
     # Iff true, any errors during metrics client operations will be logged.
     log-errors = true
   }
+
+  logging {
```
mmrozek, line 540: Shouldn't we expose this config in the universal folder also? Maybe the file `logging.conf` will be useful.

src/main/resources/logback.xml
kapke: Not! I did remove appender but forgot to remove ref to it

mmrozek (approved):
LGTM

PR #671, opened 2020-09-16
Etcm 52 make sure mantis works on windows

kapke:
# Description

This PR fixes issues found when trying to run Mantis on Windows, that is:
  - generating proper `.bat` files
  - fixing the way `SecureRandom` instance is obtained
  - KeyStore tests which were too UNIX-oriented
  - Update RocksDB to latest version, so Ethereum fork doesn't need to be used

# Testing

All tests, `sbt dist` and syncing from both sbt and dist package should work

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -121,6 +121,7 @@ scalacOptions in (Compile, console) ~= (_.filterNot(
     "-Xfatal-warnings"
   )
 ))
+Global / onChangedBuildSource := ReloadOnSourceChanges
 
 Test / parallelExecution := false
 
--- a/build.sbt
+++ b/build.sbt
@@ -87,8 +87,7 @@ val root = {
     .configs(Integration, Benchmark, Evm, Ets, Snappy, Rpc)
     .settings(commonSettings: _*)
     .settings(
-      libraryDependencies ++= dep,
-      executableScriptName := name.value
+      libraryDependencies ++= dep
     )
     .settings(inConfig(Integration)(Defaults.testSettings): _*)
     .settings(inConfig(Benchmark)(Defaults.testSettings): _*)
@@ -121,11 +120,8 @@ scalacOptions in (Compile, console) ~= (_.filterNot(
     "-Xfatal-warnings"
   )
 ))
-Global / onChangedBuildSource := ReloadOnSourceChanges
 
 Test / parallelExecution := false
-
-ThisBuild / turbo := true
 
 testOptions in Test += Tests.Argument("-oDG")
 
@@ -145,7 +141,8 @@ scalastyleSources in Test ++= { (unmanagedSourceDirectories in Integration).valu
 
 // Packaging
 mainClass in Compile := Some("io.iohk.ethereum.App")
-discoveredMainClasses in Compile := Seq("io.iohk.ethereum.mallet.main.Mallet")
+Universal / executableScriptName := name.value
+discoveredMainClasses in Compile := Seq()
 // Requires the 'ant-javafx.jar' that comes with Oracle JDK
 // Enables creating an executable with the configuration files, has to be run on the OS corresponding to the desired version
 ThisBuild / jdkPackagerType := "image"
--- a/build.sbt
+++ b/build.sbt
@@ -13,14 +13,11 @@ val commonSettings = Seq(
     .Argument(TestFrameworks.ScalaTest, "-l", "EthashMinerSpec") // miner tests disabled by default
 )
 
-// Resolver for rocksDb
-resolvers += "rocksDb" at "https://dl.bintray.com/ethereum/maven/"
-
 val dep = {
   val akkaVersion = "2.6.9"
   val akkaHttpVersion = "10.2.0"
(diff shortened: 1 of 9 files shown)


build.sbt
```diff
@@ -100,8 +97,11 @@ scalacOptions in (Compile, console) ~= (_.filterNot(Set(
   "-Ywarn-unused-import",
   "-Xfatal-warnings"
 )))
+Global / onChangedBuildSource := ReloadOnSourceChanges
+
+Test / parallelExecution := false
 
-parallelExecution in Test := false
+ThisBuild / turbo := true
```
ntallar, line 104: Why should we enable this? From the description:

>sbt 1.3.0 adds “turbo” mode that enables experimental or advanced features that might require some debugging by the build user when it doesn’t work. 

The advanced features might be useful, but I'm a bit cautious on opening the door to any experimental features

build.sbt
```diff
@@ -119,11 +119,12 @@ unmanagedResourceDirectories in Compile += baseDirectory.value / "src" / "main"
 (scalastyleConfig in Test) := baseDirectory.value / "scalastyle-test-config.xml"
 scalastyleSources in Test ++= {(unmanagedSourceDirectories in Integration).value}
 
+// Packaging
 mainClass in Compile := Some("io.iohk.ethereum.App")
-
+discoveredMainClasses in Compile := Seq("io.iohk.ethereum.mallet.main.Mallet")
```
ntallar, line 124: This doesn't seem to work on my machine:

```
 % ./bin/mallet
Error: Could not find or load main class io.iohk.ethereum.mallet.main.Mallet
```

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -193,10 +193,10 @@ class RocksDbDataSource(
 
       val tableCfg = new BlockBasedTableConfig()
         .setBlockSize(blockSize)
-        .setBlockCacheSize(blockCacheSize)
+        .setBlockCache(new ClockCache(blockCacheSize))
```
ntallar, line 196: Just to understand the decision, why ClockCache and not LRUCache? From what I read LRUCache has more overhead

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
```diff
@@ -510,7 +510,7 @@ trait GenesisDataLoaderBuilder {
 
 trait SecureRandomBuilder {
   lazy val secureRandom: SecureRandom =
-    Config.secureRandomAlgo.map(SecureRandom.getInstance).getOrElse(new SecureRandom())
+    Config.secureRandomAlgo.flatMap(name => Try(SecureRandom.getInstance(name)).toOption).getOrElse(new SecureRandom())
```
ntallar, line 513: Doesn't this hide the fact that the selected algorithm is available and other one is instead being used?

ntallar (changes requested):
Both running Mantis from `sbt dist` or `sbt run` worked on my machine! (it's a linux machine)

build.sbt
kapke: One of most important things it changes is classloader layering. Enabling turbo mode didn't broke anything so my preference is to keep it enabled and disable it only when it does break something for us.

build.sbt
kapke: Yup, once I realaized that this probably needs to be adjusted, @mirkoAlic confirmed for me that mallet can be removed so there will be a separate PR soon I think

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
kapke: Exactly this - documentation says they're very similar but `ClockCache` can perform better in certain scenarios so it seem to be better option for start.

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
kapke: If I add logging about algorithm non available, will it be sufficient?

KonradStaniec (approved):
I had run mantis on macos, from pre-exsiting database and it seems to work fine. @kapke did you check it on windows ?

kapke:
I did, whole PR was prepared on Windows machine

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
ntallar: Yes, given our current handling of config errors I think that with logging with error will be enough here

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

PR #673, opened 2020-09-16
[ETCM-73] Transactional blockchain

pslaski:
# Description

The goal is to make part of the blockchain methods transactional.

- introduced DataSourceBatchUpdate
- blockchain methods which are saving to DB are transactional now
- removed NodeStorage interface (NodeStorage and CachedNodeStorage are independent now)

# Testing

check syncing and pruning

# MISC

Storing best-known block number mechanism is using AtomicReference + DB now. Is it worth to create a ticket which will simplify that?


src/main/scala/io/iohk/ethereum/db/storage/StateStorage.scala
```diff
@@ -98,7 +97,7 @@ class ReferenceCountedStateStorage(private val nodeStorage: NodeStorage,
   }
 
   override def saveNode(nodeHash: NodeHash, nodeEncoded: NodeEncoded, bn: BigInt): Unit = {
-    new FastSyncNodeStorage(nodeStorage, bn).update(Nil, Seq(nodeHash -> nodeEncoded))
+    new FastSyncNodeStorage(cachedNodeStorage, bn).updateInStorage(Nil, Seq(nodeHash -> nodeEncoded))
```
ntallar, line 100: Why was the nodeStorage replaced here by the cachedNodeStorage?

src/main/scala/io/iohk/ethereum/db/storage/ArchiveNodeStorage.scala
```diff
@@ -8,14 +8,16 @@ import io.iohk.ethereum.mpt.NodesKeyValueStorage
...
+    cachedNodeStorage.update(Nil, toUpsert)
     this
   }
 
-  override def get(key: NodeHash): Option[NodeEncoded] = nodeStorage.get(key)
+  override def get(key: NodeHash): Option[NodeEncoded] = cachedNodeStorage.get(key)
 
-  override def persist(): Unit = {}
+  override def persist(): Unit = {
+    cachedNodeStorage.forcePersist()
```
ntallar, line 21: Why not cachedNodeStorage.persist()?

src/main/scala/io/iohk/ethereum/db/storage/CachedNodeStorage.scala
```diff
@@ -0,0 +1,22 @@
...
+class CachedNodeStorage(val storage: NodeStorage, val cache: Cache[NodeHash, NodeEncoded])
+  extends CachedKeyValueStorage[NodeHash, NodeEncoded, CachedNodeStorage] {
+  override type I = NodeStorage
+
+  override def apply(cache: Cache[NodeHash, NodeEncoded], storage: NodeStorage): CachedNodeStorage = new CachedNodeStorage(storage, cache)
+}
+
+object CachedNodeStorage {
+  // special storage where we want to minimize cache layer because datasource is in memory
+  def withMinimalCache(dataSource: EphemDataSource): CachedNodeStorage =
```
ntallar, line 20: Doesn't the 0 below mean that it has no caching instead of minimal?

ntallar (changes requested):
>removed NodeStorage interface (NodeStorage and CachedNodeStorage are independent now)

Was this needed? Our MPT node handling is way too overly complicated and I don't see the benefit of that change, should we maybe keep it as is till we dedicate fully to a better design of the whole of it?

>Storing best-known block number mechanism is using AtomicReference + DB now. Is it worth to create a ticket which will simplify that?

For me yes! But if it's not a blocker for any task we'll probably kick working on it for later

src/main/scala/io/iohk/ethereum/db/storage/FastSyncNodeStorage.scala
```diff
@@ -11,21 +11,34 @@ import io.iohk.ethereum.db.storage.encoding._
   * node saved will have its reference count equal to 1.
   *
   * */
-class FastSyncNodeStorage(nodeStorage: NodesStorage, bn: BigInt) extends ReferenceCountNodeStorage(nodeStorage, bn) {
+class FastSyncNodeStorage(cachedNodeStorage: CachedNodeStorage, bn: BigInt) extends ReferenceCountNodeStorage(cachedNodeStorage, bn) {
```
KonradStaniec, line 14: FastSync does not have concept of caching it need to save data into database directly.

src/main/scala/io/iohk/ethereum/db/dataSource/DataSourceBatchUpdate.scala
```diff
@@ -0,0 +1,17 @@
+package io.iohk.ethereum.db.dataSource
+
+case class DataSourceBatchUpdate(dataSource: DataSource, updates: Seq[DataSourceUpdate] = Nil) {
+
+  def and(that: DataSourceBatchUpdate): DataSourceBatchUpdate = {
+    require(
+      this.dataSource eq that.dataSource,
+      "Transactional storage updates must be performed on the same data source"
+    )
+    DataSourceBatchUpdate(dataSource, this.updates ++ that.updates)
```
KonradStaniec, line 10: I am wondering if queue with appending or list with prepending would not be better here ? As otherwise when batching a lot of updates from list (like in FastSync) we have quite an overhead

KonradStaniec (commented):
Few comments for now.
General remarks:

src/main/scala/io/iohk/ethereum/db/dataSource/EphemDataSource.scala
```diff
@@ -9,38 +9,37 @@ class EphemDataSource(var storage: Map[ByteBuffer, Array[Byte]]) extends DataSou
     * key.drop to remove namespace prefix from the key
     * @return key values paris from this storage
     */
-  def getAll(namespace: Namespace): Seq[(IndexedSeq[Byte], IndexedSeq[Byte])] =
-    storage.toSeq.map{case (key, value) => (key.array().drop(namespace.length).toIndexedSeq, value.toIndexedSeq)}
+  def getAll(namespace: Namespace): Seq[(IndexedSeq[Byte], IndexedSeq[Byte])] = synchronized {
```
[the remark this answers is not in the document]
pslaski: the same change was introduced in our first `template` PR ;)

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
```diff
@@ -21,45 +21,16 @@ trait DataSource {
...
-    * This function updates the DataSource by deleting, updating and inserting new (key-value) pairs.
-    * It assumes that caller already properly serialized key and value.
-    * Useful when caller knows some pattern in data to avoid generic serialization.
-    *
-    * @param toRemove which includes all the keys to be removed from the DataSource.
-    * @param toUpsert which includes all the (key-value) pairs to be inserted into the DataSource.
-    *                 If a key is already in the DataSource its value will be updated.
-    * @return the new DataSource after the removals and insertions were done.
-    */
-  def updateOptimized(toRemove: Seq[Array[Byte]], toUpsert: Seq[(Array[Byte], Array[Byte])]): DataSource
```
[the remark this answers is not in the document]
pslaski: I created new `DataSourceUpdateOptimized` type for optimized execution

src/main/scala/io/iohk/ethereum/db/dataSource/DataSourceBatchUpdate.scala
pslaski: Used Array based on https://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#concatenation-performance

src/main/scala/io/iohk/ethereum/db/dataSource/EphemDataSource.scala
```diff
@@ -9,38 +9,48 @@ class EphemDataSource(var storage: Map[ByteBuffer, Array[Byte]]) extends DataSou
...
+  def update(namespace: Namespace, toRemove: Seq[Key], toUpsert: Seq[(Key, Value)]): Unit = synchronized {
+    val afterRemoval =
+      toRemove.foldLeft(storage)((storage, key) => storage - ByteBuffer.wrap((namespace ++ key).toArray))
+    val afterUpdate = toUpsert.foldLeft(afterRemoval)(
+      (storage, toUpdate) => storage + (ByteBuffer.wrap((namespace ++ toUpdate._1).toArray) -> toUpdate._2.toArray)
+    )
+    storage = afterUpdate
+  }
+
+  def updateOptimized(toRemove: Seq[Array[Byte]], toUpsert: Seq[(Array[Byte], Array[Byte])]): Unit = synchronized {
```
ntallar, line 40: Minor: should this and the update method (`def update(namespace: Namespace, toRemove: Seq[Key], toUpsert: Seq[(Key, Value)]): Unit`) be private?

src/test/scala/io/iohk/ethereum/mpt/PersistentStorage.scala
```diff
@@ -26,9 +31,18 @@ trait PersistentStorage {
...
+    override val maxHoldTime = FiniteDuration(5, TimeUnit.MINUTES)
+  }
+
   private def testExecution(testCode: MptStorage => Unit, dbPath: String, dataSource: DataSource): Unit = {
     try {
-      testCode(new SerializingMptStorage(new ArchiveNodeStorage(new NodeStorage(dataSource))))
+      testCode(new SerializingMptStorage(
+        new ArchiveNodeStorage(
+          new CachedNodeStorage(
+            new NodeStorage(dataSource), MapCache.createCache(testCacheConfig)
```
ntallar, line 44: Does this change still make sense?

src/test/scala/io/iohk/ethereum/db/storage/TransactionalKeyValueStorageSuite.scala
```diff
@@ -0,0 +1,122 @@
...
+    override def keySerializer: Int => IndexedSeq[Byte] = intSerializer
+    override def valueSerializer: Int => IndexedSeq[Byte] = intSerializer
+    override def valueDeserializer: IndexedSeq[Byte] => Int = intDeserializer
+  }
+
+  def newIntStorage(): IntStorage = new IntStorage(EphemDataSource())
+
+  test("Get ints from KeyValueStorage") {
+    forAll(Gen.listOf(intGen), Gen.listOf(intGen)) { (intsInStorage, unfilteredIntsNotInStorage) =>
+      val intsNotInStorage = unfilteredIntsNotInStorage.distinct diff intsInStorage
```
ntallar, line 34: Given how this is generated, couldn't it often be the case that intsNotInStorage is empty?

ntallar (commented):
>What do you have in mind by this simplification ? I mean the main constraint here is that it needs to be quite performant as it is frequently used thing, and we cannont update best block in db in each best block update as we do not now if node cache has been flushed. So the idea here is to udpdate the block in db only when flushing the cache, and otherwise operate on this counter.

Hmm given that, couldn't we hide the bestKnowBlock (which I would rename to inMemoryBestBlock or cachedBestBlock) in the blockchain interface? That is, instead of `saveBestKnownBlock` and `saveBestBlock` have a single `saveBestBlock`

I see the constraint that this can't be updated before flushing the nodes cache but not the performance constraint. Usually when updating the bestBlock it's because we are removing or inserting whole blocks, aren't those operations much heavier than this?

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

KonradStaniec (approved):

PR #674, opened 2020-09-16
[CHORE] BlockchainConfig as a case class

ntallar:
# Description

Changes BlockchainConfig from being a trait to a case class.

As seen on a couple of examples here, this will:
- Allow us to use the default copy to only change the parts of the configuration on tests that applies to each
- Follow our current convention of having all our configurations as case classes


src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
```diff
@@ -43,6 +43,8 @@ object ScenarioSetup {
 
 abstract class ScenarioSetup(_vm: VMImpl, scenario: BlockchainScenario) {
 
+  import BlockchainTestConfig._
```
mirkoAlic, line 46: is this needed?

src/test/scala/io/iohk/ethereum/jsonrpc/PersonalServiceSpec.scala
```diff
@@ -416,37 +416,37 @@ class PersonalServiceSpec extends FlatSpec with Matchers with MockFactory with S
...
     val txValue = 128000
 
-    val blockchainConfig = new BlockchainConfig {
-      override val eip155BlockNumber: BigInt = 12345
-      override val chainId: Byte = 0x03.toByte
+    val blockchainConfig = BlockchainConfig (
+      eip155BlockNumber = 12345,
+      chainId = 0x03.toByte,
 
       //unused
```
mirkoAlic, line 423: all the unused can be removed i guess

src/test/scala/io/iohk/ethereum/network/handshaker/EtcHandshakerSpec.scala
```diff
@@ -172,43 +172,43 @@ class EtcHandshakerSpec extends FlatSpec with Matchers  {
       override val appStateStorage: AppStateStorage = TestSetup.this.storagesInstance.storages.appStateStorage
     }
 
-    override lazy val blockchainConfig = new BlockchainConfig {
+    override lazy val blockchainConfig = BlockchainConfig (
```
mirkoAlic, line 175: Here you could use the copy given `EphemBlockchainTestSetup` has access to blockchainConfig "default".

mirkoAlic (approved):
LGTM!

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
ntallar: It is, as baseBlockchainConfig and withSkippedPoWValidationBlockchainConfig functions use the BlockchainConfig's defined there

PR #675, opened 2020-09-16
[ETCM-43][ETCM-46] Add opt-out field

ntallar:
# Description

First changes for supporting treasury (as specified [here](https://input-output.atlassian.net/wiki/spaces/EMU/pages/1681293384/Proto-Treasury))

~Depends on #674~

# Proposed Solution

- Adds opt-out field on block headers
- Gives possibility for miner to selecting whether to opt-out or not of supporting treasury

# Pending

- [x] Adds tests for validation changes
- [x] Test on the mainnet
- [x] Test on a private network with ECIP activation

# Testing

- Mantis shouldn't change it's behaviour when connected to the ETC mainnet or mordor
- Creating a private network with lower ECIP1098 number should result in only blocks after ECIP1098 having the selected opt-out


src/main/resources/chains/private.conf
```diff
@@ -11,6 +11,8 @@
     atlantis-block-number = "0"
     agharta-block-number = "0"
     phoenix-block-number = "0"
+    # FIXME: Once the testnet is up an running we should determine this value
```
mirkoAlic, line 14: maybe we should provide a new file especially for the new testnet, WDTY?

src/main/scala/io/iohk/ethereum/consensus/validators/BlockHeaderValidatorSkeleton.scala
```diff
@@ -163,7 +164,19 @@ abstract class BlockHeaderValidatorSkeleton(blockchainConfig: BlockchainConfig)
...
+    * Validates [[io.iohk.ethereum.domain.BlockHeader.optOut]] is only defined if ECIP1098 is enabled at the block's number
+    *
+    * @param blockHeader BlockHeader to validate.
+    * @return BlockHeader if valid, an [[HeaderOptOutError]] otherwise
+    */
+  private def validateOptOut(blockHeader: BlockHeader): Either[BlockHeaderError, BlockHeaderValid] = {
+    val isEcip1098Activated = blockHeader.number >= blockchainConfig.ecip1098BlockNumber
+    val isOptOutDefined = blockHeader.optOut.isDefined
+
+    if(isEcip1098Activated && isOptOutDefined) Right(BlockHeaderValid)
```
mirkoAlic, line 177: (formatting) `if (`

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
```diff
@@ -58,13 +60,19 @@ case class BlockHeader(
...
 object BlockHeader {
 
+  private val NumberOfFields = 16
+
   val emptyOmmerHash = ByteString(Hex.decode("1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347"))
 
   def getEncodedWithoutNonce(blockHeader: BlockHeader): Array[Byte] = {
     val rlpEncoded = blockHeader.toRLPEncodable match {
-      case rlpList: RLPList =>
+      case rlpList: RLPList if rlpList.items.length == (NumberOfFields - 1) =>
```
mirkoAlic, line 69: i prefer something like: `if (isPreviousToEcip1089(rlpList) =>` WDYT?

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
```diff
@@ -58,13 +60,19 @@ case class BlockHeader(
...
         RLPList(rlpList.items.dropRight(2): _*)
 
+      case rlpList: RLPList if rlpList.items.length == NumberOfFields =>
```
mirkoAlic, line 72: ditto

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
```diff
@@ -73,8 +81,17 @@ object BlockHeader {
   implicit class BlockHeaderEnc(blockHeader: BlockHeader) extends RLPSerializable {
```
mirkoAlic, line 81: A comment regarding the reason behind `optOut match {` will help future generations i guess..hehe.

src/test/scala/io/iohk/ethereum/consensus/validators/BlockHeaderValidatorSpec.scala
```diff
@@ -180,6 +177,39 @@ class BlockHeaderValidatorSpec
...
+
+  it should "mark as invalid a pre ecip1098 block opt-out with opt out defined" in new EphemBlockchainTestSetup {
+    val ecip1098BlockNumber = validBlockHeader.number * 2
+    val blockchainConfigWithECIP1098Disabled: BlockchainConfig = blockchainConfig.copy(ecip1098BlockNumber = ecip1098BlockNumber)
+    val blockHeaderValidator = new BlockValidatorWithPowMocked(blockchainConfigWithECIP1098Disabled)
+
+    val headerWithOptOutInvalidlyOn = validBlockHeader.copy(optOut = Some(true))
+
+    val validationResult = blockHeaderValidator.validate(headerWithOptOutInvalidlyOn, validParentBlockHeader)
+    validationResult shouldBe Left(HeaderOptOutError)
```
mirkoAlic, line 199: (minor) maybe we could provide differenciation between this scenario and the following one, i mean in terms of `HeaderOptOutError` description. Still not sure if is worthy, wdyt?

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
ntallar: I found a way I preferred most, to not require the NumberOfFields value, I'll soon push it

src/main/resources/chains/private.conf
ntallar: Should we still keep the configurations for a private testnet? It could get a bit messy given that we have private.conf, private.json and private-testnet.conf

We could have for our chains/ files:
- Keep private-testnet.conf (current private.conf) and private-testnet.json (current private.json), with private-testnet.conf having ecip1098 activated at block 0 as currently
- Add internal-testnet.conf and internal-tesnet.json for our testnet

We could have for our conf/ files:
- Have a testnet-changes.conf file with the testnet configuration overrides (i.e. disabling pruning and discovery)
- Have a private-testnet.conf and a internal-testnet.conf that would just select the correct chain

mirkoAlic: maybe we could discuss this in a call.

mirkoAlic (approved):
LGTM!

src/main/resources/chains/private.conf
ntallar: Conclusion from call: https://github.com/input-output-hk/mantis/pull/681

src/main/scala/io/iohk/ethereum/consensus/ConsensusConfig.scala
```diff
@@ -32,6 +33,7 @@ object ConsensusConfig extends Logger {
     final val HeaderExtraData = "header-extra-data"
     final val BlockCacheSize = "block-cashe-size"
     final val MiningEnabled = "mining-enabled"
+    final val OptOut = "opt-out"
```
mmrozek, line 36: I am wondering if `opt-out` is not too general. Maybe `treasury-opt-out` or sth like that?

ntallar: For the configuration only or do you mean for the header field as well? I was in doubt whether renaming it to treasuryOptOut or not

mmrozek: I would prefer to change it in both places. It will be more consistent

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
```diff
@@ -135,7 +135,7 @@ class GenesisDataLoader(
       extraData = genesisData.extraData,
       mixHash = genesisData.mixHash.getOrElse(zeros(hashLength)),
       nonce = genesisData.nonce,
-      optOut = None
+      treasuryOptOut = None
```
mmrozek, line 138: 👍

mmrozek (approved):
LGTM!

PR #676, opened 2020-09-16
[CHORE] Refactor dependencies, update scalacheck

mmrozek:
# Description

- Move dependencies to separate file
- update scalacheck

CAUTION: a majority of changes is caused by renaming of `PropertyChecks` trait and formatting all touched files. Please don't be afraid ;)


repo.nix
```diff
@@ -5,10 +5,119 @@
       "sbtVersion" = "1.2.8";
     }];
   "repos" = {
+    "nix-local-preloaded" = "";
     "nix-public" = "";
     "nix-rocksDb" = "";
   };
   "artifacts" = {
+    "nix-local-preloaded/file:/Users/mmrozek/.sbtix/preloaded/com/trueaccord/lenses/lenses_2.12/0.4.12/lenses_2.12-0.4.12.jar" = {
```
mirkoAlic, line 13: =/

.scalafmt.conf
```diff
@@ -1,3 +1,3 @@
-version = "2.3.0"
+version = "2.7.1"
```
ntallar, line 1: What do you think about including scalafmt checks on circleci?

mmrozek: At some point, we need to do that. But I prefer to do that in separate PR

kapke (approved):
One minor thing and look good to me!

PR #678, opened 2020-09-17
[FIX] Add mocked miner validations

ntallar:
# Description

A couple of mocked miner fixes:
- Adds the mocked miner as one of the allowed protocols
- Adds mocked pow validations so that imported blocks created by mock miner doesn't fail

# Testing

Setup:
- mantis.consensus.protocol = mocked
- mantis.consensus.mining-enabled = true
- mantis.blockchains.network = "private"
- mantis.sync.do-fast-sync = false

And use the mocked miner QA endpoint to generate blocks


mmrozek (approved):
LGTM!

PR #680, opened 2020-09-18
[Chore] Rename sbt project name back to mantis

kapke:
# Description

After finally being able to build mantis locally with nix I could verify whether that renaming is still necessary. It seems that changes from #671 fixed the issue so project can be renamed back to `mantis`


mirkoAlic (approved):
cool!

PR #681, opened 2020-09-18
[ETCM-121] Refactor configs

ntallar:
# Description

Different changes with the objectives of:
- Naming our current testnet to internal
- Allow starting Mantis on different environments without having to comment/uncomment lines of configuration files

- Removes private testnet configurations as there is no use case for them yet
- Adds testnet-internal configurations
- Lots of renamings (i.e etc.json -> etc-genesis.json, etc.conf -> etc-chain.conf)

# Up-to-discussion

- Removing "user friendly" configurations as there haven't proven to be very used or even friendly (the other changes are compatible with this if we want to have them)
- Should we enable HTTPS on the application.conf instead?

# Testing

- Sbt dist and sbt run should work as always with trying out all the different environments (the packaged binaries now list all the environments currently supported)


src/universal/bin/eckeygen
```diff
@@ -2,4 +2,4 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd $DIR/..
-./bin/mantis -- eckeygen "$@"
+./bin/mantis-core -- eckeygen "$@"
```
mirkoAlic, line 5: should be: `mantis`  (This was fixed here: https://github.com/input-output-hk/mantis/pull/671)

src/universal/bin/faucet-server
```diff
@@ -2,4 +2,4 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd $DIR/..
-./bin/mantis -Dconfig.file=./conf/faucet.conf -- faucet "$@"
+./bin/mantis-core -Dconfig.file=./conf/faucet.conf -- faucet "$@"
```
mirkoAlic, line 5: ditto

src/universal/bin/mantis-mordor
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+./bin/mantis-core -Dconfig.file=./conf/base-app.conf -Dmantis.blockchains.network=mordor "$@"
```
mirkoAlic, line 3: i prefer:
 `./bin/mantis-core -Dconfig.file=./conf/mordor.conf` with `conf/mordor.conf`:
```
include app.conf
blockchains.network = "mordor"
```

src/universal/conf/app.conf
```diff
@@ -0,0 +1,23 @@
+# This is the base configuration file for the Mantis ETC client.
+
+# This where all the default settings are defined (this file is packaged within the executable).
+# It should always go at the top.
+include "application.conf"
+
+# Enable HTTPS
```
ntallar, line 7: This was a leftover copy from network.conf, does it still make sense? Shouldn't it be enabled in the application.conf?

src/universal/conf/testmode.conf
```diff
@@ -1,3 +1,5 @@
+include "app.conf"
```
mirkoAlic, line 1: no from your PR but i noticed that in this file is missing `qa` rpc.apis option :P

mirkoAlic:
I think the is the way to go =) great job!

mirkoAlic:
please update the README -> https://github.com/input-output-hk/mantis#command-line-version

src/main/resources/chains/etc-chain.conf
```diff
@@ -0,0 +1 @@
+{ include "base-chain.conf"}
```
mmrozek, line 1: Do we need `base-chain`?

src/universal/conf/app.conf
mmrozek, line 7: I prefer to move the https config to a separate file and here add only `include https.conf`

src/main/resources/chains/etc-chain.conf
kapke: It was created as etc and eth networks share a lot of common configuration

mmrozek: Yes, but base == etc. In my opinion, we don't need it

src/universal/conf/app.conf
```diff
@@ -0,0 +1,23 @@
...
+mantis {
+  network {
+    rpc {
+      http {
+        mode = "https"
+
+        certificate-keystore-path = "conf/mantis.jks"
```
kapke, line 14: Should HTTPS be pre-configured with these values or disabled by default so user can configure it on its own?

kapke:
Misclick 😅 
Regarding launcher scripts - I think we should add `*.bat` versions of them as well.

mmrozek:
@kapke, according to your last comment, maybe the generic script will be enough?
sth like:
```
#!/bin/bash

chain="$1"
if [ -z "$chain"]
then
  echo "You need to choose chain"
else
  shift
  ./bin/mantis -Dconfig.file=./conf/"$chain".conf "$@"
fi

**Comment by kapke** [comment]
@mmrozek  I'm not sure. There are 2 issues in that approach:
  - how to name generic mantis launcher so it doesn't interfere with that one
  - there's need to write and maintain `*.bat` version of it

File: src/universal/conf/app.conf (hunk 7 above)
Comment (ntallar) [reply src/universal/conf/app.conf:14]: Imo for now I'd just disable it and have the user configure it on its own, we can later add it for development's purposes or provide some guide on to how the user should configure it

File: src/universal/conf/app.conf (hunk 4 above)
Comment (ntallar) [reply src/universal/conf/app.conf:7]: As commented [here](https://github.com/input-output-hk/mantis/pull/681#discussion_r492147771) I will remove for now this configuration

**Comment by ntallar** [comment]
Maybe it's better than nothing to have the script that Michal mentioned? (maybe we could name it mantis-app or mantis-launcher). We could add the .bat version in a separate PR.

The other alternative is just removing this script, is it used by any for now?

File: src/main/resources/chains/etc-chain.conf (hunk 6 above)
Comment (ntallar) [reply src/main/resources/chains/etc-chain.conf:1]: Hmm I'm also a bit on Michal's side in can be a bit confusing

In my opinion I'd just replicate the duplicated configuration in both etc-chain.conf and eth-chain.conf, just to have to check a single file to see all the configurations of a chain. This kinds of files will rarely be touched so I don't see that being too much of a problem

**Review by mmrozek — APPROVED** [review]
lgtm!

File: src/main/resources/chains/etc-chain.conf (hunk 6 above)
Comment (kapke) [reply src/main/resources/chains/etc-chain.conf:1]: Go for it then!

File: src/universal/bin/mantis-launcher [hunk 8]
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
kapke, line 1: It doesn't have `execute` permission

src/main/resources/chains/etc-chain.conf
ntallar: Just updated them! Please check that I haven't screwed up in the middle

I noticed some ropsten configuration wasn't properly configured but I'd leave it as is to not fall down in that rabbit hole, if we ever start providing official support for it (seems to be the intent eventually) we'll fix it then

mmrozek: I prefer to stay with `etc.conf` instead of `base.conf`. From a user perspective `mantis-launcher etc` makes more sense than `mantis-launcher base`

src/main/resources/chains/eth-chain.conf
```diff
@@ -197,6 +218,51 @@
...
 
+  # Starting nonce of an empty account. Some networks (like Morden) use different values.
+  account-start-nonce = "0"
+
+  # The ID of the accepted chain
+  chain-id = "1"
+
+  # Custom genesis JSON file path
+  # null value indicates using default genesis definition that matches the main network
+  custom-genesis-file = null
```
kapke, line 229: minor: should we maybe get rid of optional genesis path? (separate task for Piotr or Adiran maybe?)

ntallar: Totally! Here's the task for that: https://jira.iohk.io/browse/ETCM-128

kapke (approved):
LGTM!

PR #682, opened 2020-09-21
[CHORE] Imporve fast sync config params

KonradStaniec:
# Description

With those params and with #672 , fast sync appears to be fast enough to sync up to the top of the mainnet before peers prune their stats


ntallar:
Were any of this configurations taken from geth or were they all obtained from trial and error with different values?

KonradStaniec:
so:
1. Throttling main purpose wast to not be blacklisted by remote peers, but tbh i did not see such a behaviour over several runs
2. With restartSync idle set to 15s fastsync sometimes just sit idly so it does make little sense to have it such large
3. FastSyncTargetOffset is set in manner similar to geth i.e geth actually choses top block as target but declares that target is stale when its 128 blocks behind the top, as for now we can't set target block as top network block, but we can minimize the offset to reduce the possibility of remote peer pruning their stale nodes.

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

mmrozek (approved):
LGTM. The only thing I would like to fix is to change a [log](https://github.com/input-output-hk/mantis/blob/tweak/tweak-fast-sync-config-params/src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala#L600). With retry interval = 0,5 s there is a lot of garbage logged

PR #683, opened 2020-09-21
[ETCM-110] Add eth_chainId RPC method

kapke:
# Description

According to Ethereum RPC Spec, Mantis lacks `eth_chainId` method, which makes it much harder to issue transactions from Luna.

PS. This PR got so big due to auto-formatting. Changes crucial for this task are really minimal (test for eth service, test for rpc controller, insomnia entry, controller entry, json codec, eth service method).

# Testing

1. All tests should pass
2. `eth_chainId` in Insomnia should return configured value in hex format according to spec (https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/etclabscore/ethereum-json-rpc-specification/master/openrpc.json&uiSchema%5BappBar%5D%5Bui:input%5D=false)

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
@@ -2,7 +2,8 @@ package io.iohk.ethereum.jsonrpc
 
 import akka.util.ByteString
 import io.iohk.ethereum.jsonrpc.EthService._
-import io.iohk.ethereum.jsonrpc.JsonRpcController.{JsonDecoder, JsonEncoder}
+import io.iohk.ethereum.jsonrpc.JsonRpcController.JsonDecoder.NoParamsDecoder
+import io.iohk.ethereum.jsonrpc.JsonRpcController.{Codec, JsonDecoder, JsonEncoder}
 import io.iohk.ethereum.jsonrpc.JsonRpcErrors.InvalidParams
 import io.iohk.ethereum.jsonrpc.PersonalService.{SendTransactionRequest, SendTransactionResponse, SignRequest}
 import org.json4s.{Extraction, JsonAST}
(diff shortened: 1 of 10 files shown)


src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -13,28 +13,42 @@ import org.json4s.JsonDSL._
...
+      ProtocolVersionRequest()
+    )
 
     def encodeJson(t: ProtocolVersionResponse): JValue = t.value
   }
 
+  implicit val eth_chainId = new JsonDecoder[ChainIdRequest] with JsonEncoder[ChainIdResponse] {
+    def decodeJson(params: Option[JArray]) = params match {
+      case None | Some(JArray(Nil)) => Right(ChainIdRequest())
+      case other => Left(InvalidParams("eth_chainId method does not take params"))
```
mmrozek, line 26: I realized that other parameterless methods don't check if `params` is empty. Maybe we could improve that in this PR. WDYT?

mmrozek (approved):
Minor comments only. LGTM!

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
kapke: If only @pslaski agrees (this will result in quite huge PR due to formatting)

pslaski: this PR is already huge so go for it ;)

src/main/scala/io/iohk/ethereum/jsonrpc/JsonRpcController.scala
```diff
@@ -23,11 +25,32 @@ object JsonRpcController {
   trait JsonDecoder[T] {
     def decodeJson(params: Option[JArray]): Either[JsonRpcError, T]
   }
+  object JsonDecoder {
+    abstract class NoParamsDecoder[T](request: => T) extends JsonDecoder[T] {
```
pslaski, line 29: ❤️

pslaski (approved):
LGTM, minor: I'm not sure if there are tests in JRC specs for the non-params case

kapke:
@pslaski From what I've checked there usually are such tests for happy path and I think it's enough since the behavior is uniform now across whole RPC API

PR #685, opened 2020-09-21
update dependencies

lemastero:
# Description

Update library dependencies.
(Current versions of monix and cats have some issues with diverging implicits so they are not updated.)

# Proposed Solution

# Testing

Tested by: `sbt clean test`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -16,8 +16,8 @@ val commonSettings = Seq(
 val malletDeps = Seq(
   Dependencies.scopt
 ).flatten ++ Seq(
-  "org.jline" % "jline" % "3.1.2",
-  "net.java.dev.jna" % "jna" % "4.5.1"
+  "org.jline" % "jline" % "3.16.0",
+  "net.java.dev.jna" % "jna" % "5.6.0"
 )
 
 val dep = {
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -16,8 +16,8 @@ val commonSettings = Seq(
 val malletDeps = Seq(
   Dependencies.scopt
 ).flatten ++ Seq(
-  "org.jline" % "jline" % "3.1.2",
-  "net.java.dev.jna" % "jna" % "4.5.1"
+  "org.jline" % "jline" % "3.16.0",
```
kapke, line 19: using the occasion: shouldn't jline and jna be picked from `project/Dependencies.scala` as well?

kapke (commented):
Remember to update `repo.nix` files as well, these are like lockfiles from other package managers, but required (and managed) by nix to have build reproducible.
To do that:
  - install nix (may get tricky on macOS Catalina 😅 )
  - enter nix shell (`nix-shell`)
  - run `sbtix-gen-all2` command

build.sbt
lemastero: Moved to Dependencies.

kapke (approved):
LGTM!
Since this touches metrics as well, I think @mmrozek should take a look too.

mmrozek:
@lemastero you have to sign all commits

lemastero:
Could not sign old commits. I have created different PR: #692

PR #691, opened 2020-09-23
[ETCM-118] Remove `OldRegularSync` from codebase

ajuszczak:
## Description

This pull requests removes implementation for `OldRegularSync` since we are not going to use it anymore and it might be misleading while talking about implementation of `RegularSync`.

## Important Changes Introduced

use-new-regular-sync has been removed from the configuration file, since only `RegularSync` is supported.


ntallar:
I'll take ownership of this PR as Adrian no longer works on Mantis

mmrozek (approved):
lgtm

PR #693, opened 2020-09-23
[ETCM-125] Fix faucet logging issue

mmrozek:
# Description

Add missing logging configuration to the faucet. 
Fix faucet dir path


serhii-shymkiv-iohk (approved):
LGTM

kapke (approved):
I don't really like that part of configuration is not scoped anymore. I get the reason though so I won't block this PR.

PR #694, opened 2020-09-23
[ETCM-44] Treasury block reward distribution

ntallar:
# Description

Adapts block reward distribution to the treasury

# Proposed Solution

Follows the specification described in [this](https://input-output.atlassian.net/wiki/spaces/EMU/pages/1681293384/Proto-Treasury#Block-execution) section:
- _If treasury contract wasn't deployed or was selfdestroyed_: follow ETC rules, 100% reward for block to miner
- _If treasury contract was deployed but the miner opted-out_: 80% reward for block to miner
- _If treasury contract was deployed and miner opted-in_: 80% reward for block to miner, 20% reward for treasury

# Testing

## Testing on mainnet

- Mantis should still remain compatible with ETC mainnet and mordor

## Testing on private network

1. Start up a node
  - Gas limit on mainnet genesis isn't enough for the txs being sent below, a value of 0x1388000000 worked for me
  - It requires an account with funds so you should create one and add it to the genesis file
  - Requires configuration of private network: mocked mining on, fast sync off, pruning off, etc
2. Deploy a contract with a function that calls selfdestroy, simpler one that can be used is:
```
pragma solidity ^0.5.0;

contract MockTreasury {
    function shutdown() public {
        // All funds from the contract are burned
        selfdestruct(address(0));
    }
}
```
3. Restart the node after:
 - Configuring the deployed contract address as the treasury address
 - Setting the ECIP1098 block number activation to a more feasible one
4. Mine enough blocks to surpass the ECIP1098 activation block
 - Block reward distribution should behave as spec'ed, both before and after the activation of ECIP1098
5. Call function to selfdestroy the contract, mine further blocks and check that block reward distribution behaves as before ECIP1098 activation

# Pending

- [x] Test on mainnet
- [x] Add useful endpoints examples to insomnia


src/test/scala/io/iohk/ethereum/ledger/LedgerTestingUtils.scala
```diff
@@ -0,0 +1,9 @@
+package io.iohk.ethereum.ledger
+
+object LedgerTestingUtils {
+  def calculateFullMinerReward(blockRewardCalculator: BlockRewardCalculator, blockNumber: BigInt, numberOfOmmers: Int): BigInt = {
```
mirkoAlic, line 4: why not something like:
```
package io.iohk.ethereum.ledger

object BlockRewardCalculatorOps {

  implicit class BlockRewardCalculatorOps(calculator: BlockRewardCalculator) {

    def calculateFullMinerReward(blockNumber: BigInt, numberOfOmmers: Int): BigInt = {
      val rewardForBlock = calculator.calculateMiningRewardForBlock(blockNumber)
      val rewardForOmmers = calculator.calculateMiningRewardForOmmers(blockNumber, numberOfOmmers)
      rewardForBlock + rewardForOmmers
    }
  }
}
```
Then, in tests:
```
import io.iohk.ethereum.ledger.BlockRewardCalculatorOps._
```

disclaimer: Maybe some adjustments are needed in order to compile.

insomnia_workspace.json
```diff
@@ -1,8 +1,8 @@
 {
   "_type": "export",
   "__export_format": 4,
-  "__export_date": "2020-09-21T12:39:47.334Z",
-  "__export_source": "insomnia.desktop.app:v2020.3.3",
+  "__export_date": "2020-09-24T17:48:32.016Z",
+  "__export_source": "insomnia.desktop.app:v2020.2.1",
```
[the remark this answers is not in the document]
ntallar: Only because it seemed I had an older version installed, I've updated and re-export the file!

src/test/scala/io/iohk/ethereum/ledger/LedgerTestingUtils.scala
ntallar: Interesting fact, same name for implicit class and object wrapping it doesn't play well together

mirkoAlic (approved):
:rocket:

src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
```diff
@@ -45,26 +46,47 @@ class BlockPreparator(
    * @return
    */
   private[ledger] def payBlockReward(block: Block, worldStateProxy: InMemoryWorldStateProxy): InMemoryWorldStateProxy = {
```
KonradStaniec, line 48: Just minor comment: I would update scaladoc, to reflect all the changes and maybe mention treasury spec.

ntallar: Just did! I'm not sure whether to mention the spec or not as it's not public yet. I think the description I added should be a good enough description without having to refer to it

KonradStaniec (approved):

PR #695, opened 2020-09-24
[ETCM-70-71] Checkpointing domain + Checkpointing configuration

pslaski:
# Description

Ported checkpoint domain and configuration

- checkpoint class
- configuration fields for checkpointing
- update block header RLP decoder / encoder according to https://input-output.atlassian.net/wiki/spaces/EMU/pages/1681686614/Checkpointing#RLP-encoding-(affects-network-messages-with-headers) 

# Testing
- sync with mainnet with a fresh database

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -242,6 +242,7 @@ mantis {
     # If false then that 20% gets burned
     # Doesn't have any effect is ecip1098 is not yet activated
     treasury-opt-out = false
+
   }
 
   # This is the section dedicated to Ethash mining.
--- a/src/main/scala/io/iohk/ethereum/consensus/ConsensusConfig.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/ConsensusConfig.scala
@@ -1,12 +1,11 @@
 package io.iohk.ethereum.consensus
 
 import akka.util.ByteString
-import com.typesafe.config.{Config ⇒ TypesafeConfig}
+import com.typesafe.config.{Config => TypesafeConfig}
 import io.iohk.ethereum.consensus.validators.BlockHeaderValidator
 import io.iohk.ethereum.domain.Address
 import io.iohk.ethereum.nodebuilder.ShutdownHookBuilder
 import io.iohk.ethereum.utils.Logger
-
 
 /**
  * Provides generic consensus configuration. Each consensus protocol implementation
--- a/src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
+++ b/src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
@@ -3,10 +3,11 @@ package io.iohk.ethereum.db.storage
 import java.nio.ByteBuffer
 
 import akka.util.ByteString
-import boopickle.Default.{ Pickle, Unpickle }
+import boopickle.Default.{Pickle, Unpickle}
+import io.iohk.ethereum.crypto.ECDSASignature
 import io.iohk.ethereum.db.dataSource.DataSource
 import io.iohk.ethereum.db.storage.BlockHeadersStorage.BlockHeaderHash
-import io.iohk.ethereum.domain.BlockHeader
+import io.iohk.ethereum.domain.{BlockHeader, Checkpoint}
 import io.iohk.ethereum.utils.ByteUtils.compactPickledBytes
 
 /**
@@ -31,53 +32,11 @@ class BlockHeadersStorage(val dataSource: DataSource) extends TransactionalKeyVa
 
 object BlockHeadersStorage {
   type BlockHeaderHash = ByteString
-  /** The following types are [[io.iohk.ethereum.domain.BlockHeader]] param types (in exact order).
-    *
-    * Mentioned params:
-    * parentHash, ommersHash, beneficiary, stateRoot, transactionsRoot, receiptsRoot, logsBloom,
-    * difficulty, number, gasLimit, gasUsed, unixTimestamp, extraData, mixHash, nonce.
-    */
-  type BlockHeaderBody = (
-    ByteString,
-    ByteString,
-    ByteString,
-    ByteString,
-    ByteString,
-    ByteString,
-    ByteString,
-    BigInt,
-    BigInt,
-    BigInt,
-    BigInt,
-    Long,
-    ByteString,
-    ByteString,
-    ByteString,
-    Option[Boolean]
-  )
 
   import boopickle.DefaultBasic._
 
   implicit val byteStringPickler: Pickler[ByteString] = transformPickler[ByteString, Array[Byte]](ByteString(_))(_.toArray[Byte])
-  implicit val blockHeaderPickler: Pickler[BlockHeader] = transformPickler[BlockHeader, BlockHeaderBody]
-  { case (ph, oh, b, sr, txr, rr, lb, d, no, gl, gu, ut, ed, mh, n, oo) =>
-    new BlockHeader(ph, oh, b, sr, txr, rr, lb, d, no, gl, gu, ut, ed, mh, n, oo)
-  }{ blockHeader => (
-    blockHeader.parentHash,
-    blockHeader.ommersHash,
-    blockHeader.beneficiary,
-    blockHeader.stateRoot,
-    blockHeader.transactionsRoot,
-    blockHeader.receiptsRoot,
-    blockHeader.logsBloom,
-    blockHeader.difficulty,
-    blockHeader.number,
-    blockHeader.gasLimit,
-    blockHeader.gasUsed,
-    blockHeader.unixTimestamp,
-    blockHeader.extraData,
-    blockHeader.mixHash,
-    blockHeader.nonce,
-    blockHeader.treasuryOptOut
-  )}
+  implicit val ecdsaSignaturePickler: Pickler[ECDSASignature] = generatePickler[ECDSASignature]
+  implicit val checkpointPickler: Pickler[Checkpoint] = generatePickler[Checkpoint]
+  implicit val blockHeaderPickler: Pickler[BlockHeader] = generatePickler[BlockHeader]
 }
--- a/src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
@@ -2,9 +2,7 @@ package io.iohk.ethereum.domain
 
 import akka.util.ByteString
 import io.iohk.ethereum.crypto.kec256
-import io.iohk.ethereum.rlp.RLPImplicitConversions._
-import io.iohk.ethereum.rlp.RLPImplicits._
-import io.iohk.ethereum.rlp.{RLPEncodeable, RLPList, RLPSerializable, rawDecode, encode => rlpEncode}
+import io.iohk.ethereum.rlp.{RLPDecoder, RLPEncodeable, RLPEncoder, RLPList, RLPSerializable, rawDecode, encode => rlpEncode}
 import org.bouncycastle.util.encoders.Hex
 
 case class BlockHeader(
@@ -23,7 +21,8 @@ case class BlockHeader(
     extraData: ByteString,
     mixHash: ByteString,
     nonce: ByteString,
-    treasuryOptOut: Option[Boolean]) {
+    treasuryOptOut: Option[Boolean],
+    checkpoint: Option[Checkpoint] = None) {
 
   override def toString: String = {
     s"""BlockHeader {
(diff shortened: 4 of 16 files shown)


src/main/resources/application.conf
```diff
@@ -242,6 +242,9 @@ mantis {
     # If false then that 20% gets burned
     # Doesn't have any effect is ecip1098 is not yet activated
     treasury-opt-out = false
+
+    # List of hex encoded public keys of Checkpoint Authorities
+    checkpoint-public-keys = []
```
ntallar, line 247: Shouldn't this be chain specific and go on the chains/ files?

src/main/scala/io/iohk/ethereum/consensus/ConsensusConfig.scala
```diff
@@ -72,14 +75,23 @@ object ConsensusConfig extends Logger {
     val blockCacheSize = config.getInt(Keys.BlockCacheSize)
     val miningEnabled = config.getBoolean(Keys.MiningEnabled)
     val optOut = config.getBoolean(Keys.TreasuryOptOut)
+    val checkpointPubKeys = readCheckpointPubKeys(config)
+    val minRequireSignatures = (Math.floor(checkpointPubKeys.length / 2) + 1).toInt
```
ntallar, line 79: As this is dependent on the checkpointPubKeys should we maybe define it as a function on the config object?

src/main/scala/io/iohk/ethereum/db/storage/BlockHeadersStorage.scala
```diff
@@ -31,53 +32,11 @@ class BlockHeadersStorage(val dataSource: DataSource) extends TransactionalKeyVa
...
-    blockHeader.gasUsed,
-    blockHeader.unixTimestamp,
-    blockHeader.extraData,
-    blockHeader.mixHash,
-    blockHeader.nonce,
-    blockHeader.treasuryOptOut
-  )}
+  implicit val ecdsaSignaturePickler: Pickler[ECDSASignature] = generatePickler[ECDSASignature]
+  implicit val checkpointPickler: Pickler[Checkpoint] = generatePickler[Checkpoint]
+  implicit val blockHeaderPickler: Pickler[BlockHeader] = generatePickler[BlockHeader]
```
ntallar, line 41: :heart:

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
```diff
@@ -102,24 +128,42 @@ object BlockHeader {
   }
 
   implicit class BlockheaderEncodableDec(val rlpEncodeable: RLPEncodeable) extends AnyVal {
+    private def encodeOptOut(encodedOptOut: RLPEncodeable): Option[Boolean] = {
```
ntallar, line 131: :+1: 

Isn't this decodeOptOut?

src/main/scala/io/iohk/ethereum/domain/Checkpoint.scala
```diff
@@ -0,0 +1,31 @@
...
+      Checkpoint(
+        signatures.items.map(ecdsaSignatureDec.decode)
+      )
+    case _ => throw new RuntimeException("Cannot decode Checkpoint")
+  }
+
+  implicit class CheckpointEnc(checkpoint: Checkpoint) extends RLPSerializable {
+    override def toRLPEncodable: RLPEncodeable = {
+      RLPList(
+        RLPList(checkpoint.signatures.map(_.toRLPEncodable): _*)
```
ntallar, line 25: Isn't here an extra RLPList? Why the 2 RLPLists? An RLPList will already be added by the optionality of the checkpoint

src/main/scala/io/iohk/ethereum/rlp/RLPImplicits.scala
```diff
@@ -107,4 +107,15 @@ object RLPImplicits {
       }
   }
 
+  implicit def optionEnc[T](implicit enc: RLPEncoder[T]): RLPEncoder[Option[T]] = {
+    case None => RLPList()
+    case Some(value) => RLPList(enc.encode(value))
+  }
+
+  implicit def optionDec[T](implicit dec: RLPDecoder[T]): RLPDecoder[Option[T]] = {
```
ntallar, line 115: Maybe we should start using this for the optOut field eventually :thinking:

ntallar (changes requested):

src/main/resources/application.conf
rtkaczyk: Do you mean the genesis file? While that sounds like a good idea, we've been talking about modifying the set of public key in order exclude a potential rogue node. Until we have a clear approach I would leave as it is.

In general I would avoid any deviations from our current reference implementation, until we have concrete improvement proposals.

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
rtkaczyk: Yes, `decode`. How is this related to checkpointing? Is that it was simply missed on a previous PR?

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
```diff
@@ -102,24 +128,42 @@ object BlockHeader {
   }
 
   implicit class BlockheaderEncodableDec(val rlpEncodeable: RLPEncodeable) extends AnyVal {
```
rtkaczyk, line 130: Not directly related to this PR, but I'm not fond of this naming: `BlockheaderEncodableDec` is by types symmetrical to `BlockHeaderEnc` (`RLPEncodable => BlockHeader` vs `BlockHeader => RLPEncodable`), so it should be named `BlockHeaderDec`.

What is currently  `BlockheaderDec` should be renamed to `BlockHeaderByteArrayDec`, especially that it's only used in tests.

Note: `Blockheader*` -> `BlockHeader*`

src/main/resources/application.conf
pslaski: he means chain specific configuration files (resources/chains/etc.conf)

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
pslaski: this is encoder/decoder for a header with checkpoint field which is compatible with the old blocks.

src/main/scala/io/iohk/ethereum/domain/Checkpoint.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  implicit val checkpointRLPDecoder: RLPDecoder[Checkpoint] = {
+    case RLPList(signatures: RLPList) =>
+      Checkpoint(
+        signatures.items.map(ecdsaSignatureDec.decode)
+      )
+    case _ => throw new RuntimeException("Cannot decode Checkpoint")
+  }
+
+  implicit class CheckpointEnc(checkpoint: Checkpoint) extends RLPSerializable {
```
rtkaczyk, line 22: Do we need this class? It would be more concise to express it as a SAM type (same as the decoder).

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
```diff
@@ -54,43 +54,69 @@ case class BlockHeader(
...
           RLPList(parentHash, ommersHash, beneficiary, stateRoot, transactionsRoot, receiptsRoot,
-            logsBloom, difficulty, number, gasLimit, gasUsed, unixTimestamp, extraData, mixHash, nonce, RLPList(encodedOptOut))
+            logsBloom, difficulty, number, gasLimit, gasUsed, unixTimestamp, extraData, mixHash, nonce, encodeOptOut(definedOptOut))
 
-        case None =>
-          // Pre ECIP1098 block, encoding works as if optOut field wasn't defined for backwards compatibility
+        case (None, Some(_)) =>
+          // Post ECIP1097 block with checkpoint, treasury disabled, block with checkpoint is encoded
+          RLPList(parentHash, ommersHash, beneficiary, stateRoot, transactionsRoot, receiptsRoot,
+            logsBloom, difficulty, number, gasLimit, gasUsed, unixTimestamp, extraData, mixHash, nonce, RLPList(), checkpoint)
```
rtkaczyk, line 116: It's a bit awkward.
What if we had a single optional field interpreted as "block header extensions"? It would be always present after ECIP-X. Then we could explicitly state optionality of the new extensions in all cases, and it would be safe for any future extensions as well.

src/main/resources/application.conf
rtkaczyk: Didn't realise that existed. I like that 👍

src/main/scala/io/iohk/ethereum/domain/Checkpoint.scala
pslaski: IMHO it's ok.
remember that this is an encoder for Checkpoint class.
the first RLPList is for signatures field
the second RLPList is for Checkpoint class
I know that option encoder also add RLPList, but it's for Option wrapper.

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
pslaski: do you mean one additional field for BlockHeader where we will pack all extensions ? I'm ok with that, but as it will touch also treasury opt, IMHO it should be done in different PR as it will add a lot of noise to the PR.

src/main/scala/io/iohk/ethereum/rlp/RLPImplicits.scala
pslaski: I think we can't as we are using custom Boolean encoding here.

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
rtkaczyk: Yes, definitely. This would be an important breaking change, so I'd like to consult that with others and plan it as proper a task. 

I'm also not sure if this is the best we can do. @ntallar WDYT?

pslaski: IMHO it would be also good to do something similar in BlockchainConfig

src/main/scala/io/iohk/ethereum/domain/Checkpoint.scala
ntallar: I see what you are saying but it doesn't seem to be the way other encodings are done as well, BlockHeaders and NewBlockHashes ignore the RLPList "required" by the class the are wrapped on and are just an rlplist of headers/hashes

It also depends on how you think of this:
1. as a helper that's just an analogous of having the header field as `checkpoint: Option[Seq[Signature]] `
2. as an integral part of the header structure. This would tend to be as you implemented, but it's not followed on the examples I mentioned

Given all that I'm more inclined to remove one of the RLPList

src/main/scala/io/iohk/ethereum/rlp/RLPImplicits.scala
ntallar: Is it custom? I copied it from how it's done on the reverse field of GetBlockHeaders

pslaski: we don't have any default one, but IMHO default Boolean could be true/false not 1/0 ;)

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
ntallar: Wouldn't a block header extensions field only move this complexity there, with the same matching on it's size as here? Why would that simplify any future extensions?

The only thing I think would simplify quite a lot the enconding/decoding design is having a block version field as with bitcoin (we could add it as the optional 17th field). Then:
- Version 1: no version field, traditional ETC block
- Version 2: treasuryOptOut as 18th field + checkpoint as 19th field
- ...

The spec didn't include that as I thought it might be an overkill, I haven't heard of future block headers changes for now

(I would also add any changes resulting from this discussion in a separate PR)

rtkaczyk: > Why would that simplify any future extensions?

Not necessarily simplify, but we would always have explicitness regarding optionality. Anyway, I do like the version field idea better. I think this is the way to go 💯

ntallar (approved):
Apart from the discussion on versioning (for a separate PR I think), LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

rtkaczyk (approved):

PR #701, opened 2020-09-28
[CHORE] Update README - add info about generic launcher script

mmrozek:
# Description

Add missing info about the generic launcher script


KonradStaniec (approved):

jmendiola222 (approved):
Even if `tesnet-internal` is temporarily I think we can update that later.

kapke (approved):

PR #702, opened 2020-09-28
[Fix] Fix handling of empty recipts int fast sync

KonradStaniec:
# Description

Two fixes:
- FastSync, while receiving empty receipts `reduce` thrown exception - ` new UnsupportedOperationException("empty.reduceLeft")`  which killed sync. In general it means that our error hanlding in actor is less than stellar, but for now just handle empty case.
- Sometimes our tests timeout in `EthashUtilsSpec`, try to fix it by rewriting recursive function in tail recursive manner and by changing number generation logic


src/main/scala/io/iohk/ethereum/blockchain/sync/FastSync.scala
```diff
@@ -360,35 +360,36 @@ class FastSync(
...
-            val reason = s"got empty receipts for known hashes: ${requestedHashes.map(h => Hex.toHexString(h.toArray[Byte]))}"
-            blacklist(peer.id, blacklistDuration, reason)
-          }
-
-          val remainingReceipts = requestedHashes.drop(receipts.size)
-          if (remainingReceipts.nonEmpty) {
-            syncState = syncState.enqueueReceipts(remainingReceipts)
-          }
+      if (receipts.isEmpty) {
+        val reason = s"got empty receipts for known hashes: ${requestedHashes.map(h => Hex.toHexString(h.toArray[Byte]))}"
```
mmrozek, line 364: Very minor: we have helper method `ByteStringUtils.hash2String`

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
```diff
@@ -145,6 +145,68 @@ class SyncControllerSpec extends AnyFlatSpec with Matchers with BeforeAndAfter w
...
+    sendBlockHeaders(firstNewBlock, newBlocks, peer1, newBlocks.size)
+
+    Thread.sleep(syncConfig.fastSyncThrottle.toMillis)
+    sendNewTargetBlock(defaultTargetBlockHeader.copy(number = defaultTargetBlockHeader.number + 1), peer1, peer1Status, handshakedPeers)
+
+    Thread.sleep(1.second.toMillis)
+    sendReceipts(newBlocks.map(_.hash), Seq(), peer1)
+
+    // Peer will be blacklisted for empty response, so wait he is blacklisted
+    Thread.sleep(6.second.toMillis)
```
mmrozek, line 193: Why 6 seconds? Maybe we could tweak some test params? This test will take a lot of time

src/test/scala/io/iohk/ethereum/consensus/ethash/EthashUtilsSpec.scala
```diff
@@ -2,18 +2,20 @@ package io.iohk.ethereum.consensus.ethash
 
 import akka.util.ByteString
 import io.iohk.ethereum.crypto.kec256
-import org.scalacheck.Arbitrary
+import org.scalacheck.{Gen}
```
mmrozek, line 5: Very minor: strange format of import

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
kapke: Or use mock clock maybe?

KonradStaniec: I reduced sleeps in case of blacklist (and blacklist config). 

I agree @kapke that mock clock would be perfect, but at current state sync state is almost black box and there is so much timing/scheduling  happening inside that those whole testsuit would require little refactor and i would prefer to create separate task to do it

mmrozek: Good idea. Maybe, for now, we could tag time-consuming tests like that and skip it by default for development purposes (and run on CI only). Now the whole test suite takes more than 30m. WDYT?

mmrozek (approved):
LGTM. Minor comment only

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
KonradStaniec: I created task for both of this possible improvements - https://jira.iohk.io/browse/ETCM-147.

But on my machine whole testsuite takes 6min so it is not that much, I am not sure why tests on CI takes that long.

pslaski (approved):

PR #707, opened 2020-09-29
Migrate CircleCI to Buildkite

sevanspowell:
# Description

- Use Nixkite to define a Buildkite pipeline in Nix.
- Modify the "test-ets.sh" script to reflect the new Buildkite
  environment.
- Remove old CircleCI config.
- Update the solidity compiler to version 0.5.1 because that's the
  version currently available in the Nix environment.


sevanspowell:
Note that the CircleCI build check will fail because I've removed all the CircleCI config. I'll disable that check when everyone is happy with this PR.

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,68 @@
+# steps:
```
mirkoAlic, line 1: comments can be remove now? or the idea is to keep them for some reason? (minor)

.circleci/config.yml
```diff
@@ -1,100 +0,0 @@
...
-      - image: rtkaczyk/mantis-circleci:v6
-    steps:
-      - checkout
-
-      - restore_cache:
-          key: mantis-{{ checksum "build.sbt" }}-{{ checksum "project/build.properties" }}-{{ checksum "project/plugins.sbt" }}-v1
-
-      - run:
-          name: FULL ETS
-          command: RUN_FULL_ETS=yes ./test-ets.sh
```
mirkoAlic, line 78: Not for the scope of this task, but we should think how we could add this into our CI flow. Current main issue is that it take to much time.

mirkoAlic (approved):
LGTM!

.buildkite/pipeline.yml
sevanspowell: Sorry @mirkoAlic, they should be removed, I'll do that now :)

PR #709, opened 2020-09-29
[ETCM-143] Redefine block header structure and encoding

ntallar:
# Description

Improves both the block header structure and the way it's encoded

Changes are strongly based on the assumption that ECIP1098 will be activated before ECIP1097. However, if that assumption is not true, it can be easily re-adjusted, for that reason I'd consider an overkill providing a full mechanism to support both orders. [1]

## Block header encoding

Block header encoding is done as if we had 3 different types of headers:
- **PreECIP1098 (treasury and checkpointing disabled)**: the block is encoded as currently on the mainnet
- **PostECIP1098 (treasury enabled but checkpointing disabled)**: the block is encoded as if it's fields were _all the current fields from ETC_ :+ (treasuryOptOut: Boolean)
- **PostECIP1097 (treasury and checkpointing enabled)**: the block is encoded as if it's fields were _all the current fields from ETC_ :+ (treasuryOptOut: Boolean) :+ (checkpoint: Option[Checkpoint])

In this way the type of the block can be determined exclusively by the number of fields the block has. Note that this (with the block number validations) is identical to determining the block header type through the block number

This mechanism is based on the fact that each ECIP activated added new fields to the block header, if future ECIPs want to also remove fields they should follow other mechanisms [1]

## Block header structure

Instead of having 3 different block headers structures, as mentioned on the encoding section, this is analogously done as an _extra fields_ field on the block header, with one possible value for each of the 3 header structures

# Testing

- Syncing to ETC mainnet and mordor testnet should continue working
- Treasury feature should continue working as expected

# Pending

- [x] Manual tests on ETC compatibility
- [x] Manual tests on treasury working as expected

# Notes

[1] Two possible mechanism that would allow ECIP1097 to be activated before ECIP1098 as well as removing fields:
1. Include an only encoding field of blockType, depending on which we know the type of the block and so appropriately select it's decoder
2. Have the decoder depend on the ECIPs configuration, and so determine the block type by the block number _(this is probably preferred as it doesn't involve further protocol changes)_


src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
```diff
@@ -52,71 +79,166 @@ case class BlockHeader(
...
+    extraData = extraData,
+    mixHash = mixHash,
+    nonce = nonce,
+    extraFields = HefPostEcip1097(treasuryOptOut, checkpoint)
+  )
+  // scalastyle:on parameter.number
+
+  sealed trait HeaderExtraFields
+  object HeaderExtraFields {
+    case object HefPreEcip1098 extends HeaderExtraFields
```
rtkaczyk, line 218: I think a better name would be `HefEmpty` - it would immediately stand out from the others and be associated with the default case, whereas the current name requires some attention to be distinguished from others.

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
```diff
@@ -52,71 +79,166 @@ case class BlockHeader(
...
+      case HefPreEcip1098 =>
+        // Pre ECIP1098 and ECIP1097 block, encoding works as if optOut and checkpoint fields weren't defined for backwards compatibility
+        rlpList.items.dropRight(2)
     }
-    rlpEncode(rlpEncoded)
+    rlpEncode(RLPList(rlpItemsWithoutNonce: _*))
   }
 
+  // scalastyle:off parameter.number
+  def buildPreECIP1098Header(parentHash: ByteString,
```
rtkaczyk, line 113: Is it worth adding all these methods? Maybe we could just have a default param `extraFields: HeaderExtraFields = HefEmpty` on the default constructor? And as for the extra fields, I don't think wrapping them in the HEF-constructor is that tedious to necessitate these methods

ntallar: I found them useful when changing the block header structure as that required less changes. Given that and that we might eventually further change it (in case we want 3 different header structures) I saw no reason to delete them once I had them

But I don't have too much of a strong opinions right now, if you have stronger opinions on deleting them I can do so

src/test/scala/io/iohk/ethereum/ObjectGenerators.scala
```diff
@@ -139,6 +141,16 @@ trait ObjectGenerators {
     td <- bigIntGen
   } yield NewBlock(Block(blockHeader, BlockBody(stxs, uncles)), td)
 
+  def extraFieldsGen: Gen[HeaderExtraFields] = for {
```
KonradStaniec, line 144: I know we have property tests for it, but maybe it is wortth while to create some specific encoding cases like in: https://github.com/input-output-hk/mantis/blob/develop/src/test/scala/io/iohk/ethereum/network/p2p/messages/ReceiptsSpec.scala#L38, it will give someone inspecting it in the future idea how encoded payload looks like.

src/main/scala/io/iohk/ethereum/consensus/validators/BlockHeaderValidatorSkeleton.scala
```diff
@@ -77,7 +78,7 @@ abstract class BlockHeaderValidatorSkeleton(blockchainConfig: BlockchainConfig)
       _ <- validateGasUsed(blockHeader)
       _ <- validateGasLimit(blockHeader, parentHeader)
       _ <- validateNumber(blockHeader, parentHeader)
-      _ <- validateOptOut(blockHeader)
+      _ <- validateExtraFields(blockHeader)
```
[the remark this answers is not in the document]
ntallar: For attack purposes or for facilitating our usage of it?

I'm not sure if the order of the validations should matter that much here as they all are quite simple and easy to check (except from the PoW validation that probably takes more). So I'm not convinced if it would be better for them to have a particular order

KonradStaniec (approved):

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
rtkaczyk: Well, I think that using a default arg for `extraFields` would result in even fewer changes. Any future changes should also require less maintenance. Do you think there is another advantage to having these factories?

ntallar: Tbh I don't like much default arg in general (exceptuating some use cases) as I see them too error prone... but given the subjectivity of this and that the PR has been open for a long time I'll change this

rtkaczyk (approved):

kapke (approved):

PR #716, opened 2020-10-01
[ETCM-75] Support for checkpoint blocks in Blockchain

pslaski:
# Description

added saving last checkpoint number during block save and rollback

- saving last checkpoint number to the memory and persisting it with best block number after cache persist during block save
- persisting last checkpoint number and best block number immediately during block remove

# Testing

unit tests

### DIFF
--- a/src/main/scala/io/iohk/ethereum/db/storage/StateStorage.scala
+++ b/src/main/scala/io/iohk/ethereum/db/storage/StateStorage.scala
@@ -18,8 +18,8 @@ trait StateStorage {
   def getBackingStorage(bn: BigInt): MptStorage
   def getReadOnlyStorage: MptStorage
 
-  def onBlockSave(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlock: Option[BigInt] => Unit): Unit
-  def onBlockRollback(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlock: Option[BigInt] => Unit): Unit
+  def onBlockSave(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlocksData: () => Unit): Unit
+  def onBlockRollback(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlocksData: () => Unit): Unit
 
   def saveNode(nodeHash: NodeHash, nodeEncoded: NodeEncoded, bn: BigInt)
   def getNode(nodeHash: NodeHash): Option[MptNode]
@@ -34,15 +34,15 @@ class ArchiveStateStorage(private val nodeStorage: NodeStorage,
     true
   }
 
-  override def onBlockSave(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlock: Option[BigInt] => Unit): Unit = {
+  override def onBlockSave(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlocksData: () => Unit): Unit = {
     if (cachedNodeStorage.persist()) {
-      updateBestBlock(None)
+      updateBestBlocksData()
     }
   }
 
-  override def onBlockRollback(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlock: Option[BigInt] => Unit): Unit = {
+  override def onBlockRollback(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlocksData: () => Unit): Unit = {
     if (cachedNodeStorage.persist()) {
-      updateBestBlock(None)
+      updateBestBlocksData()
     }
   }
 
@@ -71,21 +71,21 @@ class ReferenceCountedStateStorage(private val nodeStorage: NodeStorage,
     true
   }
 
-  override def onBlockSave(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlock: Option[BigInt] => Unit): Unit = {
+  override def onBlockSave(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlocksData: () => Unit): Unit = {
     val blockToPrune = bn - pruningHistory
 
     ReferenceCountNodeStorage.prune(blockToPrune, cachedNodeStorage, inMemory = blockToPrune > currentBestSavedBlock)
 
     if (cachedNodeStorage.persist()) {
-      updateBestBlock(None)
+      updateBestBlocksData()
     }
   }
 
-  override def onBlockRollback(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlock: Option[BigInt] => Unit): Unit = {
+  override def onBlockRollback(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlocksData: () => Unit): Unit = {
     ReferenceCountNodeStorage.rollback(bn, cachedNodeStorage, inMemory = bn > currentBestSavedBlock)
 
     if (cachedNodeStorage.persist()) {
-      updateBestBlock(None)
+      updateBestBlocksData()
     }
   }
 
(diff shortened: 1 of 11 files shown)


src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -225,12 +230,20 @@ class BlockchainImpl(
...
-      bestKnownBlock.get()
+    if (bestKnownBlockAndLatestCheckpoint.get().bestBlockNumber > bestBlockNum)
+      bestKnownBlockAndLatestCheckpoint.get().bestBlockNumber
     else
       bestBlockNum
   }
 
+  override def getLatestCheckpointBlockNumber(): BigInt = {
+    val latestCheckpointNumberInStorage = appStateStorage.getLatestCheckpointBlockNumber()
+    if (bestKnownBlockAndLatestCheckpoint.get().latestCheckpointNumber > latestCheckpointNumberInStorage)
```
rtkaczyk, line 241: Could explain in what situation these numbers can be different?

pslaski: I'm not sure every border case here, but as far as I understand the best block number is firstly saved in memory and then persisted to the storage only when it's time to persist cache. I want to save the best block number and latest checkpoint transactionally so  `getLatestCheckpointBlockNumber` behaves the same. We can have the latest checkpoint number in memory bigger than the number in storage because the cache wasn't persisted yet.

src/test/scala/io/iohk/ethereum/domain/BlockchainSpec.scala
```diff
@@ -11,6 +12,9 @@ import org.scalatest.matchers.should.Matchers
 
 class BlockchainSpec extends AnyFlatSpec with Matchers {
 
+  val checkpoint = ObjectGenerators.fakeCheckpointGen(2, 5).sample.get
+  //val keys = Seq(generateKeyPair(secureRandom))
```
rtkaczyk, line 16: Minor: remove

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
rtkaczyk: Ok, maybe add a comment about it?

rtkaczyk (approved):
Apart from minor issues LGTM!

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -290,7 +309,15 @@ class BlockchainImpl(
     evmCodeStorage.put(hash, evmCode)
 
   override def saveBestKnownBlock(number: BigInt): Unit = {
```
ntallar, line 311: In the BlockImport there are various places where this method is called, should we do the same change we did on save here and also change that call to (maybe better with an auxiliar function here):
```
    if (saveAsBestBlock && block.hasCheckpoint) {
      saveBestKnownBlockAndLatestCheckpointNumber(block.header.number, block.header.number)
    } else if (saveAsBestBlock) {
      saveBestKnownBlock(block.header.number)
    }
```

ntallar (changes requested):

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -323,20 +351,64 @@ class BlockchainImpl(
...
     }
 
+    val checkpointUpdates = maybeBlockHeader match {
+      case Some(header) =>
+        if (header.hasCheckpoint && header.number == getLatestCheckpointBlockNumber()) {
+          val prev = findPreviousCheckpointBlockNumber(header.number, header.number)
+          prev.map { num =>
+            // side effect
+            saveLatestCheckpointNumber(num)
+            appStateStorage.putLatestCheckpointBlockNumber(num)
```
KonradStaniec, line 361: I am not sure about this bit.
Lets say we  have saved `X` blocks with all the mpt nodes i.e best checkpoint and best known block is `X`. (for simplicity checkpoint and bestblock are the same)
Now we are on `X + 20` (without updating bestblock and checkpoint in db) , last checkpoint was on `X + 10`.  We start to rollback and and we find this checkpint an we save  into db. Now node dies.
After restart, best checkpoint in db  will be `X + 10` but best block will be `X` as this was last time we sved all data.
Imo if we must to update best checkpoint in db, we should also update best block.

src/main/scala/io/iohk/ethereum/db/storage/StateStorage.scala
```diff
@@ -120,19 +120,19 @@ class CachedReferenceCountedStateStorage(private val nodeStorage: NodeStorage,
...
-  override def onBlockSave(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlock: Option[BigInt] => Unit): Unit = {
+  override def onBlockSave(bn: BigInt, currentBestSavedBlock: BigInt)(updateBestBlocksData: => Unit): Unit = {
     val blockToPrune = bn - pruningHistory
     changeLog.persistChangeLog(bn)
     changeLog.getDeathRowFromStorage(blockToPrune).foreach {deathRow =>
       CachedReferenceCountedStorage.prune(deathRow, lruCache, blockToPrune)
     }
     if (CachedReferenceCountedStorage.persistCache(lruCache, nodeStorage)) {
-      updateBestBlock(None)
+      updateBestBlocksData
```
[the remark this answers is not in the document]
pslaski: it's a function argument, we can't call it `updateBestBlocksData()`

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -323,20 +358,64 @@ class BlockchainImpl(
...
       .and(totalDifficultyStorage.remove(blockHash))
       .and(receiptStorage.remove(blockHash))
       .and(maybeTxList.fold(transactionMappingStorage.emptyBatchUpdate)(removeTxsLocations))
       .and(blockNumberMappingUpdates)
+      .and(appStateStorage.putBestBlockNumber(newBestBlockNumber))
+      .and(checkpointUpdates)
       .commit()
 
     // not transactional part
+    saveBestKnownBlocks(newBestBlockNumber, prevCheckpointNumber)
```
ntallar, line 388: Should we also add this to our save function as well?

src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
```diff
@@ -197,8 +195,14 @@ class BlockImport(
       blockchain.save(block, receipts, td, saveAsBestBlock = false)
     }
 
+    val checkpointNumber = oldBranch
+      .filter(_.block.hasCheckpoint)
+      .sortBy(_.block.number)
+      .lastOption
```
ntallar, line 201: Minor: instead of sortBy and lastOption, should we maybe use max or maxBy?

ntallar (changes requested):

src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
pslaski: that was my first option, but maxBy/max are always returning value - in this case, we could have an empty list.

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
pslaski: in `save` is different logic which controls what blocks data will be saved, so it's better to use low-level methods there

src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
ntallar: Yeah, that's a pain of usage, but maybe we could have done `Try(oldBranch.collect { case b if b.block.hasCheckpoint => b.block.number }).max).toOption`?

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
ntallar: And much more error prone... I think we'll suffer from this eventually :confused:

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
pslaski: IMHO your proposition is less readable, but it's probably doing one less iteration (filter, sortBy, lastOption vs collect, max). WDYT about `.filter(_.block.hasCheckpoint).sortBy(-_.block.number).headOption` ?

kapke: What about https://typelevel.org/cats/api/cats/Foldable.html#maximumByOption[A,B](fa:F[A])(f:A=%3EB)(implicitevidence$3:cats.Order[B]):Option[A] ?

KonradStaniec (approved):

PR #722, opened 2020-10-05
[ETCM-129] prepare for Scala 2.13 replace ⇒ to =>

lemastero:
# Description

After switching to Scala 2.13 there are many errors. Fixing some is tricky. I want to start from the low-risk ones.
This PR fix errors caused by the usage of Unicode symbol ⇒

# Proposed Solution

replace ⇒ to =>

# Testing


mmrozek (approved):
LGTM!

PR #724, opened 2020-10-05
[ETCM-177] Fix Ommers pool

mirkoAlic:
# Description

Fix ommers calculation at pool level and remove addition of mined blocks that are not imported.

# Up to discussion

As a future improvement it would be nice to make the Ommers pool a more "independent" component that handle the removal  of ommers by its own.

- Query by parentBlockHash instead of blockNumber

# Testing

Unit tests should passed + we monitoring this at testnet level.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
@@ -45,8 +45,8 @@ class BlockImporter(
     start()
   }
 
-  private def idle: Receive = {
-    case Start => start()
+  private def idle: Receive = { case Start =>
+    start()
   }
 
   private def handleTopMessages(state: ImporterState, currentBehavior: Behavior): Receive = {
@@ -94,9 +94,10 @@ class BlockImporter(
   }
 
   private def pickBlocks(state: ImporterState): Unit = {
-    val msg = state.resolvingBranchFrom.fold[BlockFetcher.FetchMsg](
-      BlockFetcher.PickBlocks(syncConfig.blocksBatchSize))(
-      from => BlockFetcher.StrictPickBlocks(from, startingBlockNumber))
+    val msg =
+      state.resolvingBranchFrom.fold[BlockFetcher.FetchMsg](BlockFetcher.PickBlocks(syncConfig.blocksBatchSize))(from =>
+        BlockFetcher.StrictPickBlocks(from, startingBlockNumber)
+      )
 
     fetcher ! msg
   }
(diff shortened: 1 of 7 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -63,8 +63,6 @@ class BlockImporter(
     case MinedBlock(block) =>
       if (!state.importing) {
         importMinedBlock(block, state)
-      } else {
-        ommersPool ! AddOmmers(block.header)
```
mirkoAlic, line 67: this plus how ommers pool previously behave, it was the Testnet issue.

src/main/scala/io/iohk/ethereum/ommers/OmmersPool.scala
```diff
@@ -1,37 +1,76 @@
...
+      blockchain.getBlockHeaderByHash(hash) match {
+        case Some(bh) if (limit >= 0) => rec(bh.parentHash, limit - 1, acc ++ List(bh))
+        case _ => acc
+      }
+    }
+    rec(parentHash, generationLimit - 1, List.empty)
+  }
+
+  private def logStatus(event: String, ommers: Seq[BlockHeader]): Unit = {
+    val ommersAsString: Seq[String] = ommers.map { bh => s"[number = ${bh.number}, hash = ${bh.hashAsHexString}]" }
```
mmrozek, line 51: Minor: `lazy val`

src/test/scala/io/iohk/ethereum/ommers/OmmersPoolSpec.scala
```diff
@@ -7,55 +7,200 @@ import io.iohk.ethereum.Timeouts
...
+        block0,
+        block1Chain1,
+        block1Chain4,
+        block2Chain1
+      )
 
-    ommersPool ! AddOmmers(Block3125369.header)
-    ommersPool ! AddOmmers(Block3125369.header.copy(number = 2))
-    ommersPool ! RemoveOmmers(Block3125369.header)
+      ommersPool.!(GetOmmers(block3Chain1.parentHash))(testProbe.ref)
```
mmrozek, line 36: I prefer to use `ImplicitSender` instead of `testProbe` in cases like this

src/main/scala/io/iohk/ethereum/ommers/OmmersPool.scala
```diff
@@ -1,37 +1,76 @@
...
+      logStatus(event = s"Ommers given parent block ${Hex.toHexString(parentBlockHash.toArray)}", ommers)
       sender() ! OmmersPool.Ommers(ommers)
   }
+
+  private def collectAncestors(parentHash: ByteString, generationLimit: Int): List[BlockHeader] = {
+    assert(generationLimit > 0)
+    @tailrec
+    def rec(hash: ByteString, limit: Int, acc: List[BlockHeader]): List[BlockHeader] = {
+      blockchain.getBlockHeaderByHash(hash) match {
+        case Some(bh) if (limit >= 0) => rec(bh.parentHash, limit - 1, acc ++ List(bh))
```
mmrozek, line 43: 1. I think it will be a little bit more readable when we start with `rec(parentHash, generationLimit, List.empty)` and check if `limit > 0`
2. We don't need to wrap `List(bh)`. `acc  :+  bh` should be enough

mirkoAlic:
I merged given is needed for  experimental testnet purposes. cc @jmendiola222

PR #725, opened 2020-10-06
WIP [ETCM-129] remove com.twitter  util-collection

lemastero:
# Description

To migrate to Scala 2.13 it is neccessary to remove twiter util-collection library as it is not supported any more:

# Proposed Solution

Replace com.twitter.util.LruMap with java.util.LinkedHashMap. [Changelog for twitter util](https://twitter.github.io/util/guide/Changelog.html) says:
> We deleted LruMap because it was a very thin shim around a j.u.LinkedHashMap, where you override removeEldestEntry

Bonus side effect - mantis do not depends on:
* twitter/util-collection
* twitter/util-core
* twitter/util-function
* org.scala-lang/scala-parser-combinators

# Testing

### DIFF
--- a/src/main/scala/io/iohk/ethereum/faucet/FaucetApi.scala
+++ b/src/main/scala/io/iohk/ethereum/faucet/FaucetApi.scala
@@ -7,7 +7,7 @@ import akka.http.scaladsl.server.Directives._
 import akka.util.ByteString
 import ch.megard.akka.http.cors.scaladsl.CorsDirectives.cors
 import ch.megard.akka.http.cors.scaladsl.settings.CorsSettings
-import com.twitter.util.LruMap
+import java.util.{LinkedHashMap => LruMap}
 import io.iohk.ethereum.domain.{Address, Transaction}
 import io.iohk.ethereum.keystore.KeyStore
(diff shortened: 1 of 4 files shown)


src/main/scala/io/iohk/ethereum/faucet/FaucetApi.scala
```diff
@@ -40,7 +40,7 @@ class FaucetApi(rpcClient: RpcClient, keyStore: KeyStore, config: FaucetConfig,
 
   private def handleRequest(clientAddr: RemoteAddress, targetAddress: String) = {
     val timeMillis = clock.instant().toEpochMilli
-    val latestRequestTimestamp = latestRequestTimestamps.getOrElse(clientAddr, 0L)
+    val latestRequestTimestamp = latestRequestTimestamps.getOrDefault(clientAddr, 0L)
```
kapke, line 43: shouldn't cleaning be implemented now as mentioned in twitter util collection changelog? I see a memory leak here now since LRUMap's purpose is to clean least used entries when full.

lemastero:
Closing PR for now. It requires more work.

PR #726, opened 2020-10-07
[ETCM-126] add eth_pendingTransactions

lemastero:
# Description

Add support for Ethereum JSON-RPC endpoint [eth_pendingTransactions](https://github.com/etclabscore/ethereum-json-rpc-specification/blob/master/openrpc.json#L850-L859)

# Proposed Solution

# Testing


mmrozek:
I think it works the same as `qa_getPendingTransactions`. So probably we could remove the old method

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerSpec.scala
```diff
@@ -1901,6 +1902,71 @@ class JsonRpcControllerSpec
...
+      ),
+      Some(JInt(1))
+    )
+
+    val response: JsonRpcResponse = jsonRpcController.handleRequest(request).futureValue
+
+    response should haveResult(JArray(List()))
+  }
+
+  it should "mempool is empty" in new TestSetup {
```
mmrozek, line 1926: Mempool has transactions here

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerSpec.scala
```diff
@@ -1901,6 +1902,71 @@ class JsonRpcControllerSpec
     response should haveObjectResult("transactions" -> JArray(expectedTxs.toList))
   }
 
+  "request pending transactions and return valid response" should "mempool is empty" in new TestSetup {
```
mmrozek, line 1905: I think we should start to split this file. It is already too big. Maybe sth like `JsonRpcControllerHandlingTransactionsSpec`? WDYT?

src/main/scala/io/iohk/ethereum/jsonrpc/QAService.scala
```diff
@@ -1,32 +1,25 @@
 package io.iohk.ethereum.jsonrpc
 
-import akka.actor.ActorRef
+import akka.util.ByteString
```
mmrozek, line 3: Minor: double import

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerSpec.scala
lemastero: Yes, I highly agree. I try to do this in this PR but work it is a lot of changes and it shadows the real work done in here + somehow tests got broken. So it is better to go small steps.

I split JsonRpc tests in separate PR: https://github.com/input-output-hk/mantis/pull/734

src/main/scala/io/iohk/ethereum/jsonrpc/QAService.scala
lemastero: Good spot. Fixed.

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerSpec.scala
lemastero: Good catch, thank you it gave me some good hint on why tests failed after porting from qa scope to eth 👍

lemastero:
> I think it works the same as `qa_getPendingTransactions`. So probably we could remove the old method

Done.

insomnia_workspace.json
```diff
@@ -961,6 +961,43 @@
       "settingFollowRedirects": "global",
       "_type": "request"
     },
+    {
```
mmrozek, line 964: Please remove `qa_getPendingTransactions` from insomnia also

build.sbt
```diff
@@ -146,3 +146,13 @@ addCommandAlias(
     |;benchmark:compile
     |""".stripMargin
 )
+
+// prepare PR
```
mmrozek, line 150: 👍

mmrozek (approved):
LGTM!

insomnia_workspace.json
kapke: And remember to notify QA once this gets merged :)

kapke (approved):
LGTM!

PR #729, opened 2020-10-09
[ETCM-76] network messages for checkpointing

rtkaczyk:
# Description

Network messages including latest checkpoint number for best peer selection. Depends on #728


src/main/scala/io/iohk/ethereum/network/handshaker/EtcNodeStatusExchangeState.scala
```diff
@@ -52,10 +52,26 @@ case class EtcNodeStatusExchangeState(handshakerConfiguration: EtcHandshakerConf
...
     val bestBlockHeader = getBestBlockHeader()
     val totalDifficulty = blockchain.getTotalDifficultyByHash(bestBlockHeader.hash).get
+    val latestCheckpointNumber =
+      if (bestBlockHeader.number < blockchainConfig.ecip1097BlockNumber) None
+      else Some(blockchain.getLatestCheckpointBlockNumber())
+
+//    log.warn(
+//      s"""
+//        |
+//        |  KAPIBARA
```
pslaski, line 62: 💯

src/main/scala/io/iohk/ethereum/network/p2p/messages/CommonMessages.scala
```diff
@@ -12,45 +12,131 @@ import org.bouncycastle.util.encoders.Hex
...
+            ) =>
+          Status63(protocolVersion, networkId, totalDifficulty, bestHash, genesisHash)
+
+        case (
+              `code64`,
+              RLPList(
+                protocolVersion,
+                networkId,
+                totalDifficulty,
+                latestCheckpointNumber,
```
pslaski, line 56: would it be better to have `latestCheckpointNumber` as the last field on list, as it's something which might not exist?

src/test/scala/io/iohk/ethereum/network/p2p/messages/NewBlockSpec.scala
```diff
@@ -35,6 +35,8 @@ class NewBlockSpec extends AnyFunSuite with ScalaCheckPropertyChecks with Object
     assert(obtainEncoded2 == expectedEncoded2)
   }
 
+  ignore("Tests for NewBlock64?? - or are the ones in MessagesSerializationSpec sufficient") {}
```
pslaski, line 38: IMHO modifying newBlockGen to generate also `code64` messages should be enough

pslaski (changes requested):
just minor things

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -249,8 +250,17 @@ class BlockImporter(
...
-    val newBlocks = (blocks, totalDifficulties).mapN(NewBlock.apply)
+  private def broadcastBlocks(
+      blocks: List[Block],
+      totalDifficulties: List[BigInt]
+  ): Unit = {
+    val constructNewBlock =
+      if (blocks.lastOption.exists(_.number < blockchainConfig.ecip1097BlockNumber))
+        NewBlock63.apply _
+      else
+        NewBlock64.apply(_, _, blockchain.getLatestCheckpointBlockNumber())
```
ntallar, line 261: Shouldn't the lastCheckpoint value be dependent on the block being broadcasted? For example, if we are broadcasting C1->B2->C3 then B2 should have C1 as the last checkpoint block.

At least for now the td of every new block broadcasted was giving the weight of that broadcasted block, not of the peer, if so the lastCheckpoint field should behave in the same way

ntallar (changes requested):
I have other things to analyse but I'll do that tomorrow and add further comments!

ntallar:
I don't like much that whether we send a 63 or 64 version message depends on our block number... on eth when they added eth64 capability ([here](https://github.com/ethereum/go-ethereum/pull/20140/files) for example) they made several changes (as far as I understood at least):
- announce the new capability on handshake
- have the messages that a node sends to it's peers depend on the capability that they have, instead of our block number. If they announce etc64 then we send Status64 and if not Status63

WDYT? The main drawback I see of the current approach is that you'd never be leaving away the old messages, with what I'm proposing eventually a good chunck of the network will have moved to etc64 and you can deprecate eth63 and even mark those peers as bad ones

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
rtkaczyk: I thought that we're only broadcasting the top blocks. If that's not the case then yes, but this will be straight-forward after ETCM-263 using `blockchain.getChainWeightByHash`. Should I add a comment about fixing this in ETCM-263?

src/main/scala/io/iohk/ethereum/network/p2p/messages/CommonMessages.scala
rtkaczyk: Not sure, does it matter?

rtkaczyk:
> * announce the new capability on handshake

I like the idea, but it sounds like a big refactor. For one, where we currently construct the message we don't even who the recipient is. Separate task?

Also do we need this separate notion of _capability_? Couldn't we just take advantage of `protocolVersion`?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
ntallar: Hmm if we were to follow how it's handled for the TD, wouldn't the lastCheckpointBlockNumber be part of the `BlockData` that's the result of the block execution? (instead of using `blockchain.getChainWeightByHash` as you suggest)

Independently of what we decide, shouldn't we handle the TD and lastCheckpointBlockNumber in the same way?

>Should I add a comment about fixing this in ETCM-263?

But yeah, it seems to be quite related to that handling, so we could include this as well for that task

ntallar:
> Couldn't we just take advantage of protocolVersion?

Tbh I'm not sure what's the purpose of that protocolVersion... but from what I have seen, ETH/ETC updates of this sort are introduced as new capabilities, so we probably have no escape from that

>I like the idea, but it sounds like a big refactor. For one, where we currently construct the message we don't even who the recipient is. Separate task?

Why would it be that big? For now I'm seeing the expected changes to be:
- Modify our handshake so that the StatusExchangeState receives the capabilities of the peer in some way
- Modify our peers information to know if a peer supports etc64 capability or not

What would the benefit be of having this PR merged as is?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
rtkaczyk: >  wouldn't the lastCheckpointBlockNumber be part of the BlockData that's the result of the block execution?

Actually `BlockData` will contain `ChainWeight`. 

We will handle this the same way once `ChainWeight` is fully introduced, so I need this to continue working on ETCM-263, and we'll need to ETCM-263 to properly implement this improvement

rtkaczyk:
> Tbh I'm not sure what's the purpose of that protocolVersion... but from what I have seen, ETH/ETC updates of this sort are introduced as new capabilities, so we probably have no escape from that

Do know of an example in ETC?

> Why would it be that big? For now I'm seeing the expected changes to be:

Well, at least medium. We'll also need to change how blocks are broadcast. Currently `PeerInfo` is not available when constructing the message.

> What would the benefit be of having this PR merged as is?

It will unblock ETCM-263

ntallar:
>Do you know of an example in ETC?

There were no changes like this to ETC post-ETH fork, there are some examples before that, as eth62 and so forth (but I haven't taken a look at them)

>Currently PeerInfo is not available when constructing the message.

Couldn't we start building the message on BlockBroadcast? There we have the peer information (and I actually prefer it being built there, to decouple this network message from the syncing)

>It will unblock ETCM-263

Tbh I don't see this PR as close to the ready version as it was the case with the syncing one, and you are doing both so there won't be much parallelization if merged... I'd ideally do that only for the things that are overlapping between them, but I don't see that being the case for adapting this to how ETH/ETC is used to changing their network messages

src/main/scala/io/iohk/ethereum/network/p2p/messages/CommonMessages.scala
pslaski: not now, but IMHO as encoding to the RLP means encode to the list of things it should be easier to decode it having optional fields on the end than having them somewhere in the middle

src/main/scala/io/iohk/ethereum/network/p2p/messages/CommonMessages.scala
```diff
@@ -12,45 +12,131 @@ import org.bouncycastle.util.encoders.Hex
 
 object CommonMessages {
   object Status {
-    val code: Int = Versions.SubProtocolOffset + 0x00
+    // TODO: shouldn't we rather just use `protocolVersion` to distinguish these messages?
```
ntallar, line 15: With the latest discussions, should we remove this TODO?

src/test/scala/io/iohk/ethereum/blockchain/sync/PeersClientSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
+
+class PeersClientSpec extends AnyFlatSpec with Matchers with ScalaCheckPropertyChecks {
+  "PeerClient" should "determined the best peer based on it latest checkpoint number and total difficulty" in {
+    import Peers._
+
+    val table = Table[Map[Peer, PeerInfo], Option[Peer]](
+      ("PeerInfo map", "Expected best peer"),
```
ntallar, line 21: Minor, should we add some comment here on why each peer is selected in every situation?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -248,8 +249,22 @@ class BlockImporter(
...
-  private def broadcastBlocks(blocks: List[Block], totalDifficulties: List[BigInt]): Unit = {
-    val newBlocks = (blocks, totalDifficulties).mapN(NewBlock.apply)
+  private def broadcastBlocks(
+      blocks: List[Block],
+      totalDifficulties: List[BigInt]
+  ): Unit = {
+    val constructNewBlock = {
+      //FIXME: instead of choosing the message version based on block we should rely on the receiving
+      // peer's `Capability`. To be addressed in ETCM-280
+      if (blocks.lastOption.exists(_.number < blockchainConfig.ecip1097BlockNumber))
```
ntallar, line 259: For consistency, shouldn't this be <=?

src/main/scala/io/iohk/ethereum/network/handshaker/EtcNodeStatusExchangeState.scala
```diff
@@ -52,10 +52,15 @@ case class EtcNodeStatusExchangeState(handshakerConfiguration: EtcHandshakerConf
   private def createStatusMsg(): Status = {
     val bestBlockHeader = getBestBlockHeader()
     val totalDifficulty = blockchain.getTotalDifficultyByHash(bestBlockHeader.hash).get
+    val latestCheckpointNumber =
+      if (bestBlockHeader.number < blockchainConfig.ecip1097BlockNumber) None
```
ntallar, line 56: For consistency, shouldn't this be <=?

ntallar (changes requested):
Apart from minor comments, LGTM!

ntallar:
>But I'm not sure about this thinking I don't see this PR as close to the ready version as it was the case with the syncing one, and you are doing both so there won't be much parallelization if merged... but yeah, if you think that there will be a lot of overlapping between this and that then we can leave that refactoring for a future task

Regarding :point_up: we decided to merge this PR as is and complete the protocol changes implementation both with: https://jira.iohk.io/browse/ETCM-280 and https://jira.iohk.io/browse/ETCM-263

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
rtkaczyk: Not quite sure what you mean, in other parts of the code we have `number >= ecip1097BlockNumber`, so `number < ecip1097BlockNumber` is consistent with that.

src/main/scala/io/iohk/ethereum/network/p2p/messages/CommonMessages.scala
rtkaczyk: Well, it's not a case of optionality here. These are different messages with different codes, ie. once the code is determined a completely different structure may follow.
It's not a big deal the order doesn't really matter but these fields are closely related, in fact, after ETCM-263 they will correspond to a single type - `ChainWeight`, perhaps then it should even be encoded as a nested RLPList? 🤔

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
ntallar: My bad, you're right :facepalm:

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

pslaski (approved):

PR #736, opened 2020-10-14
[Chore] Add client id for metrics

mirkoAlic:
## Description
I found out that either prefix and client id are not working for "externals" registers. But still we need this feature for our metrics, with high priority for the deployer instances.

Screenshot(52)


mirkoAlic:
We ended up using a different thing at deployer level.

PR #738, opened 2020-10-15
[ETCM-174] Sorting of checkpoint signatures

pslaski:
# Description
Sort checkpoint signatures lexicographically to prevent having the same checkpoint block with a different hash (different order of signatures).

**Remember about changes on OBFT node side.**

- added validator of lexicographical order of checkpoint signatures

### DIFF
--- a/src/main/scala/io/iohk/ethereum/consensus/validators/BlockWithCheckpointHeaderValidator.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/validators/BlockWithCheckpointHeaderValidator.scala
@@ -13,6 +13,7 @@ class BlockWithCheckpointHeaderValidator(blockchainConfig: BlockchainConfig) {
 
   def validate(blockHeader: BlockHeader, parentHeader: BlockHeader): Either[BlockHeaderError, BlockHeaderValid] = {
     for {
+      _ <- validateLexicographicalOrderOfSignatures(blockHeader)
       _ <- validateCheckpointSignatures(blockHeader, parentHeader)
       _ <- validateEmptyFields(blockHeader)
       _ <- validateFieldsCopiedFromParent(blockHeader, parentHeader)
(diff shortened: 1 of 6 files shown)


pslaski:
that's right, will do the update

pslaski:
waiting for https://github.com/input-output-hk/mantis/pull/728

lemastero:
After margin [[ETCM-141] scalafmt ](https://github.com/input-output-hk/mantis/pull/743). This PR can be updated by git merge `etcm-174-validate-lexicographically-signatures-fmt`

Changes: 
[+55 −23](https://github.com/input-output-hk/mantis/pull/738/files)
[+62 -23](https://github.com/input-output-hk/mantis/compare/ETCM-141-scalafmt...etcm-174-validate-lexicographically-signatures-fmt?expand=1)

src/main/scala/io/iohk/ethereum/consensus/validators/BlockWithCheckpointHeaderValidator.scala
```diff
@@ -157,3 +171,9 @@ class BlockWithCheckpointHeaderValidator(blockchainConfig: BlockchainConfig) {
     else Left(CheckpointHeaderTreasuryOptOutError)
 
 }
+
+object BlockWithCheckpointHeaderValidator {
+  val NoCheckpointInHeaderError: BlockHeaderError = HeaderUnexpectedError(
```
ntallar, line 176: Very minor: is it worth having a `HeaderUnexpectedError` only for this? Maybe we could replace this by a `case object NoCheckpointInHeaderError extends BlockHeaderError`?

ntallar (changes requested):
Only a really minor change and LGTM! Should we document this change on [our checkpointing spec](https://input-output.atlassian.net/wiki/spaces/EMU/pages/1681686614/Checkpointing)?

pslaski:
I will update docs after merge

src/main/scala/io/iohk/ethereum/consensus/validators/BlockWithCheckpointHeaderValidator.scala
pslaski: I'm not sure about it. This error is used when `BlockWithCheckpointHeaderValidator` is trying to validate block without a checkpoint, what shouldn't have happened and is something unexpected. Maybe it's my subjective feeling, but when I did this change, I started to think that I promote that unexpected error to something which is expected negative behaviour. WDYT?

rtkaczyk (approved):

src/main/scala/io/iohk/ethereum/consensus/validators/BlockWithCheckpointHeaderValidator.scala
ntallar: Hmm I think we can leave it as is then

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

PR #739, opened 2020-10-15
[Chore] Add dev mode to build definition

kapke:
# Description

`-Xfatal-warnings` with `-Yno-unused-import` is pretty annoying. This PR fixes that

# Proposed Solution

Disable fatal warnings when `mantisDev` system property or `MANTIS_DEV` env variable is set to `true`

Enabled parallel test execution for unit tests

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,6 +5,9 @@ import scala.sys.process.Process
 // Necessary for the nix build, please do not remove.
 val nixBuild = sys.props.isDefinedAt("nix")
 
+// Enable dev mode: disable certain flags, etc.
+val mantisDev = sys.props.get("mantisDev").contains("true") || sys.env.get("MANTIS_DEV").contains("true")
+
 val commonSettings = Seq(
   name := "mantis",
   version := "3.0",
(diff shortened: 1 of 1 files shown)


lemastero (approved):

biandratti (approved):

build.sbt
```diff
@@ -5,6 +5,9 @@ import scala.sys.process.Process
 // Necessary for the nix build, please do not remove.
 val nixBuild = sys.props.isDefinedAt("nix")
 
+// Enable dev mode: disable certain flags, etc.
+val mantisDev = sys.props.get("mantisDev").contains("true") || sys.env.get("MANTIS_DEV").contains("true")
+
```
lemastero, line 10: There was a time when I was fanatical against fatal-warnings. Currently, I am like them turned on, except for console :)

Since we sort compiler flags. Possibly we could use this: https://github.com/DavidGregory084/sbt-tpolecat
By default, it turns off fatal-warnings in `console` and takes care of turning on relevant flags.

In fact, maybe using just this plugin will solve problems and there is no need for a special switch. (I am afraid sb would use it in CI etc.)

kapke: I find this flow working nicely for development (especially in iterative fashion, like TDD), where I see warnings without them breaking the build and then being able to recompile with fatal warnings just to see what's missing (I don't use console much due to metaspace leaks in sbt < 1.3, more bloop and IntelliJ).
I wanted to approach other flags proposed by tpolecat one by one to be able to enable them in small PRs done "in free time".

lemastero:
LGTM 👍

PR #741, opened 2020-10-15
[ECTM-212] Fix block preparation (taking into account EIP-161)

mmrozek:
# Description

Fix block generation.
BlockPreparator didn't work correctly in case of an empty account (with value == 0). The change was introduced in EIP-161

### DIFF
--- a/src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
@@ -178,17 +178,21 @@ trait Blockchain {
 
   def genesisBlock: Block = getBlockByNumber(0).get
 
-  def getWorldStateProxy(blockNumber: BigInt,
-                         accountStartNonce: UInt256,
-                         stateRootHash: Option[ByteString],
-                         noEmptyAccounts: Boolean,
-                         ethCompatibleStorage: Boolean): WS
+  def getWorldStateProxy(
+      blockNumber: BigInt,
+      accountStartNonce: UInt256,
+      stateRootHash: Option[ByteString],
+      noEmptyAccounts: Boolean,
+      ethCompatibleStorage: Boolean
+  ): WS
 
-  def getReadOnlyWorldStateProxy(blockNumber: Option[BigInt],
-                                 accountStartNonce: UInt256,
-                                 stateRootHash: Option[ByteString],
-                                 noEmptyAccounts: Boolean,
-                                 ethCompatibleStorage: Boolean): WS
+  def getReadOnlyWorldStateProxy(
+      blockNumber: Option[BigInt],
+      accountStartNonce: UInt256,
+      stateRootHash: Option[ByteString],
+      noEmptyAccounts: Boolean,
+      ethCompatibleStorage: Boolean
+  ): WS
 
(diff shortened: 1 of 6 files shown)


src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -28,25 +28,26 @@ object EvmConfig {
...
-      blockchainConfig.petersburgBlockNumber -> PetersburgConfigBuilder,
-      blockchainConfig.phoenixBlockNumber -> PhoenixConfigBuilder
+    val transitionBlockToConfigWithPriorityMapping: Map[BigInt, (Int, EvmConfigBuilder)] = Map(
+      blockchainConfig.frontierBlockNumber -> (1, FrontierConfigBuilder),
+      blockchainConfig.homesteadBlockNumber -> (2, HomesteadConfigBuilder),
+      blockchainConfig.eip150BlockNumber -> (3, PostEIP150ConfigBuilder),
+      blockchainConfig.eip160BlockNumber -> (4, PostEIP160ConfigBuilder),
+      blockchainConfig.eip161BlockNumber -> (5, PostEIP161ConfigBuilder),
+      blockchainConfig.byzantiumBlockNumber -> (6, ByzantiumConfigBuilder),
+      blockchainConfig.atlantisBlockNumber -> (6, AtlantisConfigBuilder),
```
KonradStaniec, line 38: As i understand that it is also not 100% deterministic as in case of same block and priority the chooses fork will be random one.  Maybe leave some comment about it and create and leave some ticket to refactor our blockchain configs ? (we have probably been postponing it long enough)

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -456,18 +473,20 @@ class BlockchainImpl(
     )
 
   //FIXME Maybe we can use this one in regular execution too and persist underlying storage when block execution is successful
```
KonradStaniec, line 475: maybe create ticket for this `FIXME` ? as it seems having two separate methods to retrieve execution world was a bit error prone

KonradStaniec (approved):
As we touched such general config and consensus maybe you cold run ets tests and report if every thing is fine ?

Otherwise lgtm.

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
mmrozek: Not random but first checked. So in our test config, when we set all numbers to 0 expecting to use the newest but we are getting the first in the map (so it is the oldest fork)

mmrozek:
@KonradStaniec I will run ETS and let you know

lemastero:
After margin [[ETCM-141] scalafmt ](https://github.com/input-output-hk/mantis/pull/743). This PR can be updated by git merge `etcm-212-fix-tx-mining-fmt`

[+486 −348](https://github.com/input-output-hk/mantis/pull/741/files)
[ +80  -29](https://github.com/input-output-hk/mantis/compare/ETCM-141-scalafmt...etcm-212-fix-tx-mining-fmt?expand=1)

kapke (approved):
LGTM!

PR #742, opened 2020-10-16
[ETCM-213] Reload bloom after restart

KonradStaniec:
# Description

Reload fast sync bloom filter after node restart

# Proposed Solution

Expose rockdb iterator over the node storage column and code column and consume it by putting all known keys in bloom filter

- Add `LoadabeBloomFilter` class which can load data from provided source
- Fix bug with NodeStorage. Up until now all nodes were put in Defualt rocks db column by accident, this pr fixes it and node storage now save mpt nodes in correct namespace - this fix break state and requires db cleanup to start


lemastero:
After margin [[ETCM-141] scalafmt ](https://github.com/input-output-hk/mantis/pull/743). This PR can be updated by git merge `etcm-213/relad-bloom-after-restart-fmt`

[+629 −202](https://github.com/input-output-hk/mantis/pull/742/files)
[+528 -138](https://github.com/input-output-hk/mantis/compare/ETCM-141-scalafmt...etcm-213/relad-bloom-after-restart-fmt?expand=1)

KonradStaniec:
Just fyi: Iterating over 10milion keys and values saved in db, on my machine (ssd drive) take around 8s. Now there is around 150M mpt nodes to save. So taking worst case into account i.e restart neat the finish and ~10s for 10M values, the worst time user will wait to re-lad bloom filter will be 150s which imo it is reasonable time to wait.

Take in mind that in case of loading failure (i.e not loading all nodes from db) will not make state sync unusable but will just degrade its performance due to possible large number of false positives i.e large number of unnecessary db calls. That why error handling is done is in best effort manner. We report the loading failure and error, but we proceed to start state sync without any loading restarts and recovers. We could in theory save the last loaded key before failure, and try to restart iterator from this key but imo for it is not worth it.

mmrozek (approved):
Nice job. One minor comment only

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -101,6 +106,50 @@ class RocksDbDataSource(
...
+                }
+              }
+            }
+            .onErrorHandleWith {
+              case IterationFinished => Observable.empty
+              case ex => Observable(Left(IterationError(ex)))
+            }
+        }
+      }
+  }
```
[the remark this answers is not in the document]
KonradStaniec: Yep, second `it.isValid` shold probably be in Task. I will refactor whole bit in `repeatEvalF` to be in for-comp

src/main/scala/io/iohk/ethereum/blockchain/sync/LoadableBloomFilter.scala
```diff
@@ -0,0 +1,41 @@
...
+      .consumeWith(Consumer.foldLeftTask(BloomFilterLoadingResult()) { (s, e) =>
+        e match {
+          case Left(value) => Task.now(s.copy(error = Some(value)))
+          case Right(value) => Task(bloomFilter.put(value)).map(_ => s.copy(writtenElements = s.writtenElements + 1))
+        }
+      })
+      .memoizeOnSuccess
+  }
+
+  def put(elem: A): Boolean = bloomFilter.put(elem)
```
kapke, line 21: is underlying bloom filter thread-safe? It's going to be populated from multiple threads from what I see

src/main/scala/io/iohk/ethereum/blockchain/sync/LoadableBloomFilter.scala
```diff
@@ -0,0 +1,41 @@
+package io.iohk.ethereum.blockchain.sync
+
+import com.google.common.hash.{BloomFilter, Funnel}
+import io.iohk.ethereum.blockchain.sync.LoadableBloomFilter.BloomFilterLoadingResult
+import io.iohk.ethereum.db.dataSource.RocksDbDataSource.IterationError
+import monix.eval.Task
+import monix.reactive.{Consumer, Observable}
+
+class LoadableBloomFilter[A](bloomFilter: BloomFilter[A], source: Observable[Either[IterationError, A]]) {
+  val loadFromSource: Task[BloomFilterLoadingResult] = {
```
kapke, line 10: I have mixed feelings about this setup TBH. If underlying bloom filter can be used only once loading completes - let's make a function which takes a source and returns task with ready-to-use bloom filter.

src/main/scala/io/iohk/ethereum/db/dataSource/DataSource.scala
```diff
@@ -51,6 +54,11 @@ trait DataSource {
     * This function closes the DataSource, if it is not yet closed, and deletes all the files used by it.
     */
   def destroy(): Unit
+
+  def iterate(): Observable[Either[IterationError, (Array[Byte], Array[Byte])]]
```
kapke, line 58: a comment would be nice describing what is in that tuple (is it key + value?)

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -101,6 +106,43 @@ class RocksDbDataSource(
...
+
+  private def namespaceIterator(namespace: Namespace): Resource[Task, RocksIterator] = {
+    Resource.fromAutoCloseable(Task(db.newIterator(handles(namespace))))
+  }
+
+  private def moveIterator(it: RocksIterator): Observable[Either[IterationError, (Array[Byte], Array[Byte])]] = {
+    Observable
+      .fromTask(Task(it.seekToFirst()))
+      .flatMap(_ => Observable.fromTask(Task(it.isValid)))
```
kapke, line 120: `filterEval`?

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -101,6 +106,43 @@ class RocksDbDataSource(
...
+      .filter(identity)
+      .flatMap { _ =>
+        Observable.fromTask(Task(Right(it.key(), it.value()))) ++ Observable
```
kapke, line 123: What about changing order a bit?
I think that this could be simplified to form like:
```scala
Observable
  // I'm not a fan of this, but I don't know significantly better solution
  .fromTask(Task(it.seekToFirst())) 
  .flatMap(_ =>
    Observable.repeatEvalF(for { //Or unfold maybe?
      isValid <- Task(it.isValid)
      //This could use `Task.ifM`, but its readabilty is discussable
      item <- if (isValid) Task(Right(it.key(), it.value())) else Task(Left(IterationFinished))
      _ <- Task(it.next())
    } yield item))
  .takeWhile(_.isRight))
  .onErrorHandle(ex => Left(IterationError(ex))
```
So:
 1. there's no duplicated part with getting first item
 2. code structure is more flat

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -101,6 +106,43 @@ class RocksDbDataSource(
...
+          }
+          .onErrorHandleWith {
+            case IterationFinished => Observable.empty
+            case ex => Observable(Left(IterationError(ex)))
+          }
+      }
+  }
+
+  def iterate(): Observable[Either[IterationError, (Array[Byte], Array[Byte])]] = {
+    Observable.fromResource(dbIterator).flatMap(it => moveIterator(it))
```
kapke, line 139: what about calling `it.seekToFirst` here?

src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
```diff
@@ -14,6 +14,8 @@ class AppStateStorage(val dataSource: DataSource) extends TransactionalKeyValueS
 
   val namespace: IndexedSeq[Byte] = Namespaces.AppStateNamespace
   def keySerializer: Key => IndexedSeq[Byte] = _.getBytes
+
+  def keyDeserializer: IndexedSeq[Byte] => Key = k => new String(k.toArray)
```
kapke, line 18: I'd consider specifying charset here and 2 lines below. Otherwise hard-to-track issues may arise e.g. when moving data between different OS

src/main/scala/io/iohk/ethereum/db/storage/FastSyncStateStorage.scala
```diff
@@ -35,6 +35,8 @@ class FastSyncStateStorage(val dataSource: DataSource)
 
   override def keySerializer: String => IndexedSeq[Byte] = _.getBytes
 
+  override def keyDeserializer: IndexedSeq[Byte] => String = b => new String(b.toArray)
```
kapke, line 38: same as in appstate storage - I'd add charset here

src/main/scala/io/iohk/ethereum/db/storage/KnownNodesStorage.scala
```diff
@@ -13,6 +13,7 @@ class KnownNodesStorage(val dataSource: DataSource) extends TransactionalKeyValu
 
   val namespace: IndexedSeq[Byte] = Namespaces.KnownNodesNamespace
   def keySerializer: String => IndexedSeq[Byte] = _.getBytes
+  def keyDeserializer: IndexedSeq[Byte] => String = k => new String(k.toArray)
```
kapke, line 16: ditto

src/test/scala/io/iohk/ethereum/blockchain/sync/LoadableBloomFilterSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+import monix.reactive.Observable
+
+class LoadableBloomFilterSpec extends FlatSpecBase {
+  implicit object LongFun extends Funnel[Long] {
+    override def funnel(from: Long, into: PrimitiveSink): Unit = {
+      Funnels.longFunnel().funnel(from, into)
+    }
+  }
+
+  "LoadableBloomFilter" should "load all correct elements " in testCaseM[Task] {
```
kapke, line 16: IIRC defining `Task` here should not be necessary

src/test/scala/io/iohk/ethereum/blockchain/sync/LoadableBloomFilterSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+    for {
+      source <- Task(Observable.fromIterable(Seq(Right(1L), Right(2L), Right(3L))))
+      filter <- Task.now(LoadableBloomFilter[Long](1000, source))
```
kapke, line 19: why not `filter = LoadableBloomFilter(...)` then? (and below too)

src/test/scala/io/iohk/ethereum/blockchain/sync/StateSyncSpec.scala
```diff
@@ -89,6 +91,24 @@ class StateSyncSpec
     }
   }
 
+  it should "start state sync when receiving start signal while bloom filter is loading" in new TestSetup() {
+    override def buildBlockChain(): BlockchainImpl = {
+      val storages = getNewStorages
+      //iterating 1M key and values should force scheduler actor o enqueue last received command i.e StartSyncing
+      (0 until 1000000).foreach { i =>
```
kapke, line 98: what about overriding storages in a way that will make source for bloom filter return observable which doesn't emit elements until some signal?

src/it/scala/io/iohk/ethereum/db/RockDbIteratorSpec.scala
```diff
@@ -0,0 +1,156 @@
...
+  it should "iterate over keys and values from different namespaces" in testCaseT { db =>
+    val codeStorage = new EvmCodeStorage(db)
+    val codeKeyValues = (1 to 10).map(i => (ByteString(i.toByte), ByteString(i.toByte))).toList
+
+    val nodeStorage = new NodeStorage(db)
+    val nodeKeyValues = (20 to 30).map(i => (ByteString(i.toByte), ByteString(i.toByte).toArray)).toList
+
+    for {
+      _ <- Task(codeStorage.update(Seq(), codeKeyValues).commit())
+      _ <- Task(nodeStorage.update(Seq(), nodeKeyValues))
```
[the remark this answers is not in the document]
KonradStaniec: it is not for now

src/main/scala/io/iohk/ethereum/blockchain/sync/LoadableBloomFilter.scala
KonradStaniec: yes, it is.

src/main/scala/io/iohk/ethereum/blockchain/sync/LoadableBloomFilter.scala
KonradStaniec: for now we are using underlying bloom filter that way (first load then use), as it is simplest possible scenario and loading time for etc is relatively shot i.e 2-3 min.

As underlying bloom filter is thread safe, we could probably start loading filter from storage and at the same time start state sync, but i would do this as separate task. (It could be especially useful if we were to support ETH when loading nodes would take probably more that 10min which is probably not acceptable)

That why i would leave it as it is.

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
KonradStaniec: so I like how this encoding gets rid of duplicated part, i will definitely go that way.

Although i will probably not use `.takeWhile(_.isRight))` and keep `Task.raiseError(IterationFinished)` part as i do not what `IterationFinished` to leak outside this function.

And i will probably not use `Task.ifM` , as in general i do not like it, and everytime i see it i need to check docs to be certain which case goes where.

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
KonradStaniec: I do not think that spreading iterator movements across two function is good idea. Imo it is nice to see it and have it in one place

src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
KonradStaniec: i was thinking about it, but decided to not touch it for now. But you are probably right, this pr break state eitherway so i will do it here.

src/test/scala/io/iohk/ethereum/blockchain/sync/LoadableBloomFilterSpec.scala
KonradStaniec: to be honest constant changes between `=` and `<-` just makes code unreadable for me, but sure i will change it 👍

src/main/scala/io/iohk/ethereum/db/storage/StorageStringCharset.scala
```diff
@@ -0,0 +1,7 @@
+package io.iohk.ethereum.db.storage
+
+import java.nio.charset.Charset
+
+object StorageStringCharset {
+  val UTF8Charset = Charset.forName("UTF-8")
```
kapke, line 6: Minor: there is `java.nio.StandardCharsets.UTF8` too

src/test/scala/io/iohk/ethereum/blockchain/sync/StateSyncSpec.scala
```diff
@@ -89,6 +95,47 @@ class StateSyncSpec
...
+        pruningMode = storages.pruningMode,
+        nodeStorage = storages.nodeStorage,
+        cachedNodeStorage = storages.cachedNodeStorage,
+        totalDifficultyStorage = storages.totalDifficultyStorage,
+        transactionMappingStorage = storages.transactionMappingStorage,
+        appStateStorage = storages.appStateStorage,
+        stateStorage = storages.stateStorage
+      ) {
+        override def mptStateSavedKeys(): Observable[Either[IterationError, ByteString]] = {
+          Observable.repeatEvalF(Task(Right(ByteString(1)))).takeWhile(_ => !loadingFinished)
```
kapke, line 121: minor: `Observable.repeat`?

kapke (approved):
LGTM!

PR #743, opened 2020-10-16
[ETCM-141] scalafmt

lemastero:
# Description

Add [scalafmt plugin](https://github.com/scalameta/sbt-scalafmt) and format all code.


rtkaczyk:
Should we add `scalafmtCheckAll` to buildkite?

.buildkite/pipeline.nix
```diff
@@ -54,6 +54,14 @@ in
       '';
     };
 
+    scalafmt = commonAttrs // {
+      dependsOn = [ compile ];
```
kapke, line 58: Why not other way around? Checking formatting is much more lightweight than compiling.

lemastero: Nice idea, changed, and now compile depends on formatting.

kapke (approved):
Let's get this merged as soon as CI passes

PR #746, opened 2020-10-20
[ETCM-202] Provide Mantis command-line interface

mmrozek:
# Description

CLI to mantis. Options:
- generating a new private key
- deriving address from private key
- generating allocs from private keys or addresses


serhii-shymkiv-iohk (approved):
* Please update the `README` too
* Other than that LGTM

src/main/scala/io/iohk/ethereum/cli/CliCommands.scala
```diff
@@ -0,0 +1,74 @@
...
+import io.iohk.ethereum.utils.ByteStringUtils
+import java.security.SecureRandom
+import org.bouncycastle.util.encoders.Hex
+
+object CliCommands {
+
+  val generatePrivateKeyCommand = "generate-private-key"
+  val deriveAddressCommand = "derive-address"
+  val generateAllocsCommand = "generate-allocs"
+  val useAddressesFlag = "useAddresses"
```
kapke, line 17: why not replace it with either `--address=<address>` or `--key=<key` (or in plural)? In that way it would be a bit easier and always clear what is passed without reaching for bool flags

src/main/scala/io/iohk/ethereum/App.scala
```diff
@@ -17,6 +18,7 @@ object App extends Logger {
     val mallet = "mallet"
     val faucet = "faucet"
     val ecKeyGen = "eckeygen"
+    val cli = "mantis-cli"
```
kapke, line 21: why not just `cli` if not separate launcher is created?

src/test/scala/io/iohk/ethereum/cli/CliCommandsSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+      )
+      .right
+      .value shouldBe s""""alloc": {$address: { "balance": $requestedBalance }, $address3: { "balance": $requestedBalance }, $address2: { "balance": $requestedBalance }}"""
+  }
+
+}
+
+object Fixture {
+
+  def argument(name: String, value: Option[Any] = None): String = s"--$name${value.fold("")(v => s"=${v.toString}")}"
```
kapke, line 87: very minor in current state - that `Any` there, it's fine now, but I'd consider using just `String` instead so any conversion needs to happen explicitly

kapke (approved):
LGTM!

PR #747, opened 2020-10-20
[FIX Fix bug in pivot block selector

KonradStaniec:
# Description

Fixes a bug in pivot block selector

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/PivotBlockSelector.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/PivotBlockSelector.scala
@@ -109,13 +109,15 @@ class PivotBlockSelector(
       timeout: Cancellable,
       headers: Map[ByteString, BlockHeaderWithVotes]
   ): Unit = {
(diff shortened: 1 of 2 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/PivotBlockSelector.scala
```diff
@@ -208,9 +210,15 @@ object PivotBlockSelector {
...
 
   implicit class SortableHeadersMap(headers: Map[ByteString, BlockHeaderWithVotes]) {
-    def mostVotedHeader: BlockHeaderWithVotes = headers.maxBy { case (_, headerWithVotes) =>
-      headerWithVotes.votes
-    }._2
+    def mostVotedHeader: Option[BlockHeaderWithVotes] = {
+      if (headers.isEmpty) {
+        None
+      } else {
+        Some(headers.maxBy { case (_, headerWithVotes) =>
```
pslaski, line 217: maybe cats maximumOption would be useful here?

pslaski (changes requested):

mmrozek (approved):
Nice catch!

pslaski (approved):

PR #735, opened 2020-10-14
Etcm 175 sync reporting issues

kapke:
# Description

Update `eth_syncing` method to standarized output (https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/etclabscore/ethereum-json-rpc-specification/master/openrpc.json&uiSchema%5BappBar%5D%5Bui:input%5D=false)) and move sync progress reporting into sync module. 

# Important Changes Introduced

A `SyncProtocol` is used now for high-level communication between sync controller and specific implementations as well as outside sync controller. Probably it makes sense to further extend that change and hide it behind a well-typed task-returning interfaces.

State breakage is related to changed datatype for tracking state nodes progress - from `Int` to `Long` (as in sync state scheduler)

# Testing

1. All automatic tests should pass (most notably unit and it ones)
2. When sync is actually happening (data is being actively fetched from network) - `eth_syncing` should return full response, including information about state nodes.

PR #749, opened 2020-10-20
[ETCM-211] Tracking of headers/bodies requests that will be ignored due to invalidation

ntallar:
# Description

Address the cause of CPU and network spikes on the testnet

# Proposed Solution

1. Start tracking when fetcher should start from scratch but there are requests in progress. This would allow the fetcher to ignore the responses when received and only then (when there are no requests in progress) send a new request
2. Receiving a new block should never cancel a request in progress (as it was on one scenario)
3. Document better syncing behavior

# Scenario addressed

1. Temporarily add logging for when a best block number is saved on the db ([here](https://github.com/input-output-hk/mantis/blob/ef2e58626d2103713d62bbc9696693752af6425f/src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala#L24))
2. Start node 3 and with mocked miner mine 500 blocks. Check with the log in 1. that a number close to 500 of blocks was saved on the db, if not continue mining till that's the case
3. Stop node 3
4. Start node 2 and with mocked miner mine 300 blocks
5. Start node 1 and connect it to node 2, await till it synced all it's blocks
6. Start node 3 and connect it to node 1
7. Mine enough blocks on node 3 so that it has more blocks than the best block number that had been saved in node 2.
8. The new blocks from 7. should have triggered that node 1 starts syncing from node 3.

Eventually the invalidation of local blocks due to unknown branch should have happened while a request was in progress, causing a new request to be sent, having at that moment 2 of them in parallel. As the sync doesn't handle 2 requests in parallel for the same type of data, eventually it should cause node 1 to have a lot of requests in parallel, using up a lot of network and cpu resources.

_Note that the step 1. and related checks are needed to prevent falling into the issue: https://jira.iohk.io/browse/ETCM-246
Note that step 8 is needed due to: https://jira.iohk.io/browse/ETCM-248_

Indicators for the issue ocurring:
- CPU usage should increase abnormally
- Logs like:
```
2020-10-20 15:27:47,315 [i.i.e.blockchain.sync.PeersClient] - Peer (PeerId(127.0.0.1:35232)) would be blacklisted (reason: Given headers are not sequence with already fetched ones), but blacklisting duration is zero
2020-10-20 15:27:47,315 [i.i.e.blockchain.sync.PeersClient] - Selected peer PeerId(127.0.0.1:35232) with address 127.0.0.1
2020-10-20 15:27:47,315 [i.i.e.blockchain.sync.PeersClient] - Peer (PeerId(127.0.0.1:35232)) would be blacklisted (reason: Given headers are not sequence with already fetched ones), but blacklisting duration is zero
2020-10-20 15:27:47,315 [i.i.e.blockchain.sync.PeersClient] - Selected peer PeerId(127.0.0.1:35232) with address 127.0.0.1
2020-10-20 15:27:47,315 [i.i.e.blockchain.sync.PeersClient] - Peer (PeerId(127.0.0.1:35232)) would be blacklisted (reason: Given headers are not sequence with already fetched ones), but blacklisting duration is zero
2020-10-20 15:27:47,315 [i.i.e.blockchain.sync.PeersClient] - Selected peer PeerId(127.0.0.1:35232) with address 127.0.0.1
2020-10-20 15:27:47,315 [i.i.e.blockchain.sync.PeersClient] - Peer (PeerId(127.0.0.1:35232)) would be blacklisted (reason: Given headers are not sequence with already fetched ones), but blacklisting duration is zero
2020-10-20 15:27:47,315 [i.i.e.blockchain.sync.PeersClient] - Selected peer PeerId(127.0.0.1:35232) with address 127.0.0.1
2020-10-20 15:27:47,317 [i.i.e.b.sync.regular.BlockFetcher] - Fetched 100 headers starting from block Some(359)
2020-10-20 15:27:47,317 [i.i.e.b.sync.regular.BlockFetcher] - Fetching headers from block 246
2020-10-20 15:27:47,317 [i.i.e.blockchain.sync.PeersClient] - Peer (PeerId(127.0.0.1:35232)) would be blacklisted (reason: Given headers are not sequence with already fetched ones), but blacklisting duration is zero
2020-10-20 15:27:47,317 [i.i.e.blockchain.sync.PeersClient] - Selected peer PeerId(127.0.0.1:35232) with address 127.0.0.1
2020-10-20 15:27:47,317 [i.i.e.b.sync.regular.BlockFetcher] - Fetched 100 headers starting from block Some(359)
2020-10-20 15:27:47,317 [i.i.e.b.sync.regular.BlockFetcher] - Fetching headers from block 246
2020-10-20 15:27:47,317 [i.i.e.b.sync.regular.BlockFetcher] - Fetched 100 headers starting from block Some(359)
2020-10-20 15:27:47,317 [i.i.e.b.sync.regular.BlockFetcher] - Fetching headers from block 246
2020-10-20 15:27:47,317 [i.i.e.b.sync.regular.BlockFetcher] - Fetched 100 headers starting from block Some(359)
2020-10-20 15:27:47,317 [i.i.e.b.sync.regular.BlockFetcher] - Fetching headers from block 246
```

# Testing

- [X] Reproducing the scenario that triggered this task and check that it's no longer there
- [X] Check that syncing continues to work, either with mainnet or mordor, just having it sync a couple thousands blocks should be enough. Synced 300000 blocks from mordor without any problems
- [x] Add unit tests on the block fetcher


src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -91,27 +95,51 @@ class BlockFetcher(
...
-
-      val newState = state.validatedHeaders(headers) match {
-        case Left(err) =>
-          peersClient ! BlacklistPeer(peer.id, err)
-          state.fetchingHeaders(false)
-        case Right(validHeaders) =>
-          state.appendHeaders(validHeaders)
-      }
+      val newState =
+        if (state.fetchingHeadersState == AwaitingHeadersToBeIgnored) {
```
KonradStaniec, line 99: Maybe I am missing somehing, but what will happen in case of request time out ? 
I see such path:

1. request headers
2. fetcher receive invalidate
3. request times out, but we are still in `AwaitingHeadersToBeIgnored` state
4. fetcher create request for new headers, from new invalidated blocks
5. fetcher receivers response witth new block headers, but it is ignored as fetcher is still in  `AwaitingHeadersToBeIgnored`

Is it correct, or there is some detail i am missing ?

ntallar: You are right! I thought I had analyzed that but I missed it

I'll update this PR with something even closer to what we do when receiving the response

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -91,27 +95,51 @@ class BlockFetcher(
...
+            case Right(validHeaders) =>
+              state.withHeaderFetchReceived.appendHeaders(validHeaders)
+          }
+        }
 
       fetchBlocks(newState)
     case RetryHeadersRequest if state.isFetchingHeaders =>
-      log.debug("Retrying request for headers")
-      fetchHeaders(state)
+      log.debug("Time-out occurred while waiting for headers")
```
KonradStaniec, line 120: more of discussion type comment. Do you think it would bring problems if we would receive this time out response later ? i.e something like:
1. We request header lets say from block with number 100
2. we receive invalidate from 100 to 80
3. we receive timeout for our request, we blacklist peer and clear state.
4. after some time (200 s by default) peer is unblacklisted , and we request headers from block with number 90 form it.
5.  we receive old response due to some network traffic problems or something like that.

Do you think it could result in something more than blacklising peer ? (for example triggering this bug with unbounded resources usage)

ntallar: Definitely, we are currently doing no validations that the headers received match the request

With that sort of validations we should be safe I think, but for now I was assuming that that case will never happen, the 30 seconds till timeout should be more than enough to prevent that, right?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -91,27 +95,51 @@ class BlockFetcher(
...
-      log.debug("Fetched {} headers starting from block {}", headers.size, headers.headOption.map(_.number))
-
-      val newState = state.validatedHeaders(headers) match {
-        case Left(err) =>
-          peersClient ! BlacklistPeer(peer.id, err)
-          state.fetchingHeaders(false)
-        case Right(validHeaders) =>
-          state.appendHeaders(validHeaders)
-      }
+      val newState =
```
kapke, line 98: I'd move this code into method on `BlockFetcherState` and leave here only logging

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
kapke: That makes me think if this whole solution shouldn't be implemented as Actor/class in between fetcher and peers client. It could track requests made and invalidate previous ones basing on strategy choosen (at-most-n, allow-all, etc.). 
This could simplify things significantly and allow code in fetcher focus on just orchestrating fetching of blocks.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -154,9 +182,13 @@ case class BlockFetcherState(
   def withPeerForBlocks(peerId: PeerId, blocks: Seq[BigInt]): BlockFetcherState =
     copy(blockProviders = blockProviders ++ blocks.map(block => block -> peerId))
 
-  def fetchingHeaders(isFetching: Boolean): BlockFetcherState = copy(isFetchingHeaders = isFetching)
+  def isFetchingHeaders: Boolean = fetchingHeadersState != NoHeadersFetched
+  def withNewHeadersFetch: BlockFetcherState = copy(fetchingHeadersState = AwaitingHeaders)
+  def withHeaderFetchReceived: BlockFetcherState = copy(fetchingHeadersState = NoHeadersFetched)
```
kapke, line 187: `no fetched` or `not fetching`?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -154,9 +182,13 @@ case class BlockFetcherState(
...
 
-  def fetchingBodies(isFetching: Boolean): BlockFetcherState = copy(isFetchingBodies = isFetching)
+  def isFetchingBodies: Boolean = fetchingBodiesState != NoBodiesFetched
+  def withNewBodiesFetch: BlockFetcherState = copy(fetchingBodiesState = AwaitingBodies)
+  def withBodiesFetchReceived: BlockFetcherState = copy(fetchingBodiesState = NoBodiesFetched)
```
kapke, line 191: ditto

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -188,11 +220,31 @@ object BlockFetcherState {
...
+    fetchingBodiesState = NoBodiesFetched,
     stateNodeFetcher = None,
     lastBlock = lastBlock,
     knownTop = lastBlock + 1,
     blockProviders = Map()
   )
+
+  trait FetchingHeadersState
+  case object NoHeadersFetched extends FetchingHeadersState
+  case object AwaitingHeaders extends FetchingHeadersState
```
kapke, line 233: what about tracking parameters for request made? It would allow to track (and at least log for now) case which Konrad mentioned - received response in a moment when we're waiting for different one

src/test/scala/io/iohk/ethereum/BlockHelpers.scala
```diff
@@ -0,0 +1,25 @@
+package io.iohk.ethereum
+
+import akka.util.ByteString
+import io.iohk.ethereum.domain.Block
+
+trait BlockHelpers {
```
kapke, line 6: Could you make it an object? I'd move helpers I extracted there: https://github.com/input-output-hk/mantis/pull/735/files#diff-840c293a5c8d271766b7ad38d72f23fdc2a5d0406db32ee60fd727cc6615ca75R1 into the same package, so less conflicts and/or code duplicates are created

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
ntallar: What would this actor do? Are you maybe thinking of something like the BlockChainDataFetcher of our other project? (only on how it requests headers and bodies)

>It could track requests

That at least would focus on :point_up:  part and allow validating that the responses received match the requests we sent

Either way I'd do that effort in parallel with this kind of patching that only touches the minimum required (and is needed for the testnet)

I'm a bit worried of how risky any design changes on our sync could be... though that BlockChainDataFetcher refactoring would be low risk I think

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
ntallar: As a handleHeaderResponse function you mean? Isn't the current structure more oriented to doing the handling of responses on the actor? Shouldn't we move the bodies/nodes response handling as well?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
ntallar: I'd do this on a separate task, one that adds the BlockChainDataFetcher class and all the validations of requests from there, wdyt?

(also for now I don't think it will be that straightforward integrating the BlockChainDataFetcher with the current actors that we have)

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
kapke: Well, `BlockFetcher` is more or less the same as `BlockchainDataFetcher` there already IMO. I think more on structure like that:
```
                           block fetcher
                      |       |                      |
headers fetcher(Fetcher)  bodies fetcher(Fetcher)  state nodes fetcher(Fetcher)
                      |       |                      |
                             peers client
```
Where `Fetcher` is that specialized class/actor that manages requests of given type and could have interface like:
```
abstract class Fetcher(multipleReqStrategy: AtMostN|AllowAll) {
  def makeRequest(msg: MessageSerializable): IO[msg.Response] //Issues a request and add it to tracked pool
  def cancelPendingRequests: IO[Unit] //Stores information about cancellation so knows which responses cannot be passed further if received
}  
```

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
kapke: After second look - let's keep it as is. It's just - the more I work with actors, more I want extract all the logic from them so it can be tested without creating that actor.

kapke (approved):
LGTM!

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
KonradStaniec: so imo details up to discussion, but i am all for doing some kinda detailed request-response tracking. Unfortunately as eth do not have any request-id for its request-response messages, such thing is ultimately necessary if we want to have possiblitly to cancel in fly requests.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
KonradStaniec: it is totally my current sentiments about actors, they should have as little logic as possible and should be only thin communication layer over main logic.

KonradStaniec (approved):

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
ntallar: Yeah, I'm having the same concerns with them

However for this case in particular I think we would fall into a too big and not understandable BlockFetcherState class if we start moving everything there, we should probably start splitting up this logic into other classes

PR #752, opened 2020-10-21
[ETCM-260] Tweak network config

KonradStaniec:
# Description

Sometimes it takes long time to find new peers. And main reason for it is that we restrict found peers to 1000 nodes where there at least 5000 in the network.

Also increases max blacklisted nodes on network level to avoid constant churn in blacklist list.

Also changes peer inoo logs to `info `as it is pretty usefull


src/main/scala/io/iohk/ethereum/blockchain/sync/BlacklistSupport.scala
```diff
@@ -12,17 +12,17 @@ trait BlacklistSupport {
 
   def scheduler: Scheduler
 
-  private val maxSize = 1000
```
mmrozek, line 15: Maybe we could get this value from config?

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -35,6 +35,7 @@ class PeerManagerActor(
     with ActorLogging
     with Stash
     with BlacklistSupport {
+  override val maxBlacklistedNodes: Int = 5000
```
mmrozek, line 38: Why do we need to override this value? Couldn't we change it in the base trait? Or maybe base trait shouldn't use any default value?

src/main/scala/io/iohk/ethereum/blockchain/sync/BlacklistSupport.scala
KonradStaniec: problem is that this trait is shared across many different components, so the question arises what config. Also some of those componets should probably use differnt values.

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
KonradStaniec: we override it as other components do not need such large value, as usually they do not see more than few peers. PeerManageActor se more peers as it is almost on the lowset level of the stack, so it will see ~5k peers from which most will be useless (not neceserilly malicious, but just not usefull).

This whole blacklisting/marking useless peers approach sucks, but here I just wanted to tweak some values to more realitics one.

One thing we could easliy imporoved is to use here value from `DiscoveryConfig.nodesLimit` as we should not have more blacklisted peers than those found in discovery. wdyt ?

mmrozek: I agree with you. We could use `nodesLimit` here. Please add a comment why such value (as you explain above)

mmrozek (approved):
lgtm

PR #753, opened 2020-10-22
[ETCM-202] Fix address derivation command

mmrozek:
# Description

Fix address derivation (address length)


src/main/scala/io/iohk/ethereum/cli/CliCommands.scala
```diff
@@ -62,8 +63,9 @@ object CliCommands {
 
   private def privKeyToAddress(privKey: Array[Byte]): String = {
```
kapke, line 64: should this be now `Address.fromPrivateKey`?

kapke (approved):
LGTM!

PR #754, opened 2020-10-22
[ETCM-173] Re-think the timestamp validation rule of checkpoint blocks

pslaski:
# Description

The goal was to make blocks with checkpoint not included in computing PoW difficulty.

# Proposed Solution

Set block with checkpoint difficulty to zero and in case of having a block with checkpoint as a parent use grandparent difficulty and timestamp gap between grandparent and great grandparent

- block with checkpoint has zero difficulty now
- the difficulty and timestamp gap used for computing new difficulty is taken from grandparents when the parent is a block with checkpoint

# Testing

Unit tests

### DIFF
--- a/src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/EthashDifficultyCalculator.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/EthashDifficultyCalculator.scala
@@ -1,10 +1,13 @@
 package io.iohk.ethereum.consensus.ethash.difficulty
 
+import akka.util.ByteString
 import io.iohk.ethereum.consensus.difficulty.DifficultyCalculator
-import io.iohk.ethereum.domain.BlockHeader
-import io.iohk.ethereum.utils.BlockchainConfig
+import io.iohk.ethereum.consensus.ethash.difficulty.EthashDifficultyCalculator._
+import io.iohk.ethereum.domain.{BlockHeader, Blockchain}
+import io.iohk.ethereum.utils.{BlockchainConfig, ByteStringUtils}
 
-class EthashDifficultyCalculator(blockchainConfig: BlockchainConfig) extends DifficultyCalculator {
+class EthashDifficultyCalculator(blockchainConfig: BlockchainConfig, grandParentsDataGetter: GrandparentsDataGetter)
+    extends DifficultyCalculator {
   import blockchainConfig._
 
   val DifficultyBoundDivision: Int = 2048
@@ -15,9 +18,15 @@ class EthashDifficultyCalculator(blockchainConfig: BlockchainConfig) extends Dif
   val ConstantinopleRelaxDifficulty: BigInt = 5000000
 
   def calculateDifficulty(blockNumber: BigInt, blockTimestamp: Long, parentHeader: BlockHeader): BigInt = {
-    lazy val timestampDiff = blockTimestamp - parentHeader.unixTimestamp
+    lazy val GrandparentsData(grandparentDifficulty, grandparentsTimestampDiff) = grandParentsDataGetter(
+      parentHeader.hash
+    )
+    val parentHeaderDifficulty = if (!parentHeader.hasCheckpoint) parentHeader.difficulty else grandparentDifficulty
+    lazy val timestampDiff = if (!parentHeader.hasCheckpoint) {
+      blockTimestamp - parentHeader.unixTimestamp
+    } else grandparentsTimestampDiff
 
-    val x: BigInt = parentHeader.difficulty / DifficultyBoundDivision
+    val x: BigInt = parentHeaderDifficulty / DifficultyBoundDivision
     val c: BigInt =
       if (blockNumber < homesteadBlockNumber) {
         if (blockTimestamp < parentHeader.unixTimestamp + 13) 1 else -1
@@ -45,7 +54,7 @@ class EthashDifficultyCalculator(blockchainConfig: BlockchainConfig) extends Dif
       } else
         0
 
-    val difficultyWithoutBomb = MinimumDifficulty.max(parentHeader.difficulty + x * c)
+    val difficultyWithoutBomb = MinimumDifficulty.max(parentHeaderDifficulty + x * c)
     difficultyWithoutBomb + extraDifficulty
   }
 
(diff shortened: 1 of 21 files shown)


src/test/scala/io/iohk/ethereum/consensus/validators/BlockHeaderValidatorSpec.scala
```diff
@@ -252,56 +253,6 @@ class BlockHeaderValidatorSpec
     )
   }
 
-  it should "properly calculate the difficulty after difficulty bomb resume (with reward reduction)" in new EphemBlockchainTestSetup {
```
rtkaczyk, line 255: Yes!

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
```diff
@@ -104,6 +104,8 @@ object BlockHeader {
 
   val EmptyOmmers: ByteString = ByteString(crypto.kec256(rlp.encode(RLPList())))
 
+  val EmptyDifficulty: BigInt = BigInt(0)
```
rtkaczyk, line 107: I'd call it `ZeroDifficulty` (since we're feeding our OCDs 😁 )

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/EthashDifficultyCalculator.scala
```diff
@@ -15,9 +18,15 @@ class EthashDifficultyCalculator(blockchainConfig: BlockchainConfig) extends Dif
   val ConstantinopleRelaxDifficulty: BigInt = 5000000
 
   def calculateDifficulty(blockNumber: BigInt, blockTimestamp: Long, parentHeader: BlockHeader): BigInt = {
-    lazy val timestampDiff = blockTimestamp - parentHeader.unixTimestamp
+    lazy val GrandparentsData(grandparentDifficulty, grandparentsTimestampDiff) = grandParentsDataGetter(
+      parentHeader.hash
+    )
+    val parentHeaderDifficulty = if (!parentHeader.hasCheckpoint) parentHeader.difficulty else grandparentDifficulty
```
rtkaczyk, line 24: Maybe rename to `ancestorDifficulty`?

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/EthashDifficultyCalculator.scala
```diff
@@ -15,9 +18,15 @@ class EthashDifficultyCalculator(blockchainConfig: BlockchainConfig) extends Dif
   val ConstantinopleRelaxDifficulty: BigInt = 5000000
 
   def calculateDifficulty(blockNumber: BigInt, blockTimestamp: Long, parentHeader: BlockHeader): BigInt = {
-    lazy val timestampDiff = blockTimestamp - parentHeader.unixTimestamp
+    lazy val GrandparentsData(grandparentDifficulty, grandparentsTimestampDiff) = grandParentsDataGetter(
```
rtkaczyk, line 21: Ok, my OCD is breaking through the roof, but I would put everything after `=` on a single new line

rtkaczyk:
I'm a bit uneasy that we leaped to implementation so quickly. Do we have any numerical analysis that suggests that this idea is in fact better, or that the current way has any practical downsides?

pslaski:
No, but I can prepare some quasi-test-benchmark to see how the difficulty will change

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/EthashDifficultyCalculator.scala
```diff
@@ -15,9 +18,15 @@ class EthashDifficultyCalculator(blockchainConfig: BlockchainConfig) extends Dif
...
+    val parentHeaderDifficulty = if (!parentHeader.hasCheckpoint) parentHeader.difficulty else grandparentDifficulty
+    lazy val timestampDiff = if (!parentHeader.hasCheckpoint) {
+      blockTimestamp - parentHeader.unixTimestamp
+    } else grandparentsTimestampDiff
 
-    val x: BigInt = parentHeader.difficulty / DifficultyBoundDivision
+    val x: BigInt = parentHeaderDifficulty / DifficultyBoundDivision
     val c: BigInt =
       if (blockNumber < homesteadBlockNumber) {
         if (blockTimestamp < parentHeader.unixTimestamp + 13) 1 else -1
```
ntallar, line 32: >parentHeader.unixTimestamp

Should this be adapted as well?

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/EthashDifficultyCalculator.scala
```diff
@@ -15,9 +18,15 @@ class EthashDifficultyCalculator(blockchainConfig: BlockchainConfig) extends Dif
   val ConstantinopleRelaxDifficulty: BigInt = 5000000
 
   def calculateDifficulty(blockNumber: BigInt, blockTimestamp: Long, parentHeader: BlockHeader): BigInt = {
```
ntallar, line 20: I've been thinking... given this scenario of B1->B2->C3->B4, that we are calculating the difficulty of B4 what would be the difference between:
1. Do as done here that the difficulty of B4 is calculated using the difference in timestamps of B2 and B1
2. Have B4 have the same difficulty as B2

?

Aren't they mostly the same? (if that's so maybe we could go with 2. as it's simpler)

ntallar (changes requested):

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/EthashDifficultyCalculator.scala
ntallar: Something else that I noticed and I don't like much is that having the `timestampDiff=B2.timestamp-B1.timestamp` means that it doesn't adapt in any way to a how long B4 takes to mine... I was thinking of something like `(B4.timestamp-B2.timestamp)/2` but it assumes that producing a checkpoint block takes the same time to produce as a regular block. Maybe the best alternative for now is just leave one of the above solutions and re-check that after evaluating the integration with OBFT

pslaski:
temporarily closed as we decided to first test current calculator on the testnet

PR #755, opened 2020-10-22
Etcm 185 db closed

biandratti:
# Description

Before we were using a volatile variable called "isClosed". When the node is shutdown, close the database, and the variable has the value "true". So usually when it receives a new request, this request validates the value of  "isClosed" and throws the exception "This RocksDbDataSource has been closed".
I reproduced this error message when I used regular sync with network "mordor".

- Previous task that add a volatile variable:
https://jira.iohk.io/browse/ETCM-50

# Interrogant

- **update function, need dbLock.readLock() or dbLock.writeLock() ?**

# Proposed Solution

I looked at the code with more details and understood that this volatile variable is not necessary. I only used “ReentrantReadWriteLock” and when I shutdown the node and closed the database, it did not unlock.

- I removed the volatile variable.
- Changed ReentrantReadWriteLock behavior.
- Added FakeRocksDbDataSource for integracion test. Then when I closed the “RocksDbDataSource”, unlocked the writelock of “ReentrantReadWriteLock”.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
@@ -2,6 +2,7 @@ package io.iohk.ethereum.blockchain.sync
 
 import akka.actor.{Actor, ActorLogging, ActorRef, PoisonPill, Props, Scheduler}
 import io.iohk.ethereum.blockchain.sync.regular.RegularSync
+import io.iohk.ethereum.consensus.blocks.CheckpointBlockGenerator
 import io.iohk.ethereum.consensus.validators.Validators
 import io.iohk.ethereum.db.storage.{AppStateStorage, FastSyncStateStorage}
 import io.iohk.ethereum.domain.Blockchain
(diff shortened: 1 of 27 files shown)


src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -116,17 +112,14 @@ class RocksDbDataSource(
...
-    this.isClosed = false
   }
 
   /**
     * This function closes the DataSource, without deleting the files used by it.
     */
   override def close(): Unit = {
     logger.debug(s"About to close DataSource in path: ${rocksDbConfig.path}")
-    assureNotClosed()
-    isClosed = true
```
KonradStaniec, line 128: I wonder if the solution to this problem should just be  calling `isClosed = true` inside the `try` block  after all close operations?

KonradStaniec (commented):
to be honest i have mixed feelings about this solution as:
 - adding fake db impl only for the sake of special unlock method sounds hacky
 - by calling lock without unlock we are just hiding potential problem i.e the fact that there could be some db requests after db close up. Imo the ultimate solution should find what component is making those requests.

PR #756, opened 2020-10-22
[ETCM-269] future task

lemastero:
# Description

Replace Future with Task in JSON RPC.

# Proposed Solution

# Testing


rtkaczyk:
When done, could you please provide a brief argument for the choice of Monix rather than cats-effect or ZIO?

build.sbt
```diff
@@ -102,6 +102,8 @@ scalacOptions in (Compile, console) ~= (_.filterNot(
   )
 ))
 
+scalafmtOnCompile := true
```
kapke, line 105: why doing this? It's explicitly discouraged on scalafmt site for good reasons IMO

src/main/scala/io/iohk/ethereum/healthcheck/Healthcheck.scala
```diff
@@ -18,20 +19,20 @@ import scala.util.{Failure, Success}
   */
 case class Healthcheck[Error, Result](
     description: String,
-    f: () => Future[Either[Error, Result]],
+    f: () => Task[Either[Error, Result]],
```
kapke, line 22: Minor - it doesn't have to be a function now

src/main/scala/io/iohk/ethereum/healthcheck/Healthcheck.scala
```diff
@@ -1,7 +1,8 @@
 package io.iohk.ethereum.healthcheck
 
-import scala.concurrent.{ExecutionContext, Future}
-import scala.util.{Failure, Success}
+import monix.eval.Task
+
+import scala.concurrent.ExecutionContext
```
kapke, line 5: Shouldn't all `ExecutionContext` disappear now from RPC code?

src/main/scala/io/iohk/ethereum/jsonrpc/AkkaTaskOps.scala
```diff
@@ -0,0 +1,20 @@
+package io.iohk.ethereum.jsonrpc
+
+import akka.actor.{Actor, ActorRef}
+import akka.util.Timeout
+import monix.eval.Task
+
+import scala.reflect.ClassTag
+
+object AkkaTaskOps { self: ActorRef =>
```
kapke, line 9: what is purpose of `self: ActorRef` here?

src/main/scala/io/iohk/ethereum/jsonrpc/CheckpointingService.scala
```diff
@@ -21,7 +22,7 @@ class CheckpointingService(
     lazy val bestBlockNum = blockchain.getBestBlockNumber()
     lazy val blockToReturnNum = bestBlockNum - bestBlockNum % req.checkpointingInterval
 
-    Future {
+    Task.fromFuture(Future {
```
kapke: Why not `Task { ... }` then?

src/main/scala/io/iohk/ethereum/jsonrpc/DebugService.scala
```diff
@@ -25,9 +24,9 @@ class DebugService(peerManager: ActorRef, etcPeerManager: ActorRef) {
     val result = for {
       ids <- getPeerIds
       peers <- Future.traverse(ids)(getPeerInfo)
-    } yield ListPeersInfoResponse(peers.flatten)
+    } yield Right(ListPeersInfoResponse(peers.flatten))
 
-    result.map(Right(_))
+    Task.fromFuture(result)
```
kapke: `Task.fromFuture` takes a by-name parameter to ensure ref transparency. Having a variable holding `Future` breaks it and may lead to weird error where something is already executed despite missing `run` on `Task`

src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
```diff
@@ -77,30 +79,30 @@ class PersonalService(
 
   private val unlockedWallets: ExpiringMap[Address, Wallet] = ExpiringMap.empty(Duration.ofSeconds(defaultUnlockTime))
 
-  def importRawKey(req: ImportRawKeyRequest): ServiceResponse[ImportRawKeyResponse] = Future {
+  def importRawKey(req: ImportRawKeyRequest): ServiceResponse[ImportRawKeyResponse] = Task.fromFuture(Future {
```
kapke: As previously: why not `Task {... }`?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -79,11 +80,15 @@ trait JsonRpcHttpServer extends Json4sSupport {
   }
 
   private def handleRequest(request: JsonRpcRequest) = {
-    complete(jsonRpcController.handleRequest(request))
+    complete(jsonRpcController.handleRequest(request).runToFuture(MonixGlobal))
```
kapke, line 83: Why not https://monix.io/api/3.2/monix/execution/Scheduler$.html#global:monix.execution.Scheduler then? That one is not available in implicit scope

src/test/scala/io/iohk/ethereum/SpecBase.scala
```diff
@@ -27,19 +27,11 @@ trait SpecBase extends TypeCheckedTripleEquals with Diagrams with Matchers { sel
     fixture.use(theTest).toIO.unsafeToFuture()
   }
 
-  def customTestCaseResourceF[T](fixture: Resource[Task, T])(theTest: T => Future[Assertion]): Future[Assertion] =
```
kapke, line 30: <3

kapke (commented):
Great job with that migration!

kapke:
One more thing, that might be useful - in service specs you might use `SpecBase` traits and for sync, blocking specs write a wrapper like done there: https://github.com/input-output-hk/mantis/pull/735/files#diff-4f204c964a1f172070b2e06bc6b410e888aa541c50fe67c44824d5028d1e82d8R59. So new specs are free to return Task and not bother about actually executing it

build.sbt
lemastero: I keep forgetting about formatting the code. I missed the warning in docs. Flag removed - I configured my IDE to do the job :)

lemastero:
> could you please provide a brief argument for the choice of Monix rather than cats-effect or ZIO?

I am strongly convinced ZIO is better.

* very detailed thread dumps
* explicit errors (you can use Task[Either[E,A]] but you need nested maps/flatmaps to handle error, ZIO expose directly methods to deal with that, and in sense, you are encouraged to say what error type is)
* capabilities of FP patterns like monad transformers, Kleisli, Contravariant, Functor, Profunctor, Arrow, Monad, MonadError, Compose are included in ZIO and hence is easier and more performant 
* you can change input/environment and that is not possible neither with Monix/cats-effect nor with Monix BIO
* and people do not block you on Twitter for random reasons :wink:
* ZIO has consistent design, always input, output and error channel not only in ZIO but in other abstractions

monix.Task because:
* it is already in the project
* it requires the least changes (when changing Future to Task you need to just change name + take care of differences in laziness and slightly different API) with ZIO there is more changes (you can model better error, because you can model input, functions become val, ...)
* fewer changes == better review 
* fewer changes == less opportunity to make error

Why monix.Task not cas-effect - I think abstracting over effect type sounds good in theory but:
* is way more complex in terms of wiring, has a more steep learning curve. I was on project with monix, I was on project with monix + CE and the latter was significantly more complex. I like fancy types in libraries/frameworks - more powerful, more general. I prefer simple code in production = easier to maintain.
* IMHO using CE to be able to switch IO monad implementation from Task to ZIO sounds good but is not great. ZIO powers come from modeling explicitly errors and input + consistent design + use data types instead of traits (speed). CE has the same design as monix and in fact, JdG was banned from typelevel libs when they were working on CE. To make CE expressive enough to handle both Monix and ZIO. It did not happen. So the only reasonable IO type implementation for CE is Task.
* switching from Task to ZIO is simpler (in theory you can import zio.Task instead of monix.eval.Task in the simplest approach) compared to CE to ZIO

I am willing to try to port code to ZIO and see by myself what would be practical benefits for mantis and come back to this discussion. Yet for now, Task seems like a good choice (stepping stone) :)

build.sbt
```diff
@@ -68,8 +68,12 @@ val root = {
       libraryDependencies ++= dep
     )
     .settings(executableScriptName := name.value)
-    .settings(inConfig(Integration)(Defaults.testSettings
-      ++ org.scalafmt.sbt.ScalafmtPlugin.scalafmtConfigSettings :+ (Test / parallelExecution := false)): _*)
+    .settings(
+      inConfig(Integration)(
+        Defaults.testSettings
```
kapke, line 73: 👍

src/main/scala/io/iohk/ethereum/healthcheck/Healthcheck.scala
lemastero: Good spot. Changed `() => Task` into `Task`

src/main/scala/io/iohk/ethereum/healthcheck/Healthcheck.scala
lemastero: Now it is not needed. Thanks.

src/main/scala/io/iohk/ethereum/jsonrpc/AkkaTaskOps.scala
lemastero: Not needed. (Leftover after the previous solution with the trait.)

src/main/scala/io/iohk/ethereum/jsonrpc/PersonalService.scala
lemastero: Done.

src/main/scala/io/iohk/ethereum/jsonrpc/CheckpointingService.scala
lemastero: This construction was removed everywhere, including PersonalService that was very very tricky.

Currently, Task.fromFuture is used:
* QAService.mineBlocks to get rid of that call we would need port Future => Task in `Consensus`; that could happen as separate PR (this PR solve more than half of the usages of Futures 334 vs 167)
* in a wrapper around ask pattern in AkkaTaskOps

src/main/scala/io/iohk/ethereum/jsonrpc/DebugService.scala
lemastero: Good point. This is fixed now.

src/test/scala/io/iohk/ethereum/SpecBase.scala
lemastero: No longer used :)

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
lemastero: Yeach previously I needed to rename globacl Scheduler from monix to avoid name class. But since c02157d1939e162ec92d38b174be10a73e2046ef there is no name clash. Fixed: https://github.com/input-output-hk/mantis/pull/756/commits/490f30a57975707055b5df49d3bbad05e015ee4b

build.sbt
lemastero: This change seems like the result of PR is not up to date (I solved some nasty conflicts yesterday, so I do not blame automatic tools like GH for bad visualization of changes :) ). There are no changes in build.sbt.

kapke (approved):
Looks good to me! 
While I do think tests could be improved a bit, I think that they can be improved separately from this PR, even if organically when working on RPC

src/main/scala/io/iohk/ethereum/jsonrpc/AkkaTaskOps.scala
```diff
@@ -0,0 +1,20 @@
...
+import scala.reflect.ClassTag
+
+object AkkaTaskOps {
+  implicit class TaskActorOps(val to: ActorRef) extends AnyVal {
+
+    def askFor[A](
+        message: Any
+    )(implicit timeout: Timeout, classTag: ClassTag[A], sender: ActorRef = Actor.noSender): Task[A] =
+      Task
+        .fromFuture((to ? message).mapTo[A])
```
mmrozek, line 17: Do we want to use strict evaluation here? Shouldn't be `Task.deferFuture`?

src/test/scala/io/iohk/ethereum/jsonrpc/DebugServiceSpec.scala
```diff
@@ -3,24 +3,31 @@ package io.iohk.ethereum.jsonrpc
...
+    with WithActorSystemShutDown
+    with Matchers
+    with MockFactory
+    with ScalaFutures {
 
   "DebugService" should "return list of peers info" in new TestSetup {
-    val result: ServiceResponse[ListPeersInfoResponse] =
-      debugService.listPeersInfo(ListPeersInfoRequest())
+    val result =
+      debugService.listPeersInfo(ListPeersInfoRequest()).runToFuture
```
mmrozek, line 30: Why `runToFuture`? To be consistent we should use `runSyncUnsafe` everywhere (and avoid `futureValue`)

src/main/scala/io/iohk/ethereum/jsonrpc/AkkaTaskOps.scala
lemastero: Very good question 🤔

src/test/scala/io/iohk/ethereum/jsonrpc/DebugServiceSpec.scala
lemastero: runSyncUnsafe  was implemented at the beginning. But this break the tests:
```
[info] DebugServiceSpec:
[info] DebugService
[info] - should return list of peers info *** FAILED *** (8 seconds, 376 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsg while waiting for PeerInfoRequest(PeerId(testProbe-4))
[info]   at scala.Predef$.assert(Predef.scala:223)
[info]   at akka.testkit.TestKitBase.expectMsg_internal(TestKit.scala:459)
[info]   at akka.testkit.TestKitBase.expectMsg(TestKit.scala:436)
[info]   at akka.testkit.TestKitBase.expectMsg$(TestKit.scala:436)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:969)
[info]   at io.iohk.ethereum.jsonrpc.DebugServiceSpec$$anon$1.<init>(DebugServiceSpec.scala:35)
[info]   at io.iohk.ethereum.jsonrpc.DebugServiceSpec.$anonfun$new$1(DebugServiceSpec.scala:28)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   ...
```
Some tests contain a mix of:
* expectations on Akka TestProbe
* sending messages to actors
* assertions at the end.

In some cases, I moved expectations and assertions at the end. But in some cases, I was not able to do it so I used conversion to Future to get async behavior.

src/main/scala/io/iohk/ethereum/jsonrpc/AkkaTaskOps.scala
lemastero: This breaks the tests, I will explore what is needed to fix that.

src/test/scala/io/iohk/ethereum/jsonrpc/DebugServiceSpec.scala
```diff
@@ -55,9 +59,7 @@ class DebugServiceSpec extends AnyFlatSpec with Matchers with MockFactory with S
     result.futureValue shouldBe Right(ListPeersInfoResponse(List.empty))
   }
 
-  trait TestSetup {
-    implicit val system: ActorSystem = ActorSystem("debug-service-test")
-
+  class TestSetup(implicit system: ActorSystem) {
```
biandratti, line 62: I like it

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -546,7 +550,7 @@ class EthService(
       reportActive()
       val bestBlock = blockchain.getBestBlock()
       val response: ServiceResponse[GetWorkResponse] =
-        getOmmersFromPool(bestBlock.hash).zip(getTransactionsFromPool).map { case (ommers, pendingTxs) =>
+        Task.parZip2(getOmmersFromPool(bestBlock.hash), getTransactionsFromPool()).map { case (ommers, pendingTxs) =>
```
biandratti, line 553: IMHO I like to use "for comprehension" more than the "zip" function. What do you think?

lemastero: parZip2 is running actions in parallel, for comprehension enforce sequential behavior. So probably parZip is better.

biandratti: excelent!

src/main/scala/io/iohk/ethereum/jsonrpc/AkkaTaskOps.scala
lemastero: Changed to `Task.deferFuture`: https://github.com/input-output-hk/mantis/pull/756/commits/7494676cc21ff01b75df74207c35f6061c110bbe

biandratti (approved):

mmrozek (approved):
LGTM!

PR #757, opened 2020-10-23
[ETCM-185] Closing db

biandratti:
# Description

We were using a volatile variable called "isClosed". When the node is shutdown, close the database, and the variable has the value "true". So usually when it receives a new request, this request validates the value of "isClosed" and throws the exception "This RocksDbDataSource has been closed".
I reproduced this error message when I used regular sync with network "mordor". And RocksDbDataSource receives at the same time a “close” request and “get” (or “update”) request for example. 

Today, before validating the ReentrantReadWriteLock variable, we validated a volatile variable called "isClosed".
So, we could have a race condition problem. For example when calling the function "update" and "close" at the same time, both validate the variable "isClosed". But at this moment,  the variable dbLock could be taken firstly for the function "close". This function closes the database and then unlocks the dbLock variable. Then the update function resumes but the database is closed.

# Interrogant

**We decided to change the behavior of variable ReentrantReadWriteLock “dbLock” from the “update” function. We began to use writeLock in place of readLock. Eventually if we detect a loss of performance for these changes, we recommend reverting this behavior and use readLock again.**

# Proposed Solution

- I changed the order to avoid race conditions.
- When the system applies shutdown, finishes the actor system and then closes the database. Usually when finalizing the actors some requests that these started remain pending without finishing. So we can see the exception “RocksDbDataSourceClosedException.
I tried with different strategies in the coordinated-shutdown of akka, and got the same result. So I decided as a workaround to change the order of shutdown, and in the end close the database. In this way the exception in my tests disappeared.

**In a future we could work in a strategy with supervisors and a strategy of coordinated-shutdown from actors system**


src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -255,7 +255,7 @@ class BlockchainImpl(
   }
 
   override def getBestBlock(): Block =
-    getBlockByNumber(getBestBlockNumber()).get
+    getBlockByNumber(getBestBlockNumber()).getOrElse(throw new RuntimeException("block not found"))
```
mirkoAlic, line 258: hmm...

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -37,16 +35,18 @@ class RocksDbDataSource(
     * @return the value associated with the passed key.
     */
   override def get(namespace: Namespace, key: Key): Option[Value] = {
-    assureNotClosed()
-    RocksDbDataSource.dbLock.readLock().lock()
+    dbLock.readLock().lock()
```
lemastero, line 38: Is the change from global lock: `RocksDbDataSource.dbLock.readLock().lock()`
to local specific to class instance: `dbLock.readLock().lock()`
Intended. Here and below?

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -37,16 +35,18 @@ class RocksDbDataSource(
...
     try {
+      assureNotClosed()
       Option(db.get(handles(namespace), readOptions, key.toArray))
     } catch {
+      case ex: RocksDbDataSourceClosedException =>
+        throw ex
       case NonFatal(e) =>
         logger.error(s"Not found associated value to a namespace: $namespace and a key: $key, cause: {}", e.getMessage)
-        throw new RuntimeException(e)
+        throw e
```
lemastero, line 47: 👍

mirkoAlic:
> Interrogant
> update function, need dbLock.readLock() or dbLock.writeLock() ?
IMHO, `dbLock.writeLock` is the "proper" way, but tbh, I'm not sure if is not going to raise some kind of bottleneck. I suggest to apply the change, but also add a good comment about the reasoning behind it. Eventually if we detect an issue we could rollback the change. WDYT?

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -160,10 +164,10 @@ class RocksDbDataSource(
     * This function closes the DataSource, without deleting the files used by it.
     */
   override def close(): Unit = {
-    logger.debug(s"About to close DataSource in path: ${rocksDbConfig.path}")
+    logger.info(s"About to close DataSource in path: ${rocksDbConfig.path}")
```
mirkoAlic, line 167: :+1:

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -175,12 +179,13 @@ class RocksDbDataSource(
       dbOptions.close()
       // 3. Free column families options
       cfOptions.close()
+      logger.info(s"DataSource closed successfully in the path: ${rocksDbConfig.path}")
     } catch {
       case NonFatal(e) =>
         logger.error("Not closed the DataSource properly, cause: {}", e)
```
mirkoAlic, line 185: At least as a temporal solution i think is preferable to have some kind of domain modelling for `DataSource` runtime exceptions, instead of just log and throw exception, in that case we could just throw our custom exceptions that would provide better context about the actual issues and avoid the log

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -24,8 +24,6 @@ class RocksDbDataSource(
     private var handles: Map[Namespace, ColumnFamilyHandle]
 ) extends DataSource {
 
-  private val logger = LoggerFactory.getLogger("rocks-db")
-
   @volatile
   private var isClosed = false
```
lemastero, line 28: WDYT about changing it to https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicReference.html
?

I have good experience with atomic refs. I remember using @volatile a very long time ago in Java and it was no reliable 🤔

kapke: Or even https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicBoolean.html :)

TBH I think that volatile is enough here since updates of this var don't involve a read (and that's the basic problem classes in `java.util.concurrent.atomic` solve - making reads and writes atomically)

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -242,6 +257,10 @@ object RocksDbDataSource {
   case object IterationFinished extends RuntimeException
   case class IterationError(ex: Throwable)
 
+  class RocksDbDataSourceClosedException(val message: String) extends IllegalStateException(message)
+
+  private val logger = LoggerFactory.getLogger("rocks-db")
```
kapke, line 262: Why not `extends Logger` then?

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -242,6 +257,10 @@ object RocksDbDataSource {
   case object IterationFinished extends RuntimeException
   case class IterationError(ex: Throwable)
 
+  class RocksDbDataSourceClosedException(val message: String) extends IllegalStateException(message)
```
kapke, line 260: what about reason field? It's usually very helpful in debugging if the real issue lies a layer below

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
kapke: From what I see it's still the global one used, it's just imported from companion object now

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
KonradStaniec: as we are now writing to this variable only after write lock is in locked state,  this could probably be even a simple var without `@volatile` annotations

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
```diff
@@ -160,10 +158,10 @@ class RocksDbDataSource(
     * This function closes the DataSource, without deleting the files used by it.
     */
   override def close(): Unit = {
-    logger.debug(s"About to close DataSource in path: ${rocksDbConfig.path}")
+    log.info(s"About to close DataSource in path: ${rocksDbConfig.path}")
     assureNotClosed()
```
KonradStaniec, line 162: I wonder if we should not check it after locking ?

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
biandratti: change reverted

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
biandratti: Yes, right

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
biandratti: For this case, it is the same. I changed it anyway, and put the same behavior that the other functions.

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
biandratti: Done

src/main/scala/io/iohk/ethereum/db/dataSource/RocksDbDataSource.scala
biandratti: I think that we need this variable as @volatile.
In case only one thread reads and writes the value of a volatile variable and other threads only read the variable, then the reading threads are guaranteed to see the latest value written to the volatile variable. 
I agree with @kapke, that volatile is enough here since updates of this var don't involve a read. Are we ok with this? Or do you think in another approach?

KonradStaniec: but we only modify this variable from the inside of write lock and read only from inside read lock, which has strictly stronger guarantees than `volatile` variable. So the variable could be simple `var` and we would still have more synchronization than with `volatile` variable.

But to be honest we can leave it `volatile`.  (I am against atomics here as having cas operations inside locks, would really be an overkill)

mirkoAlic (approved):
LGTM!

KonradStaniec (approved):

PR #758, opened 2020-10-26
[ETCM-273] Change logging behaviour and increase handshake timeout to…

mmrozek:
… investigate connection issue

# Description

Temporary change handshake timeout to check if rhe testnet is able to receive any message from outside world


src/main/resources/logback.xml
```diff
@@ -56,9 +56,9 @@
         <appender-ref ref="METRICS" />
     </root>
 
-    <logger name="io.iohk.ethereum.network.rlpx.RLPxConnectionHandler" level="INFO" />
+    <logger name="io.iohk.ethereum.network.rlpx.RLPxConnectionHandler" level="DEBUG" />
     <logger name="io.iohk.ethereum.blockchain.sync.SyncController" level="INFO" />
-    <logger name="io.iohk.ethereum.network.PeerActor" level="INFO" />
+    <logger name="io.iohk.ethereum.network.PeerActor" level="DEBUG" />
```
ntallar, line 61: Should we enable this permanently and remove this lines? (as we are setting the DEBUG option above)

ntallar, line 61: We also have some bothering logs of all the messages sent, should we remove them? Wdyt @mirkoAlic @mmrozek ?

mmrozek: I think it should be a temporary change only. When we will have more peers we will be flooded by logs from there

ntallar: I'm not sure how temporal this will be :thinking: we still have this sort of logging on the other project

mirkoAlic: I think we should re-think our log polices, IMHO:
- info: Display logs that are meaningful for someone that runs the client and more a less wants to see what is going on behind the scene.
- debug: Log everything you could
By following those definitions, i think our issue is that PeerActor or network in general is no logging relevant information at  `info` level.

ntallar: I agree, but I think the only thing that we can do about this for now is pay closer attention to the logs changed or added on new PRs (till maybe someone starts using Mantis and complains about something like that). We should probably better define what a client would like to see :thinking:

Back to the question at hand, I don't see the logs of all the messages sent/received to be so useful (we haven't required them till now), or at least we should keep them in a single line for those that don't use the JSON formatting (I don't do so locally)

mmrozek: I agree with @ntallar. I don't remember when we needed logs from PeerActor and when we will be sure after merge that node doesn't receive any message we won't need it anymore also ;)

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

We need this for our cluster, so the remainder comments will be addressed on a new PR

PR #759, opened 2020-10-26
[ETCM-275] Async node processing and downloader removal

KonradStaniec:
# Description

Changes  how nodes are processed and removes StateDownloader.

Speed of processing of each batch of nodes is highly volatile i.e at the beginning of sync it is quite fast and at the end it is quite slow, to keep the balance between number of responses queued to process in the presence of volatile number of peers, it is necessary to mark peer as active through whole mpt node request life cycle i.e 

1. ask for nodes
2. hash and verify delivered nodes
3. insert valid nodes into the running trie scheduler

Only  after this whole cycle peer is marked as free to handle another request. That way we can achieve optimal throupout through whole state sync. It also guarantees that our depth first descent won't goes too much in breadth.

Such design makes separate `StateDownloader` not necessary and even troublesome as it requires to sync up state between two actors. 

# Testing
I had already synced it to Mainnet few times.


src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
```diff
@@ -511,6 +511,78 @@ class SyncControllerSpec extends AnyFlatSpec with Matchers with BeforeAndAfter w
...
+          case EtcPeerManagerActor.GetHandshakedPeers =>
+            sender ! handshakedPeers
+            this
+
+          case SendMessage(msg: GetBlockHeadersEnc, peer) =>
+            val underlyingMessage = msg.underlyingMsg
+            if (underlyingMessage.maxHeaders == 1) {
+              // pivot block
+              sender ! MessageFromPeer(BlockHeaders(Seq(pivotHeader)), peer)
+              this
```
mmrozek, line 535: Minor: You don't mutate state, so you could move `this` after the if statement. It will simplify a logic a little bit

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
```diff
@@ -97,73 +167,298 @@ class SyncStateSchedulerActor(downloader: ActorRef, sync: SyncStateScheduler, sy
...
+      restartRequester: ActorRef
+  ): Unit = {
+    log.debug("Starting request sequence")
+    sync.persistBatch(currentState, targetBlock)
+    restartRequester ! WaitingForNewTargetBlock
+    context.become(idle(currentStats.addSaved(currentState.memBatch.size)))
+  }
+  import akka.pattern.pipe
+
+  // scalastyle:off
```
mmrozek, line 220: 😢

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
```diff
@@ -202,11 +509,225 @@ object SyncStateSchedulerActor {
...
   case class MissingNodes(missingNodes: List[SyncResponse], downloaderCapacity: Int)
 
   case class BloomFilterResult(res: BloomFilterLoadingResult)
 
+  sealed trait RequestResult
+  case class RequestData(nodeData: NodeData, from: Peer) extends RequestResult
+  case class RequestFailed(from: Peer, reason: String) extends RequestResult
+
+  sealed trait ProcessingError
+  case class Critical(er: CriticalError) extends ProcessingError
```
mmrozek, line 522: What is the difference between `Critical` and `DownloaderError` with `critical = true`?

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
```diff
@@ -2,33 +2,80 @@ package io.iohk.ethereum.blockchain.sync
...
+
+  def handleRequestResults: Receive = {
+    case ResponseReceived(peer, nodeData: NodeData, timeTaken) =>
+      log.info("Received {} state nodes in {} ms", nodeData.values.size, timeTaken)
+      context unwatch (sender())
+      self ! RequestData(nodeData, peer)
+
+    case PeerRequestHandler.RequestFailed(peer, reason) =>
+      context unwatch (sender())
+      log.debug(s"Request failed to peer {} due to {}", peer.id, reason)
```
kapke, line 73: Minor: `"Request to peer {} failed due to {}"` sounds a bit better IMO

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
```diff
@@ -97,73 +167,298 @@ class SyncStateSchedulerActor(downloader: ActorRef, sync: SyncStateScheduler, sy
...
-              newState.numberOfPendingRequests,
-              newState.queue.size()
+      syncInitiator: ActorRef,
+      nodesToProcess: Queue[RequestResult],
+      processing: Boolean,
+      restartRequested: Option[ActorRef]
+  ): Receive = handleCommonMessages orElse handleRequestResults orElse {
+    case Sync if currentState.numberOfPendingRequests > 0 && restartRequested.isEmpty =>
+      val freePeers = getFreePeers(currentDownloaderState)
+      nodesToProcess.dequeueOption match {
```
kapke, line 233: minor: since `freePeers` is always used with emptiness check this could be changed into:
`(nodesToProcess.dequeueOption, NonEmptyList.fromList(freePeers)) match {
//rest of code
`
Nice bonus of that approach:
 - no need to use `fromListUnsafe`
 - exhaustiveness checks will work (using `if` in `case` disabled exhaustiveness checker for given match expression)

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
```diff
@@ -97,73 +167,298 @@ class SyncStateSchedulerActor(downloader: ActorRef, sync: SyncStateScheduler, sy
...
+              currentState,
+              currentDownloaderState,
+              syncConfig.nodesPerRequest
             )
 
-            val (missing, state2) = sync.getMissingNodes(newState, downloaderCap)
+          requests.foreach(req => requestNodes(req))
+          processNodes(newState, currentStats, newDownloaderState, nodes).pipeTo(self)
+          context.become(
+            syncing(
```
kapke, line 251: that syncing handler tracks quite a bit of state now, maybe it makes sense to extract it to some class?

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
```diff
@@ -2,33 +2,80 @@ package io.iohk.ethereum.blockchain.sync
...
+    handshakedPeers.collect {
+      case (peer, _) if !state.activeRequests.contains(peer.id) && !isBlacklisted(peer.id) => peer
+    }
+  }
+
+  private def requestNodes(request: PeerRequest): ActorRef = {
+    log.info(s"Requesting ${request.nodes.size} from peer ${request.peer}")
+    implicit val s = scheduler
+    val handler = context.actorOf(
+      PeerRequestHandler.props[GetNodeData, NodeData](
```
kapke, line 53: Does it make sense to extract a function wrapping that actor creation and errors handling into a `Future`? Basically something merging `requestNodes` and `handleRequestResults` into a function like
```
def PeerRequesthandler.makeRequest[Req, Res](peer, message: Req): Future[Either[RequestFailure|Timeout, Res]] = {}
```

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
```diff
@@ -97,73 +167,298 @@ class SyncStateSchedulerActor(downloader: ActorRef, sync: SyncStateScheduler, sy
     }
   }
 
-  // scalastyle:off method.length
+  private def processNodes(
+      currentState: SchedulerState,
+      currentStats: ProcessingStatistics,
+      currentDownloaderState: DownloaderState,
+      requestResult: RequestResult
+  ): Future[ProcessingResult] = {
```
kapke, line 175: Does it make sense then to write this function without `Future` and wrap into Future at call site?

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
```diff
@@ -202,11 +509,225 @@ object SyncStateSchedulerActor {
...
+
+  sealed trait ResponseProcessingResult
+
+  case object UnrequestedResponse extends ResponseProcessingResult
+
+  case object NoUsefulDataInResponse extends ResponseProcessingResult
+
+  case class UsefulData(responses: List[SyncResponse]) extends ResponseProcessingResult
+
+  final case class DownloaderState(
```
kapke, line 546: Should it go to a separate file now? This one got quite big already.

src/test/scala/io/iohk/ethereum/blockchain/sync/StateSyncSpec.scala
```diff
@@ -121,7 +120,7 @@ class StateSyncSpec
         stateStorage = storages.stateStorage
       ) {
         override def mptStateSavedKeys(): Observable[Either[IterationError, ByteString]] = {
-          Observable.repeatEvalF(Task(Right(ByteString(1)))).takeWhile(_ => !loadingFinished)
+          Observable.repeat(Right(ByteString(1))).takeWhile(_ => !loadingFinished)
```
kapke, line 123: Why not `Observable.interval(1.ms).map(_ => Right(ByteString(1)).takeWhile(_ => !loadingFinished)`?

I'm not very familiar with monix's internals, but I can imagine that just `repeat` gives not much time for other stuff on processing thread.

kapke (commented):
First batch of comment, more will come tomorrow

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
KonradStaniec: Criticial stop the sync entierly as the trie is malformed for some reason, and `DownloaderError ` wihth critical only blacklist peer. I will change the naming here as it can be confusing

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
```diff
@@ -2,33 +2,80 @@ package io.iohk.ethereum.blockchain.sync
...
 import scala.concurrent.duration._
 
-class SyncStateSchedulerActor(downloader: ActorRef, sync: SyncStateScheduler, syncConfig: SyncConfig)
-    extends Actor
+class SyncStateSchedulerActor(
+    sync: SyncStateScheduler,
+    val syncConfig: SyncConfig,
+    val etcPeerManager: ActorRef,
+    val peerEventBus: ActorRef,
+    val scheduler: akka.actor.Scheduler
```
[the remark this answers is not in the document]
KonradStaniec: I would rather not do that, as this name is expected in both traits i.e `BlacklistSupport` and `PeerListSupport`, and changing it will probably introduce a lot of noise into pr.

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
KonradStaniec: great suggestion!

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
KonradStaniec: yep it will probably make code clearer

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
KonradStaniec: imo not,  as we will probably end with something weird as `self ? Request` when we will need to specify artificial timeout event hough we would be asking self. Current way is imo more natural, when the requesting actor is creating `PeerRequestHandler` by itself. 

This approach would definitely be good improvement if we would introduce `PeerClient` and then ask it for those requests, but i would do it in separate task which would do the same also for main `FastSync` (it would great step into hiding `makeRequest...` behind some `Network` interface like in the other project)

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
```diff
@@ -43,51 +86,60 @@ class SyncStateSchedulerActor(downloader: ActorRef, sync: SyncStateScheduler, sy
...
-    downloader ! GetMissingNodes(initialMissing)
-    state1
+    self ! Sync
+    initState
   }
 
-  def idle(processingStatistics: ProcessingStatistics): Receive = {
+  def idle(processingStatistics: ProcessingStatistics): Receive = handleCommonMessages orElse {
     case StartSyncingTo(root, bn) =>
       val state1 = startSyncing(root, bn)
```
kapke, line 125: `startSyncing` method is always followed by `SyncSchedulerActorState.initial`. Maybe that call could be part of `startSyncing` method?

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
```diff
@@ -97,81 +149,216 @@ class SyncStateSchedulerActor(downloader: ActorRef, sync: SyncStateScheduler, sy
...
+
+      case ProcessingResult(Left(err)) =>
+        log.debug("Received error result")
+        err match {
+          case Critical(er) =>
+            log.error(s"Critical error while state syncing ${er}, stopping state sync")
+            // TODO we should probably start sync again from new target block, as current trie is malformed or declare
+            // fast sync as failure and start normal sync from scratch
+            context.stop(self)
+          case DownloaderError(newDownloaderState, peer, description, critical) =>
```
kapke, line 337: `s/critical/blacklistable`?

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
```diff
@@ -523,59 +590,54 @@ class SyncControllerSpec extends AnyFlatSpec with Matchers with BeforeAndAfter w
...
+        probeWIthPilot: ActorRef,
+        handshakedPeers: HandshakedPeers,
+        pivotHeader: BlockHeader,
+        blockchainData: BlockchainData,
+        failedReceiptsTries: Int = 0,
+        failedBodiesTries: Int = 0,
+        onlyPivot: Boolean = false,
+        failedNodeRequest: Boolean = false
+    ): Unit = {
+      val sender = TestProbe()
```
kapke, line 627: Why not making it a method on autopilot? I can see convenience argument, but also in this way it would be easy to send a message to probe, which doesn't have autpilot installed

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncStateSchedulerActor.scala
```diff
@@ -43,135 +86,249 @@ class SyncStateSchedulerActor(downloader: ActorRef, sync: SyncStateScheduler, sy
...
-            finalizeSync(newState, targetBlock, syncInitiator)
-          } else {
+      restartRequester: ActorRef
+  ): Unit = {
+    log.debug("Starting request sequence")
+    sync.persistBatch(currentState, targetBlock)
+    restartRequester ! WaitingForNewTargetBlock
+    context.become(idle(currentStats.addSaved(currentState.memBatch.size)))
+  }
+  import akka.pattern.pipe
```
kapke, line 194: very minor - I'd prefer to have this import either locally within method or on top of the file

kapke (approved):
LGTM! If needed - I can try to sync to mainnet over the weekend to test.

mmrozek (approved):
LGTM!

KonradStaniec:
@kapke any additional syncing testing is appreciated (for now, I have tested it Locally and and on EC2 machine)

PR #760, opened 2020-10-26
[ETCM-247] Fix Fetcher updates

mirkoAlic:
# Description

According to https://jira.iohk.io/browse/ETCM-247, the desyncronization between importer and fetcher could cause several issues. 

The goal of this PR is to provide an easy fix in order to keep fetcher state updated in case a node is only importing throw new mined blocks, or new checkpoint blocks.

Also, notice that previous to this PR, if a node(node2) wants to sync with other(node1), requires that its peer(node 1) is continuously generating blocks, if is not the case is going to catch up with only the first `n` number of blocks (according to `block-headers-per-request`). By subscribing to new block headers, now fetcher can update its vision about know top and react to this.
^ issue: [Block downloading doesn't start without any new block received from the network ](https://jira.iohk.io/browse/ETCM-248)

# Proposed Solution

- Fetcher is updating last block and know top(if is the case) if a mined or checkpointed block is imported
- Fetch is subscribed to new block headers
 
# Testing 

- Regular sync IT tests should passed.
- Previous tests should work too.
- Try to reproduce the manual scenario described over ETCM-247

# TODO

Add tests for fetcher once ETCM-211 got merged (use code: ETCM-248)


src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
```diff
@@ -9,60 +9,77 @@ import org.scalatest.matchers.should.Matchers
...
+          _ <- peer2.connectToPeers(Set(peer1.node))
+          _ <- peer2.waitForRegularSyncLoadLastBlock(blockNumer)
+        } yield {
+          assert(peer1.bl.getBestBlock().hash == peer2.bl.getBestBlock().hash)
+        }
+    }
+
+    "given a previously mined blockchain" in customTestCaseResourceM(FakePeer.start2FakePeersRes()) {
+      case (peer1, peer2) =>
+        val blockHeadersPerRequest = 200
```
ntallar, line 36: Should we use here `peer2.syncConfig.blockHeadersPerRequest`?

src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
```diff
@@ -9,60 +9,77 @@ import org.scalatest.matchers.should.Matchers
...
-      _ <- peer2.importBlocksUntil(blockNumer)(IdentityUpdate)
-      _ <- peer2.startRegularSync().delayExecution(50.milliseconds)
+      _ <- peer1.startRegularSync()
+      _ <- peer2.startRegularSync()
       _ <- peer1.importBlocksUntil(blockNumer)(IdentityUpdate)
-      _ <- peer1.startRegularSync().delayExecution(50.milliseconds)
-      _ <- peer2.mineNewBlock(10)(IdentityUpdate).delayExecution(500.milliseconds)
-      _ <- peer2.mineNewBlock(10)(IdentityUpdate).delayExecution(500.milliseconds)
-      _ <- peer2.mineNewBlock(10)(IdentityUpdate).delayExecution(500.milliseconds)
+      _ <- peer2.importBlocksUntil(blockNumer)(IdentityUpdate)
```
ntallar, line 77: Isn't it prone to errors to have `importBlocksUntil` after starting the regular sync? That function goes directly into the db... I'm not sure how that could affect how syncing behaves

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -183,6 +186,23 @@ class BlockFetcher(
...
+    case MessageFromPeer(BlockHeaders(headers), _) =>
+      headers.lastOption.map { bh =>
+        log.debug(s"Candidate for new top at block ${bh.number}, current know top ${state.knownTop}")
+        val newState = state.withPossibleNewTopAt(bh.number)
+        fetchBlocks(newState)
+      }
+    //keep fetcher state updated in case new checkpoint block or mined block was imported
+    case LastBlockChanged(blockNr) => {
+      log.debug(s"New last block $blockNr imported from the inside")
+      val newState = state.withLastBlock(blockNr).withPossibleNewTopAt(blockNr)
```
ntallar, line 201: Last block seems to also involve the [last headers fetched](https://github.com/input-output-hk/mantis/blob/70c8dc5068066e9a04cbea82d77e505ccb91fb1b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala#L60), what if we are already in progress of fetching blocks from the network with block number higher than the one imported (as it's likely the case for checkpoint blocks)? Doesn't this break the consistency of the fetcher?

Maybe the new last block should be `blockNr.max(state.lastBlock)` or sth like that

ntallar (changes requested):

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
mirkoAlic: good catch

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -232,6 +255,9 @@ class BlockImporter(
           case ChainReorganised(oldBranch, newBranch, totalDifficulties) =>
             updateTxPool(newBranch, oldBranch)
             broadcastBlocks(newBranch, totalDifficulties)
+            if (informFetcherOnLastBlockChanged) {
```
kapke: Minor - after changes introduced in ETCM-175 (on develop already), it should be enough to call fetcher from `RegularSync` actor after receiving update about imported block.

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

PR #762, opened 2020-10-27
[Fix] Non checkpoint blocks after ECIP1097 should have HefPostEcip1097 extra fields

ntallar:
# Description

Fixes how extra fields from blocks are generated after ECIP1097 activation.

# Proposed Solution

We were currently not considering ECIP1097 (checkpoiting) when generating blocks, so they were still generated with HefECIP1098 (treasury), and no checkpoint field then. This made all post ECIP1097 blocks generated to marked as invalid.

# Scenario that failed

Configure Mantis with:
- Mocked mining on
- ECIP1098 block number: 5
- ECIP1097 block number: 10

Start and mine more than 10 blocks, on develop, the block 10 generated would fail

# Testing

- Unit tests added to block generator
- Follow the scenario described above, it should no longer fail

### DIFF
--- a/src/main/resources/chains/etc-chain.conf
+++ b/src/main/resources/chains/etc-chain.conf
@@ -85,6 +85,7 @@
 
   # Checkpointing fork block number
   # https://ecips.ethereumclassic.org/ECIPs/ecip-1097
+  # Has to be equal or greater than ecip1098-block-number
   ecip1097-block-number = "1000000000000000000"
 
   # DAO fork configuration (Ethereum HF/Classic split)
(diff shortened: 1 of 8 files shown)


pslaski (approved):

pslaski (changes requested):
I tried the scenario when isECIP1097 is activated, but not isECIP1098 and mocked mined block is not passing a validation. The problem is `validateExtraFields` method which assumes that isECIP1097 is always activated with isECIP1098

ntallar:
For now ECIP1098 has to always be activated before ECIP1097 and we are assuming for now that that will be the behaviour on any testnet used, as was described here: #709

I'll add a comment on the ECIPs configuration on that assumption

pslaski (approved):

rtkaczyk (approved):

PR #764, opened 2020-10-29
[ETCM-109] ECIP-1099 implementation

mmrozek:
# Description

https://ecips.ethereumclassic.org/ECIPs/ecip-1099

### DIFF
--- a/src/main/resources/chains/mordor-chain.conf
+++ b/src/main/resources/chains/mordor-chain.conf
@@ -87,6 +87,11 @@
   # https://ecips.ethereumclassic.org/ECIPs/ecip-1097
   ecip1097-block-number = "1000000000000000000"
 
+
+  # Epoch calibration block number
+  # https://ecips.ethereumclassic.org/ECIPs/ecip-1099
+  ecip1099-block-number = "2520000"
+
   # DAO fork configuration (Ethereum HF/Classic split)
   # https://blog.ethereum.org/2016/07/20/hard-fork-completed/
   dao = null
--- a/src/main/scala/io/iohk/ethereum/consensus/ethash/EthashMiner.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/ethash/EthashMiner.scala
@@ -56,12 +56,12 @@ class EthashMiner(
 
   def processMining(): Unit = {
     val parentBlock = blockchain.getBestBlock()
-    val epoch = EthashUtils.epoch(parentBlock.header.number.toLong + 1)
-
+    val blockNumber = parentBlock.header.number.toLong + 1
+    val epoch = EthashUtils.epoch(blockNumber, blockCreator.blockchainConfig.ecip1099BlockNumber.toLong)
     val (dag, dagSize) = (currentEpoch, currentEpochDag, currentEpochDagSize) match {
       case (Some(`epoch`), Some(dag), Some(dagSize)) => (dag, dagSize)
       case _ =>
-        val seed = EthashUtils.seed(epoch)
+        val seed = EthashUtils.seed(blockNumber)
         val dagSize = EthashUtils.dagSize(epoch)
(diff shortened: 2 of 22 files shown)


src/main/scala/io/iohk/ethereum/consensus/ethash/EthashMiner.scala
```diff
@@ -56,12 +56,12 @@ class EthashMiner(
 
   def processMining(): Unit = {
     val parentBlock = blockchain.getBestBlock()
-    val epoch = EthashUtils.epoch(parentBlock.header.number.toLong + 1)
-
+    val blockNumber = parentBlock.header.number.toLong + 1
+    val epoch = EthashUtils.epoch(blockNumber, blockCreator.blockchainConfig.ecip1099BlockNumber.toLong)
```
kapke, line 60: why not accepting `BigInt` there? I mean - if domain uses `BigInt` for block numbers, why shouldn't miner?

kapke (approved):
Overall looks good to me!

src/main/scala/io/iohk/ethereum/consensus/ethash/EthashMiner.scala
mmrozek: Because we are using only `Long`s in `EthashUtils`. I am not sure if we do that because of performance or sth else, but I don't want to change that in this pr.

src/main/resources/chains/mordor-chain.conf
```diff
@@ -87,6 +87,11 @@
   # https://ecips.ethereumclassic.org/ECIPs/ecip-1097
   ecip1097-block-number = "1000000000000000000"
 
+
+  # Epoch calibration block number
+  # https://ecips.ethereumclassic.org/ECIPs/ecip-1099
+  ecip1099-block-number = "2520000"
```
jmendiola222, line 93: this has already been reached on Mordor, right? Have we synced and try it there already?

mmrozek: Good point. I will sync with the Mordor before the merge

src/main/scala/io/iohk/ethereum/consensus/ethash/validators/EthashBlockHeaderValidator.scala
```diff
@@ -41,11 +42,11 @@ class EthashBlockHeaderValidator(blockchainConfig: BlockchainConfig)
...
-    def getPowCacheData(epoch: Long): PowCacheData = {
+    def getPowCacheData(epoch: Long, seed: ByteString): PowCacheData = {
       Option(powCaches.get(epoch)) match {
         case Some(pcd) => pcd
         case None =>
-          val data = new PowCacheData(cache = EthashUtils.makeCache(epoch), dagSize = EthashUtils.dagSize(epoch))
+          val data = new PowCacheData(cache = EthashUtils.makeCache(epoch, seed), dagSize = EthashUtils.dagSize(epoch))
 
           val keys = powCaches.keySet().asScala
           val keysToRemove = keys.toSeq.sorted.take(keys.size - MaxPowCaches + 1)
```
KonradStaniec, line 52: I wonder if it won't make problems during transition. i.e

1. lets say we have epoch `358` and `359` here. 
2. we enter fork block `ecip-1099` so next epoch is `180` instead of `360`
3. so we put dag cache for this epoch in cache, and we have there dag for epoch `180` and `359`  (as we have limit of 2 dag caches max, and we always removes oldest one)
4. after some blocks we arive at epoch `181`. And now we will remove dag cache for epoch `180` instead on `359` .  
It means dag dag cache for epoch `359` will be with us for long time (until we get to new epoch `360`), and if rollbacks happen accross new epoch boundaries we will need to recreate dag caches. (we have one stale in there, and we have max size of 2 dag caches )

mmrozek:
@jmendiola222, I synced with the Mordor top

KonradStaniec (approved):

PR #765, opened 2020-10-29
[FIX] Fix the way AsyncConfig is built

kapke:
# Description

Async config was built improperly causing to node to fail

# Testing

Run the node - it should not break now.


KonradStaniec (approved):

mmrozek (approved):
lgtm

PR #767, opened 2020-10-29
[ETCM-284] Pay no money to coinbase address when processing checkpoint blocks

pslaski:
# Description

All checkpoint blocks will currently be marked as invalid, as:

the block state root is set as the same as the parent
the resulting state root includes money provided to the coinbase address

# Proposed Solution

The solution is a small optimization - block with checkpoint is skipping execution of transactions and paying the reward

- block with checkpoint is not executing transactions, rewards and after execution validation
- [bonus] added information about treasury and checkpoint to the json BlockResponse (I found it useful for testing)

# Testing

unit tests + local tests with mocked miner and QA endpoints

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/BlockResponse.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/BlockResponse.scala
@@ -1,7 +1,10 @@
 package io.iohk.ethereum.jsonrpc
 
 import akka.util.ByteString
-import io.iohk.ethereum.domain.{Block, BlockHeader, BlockBody}
+import io.iohk.ethereum.crypto.ECDSASignature
+import io.iohk.ethereum.domain.{Block, BlockBody, BlockHeader}
+
+case class CheckpointResponse(signatures: Seq[ECDSASignature], signers: Seq[ByteString])
 
 case class BlockResponse(
     number: BigInt,
@@ -21,6 +24,8 @@ case class BlockResponse(
     gasLimit: BigInt,
     gasUsed: BigInt,
     timestamp: BigInt,
+    checkpoint: Option[CheckpointResponse],
+    treasuryOptOut: Option[Boolean],
     transactions: Either[Seq[ByteString], Seq[TransactionResponse]],
     uncles: Seq[ByteString]
 )
@@ -41,6 +46,12 @@ object BlockResponse {
       else
         Left(block.body.transactionList.map(_.hash))
 
+    val checkpoint = block.header.checkpoint.map { checkpoint =>
+      val signers = checkpoint.signatures.flatMap(_.publicKey(block.header.parentHash))
+
+      CheckpointResponse(checkpoint.signatures, signers)
+    }
+
     BlockResponse(
       number = block.header.number,
       hash = if (pendingBlock) None else Some(block.header.hash),
@@ -59,6 +70,8 @@ object BlockResponse {
       gasLimit = block.header.gasLimit,
       gasUsed = block.header.gasUsed,
       timestamp = block.header.unixTimestamp,
+      checkpoint = checkpoint,
+      treasuryOptOut = block.header.treasuryOptOut,
       transactions = transactions,
       uncles = block.body.uncleNodesList.map(_.hash)
     )
--- a/src/test/scala/io/iohk/ethereum/checkpointing/CheckpointingTestHelpers.scala
+++ b/src/test/scala/io/iohk/ethereum/checkpointing/CheckpointingTestHelpers.scala
@@ -2,43 +2,10 @@ package io.iohk.ethereum.checkpointing
 
 import akka.util.ByteString
 import io.iohk.ethereum.crypto.ECDSASignature
-import io.iohk.ethereum.domain.BlockHeader.HeaderExtraFields.HefPostEcip1097
-import io.iohk.ethereum.domain._
-import io.iohk.ethereum.ledger.BloomFilter
-import org.bouncycastle.crypto.AsymmetricCipherKeyPair
 import io.iohk.ethereum.crypto.ECDSASignatureImplicits.ECDSASignatureOrdering
+import org.bouncycastle.crypto.AsymmetricCipherKeyPair
 
 object CheckpointingTestHelpers {
-  def createBlockWithCheckpoint(
-      parentHeader: BlockHeader,
-      checkpoint: Checkpoint
-  ): Block = {
-    Block(createBlockHeaderWithCheckpoint(parentHeader, checkpoint), BlockBody(Nil, Nil))
-  }
-
-  def createBlockHeaderWithCheckpoint(
-      parentHeader: BlockHeader,
-      checkpoint: Checkpoint
-  ): BlockHeader = {
-    BlockHeader(
-      parentHash = parentHeader.hash,
-      beneficiary = BlockHeader.EmptyBeneficiary,
-      stateRoot = parentHeader.stateRoot,
-      ommersHash = BlockHeader.EmptyOmmers,
-      transactionsRoot = BlockHeader.EmptyMpt,
-      receiptsRoot = BlockHeader.EmptyMpt,
-      logsBloom = BloomFilter.EmptyBloomFilter,
-      difficulty = parentHeader.difficulty,
-      number = parentHeader.number + 1,
-      gasLimit = parentHeader.gasLimit,
-      gasUsed = UInt256.Zero,
-      unixTimestamp = parentHeader.unixTimestamp + 1,
-      extraData = ByteString.empty,
-      mixHash = ByteString.empty,
-      nonce = ByteString.empty,
-      extraFields = HefPostEcip1097(treasuryOptOut = false, checkpoint = Some(checkpoint))
-    )
-  }
 
   def createCheckpointSignatures(
       keys: Seq[AsymmetricCipherKeyPair],
--- a/src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerEthSpec.scala
+++ b/src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerEthSpec.scala
@@ -12,12 +12,8 @@ import io.iohk.ethereum.crypto.kec256
 import io.iohk.ethereum.domain._
 import io.iohk.ethereum.jsonrpc.EthService._
 import io.iohk.ethereum.jsonrpc.FilterManager.LogFilterLogs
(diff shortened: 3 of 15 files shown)


src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
```diff
@@ -51,61 +51,66 @@ class BlockPreparator(
     *            Pay 20% of it to the treasury contract
     *  2. Miner is payed a reward for the inclusion of ommers
     *  3. Ommers's miners are payed a reward for their inclusion in this block
+    *  4. There is no reward for block with checkpoint
     *
     * @param block the block being processed
     * @param worldStateProxy the initial state
     * @return the state after paying the appropriate reward to who corresponds
     */
+  // scalastyle:off method.length
```
rtkaczyk, line 60: Will we ever grow the balls to remove this useless tool altogether?

src/test/scala/io/iohk/ethereum/ledger/BlockExecutionSpec.scala
```diff
@@ -89,6 +90,23 @@ class BlockExecutionSpec extends AnyWordSpec with Matchers with ScalaCheckProper
...
+        val mockValidators = MockValidatorsAlwaysSucceed
+        val newConsensus: TestConsensus = consensus.withVM(vm).withValidators(mockValidators)
+        val blockValidation = new BlockValidation(newConsensus, blockchain, BlockQueue(blockchain, syncConfig))
+        val blockExecution =
+          new BlockExecution(blockchain, blockchainConfig, newConsensus.blockPreparator, blockValidation)
+
+        val (blocks, error) = blockExecution.executeBlocks(List(blockWithCheckpoint), defaultBlockHeader.difficulty)
+
+        blocks.size shouldBe 1
+        blocks.head.block shouldBe blockWithCheckpoint
```
rtkaczyk, line 107: Should we check that world state is unchanged? Can we do that here?

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerEthSpec.scala
```diff
@@ -110,6 +106,48 @@ class JsonRpcControllerEthSpec
...
+    )
+    val response = jsonRpcController.handleRequest(request).futureValue
+
+    val expectedBlockResponse =
+      Extraction.decompose(BlockResponse(blockToRequest, fullTxs = false, totalDifficulty = Some(blockTd)))
+
+    response should haveResult(expectedBlockResponse)
+  }
+
+  it should "handle eth_getBlockByHash request (block with treasuryOut)" in new JsonRpcControllerFixture {
```
rtkaczyk, line 130: I'd prefer to be consistent with our naming so `treasuryOptOut` (besides `treasuryOut` sounds weird). This related to several val names and string occurrences.

src/test/scala/io/iohk/ethereum/ledger/BlockExecutionSpec.scala
pslaski: we can't as all intermediate world states are created and hidden inside components.
But we know that state is unchanged because the previous failing state root validation is running and passing normally (it is not mocked, `MockValidatorsAlwaysSucceed` is the validator of header and body)

src/main/scala/io/iohk/ethereum/jsonrpc/BlockResponse.scala
```diff
@@ -21,6 +24,8 @@ case class BlockResponse(
     gasLimit: BigInt,
     gasUsed: BigInt,
     timestamp: BigInt,
+    checkpoint: Option[CheckpointResponse],
+    treasuryOptOut: Option[Boolean],
```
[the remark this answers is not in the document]
rtkaczyk: Adding new fields to a JSON object should never break compatibility

src/test/scala/io/iohk/ethereum/ledger/BlockExecutionSpec.scala
rtkaczyk: > we can't as all intermediate world states are created and hidden inside components.

But blockchain is not mocked and accessible here. Could we use that?

> But we know that state is unchanged because the previous failing state root validation is running and passing normally (it is not mocked, `MockValidatorsAlwaysSucceed` is the validator of header and body)

I'm not quite sure which part of the code you are referring to

src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
```diff
@@ -51,61 +51,66 @@ class BlockPreparator(
...
-        val worldAfterTreasuryReward =
-          increaseAccountBalance(treasuryAddress, UInt256(treasuryReward))(worldAfterMinerReward)
-
-        log.debug(
-          s"Paying block $blockNumber reward of $minerReward to miner with address $minerAddress" +
-            s"paying treasury reward of $treasuryReward to treasury with address $treasuryAddress"
-        )
-        worldAfterTreasuryReward
-      }
+    if (block.hasCheckpoint) worldStateProxy
```
rtkaczyk, line 65: 1. I wonder if we should move this condition outside to the `prepareBlock` method, so as to avoid running `persist` on an unchanged world state?

2. Since `BlockExecutionSpec` is possibly not the best place to test that the account is unchanged, can we add an appropriate test to `BlockPreparatorSpec`?

src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
```diff
@@ -56,6 +56,7 @@ class BlockPreparator(
     * @param worldStateProxy the initial state
     * @return the state after paying the appropriate reward to who corresponds
     */
+  // scalastyle:off method.length
```
ntallar, line 59: Should we remove this now this method remains unchanged?

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerFixture.scala
```diff
@@ -113,6 +115,11 @@ class JsonRpcControllerFixture(implicit system: ActorSystem)
     unixTimestamp = 0
   )
 
+  val checkpoint = ObjectGenerators.fakeCheckpointGen(2, 5).sample.get
+  val blockWithCheckpoint = CheckpointingTestHelpers.createBlockWithCheckpoint(Fixtures.Blocks.Block3125369.header, checkpoint)
```
ntallar, line 119: I thought we had removed this as it's duplicated with `CheckpointBlockGenerator`

src/test/scala/io/iohk/ethereum/ledger/BlockExecutionSpec.scala
```diff
@@ -89,6 +90,28 @@ class BlockExecutionSpec extends AnyWordSpec with Matchers with ScalaCheckProper
...
+        val blockWithCheckpoint = CheckpointingTestHelpers.createBlockWithCheckpoint(validBlockParentHeader, checkpoint)
+        val treasuryAccountBefore = blockchain.getAccount(blockchainConfig.treasuryAddress, blockWithCheckpoint.number)
+
+        val mockValidators = MockValidatorsAlwaysSucceed
+        val newConsensus: TestConsensus = consensus.withVM(vm).withValidators(mockValidators)
+        val blockValidation = new BlockValidation(newConsensus, blockchain, BlockQueue(blockchain, syncConfig))
+        val blockExecution =
+          new BlockExecution(blockchain, blockchainConfig, newConsensus.blockPreparator, blockValidation)
+
+        val (blocks, error) = blockExecution.executeBlocks(List(blockWithCheckpoint), defaultBlockHeader.difficulty)
```
ntallar, line 105: Should we rename this method to `executeAndValidateBlocks` for consistency?

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerEthSpec.scala
```diff
@@ -110,6 +106,48 @@ class JsonRpcControllerEthSpec
...
+    )
+    val response = jsonRpcController.handleRequest(request).futureValue
+
+    val expectedBlockResponse =
+      Extraction.decompose(BlockResponse(blockToRequest, fullTxs = false, totalDifficulty = Some(blockTd)))
+
+    response should haveResult(expectedBlockResponse)
+  }
+
+  it should "handle eth_getBlockByHash request (block with treasuryOptOut)" in new JsonRpcControllerFixture {
```
ntallar, line 130: It's odd that the test from this file are integration tests, instead of the usual per service and jrc

ntallar (changes requested):
Minor changes and LGTM!

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerEthSpec.scala
pslaski: it's because we have only one JRC service and here we are ~EthJRC tests

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerFixture.scala
rtkaczyk: I think it's my bad that it wasn't removed in the end. But most of its usages have been replaced with CheckpointBlockGenerator so it should be easy to remove now

src/test/scala/io/iohk/ethereum/checkpointing/CheckpointingTestHelpers.scala
```diff
@@ -9,12 +9,6 @@ import org.bouncycastle.crypto.AsymmetricCipherKeyPair
 import io.iohk.ethereum.crypto.ECDSASignatureImplicits.ECDSASignatureOrdering
 
 object CheckpointingTestHelpers {
-  def createBlockWithCheckpoint(
-      parentHeader: BlockHeader,
-      checkpoint: Checkpoint
-  ): Block = {
-    Block(createBlockHeaderWithCheckpoint(parentHeader, checkpoint), BlockBody(Nil, Nil))
-  }
```
rtkaczyk, line 17: Could we also remove the one below?

rtkaczyk (approved):

ntallar (approved):
LGTM! Was able to generate checkpoint blocks without any problem
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

PR #771, opened 2020-11-03
[ETCM-283] Add block bodies validation in block fetcher

mirkoAlic:
## Description
Our current BlockFetcher doesn't do any validation on the BlockHeaders and BlockBodies messages received and just match them.

## Proposed solution
Validate the matching between new bodies received and expected headers, if there isn't a match fetcher will complain about the peer that send those bodies.

## Important Changes Introduced

~`blockchain.sync.regular.BlockFetcherState`  now is coupled to `consensus.validators.std.StdBlockValidator` 
^ Ideally we shouldn't do such thing!~

## TODO
~add test regarding a sub ordered sequence of bodies received~


src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -143,10 +143,15 @@ class BlockFetcher(
           log.debug("Received {} block bodies that will be ignored", bodies.size)
           state.withBodiesFetchReceived
         } else {
-          log.debug("Fetched {} block bodies", bodies.size)
-          state.withBodiesFetchReceived.addBodies(peer, bodies)
+          state.validateBodies(bodies) match {
```
mirkoAlic, line 146: Open question: it is possible to received from the peer empty bodies at this stage? should we complain about that instead of treat it as part of the "happy" path?

kapke: Are you asking about empty "BlockBodies" message? IIRC there's no guarantee that bodies will be fetched from the same peer and peer is free to respond only with part of data (and for sure it happens quite often with headers)

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -101,14 +105,36 @@ case class BlockFetcherState(
...
-    val blocks = matching.zip(bodies).map((Block.apply _).tupled)
+  def validateBodies(receivedBodies: Seq[BlockBody]): Either[String, Seq[Block]] =
+    bodiesAreOrderedSubsetOfRequested(waitingHeaders.toList, receivedBodies)
+      .toRight(
+        "Received unrequested bodies"
+      )
+
+  // Checks that the received block bodies are an ordered subset of the ones requested
+  @tailrec
+  private def bodiesAreOrderedSubsetOfRequested(
```
kapke, line 116: `matchBodiesWithHeaders`?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -101,14 +105,36 @@ case class BlockFetcherState(
...
+      requestedHeaders: Seq[BlockHeader],
+      respondedBodies: Seq[BlockBody],
+      matchedBlocks: Seq[Block] = Nil
+  ): Option[Seq[Block]] =
+    (requestedHeaders, respondedBodies) match {
+      case (Seq(), _ +: _) => None
+      case (_, Seq()) => Some(matchedBlocks)
+      case (header +: remainingHeaders, body +: remainingBodies) =>
+        val doMatch = blockValidator.validateHeaderAndBody(header, body).isRight
```
kapke, line 125: minor: pattern match/fold instead?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -143,10 +145,15 @@ class BlockFetcher(
...
+            case Left(err) =>
+              peersClient ! BlacklistPeer(peer.id, err)
+              state.withBodiesFetchReceived
+            case Right(newBlocks) =>
+              log.debug("Fetched {} block bodies", newBlocks.size)
```
kapke, line 153: I'd probably change things here a bit:
1. add logging about fetched bodies inconditionally
2. add logging about matching result 
3. change call to validate and then do sth into "addBodies", which validates bodies when adding them and results with new state, possible error and maybe information of how many bodies were matched with headers to blocks

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
mirkoAlic: tbh, i copied this method from the other project.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
KonradStaniec: I would maybe add comment here about this fact, that there is no guarantees about bodies being fetched from the same peer, as it can lead to scenario that the legitimate peer will send empty blockbodies response as it is on different branch.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
mirkoAlic: given is a private method, i prefer to describe exactly what is doing.

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
```diff
@@ -124,6 +129,95 @@ class BlockFetcherSpec extends TestKit(ActorSystem("BlockFetcherSpec_System")) w
...
+
+      val getBlockBodiesRequest = GetBlockBodies(chain.map(_.hash))
+      peersClient.expectMsgPF() { case PeersClient.Request(msg, _, _) if msg == getBlockBodiesRequest => () }
+
+      // This response will be invalid given we are using a special validator!
+      val getBlockBodiesResponse = BlockBodies(chain.map(_.body))
+      peersClient.reply(PeersClient.Response(fakePeer, getBlockBodiesResponse))
+
+      peersClient.expectMsgClass(classOf[BlacklistPeer])
+      peersClient.expectMsgPF() { case PeersClient.Request(msg, _, _) if msg == getBlockBodiesRequest => () }
```
kapke, line 156: what about test like:
  - fetch one batch of blocks
  - for the second batch send different bodies (so even proper validator fails)
  - pick blocks and verify that only these from first batch are responded

Basically same issue with these tests - lack of explicit assertions or at least comments, why it's enough to check for certain message/lack of it.

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
```diff
@@ -124,6 +129,95 @@ class BlockFetcherSpec extends TestKit(ActorSystem("BlockFetcherSpec_System")) w
...
+      val chain = BlockHelpers.generateChain(syncConfig.blockHeadersPerRequest, FixtureBlocks.Genesis.block)
+
+      val getBlockHeadersResponse = BlockHeaders(chain.map(_.header))
+      peersClient.reply(PeersClient.Response(fakePeer, getBlockHeadersResponse))
+
+      val getBlockBodiesRequest1 = GetBlockBodies(chain.map(_.hash))
+      peersClient.expectMsgPF() { case PeersClient.Request(msg, _, _) if msg == getBlockBodiesRequest1 => () }
+
+      // It will receive all the requested bodies, but splitted in 2 parts.
+      val (subChain1, subChain2) = chain.splitAt(syncConfig.blockHeadersPerRequest / 2)
```
kapke, line 176: block bodies per request? I know these are same values in tests, but the right config value should be used anyway I think

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
```diff
@@ -124,6 +129,95 @@ class BlockFetcherSpec extends TestKit(ActorSystem("BlockFetcherSpec_System")) w
...
+        GetBlockHeaders(Left(subChain1.last.number + 1), syncConfig.blockHeadersPerRequest, skip = 0, reverse = false)
+      peersClient.expectMsgPF() { case PeersClient.Request(msg, _, _) if msg == getBlockHeadersRequest2 => () }
+
+      val getBlockBodiesRequest2 = GetBlockBodies(subChain2.map(_.hash))
+      peersClient.expectMsgPF() { case PeersClient.Request(msg, _, _) if msg == getBlockBodiesRequest2 => () }
+
+      val getBlockBodiesResponse2 = BlockBodies(subChain2.map(_.body))
+      peersClient.reply(PeersClient.Response(fakePeer, getBlockBodiesResponse2))
+
+      peersClient.expectNoMessage()
```
kapke, line 191: Is there a better way? Maybe try to pick blocks or sth like that?

kapke (dismissed):
Tests need a bit more comments, besides that the code looks good to me!

KonradStaniec (approved):

mirkoAlic:
@kapke i will wait til your pr https://github.com/input-output-hk/mantis/pull/787 is merged in order to polish the tests and adjust my code, given there will be some little overlapping between: `addBodies` <-> `receiveBlocks`.

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
mirkoAlic: The main issue i see with the test you propose is that i would need the real block validator, and that make things more complex. So i will opt for improve current test assertions if you are ok with that.

mirkoAlic: (main issue is current impl of `bodiesAreOrderedSubsetOfRequested`, given i can't use a simple mock validator to match headers and bodies)

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
```diff
@@ -124,6 +125,123 @@ class BlockFetcherSpec extends TestKit(ActorSystem("BlockFetcherSpec_System")) w
...
+      peersClient.reply(PeersClient.Response(fakePeer, getBlockBodiesResponse))
+
+      // Fetcher should blacklist the peer and retry asking for the same bodies
+      peersClient.expectMsgClass(classOf[BlacklistPeer])
+      peersClient.expectMsgPF() { case PeersClient.Request(msg, _, _) if msg == getBlockBodiesRequest => () }
+
+      // Fetcher should not enqueue any new block
+      importer.send(blockFetcher, PickBlocks(syncConfig.blocksBatchSize))
+      importer.ignoreMsg({ case BlockImporter.NotOnTop => true })
+      importer.expectNoMessage()
```
kapke, line 157: :( Can at least the duration be adjusted to not wait whole 5 seconds?

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
kapke: can this `expectNoMessageCall` be removed now?

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
```diff
@@ -124,6 +125,123 @@ class BlockFetcherSpec extends TestKit(ActorSystem("BlockFetcherSpec_System")) w
...
+      peersClient.expectMsgPF() { case PeersClient.Request(msg, _, _) if msg == getBlockHeadersRequest2 => () }
+
+      val getBlockBodiesRequest2 = GetBlockBodies(subChain2.map(_.hash))
+      peersClient.expectMsgPF() { case PeersClient.Request(msg, _, _) if msg == getBlockBodiesRequest2 => () }
+
+      // We receive empty bodies instead of the second part
+      val getBlockBodiesResponse2 = BlockBodies(List())
+      peersClient.reply(PeersClient.Response(fakePeer, getBlockBodiesResponse2))
+
+      peersClient.expectNoMessage()
```
kapke, line 235: ditto

kapke (approved):
The tests look much better now!

PR #773, opened 2020-11-04
[ETCM-193] Make target time between blocks configurable

pslaski:
# Description

Due to OBFT constraints, our internal testnet will have a higher target time between blocks than ETC, so we should make the block rate configurable for changing it on the different environments.

# Proposed Solution
Introduced new Difficulty calculator which will try to compute difficulty to target configured PoW time

- new blockchain configuration key - `pow-target-time` (default is null which means that normal ethash calculator will be used)

### DIFF
--- a/src/ets/scala/io/iohk/ethereum/ets/blockchain/EthashTestBlockHeaderValidator.scala
+++ b/src/ets/scala/io/iohk/ethereum/ets/blockchain/EthashTestBlockHeaderValidator.scala
@@ -1,7 +1,6 @@
 package io.iohk.ethereum.ets.blockchain
 
 import io.iohk.ethereum.consensus.difficulty.DifficultyCalculator
-import io.iohk.ethereum.consensus.ethash.difficulty.EthashDifficultyCalculator
 import io.iohk.ethereum.consensus.ethash.validators.EthashBlockHeaderValidator
 import io.iohk.ethereum.consensus.validators.{ BlockHeaderError, BlockHeaderValid, BlockHeaderValidatorSkeleton }
 import io.iohk.ethereum.domain.BlockHeader
@@ -14,7 +13,7 @@ class EthashTestBlockHeaderValidator(blockchainConfig: BlockchainConfig) extends
   // we need concurrent map since validators can be used from multiple places
   protected val powCaches: java.util.concurrent.ConcurrentMap[Long, PowCacheData] = new java.util.concurrent.ConcurrentHashMap[Long, PowCacheData]()
 
-  protected def difficulty: DifficultyCalculator = new EthashDifficultyCalculator(blockchainConfig)
+  protected def difficulty: DifficultyCalculator = DifficultyCalculator(blockchainConfig)
 
   def validateEvenMore(blockHeader: BlockHeader, parentHeader: BlockHeader): Either[BlockHeaderError, BlockHeaderValid] =
     Right(BlockHeaderValid)
--- a/src/main/resources/chains/etc-chain.conf
+++ b/src/main/resources/chains/etc-chain.conf
@@ -3,6 +3,10 @@
   # 1 - mainnet, 3 - ropsten, 7 - mordor
   network-id = 1
 
+  # Possibility to set Proof of Work target time for testing purposes.
+  # null means that the standard difficulty calculation rules are used
+  pow-target-time = null
+
   # Frontier block number
   frontier-block-number = "0"
 
--- a/src/main/resources/chains/eth-chain.conf
+++ b/src/main/resources/chains/eth-chain.conf
@@ -3,6 +3,10 @@
   # 1 - mainnet, 3 - ropsten, 7 - mordor
   network-id = 1
 
+  # Possibility to set Proof of Work target time for testing purposes.
+  # null means that the standard difficulty calculation rules are used
+  pow-target-time = null
+
   # Frontier block number
   frontier-block-number = "0"
 
--- a/src/main/resources/chains/mordor-chain.conf
+++ b/src/main/resources/chains/mordor-chain.conf
@@ -3,6 +3,10 @@
   # 1 - mainnet, 3 - ropsten, 7 - mordor
   network-id = 7
 
+  # Possibility to set Proof of Work target time for testing purposes.
+  # null means that the standard difficulty calculation rules are used
+  pow-target-time = null
+
   # Frontier block number
   frontier-block-number = "0"
 
--- a/src/main/resources/chains/ropsten-chain.conf
+++ b/src/main/resources/chains/ropsten-chain.conf
@@ -3,6 +3,10 @@
   # 1 - mainnet, 3 - ropsten, 7 - mordor
   network-id = 3
 
+  # Possibility to set Proof of Work target time for testing purposes.
+  # null means that the standard difficulty calculation rules are used
+  pow-target-time = null
+
   # Frontier block number
   frontier-block-number = "0"
 
--- a/src/main/resources/chains/test-chain.conf
+++ b/src/main/resources/chains/test-chain.conf
@@ -3,6 +3,10 @@
   # 1 - mainnet, 7 - mordor
   network-id = 1
 
+  # Possibility to set Proof of Work target time for testing purposes.
+  # null means that the standard difficulty calculation rules are used
+  pow-target-time = null
+
   # Frontier block number
   frontier-block-number = "0"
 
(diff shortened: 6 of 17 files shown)


src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/ConfiguredDifficultyCalculator.scala
```diff
@@ -0,0 +1,36 @@
...
+    *   [l*i, l*(i+1) ) => c = 1-i, difficulty decreases
+    *
+    *  example:
+    *  powTargetTime := 60 seconds
+    *  l := 30 seconds
+    *   [0, 0.5 min)         => difficulty increases
+    *   [0.5 min, 1.5 min)   => difficulty stays the same (the average should be powTargetTime)
+    *   [1.5 min, +infinity) => difficulty decreases
+    */
+  val LowerBoundExpectedRatio: Long = powTargetTime / 2
```
pslaski, line 26: thinking about it now, I did too simple assumption here. WDYT guts, should we configure also the deviation of target time or we should assume that deviation is X % of target time?

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/ConfiguredDifficultyCalculator.scala
```diff
@@ -0,0 +1,36 @@
+package io.iohk.ethereum.consensus.ethash.difficulty
+
+import io.iohk.ethereum.consensus.difficulty.DifficultyCalculator
+import io.iohk.ethereum.domain.BlockHeader
+
+class ConfiguredDifficultyCalculator(powTargetTime: Long) extends DifficultyCalculator {
+
+  val DifficultyBoundDivision: Int = 2048
+  val FrontierTimestampDiffLimit: Int = -99
+  val MinimumDifficulty: BigInt = 131072
```
ntallar, line 10: Should this constants be shared in some way with EthashDifficultyCalculator?

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/ConfiguredDifficultyCalculator.scala
ntallar, line 26: Why `powTargetTime / 2` here? Doesn't that give the ranges:
```
[0, time/2) = [0, 22.5)       => difficulty increases
[time/2, time) = [22.5, 45)   => difficulty stays the same
[time, 3/2 time) = [45, 67.5) => difficulty decreases
```
Meaning that the block time would be around: 37 seconds?

On the ETC calculator they are using `3/2 * powTargetTime` (LowerBoundExpectedRatio=10, powTargetTime=15) as their LowerBoundExpectedRatio, shouldn't we use the the same?

ntallar: Hmm I'm not sure if this deviation will bother us too much, neither if it's easy to change this formulas without screwing up... I wouldn't mind leaving it as is for now

ntallar (changes requested):

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/ConfiguredDifficultyCalculator.scala
```diff
@@ -0,0 +1,34 @@
...
+    *   [l*i, l*(i+1) ) => c = 1-i, difficulty decreases
+    *
+    *  example:
+    *  powTargetTime := 45 seconds
+    *  l := 30 seconds
+    *   [0, 0.5 min)         => difficulty increases
+    *   [0.5 min, 1 min)   => difficulty stays the same (the average should be powTargetTime)
+    *   [1 min, +infinity) => difficulty decreases
+    */
+  private val LowerBoundExpectedRatio: Long = (powTargetTime / 1.5).toLong
```
rtkaczyk, line 24: A nazi comment: capitalised names are for constants and this is not a constant if it depends on the class parameter

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/ConfiguredDifficultyCalculator.scala
```diff
@@ -0,0 +1,34 @@
+package io.iohk.ethereum.consensus.ethash.difficulty
+
+import io.iohk.ethereum.consensus.difficulty.DifficultyCalculator
+import io.iohk.ethereum.domain.BlockHeader
+
+class ConfiguredDifficultyCalculator(powTargetTime: Long) extends DifficultyCalculator {
```
rtkaczyk, line 6: Not fond of the name, `EthashDiffCalc` is also configured. Maybe `TargetTimeDiffCalc` or simply `QaDiffCalc`?

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/ConfiguredDifficultyCalculator.scala
```diff
@@ -0,0 +1,34 @@
...
+    */
+  private val LowerBoundExpectedRatio: Long = (powTargetTime / 1.5).toLong
+
+  def calculateDifficulty(blockNumber: BigInt, blockTimestamp: Long, parentHeader: BlockHeader): BigInt = {
+    val timestampDiff = blockTimestamp - parentHeader.unixTimestamp
+
+    val x: BigInt = parentHeader.difficulty / DifficultyBoundDivision
+    val c: BigInt = math.max(1 - (timestampDiff / LowerBoundExpectedRatio), FrontierTimestampDiffLimit)
+
+    MinimumDifficulty.max(parentHeader.difficulty + x * c)
```
rtkaczyk, line 32: How did we arrive at this formula? I would have expected it to be parameterised by something like:
- mu - the expected block time
- sigma - std deviation, where we don't change the diff
- delta - rate of change outside of std deviation range

Perhaps something like: https://en.wikipedia.org/wiki/Beta_distribution could be used. The formula we have is eerily similar ETH formula, but I think we are allowed more freedom with this, no?

Disclaimer: these are very rough proposals and are not to be taken word-for-word. But I think something along those lines would give us more control. It's not a blocker if such control is not required.

ntallar (changes requested):

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/ConfiguredDifficultyCalculator.scala
ntallar: (I don't like QaDiffCalc as this will be used on our testnet and I'm not sure for how long, not just testing purposes)

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/ConfiguredDifficultyCalculator.scala
ntallar: This was inspired by what I had done on this regard on the other project so I'll share my thoughts. As you said this formula is the ETH formula with simplifications, I didn't want to deviate too much from it just in case I'd break something (for example, if we are more aggressive with when to increase/decrease difficulty it might lead to an divergent time between blocks).

Either way, the only reason I see for maintaining the current formula is so that this doesn't become a blocker for deploying checkpointing on our testnet, if that isn't at risk we could go for another better one

(of course, if this formula doesn't work as expected we should definitely change it!)

pslaski: I would go now with the current formula for three reasons: 1. we need this fast for testnet deployment 2. It is the same formula used in other project and its upgraded version is in mantis now. 3. Developing new formula only for testnet without having in mind using it in mainnet is not worth a time now imho.

Maybe we can think about it in the future? (new task)

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

src/main/scala/io/iohk/ethereum/consensus/ethash/difficulty/ConfiguredDifficultyCalculator.scala
ntallar: I would try it out on the testnet and only create a follow-up task if we detect there that's not enough (or that it would be good to have something better)

lemastero (approved):

PR #775, opened 2020-11-04
EthService performance fix for AtomicReference update loops

dmitry-worker:
# Description

1. `updateAndGet()` method used to remove obsolete hashes contains `do{ map.filter() } while()` nested loop and has overall O(n^2) complexity.
2. `mapValues` should be avoided as its memory issue is fixed only in scala 2.13. Also, one shouldn't use mapValues if only values needed.
3. `AtomicReference` is redundant, because `compareAndSet`-based method result (in p.1) is not used in code.

# Proposed Solution

Remove `updateAndGet` to save CPU cycles, remove `AtomicReference` since all the changes in `TrieMap` are atomic.
Introduce `TrieMap` to handle concurrent writes and save memory spent on `Map` instances, that were created on each `hashRate` read request (with obsolete elements)


mirkoAlic (approved):
:rocket:

PR #776, opened 2020-11-05
[ETCM-286] Adjustments to rpc tests

mmrozek:
# Description

Fix rpc tests


mirkoAlic:
Maybe we could add this as part of the buildkite flow?

src/rpcTest/scala/io/iohk/ethereum/rpcTest/RpcApiTests.scala
```diff
@@ -899,7 +899,7 @@ class RpcApiTests extends AnyFlatSpec with Matchers with Logger {
     nonExistingFilterChanges.getLogs.asScala.size shouldEqual 0
   }
 
-  it should "eth_newBlockFilter" taggedAs (PrivNet, MainNet) in new ScenarioSetup {
+  it should "eth_newBlockFilter" taggedAs (PrivNet) in new ScenarioSetup {
```
mirkoAlic, line 902: i'm curious, why MainNet is not needed anymore?

mmrozek: The test is flaky when used in such a big network as mainnet. I am not sure if it is because of reorgs, too big response or sth else. Testing `PrivNet` tests the required behaviour so it should be enough

serhii-shymkiv-iohk (approved):
LGTM

PR #777, opened 2020-11-05
Fixed concurrency issue that allowed multiple miners instantiation.

dmitry-worker:
# Description

Concurrent `EthashConsensus.startMiningProcess()` method invocation can lead to several miner actors.
There's no thread mutual exclusion and atomicity between `AtomicReference.get()` and `AtomicReference.set()` calls.

# Proposed Solution

`synchronized` or `ReentrantReadWrite` is recommended to ensure that only one miner is created.
`AtomicReference` is redundant, since no CAS-related methods are used.
Solution assumes that no thread synchronization is needed when miner already exists.


dmitry-worker:
> Good catch, but I think the simpler solution is to use `atomicMiner.getAndUpdate`. WDYT?
Thanks! 

I was actually about to implement one of CompareAndSet-based methods, like `getAndUpdate`. 
But one thing stops me doing that: the side effect that initializes new `Actor` in the `ActorSystem`

While `getAndUpdate` will guarantee us that minerRef equals our newly created miner Actor at some point - it cannot guarantee that no other miners are instantiated as well. So, only mutual exclusion helped me to solve this :(

src/main/scala/io/iohk/ethereum/consensus/ethash/EthashConsensus.scala
```diff
@@ -45,29 +44,31 @@ class EthashConsensus private (
...
   override def sendMiner(msg: MinerProtocol): Future[MinerResponse] = {
     import akka.pattern.ask
-    atomicMiner
-      .get()
+    minerRef
       .map(_.ask(msg).mapTo[MinerResponse])
       .getOrElse(Future.successful(MinerNotExist))
   }
 
+  private[this] val mutex = new Object
```
lemastero, line 58: locks/mutex/semaphores are low level, declarative programming approach to concurrencly. They are hard to maintain and super hard to write correctly.

Would it be possible to use sth like STM here. Maybe Ref from cats effect: https://typelevel.org/cats-effect/concurrency/ref.html

src/main/scala/io/iohk/ethereum/consensus/ethash/EthashConsensus.scala
```diff
@@ -45,29 +44,31 @@ class EthashConsensus private (
     blockchainConfig = blockchainConfig
   )
```
lemastero, line 46: Is it possible to write a test that will show the problem/solution?

PR #778, opened 2020-11-05
Fixed MapCache performance issues and made it thread-safe.

dmitry-worker:
# Description

1. `MapCache` methods were using instance-level synchronization (`this.synchronized`), that impaired its performance
2. `MapCache.clear()` method surprisingly was not synchronized

# Proposed Solution

As soon as required concurrency level for `MapCache` is exclusive write + multiple reads - a `ReentrantLock` was introduced.
Syntactic-sugar helper methods `read` and `write` have wrapped all `MapCache` methods dealing with its data, so now it's completely thread-safe.

Also `FiniteDuration` is now computed only once, using nanotime difference between `lastClear` and `now`


src/main/scala/io/iohk/ethereum/db/cache/MapCache.scala
```diff
@@ -1,53 +1,56 @@
...
+    lastClear = System.nanoTime()
     cache.clear()
   }
 
-  override def shouldPersist: Boolean = {
+  override def shouldPersist: Boolean = read {
     cache.size > config.maxSize || isTimeToClear
   }
 
   private def isTimeToClear: Boolean = {
```
mirkoAlic, line 52: maybe: `threadUnsafeIsTimeToClear`

src/main/scala/io/iohk/ethereum/db/cache/MapCache.scala
```diff
@@ -1,53 +1,56 @@
...
+  @volatile private[this] var lastClear = System.nanoTime()
 
-  override def update(toRemove: Seq[K], toUpsert: Seq[(K, V)]): Cache[K, V] = {
-    this.synchronized {
-      toRemove.foreach(key => cache -= key)
-      toUpsert.foreach(element => cache += element._1 -> element._2)
-    }
+  private[this] val lock = new ReentrantReadWriteLock()
+
+  private[this] def read[R](f: => R): R = {
```
mirkoAlic, line 19: `threadSafeRead` ?

src/main/scala/io/iohk/ethereum/db/cache/MapCache.scala
```diff
@@ -1,53 +1,56 @@
...
+    lock.readLock().lock()
+    try { f } finally lock.readLock().unlock()
+  }
+
+  private[this] def write[R](f: => R): R = {
```
mirkoAlic, line 24: `threadSafeWrite` ?

src/main/scala/io/iohk/ethereum/db/cache/MapCache.scala
```diff
@@ -1,53 +1,56 @@
...
 import io.iohk.ethereum.utils.Config.NodeCacheConfig
 
 import scala.collection.{Seq, mutable}
 import scala.concurrent.duration.FiniteDuration
 
 //TODO EC-492 Investigate more carefully possibility of having read cache in front of db
-// class is not entirely thread safe
-// All updates need to be atomic and visible in respect to get, as get may be called from other threads.
-// Other methods are only called from actor context, and all updates are always visible to them
+// class is now entirely thread safe
```
mirkoAlic, line 12: i think this comment can be removed, and make it explicit at code by renaming the class to: `ThreadSafeMapCache`

src/main/scala/io/iohk/ethereum/db/cache/MapCache.scala
```diff
@@ -1,53 +1,56 @@
...
+
+  private[this] def write[R](f: => R): R = {
+    lock.writeLock().lock()
+    try { f } finally lock.writeLock().unlock()
+  }
+
+  override def update(toRemove: Seq[K], toUpsert: Seq[(K, V)]): Cache[K, V] = write {
+    toRemove.foreach(key => cache -= key)
+    toUpsert.foreach(element => cache += element._1 -> element._2)
     this
```
mirkoAlic, line 32: is `this` ok? :drum:

src/main/scala/io/iohk/ethereum/db/cache/MapCache.scala
dmitry-worker: Hmm.. Cache is something that is shared... I believe its thread safety should be assumed by default. Maybe we shouldn't rename it.

src/main/scala/io/iohk/ethereum/db/cache/MapCache.scala
dmitry-worker: :+1: or maybe `lockForReading`?

src/main/scala/io/iohk/ethereum/db/cache/MapCache.scala
dmitry-worker: Sure, since `update` method is an implementation of `trait SimpleMap` and should return a covariant cache type...

src/main/scala/io/iohk/ethereum/db/cache/MapCache.scala
dmitry-worker: I would agree if it was actually unsafe, but the only variable that's read inside is `volatile` and all the threads get notified of its changes.

src/main/scala/io/iohk/ethereum/db/cache/MapCache.scala
dmitry-worker: Maybe `lockForWriting`, as with reading?

src/main/scala/io/iohk/ethereum/db/cache/MapCache.scala
mirkoAlic: `lockForReading` and `lockForWriting` sound cool! go for it, and i will approve the PR =)

KonradStaniec (commented):
good catch about synchronising `.clear` method. 

I have mixed feelings about `ReentrantReadWriteLock()` in general it is more performant than simpler `synchornised` only in very specific scenarios.  Did you try to do any performance evaluations of this change ? (usual performance benchmark is to setup one master node with 1M block from the network, and then to setup second node to sync up locally only from this master node. This benchmark measures what is most important to use i.e syncing throuput.)

dmitry-worker:
Thanks, @KonradStaniec!

> I have mixed feelings about `ReentrantReadWriteLock()` in general it is more performant than simpler `synchornised` only in very specific scenarios.

Yes, indeed, it outperforms `synchronized` only in a highly concurrent environment. 
In my case - 20 writer threads and more. 

```
MapCache Writers:   1;    Readers:   5.    Elapsed millis:      831
OldCache Writers:   1;    Readers:   5.    Elapsed millis:      326
MapCache Writers:   5;    Readers:  10.    Elapsed millis:     1768
OldCache Writers:   5;    Readers:  10.    Elapsed millis:      948
MapCache Writers:  10;    Readers:  20.    Elapsed millis:     3409
OldCache Writers:  10;    Readers:  20.    Elapsed millis:     2487
MapCache Writers:  20;    Readers:  50.    Elapsed millis:     8619
OldCache Writers:  20;    Readers:  50.    Elapsed millis:     8951
```
I have added my test suite also, feel free to experiment on thread count and test conditions. 
If the thread count is not so big, I can close this PR, and create a new one, with only `clear` method synchronized.

KonradStaniec:
our thread workload looks as follows: at most one writer and many readers (where many is highly dependent on who is using a node). Aalso our main bottlneck when syncing is IO and not any synchronization issues, context switches etc. 

So i would say its enough for this pr to make this class thread safe without introducing readwrite lock.

dmitry-worker:
I have forgotten to add some sync to getValues() to OldCache when I ran the tests. Synchronized lead is no longer that big on a small thread quantity, but in highly concurrent environment, RWLock performance is overwhelming.

```
MapCache Writers:   1;    Readers:   5.    Elapsed millis:      879
OldCache Writers:   1;    Readers:   5.    Elapsed millis:      530
MapCache Writers:   5;    Readers:  10.    Elapsed millis:     1663
OldCache Writers:   5;    Readers:  10.    Elapsed millis:     1574
MapCache Writers:  10;    Readers:  20.    Elapsed millis:     3294
OldCache Writers:  10;    Readers:  20.    Elapsed millis:     3951
MapCache Writers:  20;    Readers:  50.    Elapsed millis:     9726
OldCache Writers:  20;    Readers:  50.    Elapsed millis:    18556
```

PR #782, opened 2020-11-06
[ETCM-301] Fix block preparation

mmrozek:
# Description

1. Change type of  `stateRoot` from `Option` to `ByteString` in `WorldStateProxy` to avoid mistakes
2. Add a parent as a parameter of `prepareBlock` method.

This change should also solve the issue described in [ETCM-329](https://jira.iohk.io/browse/ETCM-329)


src/main/scala/io/iohk/ethereum/ledger/BlockExecution.scala
```diff
@@ -67,7 +67,13 @@ class BlockExecution(
     * @param block the block with transactions to run
     */
   private[ledger] def executeBlockTransactions(block: Block): Either[BlockExecutionError, BlockResult] = {
-    val parentStateRoot = blockchain.getBlockHeaderByHash(block.header.parentHash).map(_.stateRoot)
+    val parentStateRoot =
+      blockchain
+        .getBlockHeaderByHash(block.header.parentHash)
+        .map(_.stateRoot)
+        .getOrElse(
+          throw new IllegalStateException("Cannot find parent")
```
dmitry-worker, line 75: Maybe it would be better to return `Left(BlockExecutionError)` even in that case?

src/snappy/scala/io/iohk/ethereum/snappy/SnappyTest.scala
```diff
@@ -66,9 +66,10 @@ class SnappyTest extends AnyFreeSpec with Matchers with Logger {
...
 
-      case None =>
-        // this seems to discard failures, for better errors messages we might want to implement a different method (simulateBlock?)
-        val result = blockPreparator.prepareBlock(block)
-        Right(result.blockResult.receipts)
+      case None => throw new NotImplementedError()
+//        // this seems to discard failures, for better errors messages we might want to implement a different method (simulateBlock?)
+//        val result = blockPreparator.prepareBlock(block)
+//        Right(result.blockResult.receipts)
+//        FIXME DO WE NEED THAT?
```
KonradStaniec, line 73: Maybe we could create task to delete those snappy tests? (I highly doubt anybody will ever use it)

src/main/scala/io/iohk/ethereum/ledger/BlockExecution.scala
```diff
@@ -67,7 +67,13 @@ class BlockExecution(
     * @param block the block with transactions to run
     */
   private[ledger] def executeBlockTransactions(block: Block): Either[BlockExecutionError, BlockResult] = {
```
KonradStaniec, line 69: maybe it would be possible to pass parent header here as param , to avoid this `blockchain.getBlockHeaderByHash(block.header.parentHash` ?

src/snappy/scala/io/iohk/ethereum/snappy/SnappyTest.scala
mmrozek: Good idea. I just created https://jira.iohk.io/browse/ETCM-349

src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
```diff
@@ -376,12 +376,11 @@ class BlockPreparator(
     }
   }
 
-  def prepareBlock(block: Block): BlockPreparationResult = {
-    val parentStateRoot = blockchain.getBlockHeaderByHash(block.header.parentHash).map(_.stateRoot)
+  def prepareBlock(block: Block, parent: BlockHeader): PreparedBlock = {
```
ntallar, line 379: I think this addresses https://jira.iohk.io/browse/ETCM-329 but only if the mocked miner doesn't use any txs, if you:
1. Mine 30 blocks, i.e. with numbers 1-30
2. Set to mine 30 blocks having block 1 as it's parent
3. Send a tx before step 2. completes

It will fail as in ETCM-329... we had solved that in the other project by passing here the `parentState: InMemoryWorldState` instead of having to fetch it here

src/main/scala/io/iohk/ethereum/ledger/BlockPreparator.scala
```diff
@@ -376,22 +376,30 @@ class BlockPreparator(
...
 
     prepared match {
       case (execResult @ BlockResult(resultingWorldStateProxy, _, _), txExecuted) =>
         val worldToPersist = payBlockReward(block, resultingWorldStateProxy)
         val worldPersisted = InMemoryWorldStateProxy.persistState(worldToPersist)
-        BlockPreparationResult(
+        PreparedBlock(
           block.copy(body = block.body.copy(transactionList = txExecuted)),
           execResult,
           worldPersisted.stateRootHash,
```
ntallar, line 405: Trying the scenario I mentioned I found a bug: the line below should be `worldPersisted`, the stateRootHash seems to be different if not

ntallar (changes requested):
Apart from the below comments, LGTM!

Do you think there's any test that could be added for any of the two issues?

src/test/scala/io/iohk/ethereum/jsonrpc/EthServiceSpec.scala
```diff
@@ -1345,7 +1351,7 @@ class EthServiceSpec
     val fakeWorld = blockchain.getReadOnlyWorldStateProxy(
       None,
       UInt256.Zero,
-      None,
+      ByteString.empty,
```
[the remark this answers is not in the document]
mmrozek: It doesn't matter in this case

mmrozek:
@ntallar I don't think that we are able to add some reliable test to check that. I can't imagine any

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

KonradStaniec (approved):

PR #784, opened 2020-11-06
[Fix] Propagation improvements

ntallar:
# Description

Propagation improvements required for the testnet

# Proposed Solution

- Changed max-xxx-peers configurations to match the other project. Note that the effect of them might be a bit unreliable due to our bad usage of them, this will be improved a bit with #763 
- Increased branch-resolution-request-size to 30 to be more resilient to reorgs
- Increased timestamp of the genesis block and it's difficulty, for the testnet to be closer to the 15 second target time per blocks. Without this, the difficult moved toward that direction, but it would take quite a while to reach there.
- Removed validation of NewBlockHashes message so that know top is always updated when one is received, as is the case with NewBlock messages

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -119,13 +119,13 @@ mantis {
       max-mpt-components-per-message = 200
 
       # Maximum number of peers this node can connect to
-      max-outgoing-peers = 80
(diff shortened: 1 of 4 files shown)


mirkoAlic (approved):
LGTM!

PR #785, opened 2020-11-07
Added some laziness to eliminate a risk of importing invalid blocks

dmitry-worker:
# Description

There's a significant chance of adding blocks prior to any validtion completed on them.
It happens because scala `Future` is eagerly evaluated, so `validationResult` and `importResult` start simultaneously on different execution contexts! 

A simple testcase:

```
      implicit val ec = global

      /*lazy */ val slowFuture = Future { Thread.sleep(1000); println("hello") }
      /*lazy */ val fastFuture = Future( println("world"))

      val res = for {
        x <- slowFuture
        y <- fastFuture
      } yield println("!")

      Await.result(res, 2.seconds)
```

will end up printing:

```
world
hello
!
```

Read more here:
https://stackoverflow.com/questions/57403756/regarding-scala-futures-being-eager

And here:
https://svejcar.dev/posts/2019/11/14/troubles-with-future/

# Proposed Solution

Transform `val` to a `def` 
Or add `lazy` modifier.
Or inline values into for comprehension.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
+++ b/src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
@@ -24,8 +24,8 @@ class BlockImport(
       currentBestBlock: Block,
       currentTd: BigInt
   )(implicit blockExecutionContext: ExecutionContext): Future[BlockImportResult] = {
-    val validationResult = Future(blockValidation.validateBlockBeforeExecution(block))(validationContext)
-    val importResult = Future(importBlockToTop(block, currentBestBlock.header.number, currentTd))(blockExecutionContext)
+    lazy val validationResult = Future(blockValidation.validateBlockBeforeExecution(block))(validationContext)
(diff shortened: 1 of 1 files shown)


KonradStaniec:
hey @dmitry-worker thanks for pr, but this is as intended.
If you look further how the validation error is handled in `handleImportTopValidationError` you will see that if validation fails, all state and block changes are rollbacked.

dmitry-worker:
Thanks, @KonradStaniec! Would you prefer to close this PR or should I add some improvement? 
(I am also in doubt when I think about possible race conditions in handling import error and adding a block into a blockchain. Especially the case when different threads are trying to add and to remove a block from the storage).

KonradStaniec:
I would close this pr for now.

As for race conditions, there should be none as there should be always at most one thread doing importing or removing. i.e we are either saving new block or removing old ones, we do not do this concurrently. Even in this case, we first imported block and and only after it we remove this block.

The only race conditions we deal with currently are between miner and importer, but this somewhat different matter.

PR #786, opened 2020-11-09
[ETCM-302] Add more debug log to block number handling

KonradStaniec:
# Description
 Added more logs in blockchain handling of current block numbers


src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
```diff
@@ -150,6 +151,7 @@ class BlockImport(
       parentHash = parent.header.parentHash
       parentWeight <- blockchain.getChainWeightByHash(parentHash)
     } yield {
+      log.debug("Removing blocks starting from {}", bestNumber)
```
mmrozek, line 154: Shouldn't we log also `parentHash`?

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

PR #787, opened 2020-11-09
Etcm 290 invalid branches built by fetcher

kapke:
# Description

One of reasons for forks on cluster deployments is the fact that `BlockFetcher` builds block sequences which are invalid chanins, making them effectively impossible to import and thus - only local miner is able to provide blocks which are valid chain extension.

# Proposed Solution

Validate header chains as they are appended to queue, to ensure they are possible to be imported.

Thing that was out of scope, but seemed to be affecting whole syncing story a lot - when receiving new blocks remove check for fetching. 

# Testing

Already tested regular sync of whole chain on Mordor - no issues at all.
Spun up a local cluster with 5 miners couple of times:
  - got one miner fork around block 400, later OS killed 2 containers due to lack of resources (including the forked one), rest 3 continued to work without any problems or forks for over 4000 blocks
  - another time - local cluster with 5 miners - no issues at all for ~200 blocks

## Update

After resolving issue with fetcher being stuck on headers from rejected fork:
  - local cluster of 5 miners - no issues at all for ~2500 blocks
  - sync to Mordor - caught top and stayed on it over a night without visible issues

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
@@ -114,12 +114,12 @@ class BlockFetcher(
         } else {
           log.debug("Fetched {} headers starting from block {}", headers.size, headers.headOption.map(_.number))
 
-          state.validatedHeaders(headers) match {
+          state.appendHeaders(headers) match {
             case Left(err) =>
               peersClient ! BlacklistPeer(peer.id, err)
               state.withHeaderFetchReceived
-            case Right(validHeaders) =>
-              state.withHeaderFetchReceived.appendHeaders(validHeaders)
+            case Right(updatedState) =>
+              updatedState.withHeaderFetchReceived
           }
         }
 
@@ -216,12 +216,16 @@ class BlockFetcher(
         val newState = state.appendNewBlock(block, peerId)
         supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
         context become started(newState)
-        // waiting for some bodies but we don't have this header yet - at least we can use new block header
+        // waiting for some bodies but we don't have this header yet - at least we can try to use new block header
       } else if (newBlockNr == state.nextToLastBlock && !state.isFetchingHeaders) {
-        log.debug("Waiting for bodies. Add only headers")
-        val newState = state.appendHeaders(List(block.header))
-        supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
-        fetchBlocks(newState)
+        log.debug("Waiting for bodies. Try to add headers at least")
+        state.appendHeaders(List(block.header)) match {
+          case Left(error) =>
+            log.debug("{}", error)
+          case Right(newState) =>
+            supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
+            fetchBlocks(newState)
+        }
         // we're far from top
       } else if (newBlockNr > nextExpectedBlock) {
         log.debug("Far from top")
@@ -308,7 +312,6 @@ class BlockFetcher(
 
   private def fetchBodies(state: BlockFetcherState): Unit = {
     log.debug("Fetching bodies")
-
 
     val hashes = state.takeHashes(syncConfig.blockBodiesPerRequest)
     requestBlockBodies(hashes) pipeTo self
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
@@ -116,7 +116,7 @@ class BlockFetcher(
 
           state.appendHeaders(headers) match {
             case Left(err) =>
-              log.info("{}", err)
+              log.info("Dismissed received headers due to: {}", err)
               state.withHeaderFetchReceived
             case Right(updatedState) =>
               updatedState.withHeaderFetchReceived
@@ -191,6 +191,7 @@ class BlockFetcher(
       supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
       fetchBlocks(newState)
     case MessageFromPeer(NewBlock(_, block, _), peerId) =>
+      //TODO ETCM-389: Handle mined, checkpoint and new blocks uniformly
       log.debug("Received NewBlock {}", block.idTag)
       val newBlockNr = block.number
       val nextExpectedBlock = state.lastFullBlockNumber + 1
@@ -260,7 +261,7 @@ class BlockFetcher(
       .getOrElse(fetcherState)
 
   private def fetchHeaders(state: BlockFetcherState): Unit = {
-    val blockNr = state.nextToLastBlock
+    val blockNr = state.nextBlockToFetch
     val amount = syncConfig.blockHeadersPerRequest
 
     fetchHeadersFrom(blockNr, amount) pipeTo self
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
@@ -117,7 +117,7 @@ class BlockFetcher(
           state.appendHeaders(headers) match {
             case Left(err) =>
               log.info("{}", err)
-              peersClient ! BlacklistPeer(peer.id, err)
+//              peersClient ! BlacklistPeer(peer.id, err)
               state.withHeaderFetchReceived
             case Right(updatedState) =>
               updatedState.withHeaderFetchReceived
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
@@ -123,6 +123,8 @@ class BlockFetcher(
           }
         }
 
+      log.info("{}", newState.waitingHeaders.map(_.idTag))
+
       //First successful fetch
       if (state.waitingHeaders.isEmpty) {
         supervisor ! ProgressProtocol.StartedFetching
@@ -146,6 +148,8 @@ class BlockFetcher(
           log.debug("Fetched {} block bodies", bodies.size)
           state.withBodiesFetchReceived.addBodies(peer, bodies)
         }
+
+      log.info("{}", newState.readyBlocks.map(_.idTag))
 
       fetchBlocks(newState)
     case RetryBodiesRequest if state.isFetchingBodies =>
@@ -245,6 +249,10 @@ class BlockFetcher(
       log.debug(s"New last block $blockNr imported from the inside")
       val newLastBlock = blockNr.max(state.lastBlock)
       val newState = state.withLastBlock(newLastBlock).withPossibleNewTopAt(blockNr)
+
+      log.info("{}", newState.status)
+      log.debug("{}", newState.statusDetailed)
+
       fetchBlocks(newState)
     }
   }
@@ -300,6 +308,7 @@ class BlockFetcher(
 
   private def fetchBodies(state: BlockFetcherState): Unit = {
     log.debug("Fetching bodies")
+
 
     val hashes = state.takeHashes(syncConfig.blockBodiesPerRequest)
     requestBlockBodies(hashes) pipeTo self
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
@@ -117,7 +117,6 @@ class BlockFetcher(
           state.appendHeaders(headers) match {
             case Left(err) =>
               log.info("{}", err)
-//              peersClient ! BlacklistPeer(peer.id, err)
               state.withHeaderFetchReceived
             case Right(updatedState) =>
               updatedState.withHeaderFetchReceived
@@ -145,7 +144,7 @@ class BlockFetcher(
           state.withBodiesFetchReceived
         } else {
           log.debug("Fetched {} block bodies", bodies.size)
-          state.withBodiesFetchReceived.addBodies(peer, bodies)
+          state.withBodiesFetchReceived.addBodies(peer.id, bodies)
         }
 
       fetchBlocks(newState)
@@ -195,38 +194,19 @@ class BlockFetcher(
       log.debug("Received NewBlock {}", block.idTag)
       val newBlockNr = block.number
       val nextExpectedBlock = state.lastFullBlockNumber + 1
-      // we're on top, so we can pass block directly to importer
-      if (newBlockNr == nextExpectedBlock && state.isOnTop) {
+
+      if (state.isOnTop && newBlockNr == nextExpectedBlock) {
         log.debug("Passing block directly to importer")
         val newState = state.withPeerForBlocks(peerId, Seq(newBlockNr)).withKnownTopAt(newBlockNr)
         state.importer ! OnTop
         state.importer ! ImportNewBlock(block, peerId)
         supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
         context become started(newState)
-        // there are some blocks waiting for import but it seems that we reached top on fetch side so we can enqueue new block for import
-      } else if (newBlockNr == nextExpectedBlock && !state.isFetching && state.waitingHeaders.isEmpty) {
-        log.debug("Enqueueing new block for import")
-        val newState = state.appendNewBlock(block, peerId)
-        supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
-        context become started(newState)
-        // waiting for some bodies but we don't have this header yet - at least we can try to use new block header
-      } else if (newBlockNr == state.nextToLastBlock && !state.isFetchingHeaders) {
-        log.debug("Waiting for bodies. Trying to add headers at least")
-        state.appendHeaders(List(block.header)) match {
-          case Left(error) =>
-            log.debug(error)
-          case Right(newState) =>
-            supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
-            fetchBlocks(newState)
-        }
-        // we're far from top
-      } else if (newBlockNr > nextExpectedBlock) {
-        log.debug("Received block is too new")
-        val newState = state.withPossibleNewTopAt(newBlockNr)
+      } else {
+        log.debug("Ignoring received block as it doesn't match local state or fetch side is not on top")
+        val newState = state.withPossibleNewTopAt(block.number)
         supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
         fetchBlocks(newState)
-      } else {
-        log.debug("Had to ignore received block {}", block.idTag)
       }
     case BlockImportFailed(blockNr, reason) =>
       val (peerId, newState) = state.invalidateBlocksFrom(blockNr)
@@ -244,13 +224,12 @@ class BlockFetcher(
         fetchBlocks(newState)
       }
     //keep fetcher state updated in case new checkpoint block or mined block was imported
-    case InternalLastBlockImport(blockNr) => {
+    case InternalLastBlockImport(blockNr) =>
       log.debug(s"New last block $blockNr imported from the inside")
       val newLastBlock = blockNr.max(state.lastBlock)
       val newState = state.withLastBlock(newLastBlock).withPossibleNewTopAt(blockNr)
 
       fetchBlocks(newState)
-    }
   }
 
   private def handlePickedBlocks(
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
@@ -124,8 +124,6 @@ class BlockFetcher(
           }
         }
 
-      log.info("{}", newState.waitingHeaders.map(_.idTag))
-
       //First successful fetch
       if (state.waitingHeaders.isEmpty) {
         supervisor ! ProgressProtocol.StartedFetching
@@ -250,9 +248,6 @@ class BlockFetcher(
       log.debug(s"New last block $blockNr imported from the inside")
       val newLastBlock = blockNr.max(state.lastBlock)
       val newState = state.withLastBlock(newLastBlock).withPossibleNewTopAt(blockNr)
-
-      log.info("{}", newState.status)
-      log.debug("{}", newState.statusDetailed)
 
       fetchBlocks(newState)
     }
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
@@ -116,6 +116,7 @@ class BlockFetcher(
 
           state.appendHeaders(headers) match {
             case Left(err) =>
+              log.info("{}", err)
               peersClient ! BlacklistPeer(peer.id, err)
               state.withHeaderFetchReceived
             case Right(updatedState) =>
@@ -200,7 +201,7 @@ class BlockFetcher(
       val newBlockNr = block.number
       val nextExpectedBlock = state.lastFullBlockNumber + 1
 
-      log.debug("Received NewBlock nr {}", newBlockNr)
+      log.debug("Received NewBlock {}", block.idTag)
 
       // we're on top, so we can pass block directly to importer
       if (newBlockNr == nextExpectedBlock && state.isOnTop) {
@@ -228,10 +229,14 @@ class BlockFetcher(
         }
         // we're far from top
       } else if (newBlockNr > nextExpectedBlock) {
-        log.debug("Far from top")
-        val newState = state.withKnownTopAt(newBlockNr)
+        log.debug("Received block is too new")
+        val newState = state.withPossibleNewTopAt(newBlockNr)
         supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
         fetchBlocks(newState)
+      } else {
+        log.debug("Had to ignore received block {}", block.idTag)
+        log.info("{}", state.status)
+        log.debug("{}", state.statusDetailed)
       }
     case BlockImportFailed(blockNr, reason) =>
       val (peerId, newState) = state.invalidateBlocksFrom(blockNr)
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
@@ -150,8 +150,6 @@ class BlockFetcher(
           state.withBodiesFetchReceived.addBodies(peer, bodies)
         }
 
-      log.info("{}", newState.readyBlocks.map(_.idTag))
-
       fetchBlocks(newState)
     case RetryBodiesRequest if state.isFetchingBodies =>
       log.debug("Time-out occurred while waiting for bodies")
@@ -193,19 +191,15 @@ class BlockFetcher(
         case Left(_) => state
         case Right(validHashes) => state.withPossibleNewTopAt(validHashes.lastOption.map(_.number))
       }
-
       supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
-
       fetchBlocks(newState)
     case MessageFromPeer(NewBlock(_, block, _), peerId) =>
+      log.debug("Received NewBlock {}", block.idTag)
       val newBlockNr = block.number
       val nextExpectedBlock = state.lastFullBlockNumber + 1
-
-      log.debug("Received NewBlock {}", block.idTag)
-
       // we're on top, so we can pass block directly to importer
       if (newBlockNr == nextExpectedBlock && state.isOnTop) {
-        log.debug("Pass block directly to importer")
+        log.debug("Passing block directly to importer")
         val newState = state.withPeerForBlocks(peerId, Seq(newBlockNr)).withKnownTopAt(newBlockNr)
         state.importer ! OnTop
         state.importer ! ImportNewBlock(block, peerId)
@@ -213,16 +207,16 @@ class BlockFetcher(
         context become started(newState)
         // there are some blocks waiting for import but it seems that we reached top on fetch side so we can enqueue new block for import
       } else if (newBlockNr == nextExpectedBlock && !state.isFetching && state.waitingHeaders.isEmpty) {
-        log.debug("Enqueue new block for import")
+        log.debug("Enqueueing new block for import")
         val newState = state.appendNewBlock(block, peerId)
         supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
         context become started(newState)
         // waiting for some bodies but we don't have this header yet - at least we can try to use new block header
       } else if (newBlockNr == state.nextToLastBlock && !state.isFetchingHeaders) {
-        log.debug("Waiting for bodies. Try to add headers at least")
+        log.debug("Waiting for bodies. Trying to add headers at least")
         state.appendHeaders(List(block.header)) match {
           case Left(error) =>
-            log.debug("{}", error)
+            log.debug(error)
           case Right(newState) =>
             supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
             fetchBlocks(newState)
@@ -235,8 +229,6 @@ class BlockFetcher(
         fetchBlocks(newState)
       } else {
         log.debug("Had to ignore received block {}", block.idTag)
-        log.info("{}", state.status)
-        log.debug("{}", state.statusDetailed)
       }
     case BlockImportFailed(blockNr, reason) =>
       val (peerId, newState) = state.invalidateBlocksFrom(blockNr)
@@ -249,7 +241,7 @@ class BlockFetcher(
     //ex. After a successful handshake, fetcher will receive the info about the header of the peer best block
     case MessageFromPeer(BlockHeaders(headers), _) =>
       headers.lastOption.map { bh =>
-        log.debug(s"Candidate for new top at block ${bh.number}, current know top ${state.knownTop}")
+        log.debug(s"Candidate for new top at block ${bh.number}, current known top ${state.knownTop}")
         val newState = state.withPossibleNewTopAt(bh.number)
         fetchBlocks(newState)
       }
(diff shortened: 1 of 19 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -88,6 +89,9 @@ case class BlockFetcherState(
         .ensure("Given headers are not sequence with already fetched ones")(_.head.number <= nextToLastBlock)
         .ensure("Given headers aren't better than already fetched ones")(_.last.number > lastBlock)
         .ensure("Given headers should form a sequence without gaps")(HeadersSeq.areChain)
+        .ensure("Given headers do not form a chain with already stored ones")(headers =>
+          (waitingHeaders.lastOption, headers.headOption).mapN(_ isParentOf _).getOrElse(true)
```
ntallar, line 93: Shouldn't we remove the conditions
- Given headers are not sequence with already fetched ones
- Given headers aren't better than already fetched ones
- Given headers do not form a chain with already stored ones

? The response of a peer isn't invalid if they don't fulfill this

If its a condition required by our waitingHeaders structure maybe we should guarantee this when appending them

Or at least we should stop blacklisting the peer if these validations fail, maybe we could have 2 separate errors, one for consistency of our structures and the other one for invalid responses? I'd separate it into two functions if possible

(note that the above overlaps with #771  )

ntallar (changes requested):

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -72,12 +71,14 @@ case class BlockFetcherState(
...
-    withPossibleNewTopAt(headers.lastOption.map(_.number))
-      .copy(
-        waitingHeaders = waitingHeaders ++ headers.filter(_.number > lastBlock).sortBy(_.number),
-        lastBlock = HeadersSeq.lastNumber(headers).getOrElse(lastBlock)
-      )
+  def appendHeaders(headers: Seq[BlockHeader]): Either[String, BlockFetcherState] =
+    validatedHeaders(headers).map(validHeaders => {
+      withPossibleNewTopAt(headers.lastOption.map(_.number))
+        .copy(
+          waitingHeaders = waitingHeaders ++ headers.filter(_.number > lastBlock).sortBy(_.number),
```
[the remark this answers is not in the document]
kapke: Since this filter was a thing, which caused inconsistencies I removed it. So far tests are passing, I'm going to run cluster and sync to Mordor to see whether it affects stability in any way. I can image one case where it does, but I'm not sure whether it's really an issue

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
kapke: But these conditions are also to ensure consistency of internal state. Probably it makes more sense to change these errors into an ADT and make a decision about blacklisting basing on the result. I'm not sure if this PR is the place to do that change. WDYT?

ntallar: Given that we have blacklisting disabled I don't think it's critical to address this right away, we could do it after #771 gets merged (or even have this discussion there)

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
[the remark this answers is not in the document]
ntallar: I never liked that filter too much :thinking: that filtering will be done by the importer either way, or maybe could be moved to when the fetcher is queried (to decouple them further)

ntallar (changes requested):

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -52,7 +51,7 @@ case class BlockFetcherState(
 
   def hasFetchedTopHeader: Boolean = lastBlock == knownTop
 
-  def isOnTop: Boolean = !isFetching && hasFetchedTopHeader && hasEmptyBuffer
+  def isOnTop: Boolean = hasFetchedTopHeader && hasEmptyBuffer
```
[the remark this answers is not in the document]
kapke: not really, since `hasEmptyBuffer` covers that
That condition made sure that fetcher state wrt best known block does not change due to some pending request. I don't think it matters though - the fetcher can be on top even if it has still some pending request.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
kapke: I thought about it a bit (from different perspective though) and I came to conclusion that we need one of those (at least now):
  - build proper full branches on fetcher side
    That's because if mined block is accepted in other nodes - we'll receive it "properly" through network layer anyway and once we import branch got from fetcher - it'll be discarded (if lower/the same as the mined one) or used (if better blocks appeared on the network). This is essential from point of view of catching the network as it allows to have miner enabled and still be able to process blocks from network quite fast without big changes within code. That approach makes it a bit questionable to update `bestBlock` in fetcher state on receiving imported block, but it is still quite important when near top, otherwise node would fork too easy
  - clean fetcher state after receiving internal block (checkpoint or mined one) - while the easiest way to keep things in sync, etc. it would result in awful performance when catching the network (e.g. after turning node off for a night), because miner could effectively livelock/DoS the fetcher (especially in poor network conditions)

Given the above - I think that current approach present in code (building proper branches on fetcher side + disabling blacklisting due to useless responses) is the best until we extract component for managing tentative branches (where we could simply build multiple partial branches when near top, etc.)

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -188,42 +189,44 @@ class BlockFetcher(
...
         // there are some blocks waiting for import but it seems that we reached top on fetch side so we can enqueue new block for import
       } else if (newBlockNr == nextExpectedBlock && !state.isFetching && state.waitingHeaders.isEmpty) {
-        log.debug("Enqueue new block for import")
+        log.debug("Enqueueing new block for import")
         val newState = state.appendNewBlock(block, peerId)
         supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
         context become started(newState)
-        // waiting for some bodies but we don't have this header yet - at least we can use new block header
+        // waiting for some bodies but we don't have this header yet - at least we can try to use new block header
       } else if (newBlockNr == state.nextToLastBlock && !state.isFetchingHeaders) {
```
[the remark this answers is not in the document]
kapke: I think so, it may be the case where we fill the block gap between fetcher and actual blockchain without much hassle just by receiving some new block from the network

kapke:
@ntallar As you can see in the code, there is blacklisting disabled when receiving headers response. I think I would:
  - create a JIRA ticket to revise whole response validation story and its relation to blacklisting
  - make blacklisting tests ignored and marked with that ticket
  - remove blacklisting line

WDYT?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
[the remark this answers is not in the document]
ntallar: But shouldn't the condition be `state.waitingHeaders.lastOption.map(_.number).getOrElse(state.nextToLastBlock)`, if not the append will fail

Why not keep it simple and attempt appending it all the time?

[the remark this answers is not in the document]
kapke: I'm not sure I get your point - do you want to simply remove this branch or merge all remaining branches to single, where we try to append header of new block unconditionally and then see what happens?

I like the latter approach TBH.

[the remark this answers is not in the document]
kapke: And on top of that it would even be quite simple to try to find suffix of headers that can be appended to queue so even response from pending request could be at least partially used. (In future/separate task of course)

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
ntallar: You mean then that we should close this PR? There's no way of reaching a better shape of it? I think there are some minor improvements that we can do without dropping the current approach

Either way, I'd definitely go for `build proper full branches on fetcher side` approach as the other one will lead to even more nasty workarounds. I think that maybe that maybe we could add the TentativeCompleteBranches only without removing the BlockQueue... but whatever we do about it will be on December

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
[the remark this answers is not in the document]
ntallar: >not really, since hasEmptyBuffer covers that

Does it? If you just send a request, wouldn't the buffer still be empty?

> the fetcher can be on top even if it has still some pending request

Should we add a comment on what actually being on top means then for the fetcher?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -88,6 +90,9 @@ case class BlockFetcherState(
         .ensure("Given headers are not sequence with already fetched ones")(_.head.number <= nextToLastBlock)
         .ensure("Given headers aren't better than already fetched ones")(_.last.number > lastBlock)
         .ensure("Given headers should form a sequence without gaps")(HeadersSeq.areChain)
+        .ensure("Given headers do not form a chain with already stored ones")(headers =>
+          (waitingHeaders.lastOption, headers.headOption).mapN(_ isParentOf _).getOrElse(true)
```
ntallar, line 94: I was thinking... should we maybe check this condition to see if there's any way of concatenating both branches? This condition might be false only because we received a header from a NewBlock

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -258,7 +258,7 @@ class PeerActor[R <: HandshakeResult](
         handleTerminated(rlpxConnection, 0, Handshaked) orElse {
 
           case RLPxConnectionHandler.MessageReceived(message) =>
-            log.debug(s"Received message: {} from $peerId", message)
+            log.debug(s"Received message: {} from $peerId", message.getClass.getName)
```
ntallar, line 261: Noooooo, this logs were useful on tesnet debugging

ntallar:
>@ntallar As you can see in the code, there is blacklisting disabled when receiving headers response. I think I would:

I agree with that plan! Maybe something like #771 but for the headers

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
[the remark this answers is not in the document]
ntallar: You are right!

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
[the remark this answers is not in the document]
ntallar: The latter as well! That would simplify quite a lot this code

>And on top of that it would even be quite simple to try to find suffix of headers that can be appended to queue so even response from pending request could be at least partially used. (In future/separate task of course)

Hmm I'd prefer adding it here but you're right that it's not as straightforward (ideally we would add some comparison between the branch being appended and the one in memory), could you create a task for it?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
ntallar: Given [this](https://github.com/input-output-hk/mantis/pull/787#discussion_r520688749) conversation, I think this comment doesn't apply

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
kapke: And disruptive when running regular sync on mordor with debug level ;)

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
kapke: `build proper full branches on fetcher side` - That's the approach implemented in this PR

kapke:
It seems that current state of this branch introduces yet another reason for forks - a scenario where:
  - fetcher receives a branch
  - some of these blocks get rejected by rest of the network
  - fetcher gets stuck because it's unable to fetch a branch which follows chain in queue

I see 2 solutions to that problem, the question is which seems to be better:
  - just make sure that the chain is contiguous in terms of block numbers - ledger will import whatever possible and then force to fetcher to clear state up in case of no longer continued branch (at cost of blacklisting honest peer)
  - count number of rejected responses and if it passes some threshold - reset fetcher state (or even restart the actor) so that invalid state is being invalidated and the accepted fork should be automatically picked

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
ntallar: Oh I thought you meant doing something like the TBM with that

But then, wdyt about removing the conditions: `_.last.number > lastBlock`, `_.head.number <= nextToLastBlock`? Maybe I'm missing the point of having them

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
ntallar: Totally, I usually remove this logs when local testing xD maybe our cluster should eventually have a different logback than other nodes?

But given that we don't have that in place and that this logs are useful on the testnet we'll have to stick with this disruptive logs, we can create a future task to do in conjunction with devops to bring a separate logback to the cluster or thing of another alternative

ntallar:
Wouldn't that scenario evolve into:
1. BlockImporter imports the blocks from fetcher so he removes it from there
2. Fetcher would then be able to fetch the blocks from the fork (though they would be unconcatenable)
3. When the blocks fail to be imported the `InvalidateBlocksFrom` message would kick in and result in the fetcher going back... eventually being able to solve the fork

The only problem I see is that blocks being mined in the middle are quite disruptive, as they invalidate the invalidation of the `InvalidateBlocksFrom` message. Maybe the fetcher should keep both
- a last block: with the last block number imported
- a block number to fetch

wdyt?

Both your solutions are a bit too hacky imho, this code is already too complicated to add more workarounds:
1. I'm not sure which of the components from the ledger rely on the blocks passed being a branch, something could break in the middle if you break that assumption.
2. This is the safest of the two, though it only hides the problem as we are only restarting periodically

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -188,42 +189,44 @@ class BlockFetcher(
         case Left(_) => state
         case Right(validHashes) => state.withPossibleNewTopAt(validHashes.lastOption.map(_.number))
       }
-
       supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
-
       fetchBlocks(newState)
     case MessageFromPeer(NewBlock(_, block, _), peerId) =>
```
mirkoAlic, line 194: can we rewrite this handling before take decisions about a refactor, i mean, every `if / if else` below uses different things, and is really hard to reason about it.. i would prefer to have something like:
```
// use a variable to reflect this condition instead of adding a comment
if (newBlockNr == nextExpectedBlock) {
  if(state.isOnTop) {
    ...
  } else {
     ...
  }
```
} else if (newBlockNr > nextExpectedBlock) {
  ...
} else {
 log.info("older block")
}

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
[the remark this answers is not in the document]
mirkoAlic: IMHO, if we are still fetching things, we should be able to not use current headers from new block, i mean, other stuff should be sufficient to receive the missing info later on.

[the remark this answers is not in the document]
mirkoAlic: tbh, i think that this handling is trying to do a lot at the same time, and feels like the reason behind that is because is trying to patch other parts of the process.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -188,42 +189,44 @@ class BlockFetcher(
...
         // we're far from top
       } else if (newBlockNr > nextExpectedBlock) {
-        log.debug("Far from top")
-        val newState = state.withKnownTopAt(newBlockNr)
+        log.debug("Received block is too new")
+        val newState = state.withPossibleNewTopAt(newBlockNr)
         supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
         fetchBlocks(newState)
+      } else {
+        log.debug("Had to ignore received block {}", block.idTag)
```
mirkoAlic, line 229: older block?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -114,12 +114,13 @@ class BlockFetcher(
         } else {
           log.debug("Fetched {} headers starting from block {}", headers.size, headers.headOption.map(_.number))
 
-          state.validatedHeaders(headers) match {
+          state.appendHeaders(headers) match {
             case Left(err) =>
-              peersClient ! BlacklistPeer(peer.id, err)
+              log.info("{}", err)
+//              peersClient ! BlacklistPeer(peer.id, err)
```
mirkoAlic, line 120: please add a comment and the task link

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
mirkoAlic: i'm agree with Nico, i think the 2 first validations are no longer needed.

kapke:
@ntallar Yes, making validations less restrictive on fetcher side (that is requiring received part to be a chain and requiring it to create a chain with queued blocks/headers only with regard to block numbers) should led to that scenario.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
kapke: I see, restoring then

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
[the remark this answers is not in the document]
kapke: I totally agree, other reason is that it was created only with following the network in mind and handling mining blocks was added at the very end as an afterthought just to enable local development.

ntallar:
>Please note that such approach to handling irrecoverable errors is at core of actors (which we use here) as well as is quite effective at keeping system operational.

I'm sorry but I'm still not convinced. It's definitely a good last resource in case we don't manage to fix an issue, but this hasn't yet become a blocker

>@ntallar Not really. In my case fetcher got stuck because of having a header from non-existing fork and being unable to make any progress.

I see now the issue you mentioned! Shouldn't we do a better handling of our BlockBodies response to address it? A simple alternative is dropping headers from the waitingHeaders when a body of a header is requested but not received. Maybe it can be a next step to #771

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -100,15 +106,19 @@ case class BlockFetcherState(
...
-    val blocks = matching.zip(bodies).map((Block.apply _).tupled)
-
-    withPeerForBlocks(peer.id, blocks.map(_.header.number))
-      .copy(
-        readyBlocks = readyBlocks.enqueue(blocks),
-        waitingHeaders = waiting
-      )
+  def addBodies(peerId: PeerId, bodies: Seq[BlockBody]): BlockFetcherState = {
+    if (bodies.isEmpty) {
+      copy(waitingHeaders = Queue.empty)
```
mirkoAlic, line 111: why do the clean up and not return just `this`?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -114,12 +114,12 @@ class BlockFetcher(
         } else {
           log.debug("Fetched {} headers starting from block {}", headers.size, headers.headOption.map(_.number))
 
-          state.validatedHeaders(headers) match {
+          state.appendHeaders(headers) match {
             case Left(err) =>
-              peersClient ! BlacklistPeer(peer.id, err)
+              log.info("{}", err)
```
ntallar, line 119: Minor: maybe something like: "Dismissing received headers due to: {}"

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -68,26 +67,33 @@ case class BlockFetcherState(
...
+    validatedHeaders(headers.sortBy(_.number)).map(validHeaders => {
+      val lastNumber = HeadersSeq.lastNumber(validHeaders)
+      withPossibleNewTopAt(lastNumber)
+        .copy(
+          waitingHeaders = waitingHeaders ++ validHeaders,
+          lastBlock = lastNumber.getOrElse(lastBlock)
+        )
+    })
 
   def validatedHeaders(headers: Seq[BlockHeader]): Either[String, Seq[BlockHeader]] =
```
ntallar, line 87: Should we make this function private as it's no longer used from outside?

Maybe we could add also:
- add comment that the purpose of it is to check consistency with the fetcher state
- add a FIXME on the fact that this should eventually be split into validations of the request/response matching and blacklisting for that case

?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -188,40 +188,23 @@ class BlockFetcher(
         case Left(_) => state
         case Right(validHashes) => state.withPossibleNewTopAt(validHashes.lastOption.map(_.number))
       }
-
       supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
-
       fetchBlocks(newState)
     case MessageFromPeer(NewBlock(_, block, _), peerId) =>
+      log.debug("Received NewBlock {}", block.idTag)
```
ntallar, line 194: :+1:

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -188,40 +188,23 @@ class BlockFetcher(
...
-      log.debug("Received NewBlock nr {}", newBlockNr)
-
-      // we're on top, so we can pass block directly to importer
-      if (newBlockNr == nextExpectedBlock && state.isOnTop) {
-        log.debug("Pass block directly to importer")
+      if (state.isOnTop && newBlockNr == nextExpectedBlock) {
+        log.debug("Passing block directly to importer")
         val newState = state.withPeerForBlocks(peerId, Seq(newBlockNr)).withKnownTopAt(newBlockNr)
         state.importer ! OnTop
         state.importer ! ImportNewBlock(block, peerId)
```
ntallar, line 202: Maybe this is too big of a change... but with our changes below we are starting (here) to treat NewBlock as mined and checkpoint blocks, then why all the isOnTop handling on the BlockImporter?

It's only used for this message, but if it worked with mined blocks then why not here as well?

(for this PR maybe only leaving a comment would be enough)

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -68,26 +67,33 @@ case class BlockFetcherState(
       .orElse(waitingHeaders.headOption.map(_.number))
       .getOrElse(lastBlock)
 
-  def nextToLastBlock: BigInt = lastBlock + 1
+  def nextToLastBlock: BigInt = waitingHeaders.lastOption
```
ntallar, line 70: Should we rename this as blockHeaderToFetch or something like that? (also adding some comments would be nice)

ntallar (changes requested):

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
ntallar: I agree with mk, this is inconsistent with the fact that we are keeping the `waitingHeaders = waiting` below

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
```diff
@@ -124,6 +124,78 @@ class BlockFetcherSpec extends TestKit(ActorSystem("BlockFetcherSpec_System")) w
...
+      val refForAnswerFirstBodiesReq = peersClient.expectMsgPF() {
+        case PeersClient.Request(msg, _, _) if msg == firstGetBlockBodiesRequest => peersClient.lastSender
+      }
+
+      // Block 5 is mined (we could have reached this stage due to invalidation messages sent to the fetcher)
+      val minedBlock = alternativeSecondBlocksBatch.drop(5).head
+      val minedBlockNumber = minedBlock.number
+      blockFetcher ! InternalLastBlockImport(minedBlockNumber)
+
+      // Answer pending requests: first block bodies request + second block headers request
```
mirkoAlic, line 170: maybe i'm wrong, but this comment doesn't match with the lines below..

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
```diff
@@ -124,6 +124,78 @@ class BlockFetcherSpec extends TestKit(ActorSystem("BlockFetcherSpec_System")) w
...
+      // Second bodies request
+      val refForAnswerSecondBodiesReq = peersClient.expectMsgPF() { case PeersClient.Request(GetBlockBodies(_), _, _) =>
+        peersClient.lastSender
+      }
+      peersClient.send(
+        refForAnswerSecondBodiesReq,
+        PeersClient.Response(fakePeer, BlockBodies(alternativeSecondBlocksBatch.drop(6).map(_.body)))
+      )
+
+      importer.send(blockFetcher, PickBlocks(100))
```
mirkoAlic, line 191: 100 == `firstBlocksBatch.size + secondBlocksBatch.size` ?

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
```diff
@@ -124,6 +124,78 @@ class BlockFetcherSpec extends TestKit(ActorSystem("BlockFetcherSpec_System")) w
...
+        case PeersClient.Request(msg, _, _) if msg == secondGetBlockHeadersRequest => peersClient.lastSender
+      }
+
+      // First bodies request
+      val firstGetBlockBodiesRequest = GetBlockBodies(firstBlocksBatch.map(_.hash))
+      val refForAnswerFirstBodiesReq = peersClient.expectMsgPF() {
+        case PeersClient.Request(msg, _, _) if msg == firstGetBlockBodiesRequest => peersClient.lastSender
+      }
+
+      // Block 5 is mined (we could have reached this stage due to invalidation messages sent to the fetcher)
```
mirkoAlic, line 165: i think `minedBlock.number == 16`

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
```diff
@@ -124,6 +124,78 @@ class BlockFetcherSpec extends TestKit(ActorSystem("BlockFetcherSpec_System")) w
...
+      }
+      peersClient.send(
+        refForAnswerSecondBodiesReq,
+        PeersClient.Response(fakePeer, BlockBodies(alternativeSecondBlocksBatch.drop(6).map(_.body)))
+      )
+
+      importer.send(blockFetcher, PickBlocks(100))
+      importer.ignoreMsg({ case BlockImporter.NotOnTop => true })
+      importer.expectMsgPF() { case BlockFetcher.PickedBlocks(blocks) =>
+        val headers = blocks.map(_.header).toList
```
mirkoAlic, line 194: current chain: (using only block numbers)
`16, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14`

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
```diff
@@ -124,6 +124,78 @@ class BlockFetcherSpec extends TestKit(ActorSystem("BlockFetcherSpec_System")) w
       peersClient.expectMsgClass(classOf[BlacklistPeer])
       peersClient.expectMsgPF() { case PeersClient.Request(msg, _, _) if msg == firstGetBlockHeadersRequest => () }
     }
+
+    "should ensure blocks passed to importer are always forming chain" in new TestSetup {
```
mirkoAlic, line 128: Could this test be more similar to the ones above? I mean,  instead of invalidation, here we are mining block in the middle...I think this sequence should be enough:
- fetcher request headers
- we mine a new block (block 1  from a fork)
- fetcher should dismiss pending headers response
- fetcher should ask for new headers (from new top)
- fetcher should receive those blocks..

mirkoAlic, line 128: Could this test be more similar to the ones above? I mean, instead of invalidation, here we are mining block in the middle...I think this sequence should be enough:

- fetcher request headers
-  we mine a new block (block 1 from a fork)
- fetcher should receive those blocks bodies
- the resulted chain shouldn't have gaps.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
kapke: How are we going to invalidate headers from rejected chain then? I thought that this is the way we agreed on to solve that scenario: https://github.com/input-output-hk/mantis/pull/787#issuecomment-726177268.

I don't think this is really an inconsistency since they are different cases (no bodies means either given node is not synced up or it simply doesn't have given blocks for any reason).

I can make it clearer by adding comment or turning the return type into an either (plus comment), but I do think that this is the way to go (and works really well as tests from last weekend show)

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
kapke: I'll leave the comment. Making it right and consistent with handling mined and checkpoint blocks will be a much bigger task.

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
kapke: I'm not sure. Whole issue was that if block was mined between 2 headers requests, gap was created because not all of headers were added to queue. Scenario you described seem to be more appropriate to check wrt received block bodies validation.

In scenario you described we still would need to either prepare fetcher somehow with some branch - which essentially would evolve it to the same as it is now.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
mirkoAlic: Fair enough! i think just adding a comment like *this is enough:
*"no bodies means either given node is not synced up or it simply doesn't have given blocks for any reason. Faced with this scenario we decide to give up about those requested headers"

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
mirkoAlic: ok, no requested changes for now then! =)

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
kapke: Done exactly that thing ;)

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -100,15 +116,26 @@ case class BlockFetcherState(
...
-    val (matching, waiting) = waitingHeaders.splitAt(bodies.length)
-    val blocks = matching.zip(bodies).map((Block.apply _).tupled)
-
-    withPeerForBlocks(peer.id, blocks.map(_.header.number))
-      .copy(
-        readyBlocks = readyBlocks.enqueue(blocks),
-        waitingHeaders = waiting
-      )
+  /**
+    * Matches bodies with headers in queue and adding matched bodies to the blocks.
```
mirkoAlic, line 120: :1st_place_medal:

mirkoAlic (approved):
LGTM!

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
ntallar: Agree as well! Had a whole chat with mk where he convinced me of this being the approach (I expected something like this but implemented more closer to what we have on the other project, though this approach fits better with having requests in parallel)

ntallar (approved):
LGTM! Love the comments :heart: 
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

PR #790, opened 2020-11-11
[FIX] Return the last blocks when fetching the ommers ancestors from memory

ntallar:
# Description

Addresses issue on testnet

## Scenario

- Importing a chain of 12 blocks (1858 - 1870)
- Block 1870 had an ommer of number 1865 (it's chain ancestor is then block 1864)

When importing block 1970:
1. Blocks 1858-1869 were returned from the BlockQueue when fetching the branch from it
2. Taking 6 blocks from it resulted in blocks: 1858-1863

The above chain didn't have the ommer ancestor, so the validation of them failed with `OmmersAncestorsError`. As the above mentioned chain was followed by 3/5 miners they forked irrecoverably.

# Proposed Solution

Changing `Taking 6 blocks from it resulted in blocks: 1858-1863` to `Taking the last 6 blocks from it resulted in blocks: 1864-1869`

# Testing

Due to the complexity of testing this scenario and the time requirements for the having the internal testnet running no test was added.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/ledger/BlockValidation.scala
+++ b/src/main/scala/io/iohk/ethereum/ledger/BlockValidation.scala
@@ -20,17 +20,20 @@ class BlockValidation(consensus: Consensus, blockchain: Blockchain, blockQueue: 
   }
 
   private def getNBlocksBackFromChainOrQueue(hash: ByteString, n: Int): List[Block] = {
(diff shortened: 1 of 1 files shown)


mirkoAlic (approved):
LGTM! please create a task for future analysis over block queue in order to see how we could start adding tests for this kind of things.

PR #793, opened 2020-11-13
[ETCM-280] Use capabilities when sending checkpoint related network messages

pslaski:
# Description

The version of the message to send should be determined by peer capabilities

# Proposed Solution
During the handshake, the peer is introducing its capabilities to supported protocols. Base on that mantis is deciding which version of message use.

- new protocol version 64 with new NewBlock and Status messages (can be configured by `network.protocol-version`)
- decouple NewBlock and Status messages from our services

# Testing

Enable PV64 and try to sync with PV63 client and PV64 client


src/main/scala/io/iohk/ethereum/network/handshaker/EtcHelloExchangeState.scala
```diff
@@ -24,10 +26,14 @@ case class EtcHelloExchangeState(handshakerConfiguration: EtcHandshakerConfigura
 
   override def applyResponseMessage: PartialFunction[Message, HandshakerState[PeerInfo]] = { case hello: Hello =>
     log.debug("Protocol handshake finished with peer ({})", hello)
-    if (hello.capabilities.contains(Capability("eth", Versions.PV63.toByte)))
-      EtcNodeStatusExchangeState(handshakerConfiguration)
+    if (handshakerConfiguration.protocolVersion == Versions.PV64 && hello.capabilities.contains(Eth64Capability))
```
ntallar, line 29: Should we add a comment on our `protocol-version` from the application.conf on how it impacts on this, the problems with eth64 and how it should be used?

src/main/scala/io/iohk/ethereum/network/handshaker/EtcNodeStatusExchangeState.scala
```diff
@@ -21,49 +18,16 @@ case class EtcNodeStatusExchangeState(handshakerConfiguration: EtcHandshakerConf
       timeout = peerConfiguration.waitForStatusTimeout
     )
 
-  def applyResponseMessage: PartialFunction[Message, HandshakerState[PeerInfo]] = { case remoteStatus: Status =>
```
ntallar, line 24: There seems to be quite repeated code between `EtcNodeStatus64ExchangeState` and `EtcNodeStatus64ExchangeState` could we maybe add a function here `applyRemoteStatusMessage: RemoteStatus => HandshakerState[PeerInfo]`? The states would then only convert Status to RemoteStatus and call it

ntallar (changes requested):
I like how this is all looking! I still have pending taking a look at the test changes, some comments before that:

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV64.scala
```diff
@@ -0,0 +1,137 @@
...
+        s"bestHash: ${Hex.toHexString(bestHash.toArray[Byte])}, " +
+        s"genesisHash: ${Hex.toHexString(genesisHash.toArray[Byte])}," +
+        s"}"
+
+    override def code: Int = Status.code
+  }
+
+  object NewBlock {
+
+    val code: Int = Versions.SubProtocolOffset + 0x07
```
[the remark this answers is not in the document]
pslaski: IMHO we should have even only one place with all codes - the code determines the msg type, not msg version, so every NewBlock version will have the same code. But I wasn't sure to add it in this PR, but now I think I will do it.

src/main/scala/io/iohk/ethereum/network/handshaker/EtcHelloExchangeState.scala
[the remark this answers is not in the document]
pslaski: I agree - current PV64 version should be temporary, only for our testnet usage and ultimately it should be PV65/PV66 (I'm not sure what is last ETC PV)

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV64.scala
```diff
@@ -0,0 +1,132 @@
...
+
+      override def toRLPEncodable: RLPEncodeable = {
+        import msg._
+        RLPList(
+          protocolVersion,
+          networkId,
+          chainWeight.totalDifficulty,
+          bestHash,
+          genesisHash,
+          chainWeight.lastCheckpointNumber
```
rtkaczyk, line 31: This is really minor, but I see you keep insisting on having this as last field in the RLP. I really don't see the point in this. This is not an extension, all the fields could be totally different. It would make much more sense to me to keep relevant fields close together.

src/test/scala/io/iohk/ethereum/network/EtcPeerManagerSpec.scala
```diff
@@ -108,14 +132,29 @@ class EtcPeerManagerSpec extends AnyFlatSpec with Matchers {
...
     //when
     peersInfoHolder ! MessageFromPeer(newBlock, peer1.id)
 
     //then
     requestSender.send(peersInfoHolder, PeerInfoRequest(peer1.id))
-    requestSender.expectMsg(PeerInfoResponse(Some(peer1Info.withChainWeight(newBlock.chainWeight))))
+    requestSender.expectMsg(PeerInfoResponse(Some(peer1Info.withChainWeight(ChainWeight.totalDifficultyOnly(newBlock.totalDifficulty)))))
+  }
+
+  it should "update the peer total difficulty when receiving a PV64.NewBlock" in new TestSetup {
```
rtkaczyk, line 145: total diff -> chain weight ?

src/main/scala/io/iohk/ethereum/network/p2p/MessageDecoders.scala
```diff
@@ -38,45 +35,59 @@ object NetworkMessageDecoder extends MessageDecoder {
...
+      case (_, Codes.NewBlockCode) => payload.toNewBlock
 
       case (PV61, t) => handlePV61(t, payload)
 
-      case (PV62 | PV63, pv62.NewBlockHashes.code) => payload.toNewBlockHashes
-      case (PV62 | PV63, pv62.GetBlockHeaders.code) => payload.toGetBlockHeaders
-      case (PV62 | PV63, pv62.BlockHeaders.code) => payload.toBlockHeaders
-      case (PV62 | PV63, pv62.GetBlockBodies.code) => payload.toGetBlockBodies
-      case (PV62 | PV63, pv62.BlockBodies.code) => payload.toBlockBodies
+      case (PV62 | PV63, Codes.NewBlockHashesCode) => payload.toNewBlockHashes
```
rtkaczyk, line 55: Shouldn't these messages be handled for PV64 as well? This code is getting a bit out of control so I would suggest to refactor as follows: 

```
def fromBytes(protocolVersion, msgCode, payload) = {
  protocolVersion match {
     case PV64 => handlePV64(msgCode, payload)
     case PV63 => handlePV63(msgCode, payload)
     ...
     case PV60 => handlePV60(msgCode, payload)
     case _ => fail()
  }
}

private def handlePV64(msgCode, payload) = msgCode match {
  case pv64msg1 => decode(payload)
  case pv64msg2 => decode(payload)
  case msgObsoletedByPV64 => fail() // not sure if we need this case for any PV**
  ...
  case _ => handlePV63(msg, payload)
}

private def handlePV63(msgCode, payload) = msgCode match {
  case pv63msg1 => decode(payload)
  case pv63msg2 => decode(payload)
  ...
  case _ => handlePV62(msg, payload)
}

...

private def handlePV60(msgCode, payload) = msgCode match {
  case pv60msg1 => decode(payload)
  case pv60msg2 => decode(payload)
  ...
  case _ => fail()
}
```

This introduces [PV60 as the version for common messages](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#eth60-and-below). Perhaps we should even rename `CommonMessages` to `PV60` as this would give us more consistent naming?

pslaski: In my initial version, Status and NewBlock were part of PV63, but I decided later to move them to CommonMessages again. The true is PV61 Status is different than other ones, but we have not supported it and just use that common Status everywhere besides new PV. I don't know if we want to clean it up or not? @ntallar ? If yes, we can do this in the different PR, as this one is already big one.

src/main/scala/io/iohk/ethereum/network/p2p/MessageDecoders.scala
```diff
@@ -38,45 +35,87 @@ object NetworkMessageDecoder extends MessageDecoder {
 // scalastyle:off
 object EthereumMessageDecoder extends MessageDecoder {
 
-  override def fromBytes(msgCode: Int, payload: Array[Byte], protocolVersion: Version): Message =
+  private type MessageHandler = PartialFunction[(Int, Array[Byte]), Message]
```
rtkaczyk, line 38: How is this approach better/cleaner than what I proposed?

pslaski: It is far more composable - with your approach I can't just do `case _ => handlePV63(msg, payload)` because it isn't true, we are handling NewBlock/Status differently than PV63, so it ends up with copying all those PV63 and PV62 messages or doing similar partition as I did and instead of `orElse` use `case _ => nextHandler(msg, payload)` where nextHandler will have another call with wildcard etc. For me it's just the perfect fit for PartialFunction

rtkaczyk: > I can't just do case _ => handlePV63(msg, payload) because it isn't true

What isn't true?

pslaski: ok you right, I can do this with simple matching, I forgot that those new messages would be matched firstly and the bad path is not reachable

src/main/scala/io/iohk/ethereum/network/p2p/MessageDecoders.scala
```diff
@@ -38,45 +35,87 @@ object NetworkMessageDecoder extends MessageDecoder {
...
+
+  override def fromBytes(msgCode: Int, payload: Array[Byte], protocolVersion: Version): Message = {
     (protocolVersion, msgCode) match {
       //wire protocol
       case (_, Hello.code) => payload.toHello
+      case (PV61, _) => pv61Handler(msgCode, payload)
+      case (PV62, _) => pv62Handler(msgCode, payload)
+      case (PV63, _) => pv63Handler(msgCode, payload)
+      case (PV64, _) => pv64Handler(msgCode, payload)
```
ntallar, line 47: Won't this mean that if a peer (from mainnet or mordor) supports ETH/ETC's PV64 we'll fail to decode their NewBlock and Status messages?

Wouldn't it be simpler if we rename our PV64 to PV66? Using PV64 might lead to other confusions like this

ntallar (changes requested):

src/main/scala/io/iohk/ethereum/network/p2p/messages/package.scala
```diff
@@ -3,11 +3,29 @@ package io.iohk.ethereum.network.p2p
...
+  object Codes {
+    val StatusCode: Int = ProtocolVersions.SubProtocolOffset + 0x00
+    val NewBlockHashesCode: Int = ProtocolVersions.SubProtocolOffset + 0x01
+    val SignedTransactionsCode: Int = ProtocolVersions.SubProtocolOffset + 0x02
+    val GetBlockHeadersCode: Int = ProtocolVersions.SubProtocolOffset + 0x03
+    val BlockHeadersCode: Int = ProtocolVersions.SubProtocolOffset + 0x04
+    val GetBlockBodiesCode: Int = ProtocolVersions.SubProtocolOffset + 0x05
+    val BlockBodiesCode: Int = ProtocolVersions.SubProtocolOffset + 0x06
+    val NewBlockCode: Int = ProtocolVersions.SubProtocolOffset + 0x07
+    // This message is removed in PV62 and this code is reused in PV65 with different msg type
```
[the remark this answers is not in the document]
rtkaczyk: Do you see any problem with that? In fact this is what we are doing for `Status` and `NewBlock`. Just because we are calling them the same (names are irrelevant on the protocol level) doesn't change the fact that they are different messages :)

src/main/scala/io/iohk/ethereum/network/p2p/MessageDecoders.scala
pslaski: yes it is, but having the assumption that PV_X includes PV_X-1, we cannot use PV66 too

src/main/scala/io/iohk/ethereum/network/p2p/messages/package.scala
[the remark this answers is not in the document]
pslaski: I don't know, maybe they know that that message is so old and used only in one protocol version that it was prettier to not block that code forever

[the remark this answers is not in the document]
ntallar: Nono, I was just complaining about why ETH did this:
>     // This message is removed in PV62 and this code is reused in PV65 with different msg type

src/test/scala/io/iohk/ethereum/network/p2p/MessageSpec.scala
```diff
@@ -1,77 +1,223 @@
...
+import io.iohk.ethereum.network.p2p.messages.Capability.Capabilities._
+import io.iohk.ethereum.network.p2p.messages.CommonMessages.SignedTransactions
 import io.iohk.ethereum.network.p2p.messages._
+import io.iohk.ethereum.nodebuilder.SecureRandomBuilder
 import org.bouncycastle.util.encoders.Hex
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 
-class MessageSpec extends AnyFlatSpec with Matchers {
+class MessageSpec extends AnyFlatSpec with Matchers with SecureRandomBuilder {
```
ntallar, line 14: I like how this is looking now! Minor, isn't this now a MessagesDecoderSpec?

ntallar (changes requested):
Only major comment regarding the block broadcasting compatibility issue

Apart from this comments, LGTM!

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockBroadcast.scala
```diff
@@ -18,35 +20,39 @@ class BlockBroadcast(val etcPeerManager: ActorRef, syncConfig: SyncConfig) {
...
     newBlock.block.header.number > peerInfo.maxBlockNumber ||
       newBlock.chainWeight > peerInfo.chainWeight
 
-  private def broadcastNewBlock(newBlock: NewBlock, peers: Set[Peer]): Unit =
-    obtainRandomPeerSubset(peers).foreach { peer =>
-      etcPeerManager ! EtcPeerManagerActor.SendMessage(newBlock, peer.id)
+  private def broadcastNewBlock(blockToBroadcast: BlockToBroadcast, peers: Map[Peer, PeerInfo]): Unit =
+    obtainRandomPeerSubset(peers.keySet).foreach { peer =>
+      val message: MessageSerializable =
+        if (peers(peer).remoteStatus.protocolVersion == ProtocolVersions.PV64) blockToBroadcast.as64
```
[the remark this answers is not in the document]
pslaski: It won't as we set Remote status PV to PV64 only when `handshakerConfiguration.protocolVersion == ProtocolVersions.PV64 && hello.capabilities.contains(Etc64Capability)` which means that normal eth PV64 client will still use old PV63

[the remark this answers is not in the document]
ntallar: But aren't we setting the remote status to exactly what's being returned by the other clients? (not for Mantis, but when connected with a geth client for example, that has eth64 implemented)

src/main/scala/io/iohk/ethereum/network/EtcPeerManagerActor.scala
```diff
@@ -229,10 +234,38 @@ class EtcPeerManagerActor(
...
 object EtcPeerManagerActor {
 
-  val msgCodesWithInfo: Set[Int] = Set(BlockHeaders.code, NewBlock.code63, NewBlock.code64, NewBlockHashes.code)
+  val msgCodesWithInfo: Set[Int] = Set(Codes.BlockHeadersCode, Codes.NewBlockCode, Codes.NewBlockHashesCode)
+
+  /**
+    * RemoteStatus was created to decouple status information from protocol status messages
+    * (they are different versions of Status msg)
+    */
+  case class RemoteStatus(
```
ntallar, line 243: Should we add a toString? The current way it's being printed is:
```
2020-11-24 14:26:11,053 [i.i.e.n.h.EtcNodeStatus64ExchangeState] - Peer returned status (RemoteStatus(64,1,ChainWeight(0,262144),ByteString(-12, -125, 74, 124, 9, 60, 73, -55, -58, 101, 74, -3, 83, -97, 106, -91, -82, 55, -70, -51, -63, -39, 36, -14, -116, -15, -86, 13, 48, -82, -61, -110),ByteString(-12, -125, 74, 124, 9, 60, 73, -55, -58, 101, 74, -3, 83, -97, 106, -91, -82, 55, -70, -51, -63, -39, 36, -14, -116, -15, -86, 13, 48, -82, -61, -110)))
```

ntallar (commented):
Apart from the two comments above, LGTM!

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockBroadcast.scala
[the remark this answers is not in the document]
pslaski: no, check https://github.com/input-output-hk/mantis/pull/793/files#diff-1d7ee333b204994a51c452394c8e4731125ff7405048d12ba54e637f4923795bR30 the PV64 will be used only for etc64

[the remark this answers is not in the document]
pslaski: `we setting the remote status to exactly what's being returned by the other clients` - you are right, but remember that we are not sending eth64 capability, so other clients should send eth63 protocol version in their Status msg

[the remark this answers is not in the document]
ntallar: After discussions with Pawel and some research we found out that the protocolVersion is calculated based on the capabilities shared by both peers, so there's no issue with the current approach

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

src/main/resources/application.conf
```diff
@@ -38,6 +38,8 @@ mantis {
 
   network {
     # Ethereum protocol version
+    # Supported versions:
+    # 63, 64 (experimental version which enables usage of messages with checkpointing information. In the future after ETCM-355, ETCM-356, it will be 66 probably)
     protocol-version = 63
```
ntallar, line 43: One last request for this PR, could you set protocol-version to 64 to the `testnet-internal-nomad.conf`?

It would be compliant with the fact that that testnet already has ECIP 1097 enabled

mirkoAlic:
I reviewed this in the following one(https://github.com/input-output-hk/mantis/pull/810), where i wrote some comments, but in general, the code LGTM!

src/main/scala/io/iohk/ethereum/network/p2p/MessageDecoders.scala
```diff
@@ -38,45 +36,72 @@ object NetworkMessageDecoder extends MessageDecoder {
...
+      case pv => throw new RuntimeException("Unknown protocol version: " + pv)
+    }
+  }
 
-      case (PV62 | PV63, pv62.NewBlockHashes.code) => payload.toNewBlockHashes
-      case (PV62 | PV63, pv62.GetBlockHeaders.code) => payload.toGetBlockHeaders
-      case (PV62 | PV63, pv62.BlockHeaders.code) => payload.toBlockHeaders
-      case (PV62 | PV63, pv62.GetBlockBodies.code) => payload.toGetBlockBodies
-      case (PV62 | PV63, pv62.BlockBodies.code) => payload.toBlockBodies
+  private def handleCommonMessages(msgCode: Int, payload: Array[Byte]): Message = {
```
rtkaczyk, line 49: Why do we not want to call them PV60?

src/main/scala/io/iohk/ethereum/network/p2p/MessageDecoders.scala
```diff
@@ -38,45 +36,72 @@ object NetworkMessageDecoder extends MessageDecoder {
...
-      case (_, SignedTransactions.code) => payload.toSignedTransactions
-      case (_, NewBlock.code63 | NewBlock.code64) => payload.toNewBlock(msgCode)
-
-      case (PV61, t) => handlePV61(t, payload)
+  override def fromBytes(msgCode: Int, payload: Array[Byte], protocolVersion: Version): Message = {
+    protocolVersion match {
+      case PV61 => handlePV61(msgCode, payload)
+      case PV62 => handlePV62(msgCode, payload)
+      case PV63 => handlePV63(msgCode, payload)
+      case PV64 => handlePV64(msgCode, payload)
```
rtkaczyk, line 44: Very minor optimization: higher PVs will be matched more often, so they could matched first

src/main/scala/io/iohk/ethereum/network/p2p/MessageDecoders.scala
```diff
@@ -38,45 +36,72 @@ object NetworkMessageDecoder extends MessageDecoder {
...
-
-      //FIXME: I still have an issue with protocolVersion, we are use PV62 and PV63 and code names for Status and NewBlock
-      // suggest that there is PV64, but there isn't
-      //common
-      case (_, Status.code63 | Status.code64) => payload.toStatus(msgCode)
-      case (_, SignedTransactions.code) => payload.toSignedTransactions
-      case (_, NewBlock.code63 | NewBlock.code64) => payload.toNewBlock(msgCode)
-
-      case (PV61, t) => handlePV61(t, payload)
+  override def fromBytes(msgCode: Int, payload: Array[Byte], protocolVersion: Version): Message = {
```
rtkaczyk, line 39: This looks very tidy now <3

src/main/scala/io/iohk/ethereum/network/p2p/MessageDecoders.scala
```diff
@@ -1,35 +1,33 @@
...
   override def fromBytes(msgCode: Int, payload: Array[Byte], protocolVersion: Version): Message =
-    (protocolVersion, msgCode) match {
-      case (_, Disconnect.code) => payload.toDisconnect
-      case (_, Ping.code) => payload.toPing
-      case (_, Pong.code) => payload.toPong
+    msgCode match {
+      case Disconnect.code => payload.toDisconnect
+      case Ping.code => payload.toPing
+      case Pong.code => payload.toPong
+      case Hello.code => payload.toHello
```
rtkaczyk, line 30: <3

src/main/scala/io/iohk/ethereum/network/p2p/MessageDecoders.scala
rtkaczyk: I think `CommonMessages` may be misleading. Since PV64 overrides `Status` and `NewBlock`, they are not common anymore

pslaski: https://github.com/input-output-hk/mantis/pull/793#discussion_r526332016 ;)

pslaski: I can refactor it in ETCM-402

ntallar (approved):

PR #794, opened 2020-11-13
ETCM-365: Version info in Hello.clientId

aakoshh:
# Description

The PR adds the conventional Ethereum node name in the `WireProtocol.Hello.clientId` field, for example: `mantis/v3.0-cd5ae33/linux-amd64/ubuntu-openjdk64bitservervm-java-11.0.9`. 

The `-Dmantis.client-identity` property can be used to splice a user identity into it to distinguish our client, perhaps by a ticket number or just belonging to an organisation. 

I didn't do as an elaborate normalisation as `besu` in the `PlatformDetector`, called from `BesuInfo`; seemed like an overkill.


src/main/resources/application.conf
```diff
@@ -1,6 +1,6 @@
 mantis {
-  # Identifier used when connecting to other clients
-  client-id = "mantis"
+  # Optionally override the client ID sent in Hello messages.
+  client-id = null
```
ntallar, line 3: What would be the use case for an user configuring this manually? Why not using only the generated in VersionInfo?

ntallar (commented):
>I didn't do as an elaborate normalisation as besu in the PlatformDetector, called from BesuInfo; seemed like an overkill.

Agree! Apart from one comment LGTM!

src/main/resources/application.conf
aakoshh: I have no idea, I just didn't want to remove it since the current "mantis" value can also be overwritten. I thought even that maybe the first part of the name should come from here, not from the project name. I think Besu supports some kind of [identity](https://github.com/hyperledger/besu/blob/c8e23c970ebef4297d6e75abec784f7024df340c/besu/src/main/java/org/hyperledger/besu/BesuInfo.java#L33) but it's unclear what that might be. 

Anyway it's up to you, we can remove it completely if you prefer.

aakoshh: Changed it so it's more like besu: if you set it it will be appended to the `mantis` part, so we can distinguish it like `mantis-etc` or whatever.

ntallar: I don't like much the name of this config, the client id would be after the augmentation, not the augmentation alone... maybe something like `client-id-augmentation`?

ntallar (commented):
Apart from a last comment, LGTM!

src/main/resources/application.conf
aakoshh: That's a bit of a mouthful :sweat_smile: 
Should we name it `client-identity` after the besu setting that inspired it?

aakoshh: Renamed, and also changed the format slightly to match besu, based on their [tests](https://github.com/hyperledger/besu/blob/72e975e8e7e415fc396bf26c515fc0f3e8d23fba/besu/src/test/java/org/hyperledger/besu/BesuInfoTest.java#L48-L54) explaining this is the ethstats friendly way.

src/main/scala/io/iohk/ethereum/utils/VersionInfo.scala
```diff
@@ -0,0 +1,48 @@
...
+    * - coregeth/v1.11.8-stable-305b5089/linux-amd64/go1.14.4
+    *
+    * Apparently ethstats expects either 4 parts or 5:
+    * - client/version/os/compiler
+    * - client/identity/version/os/compiler
+    */
+  def nodeName(maybeIdentity: Option[String] = None): String = {
+    val app = {
+      val name = BuildInfo.name
+      val id = maybeIdentity.map("/" + _).getOrElse("") //
```
ntallar, line 21: Minor: should we remove the last //?

src/main/scala/io/iohk/ethereum/utils/VersionInfo.scala
```diff
@@ -0,0 +1,48 @@
...
+      s"$name$id"
+    }
+    val version = {
+      val version = BuildInfo.version
+      val commit = BuildInfo.gitHeadCommit.map("-" + _.take(7)).getOrElse("")
```
ntallar, line 26: Minor: From the example `coregeth/v1.11.8-stable-305b5089/linux-amd64/go1.14.4`, shouldn't it be `take(8)`?

Though I'm not sure why they do something different than GH... and also that change probably has no impact so we can diverge on that

src/main/resources/application.conf
ntallar: Should we maybe copy those tests?

src/main/scala/io/iohk/ethereum/utils/VersionInfo.scala
aakoshh: Huh, thanks, I wanted to add a comment but then added it to the docstring.

src/main/scala/io/iohk/ethereum/utils/VersionInfo.scala
aakoshh: I used 7 to match github. I think the version is largely up to us, I thought it's useful since you can test something without changing the version, and as long as we can use the commit hash in Github to find the exact code version it looks useful.

src/main/resources/application.conf
aakoshh: I thought [this](https://github.com/input-output-hk/mantis/pull/794/files#diff-a00369dbb33d886362f525713ce1b0f7e1ca5a14151127b91ff44c5864e1dc67R10) test captures the gist of it.

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

PR #795, opened 2020-11-15
Fixed `clear` method synchronization

dmitry-worker:
# Description

Method `clear()` in `MapCache` was not synchronized.
`AtomicLong` was redundant because no CAS-related methods are in use.

# Proposed Solution

Add sync to `clear()`
Downgrade `val AtomicRefernce` to `@volatile var`

### DIFF
--- a/src/main/scala/io/iohk/ethereum/db/cache/MapCache.scala
+++ b/src/main/scala/io/iohk/ethereum/db/cache/MapCache.scala
@@ -14,7 +14,7 @@ import scala.concurrent.duration.FiniteDuration
 // Other methods are only called from actor context, and all updates are always visible to them
 class MapCache[K, V](val cache: mutable.Map[K, V], config: NodeCacheConfig) extends Cache[K, V] {
 
-  private val lastClear = new AtomicLong(System.nanoTime())
+  @volatile private var lastClear = System.nanoTime()
 
   override def update(toRemove: Seq[K], toUpsert: Seq[(K, V)]): Cache[K, V] = {
     this.synchronized {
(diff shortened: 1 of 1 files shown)


mmrozek (commented):
WDYT about using `ConcurrentMap` instead of synchronizing all methods?

dmitry-worker:
> WDYT about using `ConcurrentMap` instead of synchronizing all methods?

I would rather do this instead of non-elegant synchronized solution, but we might face a dangerous situation with some of cached values disappearing at the moment they were added to the cache: 

Based on bulk updated method in a `MapCache` we might want to see in cache all the values we put there:

```
  override def update(toRemove: Seq[K], toUpsert: Seq[(K, V)]): Cache[K, V] = this.synchronized {
    toRemove.foreach(key => cache -= key)
    toUpsert.foreach(element => cache += element._1 -> element._2)
    this
  }
```
But what might happen without explicit synchronization, when we leave all the thread-safety to a `concurrent.Map`?
We have a certain risk of someone `clear`ing the cache right in the middle of this `foreach` loop:
```toUpsert.foreach(element => cache += element._1 -> element._2)```

And as a result of this - only a part of values we wanted to store will be actually in the cache at the end of the method invocation!


Moreover, there's even more dangerous point that can result in a loss of a valuable information. 
Take a look at `CachedKeyValueStorage` method that utilizes a cache:
```
  def forcePersist(): Unit = {
    storage.update(Nil, cache.getValues)
       <- what will happen if cache is updated at this point?
    cache.clear()
  }
```
I suggest we have to add a new method to our cache, for example `drain`. 
That method will purge cache contents, returning them as list atomically.
Then we are guaranteed to deliver all its data to a storage.

If that idea makes sense - I'll make a new PR regarding CachedKeyValueStorage issue.

dmitry-worker:
closed due to PR#817

PR #797, opened 2020-11-16
[ETCM-366] White listing consensus

KonradStaniec:
# Description
Non-standard ethereum PoW consensus protocol, which allows restricting list of possible miners.
Main differences from basic PoW consensus protocol:

Each miner, signs header data before mining i.e prepared header without mixHash and Nonce, and appends this
signature to blockheader.extraData field. Only such prepared header is mined upon.

Each validator, checks (in addition to standard blockheader validations):
- if blockheader.extraData field has at most 97 bytes length (32 bytes of standard extraData + 65 bytes for ECDSA signature
- if signature is a valid signature over all blockheader data except: mixHash, Nonce, last 65 bytes of extraData field (those bytes are signature itself)
- if public key recovered from correct signature is contained within allowedMinersPublicKeys set defined for given chain


src/main/scala/io/iohk/ethereum/utils/BlockchainConfig.scala
```diff
@@ -141,13 +143,14 @@ object BlockchainConfig {
       ethCompatibleStorage = ethCompatibleStorage,
       bootstrapNodes = bootstrapNodes,
       checkpointPubKeys = checkpointPubKeys,
+      allowedMinersPublicKeys = allowedMinersPublicKeys,
       ecip1099BlockNumber = ecip1099BlockNumber
     )
   }
   // scalastyle:on method.length
-  private def readCheckpointPubKeys(blockchainConfig: TypesafeConfig): Set[ByteString] = {
+  private def readCheckpointPubKeys(blockchainConfig: TypesafeConfig, path: String): Set[ByteString] = {
```
kapke, line 151: `readPublicKeysSet` then?

src/test/scala/io/iohk/ethereum/consensus/ethash/RestrictedEthashSignerSpec.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
+
+class RestrictedEthashSignerSpec
+    extends AnyFlatSpec
+    with Matchers
+    with ScalaCheckPropertyChecks
+    with ObjectGenerators
+    with SecureRandomBuilder {
+  "RestrictedEthashSigner" should "sign and validate correct header" in {
+    forAll(blockHeaderGen, genKey(secureRandom)) { (header, key) =>
```
kapke, line 18: should similar tests be written against `RestrictedEthashHeaderValidator`?

src/test/scala/io/iohk/ethereum/consensus/ethash/RestrictedEthashSignerSpec.scala
```diff
@@ -0,0 +1,45 @@
...
+      val signedHeader = RestrictedEthashSigner.signHeader(header, key)
+      val keyAsBytes = ByteString.fromArrayUnsafe(
```
kapke, line 20: `io.iohk.ethereum.crypto#keyPairToByteArrays` and friends instead?

src/test/scala/io/iohk/ethereum/consensus/validators/RestrictedEthashBlockHeaderValidatorSpec.scala
```diff
@@ -0,0 +1,162 @@
...
+
+  it should "correctly validate header with valid key" in new TestSetup {
+    val blockHeaderValidator = new RestrictedEthashBlockHeaderValidator(createBlockchainConfig(Set(validKey)))
+    val validationResult = blockHeaderValidator.validate(validHeader, validParent)
+    assert(validationResult == Right(BlockHeaderValid))
+  }
+
+  it should "fail to validate header with invalid key" in new TestSetup {
+    val allowedKey = crypto.generateKeyPair(secureRandom)
+    val keyBytes = ByteString.fromArrayUnsafe(
```
kapke, line 46: ditto

src/test/scala/io/iohk/ethereum/consensus/ethash/RestrictedEthashSignerSpec.scala
```diff
@@ -0,0 +1,45 @@
...
+          .drop(1)
+      )
+      assert(RestrictedEthashSigner.validateSignature(signedHeader, Set(keyAsBytes)))
+    }
+  }
+
+  it should "fail to validate header signed with wrong key" in {
+    forAll(blockHeaderGen, genKey(secureRandom), genKey(secureRandom)) { (header, correctKey, wrongKey) =>
+      val signedHeader = RestrictedEthashSigner.signHeader(header, correctKey)
+      val wrongKeyAsBytes = ByteString.fromArrayUnsafe(
```
kapke, line 34: ditto

src/test/scala/io/iohk/ethereum/consensus/validators/RestrictedEthashBlockHeaderValidatorSpec.scala
```diff
@@ -0,0 +1,162 @@
...
+
+    // correct header is signed by different key that the one generated here
+    val blockHeaderValidator = new RestrictedEthashBlockHeaderValidator(createBlockchainConfig(Set(keyBytes)))
+    val validationResult = blockHeaderValidator.validate(validHeader, validParent)
+    assert(validationResult == Left(HeaderExtraDataError))
+  }
+
+  it should "fail to validate header re-signed by valid signer" in new TestSetup {
+    val allowedKey = crypto.generateKeyPair(secureRandom)
+    val keyBytes = ByteString.fromArrayUnsafe(
```
kapke, line 62: ditto

kapke (commented):
Mostly minor comments. Overall looks good!

src/main/scala/io/iohk/ethereum/consensus/ethash/validators/RestrictedEthashBlockHeaderValidator.scala
```diff
@@ -0,0 +1,32 @@
...
+
+class RestrictedEthashBlockHeaderValidator(blockchainConfig: BlockchainConfig)
+    extends EthashBlockHeaderValidator(blockchainConfig) {
+
+  val ExtraDataMaxSize = BlockHeaderValidator.MaxExtraDataSize + ECDSASignature.EncodedLength
+
+  override protected def validateExtraData(blockHeader: BlockHeader): Either[BlockHeaderError, BlockHeaderValid] = {
+    val tooLargeExtraData = blockHeader.extraData.length > ExtraDataMaxSize
+
+    if (tooLargeExtraData) {
```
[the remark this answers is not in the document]
KonradStaniec: I am not great fan of this as i think `if-else` is more ubiquitous accros programming world, but I see your point that this stacking could be improved upon.  I will refactor it to make it more readable.

src/main/scala/io/iohk/ethereum/consensus/ethash/validators/RestrictedEthashBlockHeaderValidator.scala
```diff
@@ -0,0 +1,32 @@
...
+import io.iohk.ethereum.crypto.ECDSASignature
+import io.iohk.ethereum.domain.BlockHeader
+import io.iohk.ethereum.utils.BlockchainConfig
+
+class RestrictedEthashBlockHeaderValidator(blockchainConfig: BlockchainConfig)
+    extends EthashBlockHeaderValidator(blockchainConfig) {
+
+  val ExtraDataMaxSize = BlockHeaderValidator.MaxExtraDataSize + ECDSASignature.EncodedLength
+
+  override protected def validateExtraData(blockHeader: BlockHeader): Either[BlockHeaderError, BlockHeaderValid] = {
```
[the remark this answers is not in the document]
KonradStaniec: I shouldn't as:
 - `RestritedEthash `should only be used in some private nets so validating `validateDaoForkExtraData` makes little sense
-  both validator require different maximal sizes restricted has max set to `97` bytes, and normal has max set to `32` bytes

src/test/scala/io/iohk/ethereum/consensus/ethash/RestrictedEthashSignerSpec.scala
KonradStaniec: tbh i am not sure that is possible (or desirable) as `RestrictedEthashBlockHeaderValidator` inherits from `EthashBlockHeaderValidator` which mean method `validate` validates proof of work therefore to property test it one should mine valid blocks which would take too much time.

Thats why I had split that way:
- property test for signing and validating to test various different inputs 
- `RestrictedEthashBlockHeaderValidatorSpec` which tests different `if-else` parts from `RestrictedEthashBlockHeaderValidator` plus the case which test against someone who would try to re-sign header after pow is mined

src/main/scala/io/iohk/ethereum/consensus/ethash/validators/RestrictedEthashBlockHeaderValidator.scala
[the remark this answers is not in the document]
kapke: Then a comment would be welcome. I mean - I see `RestrictedEthash` as ethash + whitelist of miners. If something regarding extra data changes in ethash - that change won't be reflected here (but it should IMO). 
Maybe that means that validators need to be more modular and only composed where full consensus is built? Definitely not something for this PR, but for a future improvement.

kapke (approved):
LGTM!

src/main/scala/io/iohk/ethereum/consensus/ethash/RestrictedEthashSigner.scala
```diff
@@ -0,0 +1,33 @@
...
+    val maybePubKey = for {
+      sig <- ECDSASignature.fromBytes(signature)
+      pubKeyBytes <- sig.publicKey(headerHash)
+    } yield ByteString.fromArrayUnsafe(pubKeyBytes)
+
+    maybePubKey.exists(allowedMiners.contains)
+  }
+
+  def signHeader(blockHeader: BlockHeader, keyPair: AsymmetricCipherKeyPair): BlockHeader = {
+    val encoded = getEncodedWithoutNonce(blockHeader)
```
ntallar, line 26: Why without the nonce and mixHash?

src/main/scala/io/iohk/ethereum/consensus/ethash/validators/RestrictedEthashBlockHeaderValidator.scala
```diff
@@ -0,0 +1,33 @@
...
+    extends EthashBlockHeaderValidator(blockchainConfig) {
+
+  val ExtraDataMaxSize = BlockHeaderValidator.MaxExtraDataSize + ECDSASignature.EncodedLength
+
+  private def validateSignatureAgainstAllowedMiners(
+      blockHeader: BlockHeader,
+      allowedMiners: Set[ByteString]
+  ): Either[BlockHeaderError, BlockHeaderValid] = {
+    val emptyOrValid = allowedMiners.isEmpty || RestrictedEthashSigner.validateSignature(blockHeader, allowedMiners)
+    Either.cond(emptyOrValid, BlockHeaderValid, HeaderExtraDataError)
```
ntallar, line 21: Should we maybe have a different BlockHeaderError for this case?

ntallar (changes requested):
Only some minor comments, worked well while testing it! Will take a look at the tests next week

src/main/scala/io/iohk/ethereum/consensus/ethash/blocks/RestrictedEthashBlockGeneratorImpl.scala
```diff
@@ -0,0 +1,69 @@
...
+      validators,
+      blockchain,
+      blockchainConfig,
+      consensusConfig,
+      blockPreparator,
+      difficultyCalc,
+      blockTimestampProvider
+    ) {
+
+  override def generateBlock(
```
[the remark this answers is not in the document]
KonradStaniec: It is really similar, but I wanted avoid adding refactors here. Also those are conceptually two different protocols so it makes sense for them to have different `genereateBlock` methods.

src/main/scala/io/iohk/ethereum/consensus/ethash/RestrictedEthashSigner.scala
KonradStaniec: Becouse those are modified by mining i.e mining is always done on whole block data except `nonce` and `mixHash`, and correct values are of `nonce` and `mixHash` are provided by mining result.

src/test/scala/io/iohk/ethereum/consensus/validators/RestrictedEthashBlockHeaderValidatorSpec.scala
```diff
@@ -0,0 +1,149 @@
...
+      gasUsed = 0,
+      unixTimestamp = 1605514463,
+      extraData = ByteStringUtils.string2hash(
+        "6d616e746973808fc245b89183f28ac985019992f202a73c7ab600b0aefa18dcba71a8f3576129280d56f4f499e7a8a53a047e91d73d881745b7a6ac7ca9449fc2b3bb1608921c"
+      ),
+      mixHash = ByteStringUtils.string2hash("2db10efede75cfe87b6f378d9b03e712098e8cd3759784db56d65cc9e9911675"),
+      nonce = ByteStringUtils.string2hash("a57246871d5c8bcc")
+    )
+
+    val validHeader = BlockHeader(
```
ntallar, line 128: Another hardcoded header :confused: can we use one of the Fixtures ones and maybe extend it with a signature?

src/test/scala/io/iohk/ethereum/consensus/ethash/RestrictedEthashSignerSpec.scala
ntallar: On the BlockHeaderValidatorSpec we used:
```
class BlockValidatorWithPowMocked(blockchainConfig: BlockchainConfig)
      extends BlockHeaderValidatorSkeleton(blockchainConfig) {
    override protected def difficulty: DifficultyCalculator = difficultyCalculator

    override def validateEvenMore(
        blockHeader: BlockHeader,
        parentHeader: BlockHeader
    ): Either[BlockHeaderError, BlockHeaderValid] = Right(BlockHeaderValid)
  }
```
Something like that might be useful for what you are saying

ntallar (changes requested):
Only a minor comment, apart from which LGTM!

src/main/scala/io/iohk/ethereum/consensus/ethash/RestrictedEthashSigner.scala
ntallar: Oh, because we are obtaining the correct nonce and mixHash after this... we could have done the other way around, but I see no problem between either alternative, it will always have been generated by a whitelisted miner, aka a trusted one

src/test/scala/io/iohk/ethereum/consensus/validators/RestrictedEthashBlockHeaderValidatorSpec.scala
KonradStaniec: nope, if we extend the signature the pow validation will fail. We could mock pow validation, but the whole point of this test is to test the whole validation and pow validation of extended extraData filed. I will add comment there about the fact that those are special headers (with extended extraData field) and they should only be used in those tests.

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

PR #801, opened 2020-11-17
[Chore] Update testnet internal configs

ntallar:
# Description

Updates configurations due to requirements from both gac and nomad testnets:
- Updated gac's genesis with new address
- Configured genesis + bootstrap for connecting to the nomad testnet

# Proposed Solution

As both gac and nomad testnet are diverging (different bootstrap and genesis), totally separate configurations will be used for them.

# Testing

- Start Mantis with testnet-internal-nomad configured and see that the node starts syncing to the testnet


mmrozek (approved):
LGTM. Works as expected

PR #804, opened 2020-11-20
[ETCM-197] Move all fast sync components to separate package

KonradStaniec:
# Description

Moves all fast sync components to separate package


mmrozek (commented):
Looks good, only minor things:
- `BlockBroadcast` could be moved to `regular` pkg
- we could rename `FastSyncReceiptsValidator`, `FastSyncStateStorageActor`, `FastSyncMetrics`. `FastSync` prefix isn't needed anymore

mmrozek (approved):
LGTM!

PR #805, opened 2020-11-20
[FIX] Handle timeouts on BlockFetcher

ntallar:
# Description

Address forking issue on the testnet

## Scenario

During the startup of the nomad cluster there were some connectivity problems and some connections got closed resulting in timeouts on the requests. However for mantis-5
1. He sent a request for a block header to a peer whose connection was closing (or soon to be closed)
2. The request eventually died due to a timeout (or at least that's my theory), we can see that it died as the requesters logging of PeersClient reached 0 at some moment.
3. But the BlockFetcher never received that timeout, so it was permanently stuck on `fetching headers == true`

# Proposed Solution

Recover from ask pattern timeout on the BlockFetcher by using the error fallback, which results in the request being cancelled and a new one triggered

# Testing

Added unit test

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
@@ -44,7 +44,7 @@ class BlockFetcher(
   import BlockFetcher._
 
   implicit val ec: ExecutionContext = context.dispatcher
-  implicit val timeout: Timeout = syncConfig.peerResponseTimeout + 1.second // some margin for actor communication
+  implicit val timeout: Timeout = syncConfig.peerResponseTimeout + 2.second // some margin for actor communication
 
   override def receive: Receive = idle()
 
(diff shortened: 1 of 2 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -44,7 +44,7 @@ class BlockFetcher(
   import BlockFetcher._
 
   implicit val ec: ExecutionContext = context.dispatcher
-  implicit val timeout: Timeout = syncConfig.peerResponseTimeout + 1.second // some margin for actor communication
+  implicit val timeout: Timeout = syncConfig.peerResponseTimeout + 2.second // some margin for actor communication
```
biandratti, line 47: Add value if we put this time margin as config?

ntallar: Hmm would we ever want to change this again? Maybe we could have a general mantis configuration of akka ask timeout margins (as we are using this 2 seconds in other places)

Either way I'd do that on a separate task

biandratti: That sound good

biandratti (approved):

mirkoAlic (approved):
:+1:

PR #768, opened 2020-10-29
[ETCM-251] faucet healthcheck

biandratti:
# Description

Add json rpc behavior in “faucet”. So we change httpServer to jsonRpcServer.
- We change the resource from sendFunds. 
- And add a new resource from the status of “faucet”, that today it’s return value “FaucetUnavailable”. In other tasks we complete the status from faucet.

# Proposed Solution

I added a new JsonRpc controller called "JsonRpcBaseController'', it's extended for the clases JsonRpcController and FaucetJsonRpcController.  It is the way to decouple both controllers.

# Important Changes Introduced

- Change the endpoint from the request send funds of "faucet". Before it worked with "rest", now it will work with "json rpc".
Before => [POST] {host}/faucet?address={account_id}
Today => We use json rpc with "method": "faucet_sendFunds" and account_id as parameters.
New endpoint was documented in the file "insomnia_workspace.json".

- Before  for the resource "send Funds'' we had a validation for the next request from the same ip address. We validated the limit time between requests form the same ip address. This validation was removed because it is not necessary, this sholud be part of DNS protection.

# Testing

**json Rpc regression**

PR #806, opened 2020-11-20
[ETCM-331] Add Rate Limit for rpc requests

SoseMagarian:
# Description

In a [previous refactor](https://github.com/input-output-hk/mantis/pull/768), IP limit tracking was removed due to the nature of the refactor and because this isn't a problem that mantis codebase should be handling. As a temporary solution, this PR brings back the IP limit tracking for `faucet_sendFunds` requests.

# Proposed Solution

## First approach _[Deprecated]_

Two new traits extending the `JsonRpcHttpServer` were created to separate the behavior for the node and the faucet
1. `FaucetJsonRpcHttpServer `
2. `NodeJsonRpcHttpServer`

Their child classes (`BasicJsonRpcHttpServer` and `JsonRpcHttpsServer`) also need to be extended for the node and for the faucet. The main goal is being able to handle the `route` val and extract the Ip and limit the requests by ip, but only for the faucet.

## Final Approach
Add ip limit tracking as a JsonRpc capability. It can be enabled/disabled by configuration. 
Some decisions to take into consideration: 
- Batch requests are disabled with IP-restriction enabled
- Requesting faucet's status does not add the ip to the tracked ips

# Testing

Run the faucet with a node and try to execute requests in less than the minimum time interval allowed between requests (10 seconds by default). Also, verify that `faucet_status` has no requests limit and should not affect the `faucet_sendFunds` requests.


src/main/scala/io/iohk/ethereum/jsonrpc/server/http/FaucetJsonRpcHttpServer.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  val latestTimestampCacheSize: Int
+
+  val latestRequestTimestamps: LruMap[RemoteAddress, Long]
+
+  val clock: Clock = Clock.systemUTC()
+
+  override def route: Route = cors(corsSettings) {
+    (pathEndOrSingleSlash & post) {
+      (extractClientIP & entity(as[JsonRpcRequest])) { (clientAddress: RemoteAddress, request: JsonRpcRequest) =>
```
dmitry-worker, line 31: Is it enough to check IP address only in HTTP header, or maybe we shouldn't trust them?
Maybe we have to read `AttributeKeys.remoteAddress`? What do you think?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/FaucetJsonRpcHttpServer.scala
```diff
@@ -0,0 +1,97 @@
...
+class FaucetBasicJsonRpcHttpServer(
+    jsonRpcController: JsonRpcBaseController,
+    jsonRpcHealthChecker: JsonRpcHealthChecker,
+    config: JsonRpcHttpServerConfig
+)(implicit actorSystem: ActorSystem)
+    extends BasicJsonRpcHttpServer(jsonRpcController, jsonRpcHealthChecker, config)(actorSystem)
+    with FaucetJsonRpcHttpServer {
+
+  //FIXME: these 2 variables should be retrieved from a config file
+  override val minRequestInterval: Int = 10000
```
biandratti, line 77: This config would be part of the JsonRpcHttpServerConfig. Maybe as "Rate Limiting"

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/BasicJsonRpcHttpServer.scala
```diff
@@ -11,7 +11,7 @@ import io.iohk.ethereum.utils.Logger
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.util.{Failure, Success}
 
-class BasicJsonRpcHttpServer(
+abstract class BasicJsonRpcHttpServer(
```
biandratti, line 14: I think that maybe we can invert the next class name: "BasicJsonRpcHttpServer" with "JsonRpcHttpServer". What do you think?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/FaucetJsonRpcHttpServer.scala
```diff
@@ -0,0 +1,97 @@
...
+    config.mode match {
+      case "http" =>
+        Right(new FaucetBasicJsonRpcHttpServer(jsonRpcController, jsonRpcHealthchecker, config)(actorSystem))
+      case "https" =>
+        Right(new FaucetJsonRpcHttpsServer(jsonRpcController, jsonRpcHealthchecker, config, secureRandom)(actorSystem))
+      case _ => Left(s"Cannot start JSON RPC server: Invalid mode ${config.mode} selected")
+    }
+}
+
+class FaucetBasicJsonRpcHttpServer(
```
biandratti, line 68: I don't like this class name. What do you think?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/FaucetJsonRpcHttpServer.scala
```diff
@@ -0,0 +1,97 @@
...
+    with FaucetJsonRpcHttpServer {
+
+  //FIXME: these 2 variables should be retrieved from a config file
+  override val minRequestInterval: Int = 10000
+  override val latestTimestampCacheSize: Int = 1024
+
+  override val latestRequestTimestamps = new LruMap[RemoteAddress, Long](latestTimestampCacheSize)
+}
+
+class FaucetJsonRpcHttpsServer(
```
biandratti, line 83: The same, I don't like this class name. What do you think?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/NodeJsonRpcHttpServer.scala
```diff
@@ -0,0 +1,95 @@
...
+    config.mode match {
+      case "http" => Right(new NodeBasicJsonRpcHttpServer(jsonRpcController, jsonRpcHealthchecker, config)(actorSystem))
+      case "https" =>
+        Right(new NodeJsonRpcHttpsServer(jsonRpcController, jsonRpcHealthchecker, config, secureRandom)(actorSystem))
+      case _ => Left(s"Cannot start JSON RPC server: Invalid mode ${config.mode} selected")
+    }
+
+}
+
+class NodeBasicJsonRpcHttpServer(
```
biandratti, line 80: The same, I don't like this class name. What do you think?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/NodeJsonRpcHttpServer.scala
```diff
@@ -0,0 +1,95 @@
...
+    jsonRpcController: JsonRpcBaseController,
+    jsonRpcHealthChecker: JsonRpcHealthChecker,
+    config: JsonRpcHttpServerConfig
+)(implicit actorSystem: ActorSystem)
+    extends BasicJsonRpcHttpServer(jsonRpcController, jsonRpcHealthChecker, config)(actorSystem)
+    with NodeJsonRpcHttpServer {}
+
+class NodeJsonRpcHttpsServer(
```
biandratti, line 88: The same, I don't like this class name. What do you think?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/FaucetJsonRpcHttpServer.scala
```diff
@@ -0,0 +1,97 @@
...
+        handleRequest(clientAddress, request)
+      }
+    }
+  }
+
+  def handleRequest(clientAddr: RemoteAddress, request: JsonRpcRequest): StandardRoute = {
+    val timeMillis = clock.instant().toEpochMilli
+
+    if (request.method == FaucetJsonRpcController.SendFunds) {
+      val latestRequestTimestamp = latestRequestTimestamps.getOrElse(clientAddr, 0L)
```
biandratti, line 41: This behavior can be encapsulated. Adding unit test

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/BasicJsonRpcHttpServer.scala
SoseMagarian: I agree!

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/FaucetJsonRpcHttpServer.scala
SoseMagarian: HI! Thanks for the comment! I think you're right. I'll create a different task for it (since it might escape the scope of this one -> just bringing back old behavior)

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -31,6 +36,16 @@ trait JsonRpcHttpServer extends Json4sSupport {
 
   def corsAllowedOrigins: HttpOriginMatcher
 
+  val minRequestInterval: FiniteDuration
```
biandratti, line 39: We can add in JsonRpcHttpServer the val config: JsonRpcHttpServerConfig.
So, in this way, we don't need the next params:
 
 
-   val minRequestInterval: FiniteDuration
-   val latestTimestampCacheSize: Int
-   val latestRequestTimestamps: LruMap[RemoteAddress, Long]
-   val ipTrackingEnabled: Boolean
 
Example:
 
trait JsonRpcHttpServer extends Json4sSupport {
  val jsonRpcController: JsonRpcBaseController
  val jsonRpcHealthChecker: JsonRpcHealthChecker
  val config: JsonRpcHttpServerConfig
 
And we access these values through the config.

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/InsecureJsonRpcHttpServer.scala
```diff
@@ -2,23 +2,34 @@ package io.iohk.ethereum.jsonrpc.server.http
...
-class BasicJsonRpcHttpServer(
+class InsecureJsonRpcHttpServer(
     val jsonRpcController: JsonRpcBaseController,
     val jsonRpcHealthChecker: JsonRpcHealthChecker,
     config: JsonRpcHttpServerConfig
 )(implicit val actorSystem: ActorSystem)
     extends JsonRpcHttpServer
     with Logger {
 
+  override val ipTrackingEnabled: Boolean = config.ipTrackingEnabled
```
biandratti, line 25: With the proposal of JsonRpcHttpServer, we need to add the next val, and remove the overridden values:

val config: JsonRpcHttpServerConfig

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/SecureJsonRpcHttpServer.scala
```diff
@@ -25,6 +28,14 @@ class JsonRpcHttpsServer(
     extends JsonRpcHttpServer
     with Logger {
 
+  override val ipTrackingEnabled: Boolean = config.ipTrackingEnabled
```
biandratti, line 31: (ditto InsecureJsonRpcHttpServer)
With the proposal of JsonRpcHttpServer, we need to add the next val, and remove the overridden values:

val config: JsonRpcHttpServerConfig

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -135,6 +171,13 @@ object JsonRpcHttpServer extends Logger {
         override val port: Int = rpcHttpConfig.getInt("port")
 
         override val corsAllowedOrigins = ConfigUtils.parseCorsAllowedOrigins(rpcHttpConfig, "cors-allowed-origins")
+
+        private val ipTrackingConfig = rpcHttpConfig.getConfig("ip-restriction")
```
biandratti, line 175: We can add new objects to encapsulate behavior of rate limit.

src/main/resources/application.conf
```diff
@@ -201,6 +201,19 @@ mantis {
         # Domains allowed to query RPC endpoint. Use "*" to enable requests from
         # any domain.
         cors-allowed-origins = []
+
+        # Ip Limit tracking for JSON-RPC requests
+        # Limits the amount of request the same ip can perform in a given amount of time
+        ip-restriction {
```
biandratti, line 207: Maybe we can change "ip-restriction" to "rate-limit"? What do you think?

src/universal/conf/faucet.conf
```diff
@@ -117,6 +117,19 @@ mantis {
         # Domains allowed to query RPC endpoint. Use "*" to enable requests from
         # any domain.
         cors-allowed-origins = []
+
+        # Ip Limit tracking for JSON-RPC requests
+        # Limits the amount of request the same ip can perform in a given amount of time
+        ip-restriction {
```
biandratti, line 123: ditto

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -50,14 +65,30 @@ trait JsonRpcHttpServer extends Json4sSupport {
...
   }
 
+  def handleRequest(clientAddress: RemoteAddress, request: JsonRpcRequest): StandardRoute = {
+    if (ipTrackingEnabled && request.method != FaucetJsonRpcController.Status) {
+      handleRestrictedRequest(clientAddress, request)
+    } else complete(jsonRpcController.handleRequest(request).runToFuture)
+  }
+
+  def handleRestrictedRequest(clientAddress: RemoteAddress, request: JsonRpcRequest): StandardRoute = {
+    val timeMillis = clock.instant().toEpochMilli
```
biandratti, line 83: Maybe we can move this validation to another class. What do you think?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -31,6 +36,16 @@ trait JsonRpcHttpServer extends Json4sSupport {
 
   def corsAllowedOrigins: HttpOriginMatcher
 
+  val minRequestInterval: FiniteDuration
+
+  val latestTimestampCacheSize: Int
+
+  val latestRequestTimestamps: LruMap[RemoteAddress, Long]
```
ntallar, line 43: Maybe this is included in the above comment by Maxi, but also this can be derived from the config here without having to override it on it's sons

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -50,14 +65,30 @@ trait JsonRpcHttpServer extends Json4sSupport {
...
+      (extractClientIP & entity(as[JsonRpcRequest])) { (clientAddress, request) =>
+        handleRequest(clientAddress, request)
       } ~ entity(as[Seq[JsonRpcRequest]]) { request =>
         handleBatchRequest(request)
       }
     }
   }
 
+  def handleRequest(clientAddress: RemoteAddress, request: JsonRpcRequest): StandardRoute = {
+    if (ipTrackingEnabled && request.method != FaucetJsonRpcController.Status) {
```
ntallar, line 77: Why are we only disabling this for the faucet status? Maybe it's because it's used as the healthcheck? I though the healthcheck was something separately

src/test/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServerSpec.scala
```diff
@@ -92,13 +100,119 @@ class JsonRpcHttpServerSpec extends AnyFlatSpec with Matchers with ScalatestRout
...
+
+    postRequest ~> Route.seal(mockJsonRpcHttpServerWithIpRestriction.route) ~> check {
+      status shouldEqual StatusCodes.OK
+      responseAs[String] shouldEqual """{"jsonrpc":"2.0","result":"this is a response","id":1}"""
+    }
+    postRequest ~> Route.seal(mockJsonRpcHttpServerWithIpRestriction.route) ~> check {
+      status shouldEqual StatusCodes.TooManyRequests
+    }
+
+    FakeClock.advanceTime(10)
```
ntallar, line 170: Minor: maybe we should set this to `2* config.minRequestInterval` instead of the hardcoded 10?

src/test/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServerSpec.scala
```diff
@@ -118,9 +237,45 @@ class JsonRpcHttpServerSpec extends AnyFlatSpec with Matchers with ScalatestRout
...
+      override val latestRequestTimestamps = new LruMap[RemoteAddress, Long](latestTimestampCacheSize)
+
+      def run(): Unit = ()
+
+      override def corsAllowedOrigins: HttpOriginMatcher = config.corsAllowedOrigins
+    }
+  }
+}
+
+object FakeClock extends Clock {
```
ntallar, line 268: Maybe this should be a `class FakeClock extends Clock` and on the TestSetup we create an instance of it instead of using this singleton?

ntallar (changes requested):

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -50,14 +56,29 @@ trait JsonRpcHttpServer extends Json4sSupport {
...
+    //FIXME: FaucetJsonRpcController.Status should be part of a Healthcheck request or alike.
+    // As a temporary solution, it is being excluded from the Rate Limit.
+    if (config.rateLimit.enabled && request.method != FaucetJsonRpcController.Status) {
+      handleRateLimitedRequest(clientAddress, request)
+    } else complete(jsonRpcController.handleRequest(request).runToFuture)
+  }
+
+  def handleRateLimitedRequest(clientAddress: RemoteAddress, request: JsonRpcRequest): StandardRoute = {
+    if (isRequestAvailable(clientAddress)) {
+      log.warn(s"Request limit exceeded for ip 1 ${clientAddress.toIP.getOrElse("unknown")}")
```
biandratti, line 77: 1?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/RateLimit.scala
```diff
@@ -0,0 +1,26 @@
+package io.iohk.ethereum.jsonrpc.server.http
+
+import java.time.Clock
+
+import akka.http.scaladsl.model.RemoteAddress
+import com.twitter.util.LruMap
+import io.iohk.ethereum.jsonrpc.server.http.JsonRpcHttpServer.JsonRpcHttpServerConfig
+import io.iohk.ethereum.utils.Logger
+
+trait RateLimit extends Logger {
```
biandratti, line 10: we can remove "extendes Logger", is not used.

biandratti, line 10: we can remove "extendes Logger", is not used.

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/SecureJsonRpcHttpServer.scala
```diff
@@ -1,24 +1,24 @@
...
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.util.{Failure, Success}
 
-class JsonRpcHttpsServer(
+class SecureJsonRpcHttpServer(
     val jsonRpcController: JsonRpcBaseController,
     val jsonRpcHealthChecker: JsonRpcHealthChecker,
-    config: JsonRpcHttpServerConfig,
+    override val config: JsonRpcHttpServerConfig,
```
biandratti, line 21: override?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
SoseMagarian: I added a `FIXME` for this, the faucet Status should somehow work as a healthcheck so I left it out of the rate limiting in case it's being used by some of our services. Eventually it should be part of the healthcheck or be handled more elegantly

src/test/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServerSpec.scala
```diff
@@ -92,35 +102,193 @@ class JsonRpcHttpServerSpec extends AnyFlatSpec with Matchers with ScalatestRout
...
-    }
+    val mockJsonRpcHttpServer = new FakeJsonRpcHttpServer(
+      mockJsonRpcController,
+      mockJsonRpcHealthChecker,
+      serverConfig,
+      serverConfig.corsAllowedOrigins,
+      fakeClock)
 
     val corsAllowedOrigin = HttpOrigin("http://localhost:3333")
+    val mockJsonRpcHttpServerWithCors = new FakeJsonRpcHttpServer(
```
biandratti, line 251: val mockJsonRpcHttpServerWithCors = new FakeJsonRpcHttpServer(
      jsonRpcController= mockJsonRpcController,
     jsonRpcHealthChecker=  mockJsonRpcHealthChecker,
      config= serverConfig,
      cors=  HttpOriginMatcher(corsAllowedOrigin),
      testClock =fakeClock)

src/test/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServerSpec.scala
```diff
@@ -92,35 +102,193 @@ class JsonRpcHttpServerSpec extends AnyFlatSpec with Matchers with ScalatestRout
...
+      mockJsonRpcController,
+      mockJsonRpcHealthChecker,
+      serverConfig,
+      HttpOriginMatcher(corsAllowedOrigin),
+      fakeClock)
+
+    val mockJsonRpcHttpServerWithRateLimit = new FakeJsonRpcHttpServer(
```
biandratti, line 258: ditto

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/SecureJsonRpcHttpServer.scala
SoseMagarian: Missed that, thanks!

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/RateLimit.scala
```diff
@@ -0,0 +1,25 @@
...
+
+  val latestRequestTimestamps = new LruMap[RemoteAddress, Long](config.rateLimit.latestTimestampCacheSize)
+
+  val clock: Clock = Clock.systemUTC()
+
+  def isRequestAvailable(clientAddress: RemoteAddress): Boolean = {
+    val timeMillis = clock.instant().toEpochMilli
+    val latestRequestTimestamp = latestRequestTimestamps.getOrElse(clientAddress, 0L)
+
+    val response = latestRequestTimestamp + config.rateLimit.minRequestInterval.toMillis < timeMillis
```
ntallar, line 21: Minor: maybe we should rename it: `isBelowRateLimit`?

ntallar (approved):
Only a minor comment apart from which LGTM! I played around with it and it worked as expected!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

biandratti (approved):

PR #808, opened 2020-11-20
[ETCM-135] Tx history improvements for wallet

kapke:
# Description

Move transaction history method to `mantis_` namespace, and add couple of additional informations a requested by Mantis Wallet devs.

It's a high priority because it's quite important from wallet perspective.

# Testing

1. all unit tests should pass
2. check if method complains about too large number of blocks within range
3. check whether method finds transactions sent or received by given address (only direct sender or direct receiver is supported so far), example tx on mordor: https://blockscout.com/etc/mordor/tx/0x6249ce7bae7e6c57b08f4a9333bd335a9c001b648e2b62b30c4cea8e80fa9842/internal-transactions, it's a tx from faucet to my testing address

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
@@ -4,14 +4,12 @@ import akka.util.ByteString
 import io.iohk.ethereum.jsonrpc.EthService._
 import io.iohk.ethereum.jsonrpc.JsonRpcError.InvalidParams
 import io.iohk.ethereum.jsonrpc.PersonalService.{SendTransactionRequest, SendTransactionResponse, SignRequest}
+import io.iohk.ethereum.jsonrpc.serialization.JsonEncoder.OptionToNull._
 import io.iohk.ethereum.jsonrpc.serialization.JsonMethodDecoder.NoParamsMethodDecoder
 import io.iohk.ethereum.jsonrpc.serialization.{JsonEncoder, JsonMethodCodec, JsonMethodDecoder}
 import org.json4s.JsonAST.{JArray, JBool, JString, JValue, _}
 import org.json4s.JsonDSL._
-import org.json4s.{Extraction, JsonAST, Merge}
-import JsonEncoder.OptionToNull._
-import io.iohk.ethereum.transactions.TransactionHistoryService.ExtendedTransactionData
-import org.json4s
+import org.json4s.{Extraction, JsonAST}
 
 // scalastyle:off number.of.methods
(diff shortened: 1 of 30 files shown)


src/main/scala/io/iohk/ethereum/transactions/TransactionHistoryService.scala
```diff
@@ -0,0 +1,143 @@
...
+      fromBlocks: NumericRange[BigInt]
+  ): Task[List[ExtendedTransactionData]] = {
+    val txnsFromBlocks = Observable
+      .from(fromBlocks.reverse)
+      .mapParallelOrdered(10)(blockNr => Task { blockchain.getBlockByNumber(blockNr) })(OverflowStrategy.Unbounded)
+      .collect { case Some(block) => block }
+      .concatMap { block =>
+        val getBlockReceipts = Task {
+          blockchain.getReceiptsByHash(block.hash).map(_.toVector).getOrElse(Vector.empty)
+        }.memoizeOnSuccess
```
mmrozek, line 37: `memoizeOnSuccess` won't work here

src/test/scala/io/iohk/ethereum/jsonrpc/EthServiceSpec.scala
```diff
@@ -1221,7 +1146,7 @@ class EthServiceSpec
 
     val jsonRpcConfig = JsonRpcConfig(Config.config, available)
 
-    val ethService = new EthService(
+    lazy val ethService = new EthService(
```
mmrozek, line 1149: Why? I think we should avoid laziness when starting an app (to fail fast if sth is missing)

src/main/scala/io/iohk/ethereum/transactions/TransactionHistoryService.scala
kapke: it will work on a per-block basis, won't it?

mmrozek: You're right. I missed it

mmrozek (approved):
LGTM

src/main/scala/io/iohk/ethereum/jsonrpc/EthJsonMethodsImplicits.scala
```diff
@@ -12,6 +13,7 @@ import org.json4s.{Extraction, JsonAST}
 
 // scalastyle:off number.of.methods
 object EthJsonMethodsImplicits extends JsonMethodsImplicits {
+  implicit val transactionResponseJsonEncoder: JsonEncoder[TransactionResponse] = Extraction.decompose(_)
```
lemastero, line 16: 👍

lemastero (approved):

PR #809, opened 2020-11-22
Random access for transactions in block

dmitry-worker:
# Description

Block body utilizes `List.apply(index)` to find transactions, which has linear complexity.
Expensive `zipWithIndex` for list is used to enumerate through transactions.
`List.reverse` was used to iterate backwards.

# Proposed Solution

Refactored BlockBody, introducing iterators for forward and backward traverses.
Constant-time `getTransactionByIndex` introduced.
Reduced overall memory usage and fragmentation by using Arrays.


src/main/scala/io/iohk/ethereum/domain/BlockBody.scala
```diff
@@ -3,19 +3,136 @@ package io.iohk.ethereum.domain
...
 
   lazy val numberOfUncles: Int = uncleNodesList.size
+
+  /**
+    * Works as `copy` for case classes
+    * @param transactions - new transactions for the block
+    *        Beware using array-based collections as an argument, because:
+    *  {{{
+    *  val ar1 = Array(1, 2, 3)
+    *  val seq = arr.toSeq
```
mirkoAlic: ?
```suggestion
    *  val seq = ar1.toSeq
```

src/main/scala/io/iohk/ethereum/domain/BlockBody.scala
```diff
@@ -3,19 +3,136 @@ package io.iohk.ethereum.domain
...
+    *  val ar2 = seq.toArray
+    *  ar2.update(0,0)
+    *  ar1
+    *  }}}
+    *  would return (0, 2, 3), because ar1 === ar2
```
mirkoAlic, line 29: you suggest that `ar1 === ar2` after ar2.update(0,0) ? i mean, AFAIK that is NOT true..

dmitry-worker: @mirkoAlic thanks! I just wanted to mention that `ar1` is the same instance as `ar2` here.
Hence whenever you modify the first one - the second is also updated.
`.toSeq` just wraps an `Array` by producing a `WrappedArray` instance.
And `WrappedArray.toArray` call returns the array it wraps back to the user.

src/main/scala/io/iohk/ethereum/domain/BlockBody.scala
```diff
@@ -3,19 +3,136 @@ package io.iohk.ethereum.domain
...
+    BlockBody(this.transactionList, uncles)
+  }
+
+  /**
+   * @return An inner array size.
+   */
+  override def size: Int = numberOfTxs
+
+
+  override def iterator: Iterator[SignedTransaction] = {
```
KonradStaniec: i would rename it to `transactionIteratator` (and do that with other methods which operates on signed transactions) as blockbody has also `unclesList` so it is not obvious which iterator we are getting.

KonradStaniec (commented):
My main concern with this pr is that i do not see any benchmarks here. From what I remember this code was never a bottleneck. The `SignedTransactions` collection is usually small like:
- up to 10-20 transaciton on ETC
- up to 200-300 on ETH

So if we are willing to trade off simplicity for performance (change simple case class for custom class operationg on arrays)  it would be nice to see some numbers what we are gaining.

src/main/scala/io/iohk/ethereum/domain/BlockBody.scala
```diff
@@ -3,19 +3,136 @@ package io.iohk.ethereum.domain
...
-case class BlockBody(transactionList: Seq[SignedTransaction], uncleNodesList: Seq[BlockHeader]) {
-  override def toString: String =
-    s"BlockBody{ transactionList: $transactionList, uncleNodesList: $uncleNodesList }"
+import scala.collection.immutable
+
+class BlockBody private(
+                        private val transactionList: Array[SignedTransaction],
+                        val uncleNodesList: Seq[BlockHeader]
+                      ) extends Iterable[SignedTransaction] {
+
```
aakoshh: As I mentioned on our call, I think you can get the benefit of O(1) access for indexing, without having to override `hashCode` and `equals` by using `IndexedSeq`:

```scala
case class BlockBody(transactions: IndexedSeq[SignedTransaction], uncleNodes: Seq[BlockHeader])
```

If you make sure that the case class is instantiated with an `Array` argument, it will be wrapped in a way that prevents mutability as well:
```console
scala> case class Test(xs: IndexedSeq[Int])
defined class Test

scala> val a = Test(Array(1,2,3))
a: Test = Test(WrappedArray(1, 2, 3))

scala> val b = Test(Array(1,2,3))
b: Test = Test(WrappedArray(1, 2, 3))

scala> a == b
res0: Boolean = true

scala> a.xs(0) = 0
<console>:13: error: value update is not a member of IndexedSeq[Int]
       a.xs(0) = 0
         ^
```

Personally I'd prefer this approach, letting people access the immutable field directly for filtering and enumerating, rather than providing half a dozen delegating methods.

The other thing was to not inherit `Iterable` since it wraps 2 collections and there's no obvious reason to favor one over the other.

PR #811, opened 2020-11-25
[ETCM-403] Expose application.conf and chains configs directory

mmrozek:
# Description

Copy all configs to the package conf directory - it will be easier to understand for users how to configure the client.


src/universal/conf/app.conf
```diff
@@ -1,5 +1,5 @@
 # This is the base configuration file for the Mantis ETC client.
```
mirkoAlic, line 1: maybe is a good time to get rid of this file(`app.conf`). IMO is just a useless proxy.

mirkoAlic: (disclaimer: probably out of scope)

mirkoAlic (approved):
LGTM!

faraggi (commented):
LGTM :)

PR #812, opened 2020-11-25
[ETCM-411] Add optional param to generate more keys at once

KonradStaniec:
# Description

Add possibility to generate n keys (ported from other project)

### DIFF
--- a/src/main/scala/io/iohk/ethereum/cli/CliCommands.scala
+++ b/src/main/scala/io/iohk/ethereum/cli/CliCommands.scala
@@ -7,11 +7,13 @@ import io.iohk.ethereum.crypto._
 import io.iohk.ethereum.domain.Address
 import io.iohk.ethereum.utils.ByteStringUtils
 import java.security.SecureRandom
+import io.iohk.ethereum.nodebuilder.SecureRandomBuilder
 import org.bouncycastle.util.encoders.Hex
 
-object CliCommands {
+object CliCommands extends SecureRandomBuilder {
 
   val generatePrivateKeyCommand = "generate-private-key"
+  val generateKeyPairsCommand = "generate-key-pairs"
   val deriveAddressCommand = "derive-address"
   val generateAllocsCommand = "generate-allocs"
   val balanceOption = "balance"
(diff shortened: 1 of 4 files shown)


mmrozek (commented):
Looks good, but what do you think about moving this functionality to CLI?

KonradStaniec:
one one side it makes sense, on the other side i know that Shay is using it in that format so it would be probably easier for him to use it that way without any learning curve and changes in tooling

mmrozek (approved):
Ok, so maybe create a task to do that in the future. We could handle both methods at the same time. The cli will be more readable for the users

src/main/scala/io/iohk/ethereum/crypto/EcKeyGen.scala
```diff
@@ -3,19 +3,24 @@ package io.iohk.ethereum.crypto
...
+  * to generate (default is 1).
+  * The key pairs will be printed in the format:
   *   priv-key-hex (32 bytes)
   *   pub-key-hex (64 bytes)
   *
   * Run:
-  *   ./eckeygen > mantis-datadir/node.key
+  *   ./eckeygen [n] > mantis-datadir/node.key
   *
   * to generate the private key for the node. Note that only the private key will be read upon Mantis boot,
```
mirkoAlic, line 15: (minor "Note that only the *first* private key will be read upon Mantis boot"

mirkoAlic (approved):

src/main/scala/io/iohk/ethereum/crypto/EcKeyGen.scala
KonradStaniec: actually it is more general as even if we generate one pair, only the private key is read at boot time and then whole key pair is regenerated based on this key `keyPairFromPrvKey(Hex.decode(privHex))`

src/main/scala/io/iohk/ethereum/cli/CliCommands.scala
```diff
@@ -27,6 +29,22 @@ object CliCommands {
       }
     }
 
+  private val GenerateKeyPairs: Command[String] =
```
mmrozek, line 32: Please add a test case to `CliCommandsSpec`

PR #814, opened 2020-11-25
[ETCM-415] Fix random subset of peers where block is broadcasted

ntallar:
# Description

Fixes random subset of peers not working properly

Given a set, Random.shuffle doesn't work properly (kudos to @dmitry-worker for finding this!):
- `Random.shuffle((1 to 10).toSet).take(5) == (1 to 10).toSet.take(5)`, the shuffle does nothing (this seems to happen with more than 4 elements as the HashSet implementation is used)
- `Random.shuffle((1 to 10).toSet) == Random.shuffle((1 to 5).toSet)`, adding more elements doesn't alter the result of the shuffle

This might have lead into bad propagation of blocks, increasing the chances of having forks on our testnet

# Proposed Solution

- Converted Set to Seq before shuffling
- Removed broadcast-new-block-hashes config as it's heavily used:
   1. Our current sync implementation relies on this message to have an up-to-date known top, and also triggering the fetching of new blocks
   2. Without sending this message (as when this flag was off) we are not following the Ethereum protocol


mirkoAlic (approved):
:rocket:

biandratti (approved):

PR #815, opened 2020-11-26
Add miners whitelist to testnet configs

kreisys:
- Adds a json file with a pool of 1000 public keys allowed to mine
- Includes that file into the default configuration of the nomad and gac testnets
- Sets the default `consensus.protocol` for both testnets (in `base-testnet.conf`)
  to `restricted-ethash` in order to enable whitelist.

Tested that a newly deployed gac-based testnet whose miners are not in the whitelist remains at block 0.


src/main/resources/chains/testnet-allowed-miners.json
```diff
@@ -0,0 +1,1004 @@
+{
+  "allowed-miners": [
+    "006f366775cd73455ee7f208d8dc258ba7c1d48645ef2f4aeef7024652ac10205aed36bfc594e85021acc216762e7ab75f7759a815409f37d05e863651920091",
```
mirkoAlic, line 3: long list hehe :P

src/main/resources/chains/testnet-internal-gac-chain.conf
```diff
@@ -149,3 +149,5 @@
     # empty means that everybody can mine
     allowed-miners = []
 }
+
+include "chains/testnet-allowed-miners"
```
ntallar, line 153: Should we include the same list of the nomad deployer or will different keys be generated there?

kreisys: it's a draft. only reason it's open is for hydra to pick it up.

kreisys: (the previous attempt simply didn't work; hocon documentation isn't super clear)

KonradStaniec (commented):
just an general comment: the list of public keys is taken into account only if option 
`mantis.consensus.protocol` is set to ` restricted-ethash`

kreisys:
> just an general comment: the list of public keys is taken into account only if option
> `mantis.consensus.protocol` is set to ` restricted-ethash`

Now you tell me! 😅

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

PR #820, opened 2020-11-27
[ETCM-426] Changed the gas limit strategy to a constant one

ntallar:
# Description

 Our current strategy is to always increase the gas limit of our blocks as much as possible which in our current cluster setup resulted in:
- Exposed a bug on web3, as that big a number was being properly shown by the explorer
- Means that after 20k blocks 900k million txs could be sent per block, which is very attackable

# Proposed Solution

- The genesis gas limit was set to the ETC mainnet value: 8 million
- The miners strategy was changed to always set the gas limit of blocks to the same as the parent

# For future tasks

A proper voting strategy (with a target as other clients) should be set up

# Testing

Start up the node and mine a couple blocks, check that all the blocks have a gasLimit of 8 million


mirkoAlic (approved):
LGTM!

PR #823, opened 2020-12-01
ETCM-[165,166]: Move RLP and crypto into their own sub-projects

aakoshh:
# Description

Separates out the `rlp` and `crypto` packages into their own sbt sub-projects within the same build, so that at some point they can be published as standalone libraries. Because they both depended on `ByteUtils` a separate `bytes` project was also created.

Dependencies were cut down to the minimum, but Akka remains - I think RLP the formats that got to do with `ByteString` could be moved to the `node` at some point, but it seems more pervasive in crypto. 

The node jar keeps the `mantis` name so the Universal package as well as the `Hello.id` stays the same. Added the `io.iohk` org.
```console
$ ls mantis-3.1.0/lib | grep io.iohk
io.iohk.mantis-3.1.0.jar
io.iohk.mantis-3.1.0-launcher.jar
io.iohk.mantis-bytes-3.1.0.jar
io.iohk.mantis-crypto-3.1.0.jar
io.iohk.mantis-rlp-3.1.0.jar
io.iohk.scalanet_2.12-0.4.4-SNAPSHOT.jar
io.iohk.scalanet-discovery_2.12-0.4.4-SNAPSHOT.jar
```

Also fixes coverage report paths, based on the logs these are the correct ones, but it could be different on buildkite:
```
[info] Wrote instrumentation file [/home/aakoshh/projects/iohk/mantis/target/scala-2.12/scoverage-data/scoverage.coverage]
[info] Will write measurement data to [/home/aakoshh/projects/iohk/mantis/target/scala-2.12/scoverage-data]
```


project/Dependencies.scala
```diff
@@ -2,6 +2,14 @@ import sbt._
 
 object Dependencies {
 
+  val akkaUtil: Seq[ModuleID] = {
+    val akkaVersion = "2.6.9"
```
lemastero: It looks like we two akkaVersion declared. Maybe it would make sense to have 1

rlp/src/main/scala/io/iohk/ethereum/rlp/RLPImplicits.scala
```diff
@@ -51,7 +50,7 @@ object RLPImplicits {
   implicit val bigIntEncDec = new RLPEncoder[BigInt] with RLPDecoder[BigInt] {
 
     override def encode(obj: BigInt): RLPValue = RLPValue(
-      if (obj.equals(BigInt(0))) byteToByteArray(0: Byte) else obj.toUnsignedByteArray
+      if (obj.equals(BigInt(0))) byteToByteArray(0: Byte) else bigIntToUnsignedByteArray(obj)
```
lemastero: It looks like `toUnsignedByteArray` in BigIntExtensionMethods.BigIntAsUnsigned
has the same implementation as: `bigIntToUnsignedByteArray` in RLP

Is his neccessary?

aakoshh: It is indeed a carbon copy. I brought it over because this was the only thing it used from the utils, and I didn't want to create a "mantis-utils" library for RLP to depend on, and it also didn't look like the entire `ByteUtils` should be be moved to RLP. But I'm open to suggestions.

project/Dependencies.scala
aakoshh: Sure, done.

rlp/src/main/scala/io/iohk/ethereum/rlp/RLPImplicits.scala
aakoshh: Maybe a dedicated `mantis-bytes` library? Probably crypto would use it too.

aakoshh: Created a `bytes` project, moved `bigIntToUnsignedByteArray` to `ByteUtils`, and now `BigIntExtensionMethods` calls that as well.

aakoshh: Did the same treatment for crypto.

enriquerodbe (approved):

PR #825, opened 2020-12-02
[Chore] request logged only if it doesn't have sensible information

biandratti:
# Description

request logged only if it doesn't have sensible information

# Proposed Solution

If the request method is “personal”, we only log the next legend "sensitive information"


ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

PR #826, opened 2020-12-02
[ETCM-205] Add information about checkpoints in tx history

kapke:
# Description

This PR adds information about tx being in checkpointed block or not to `mantis_getAccountTransactions`.
Also, as it is very useful - in this PR I've added a dependency on diffx - library which highlights the differences in compared data.

# Testing

All tests should pass. One can also try to use `qa_generateCheckpoint` for adding a checkpoint and checking if transactions from lower blocks are properly marked

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -54,7 +54,9 @@ object Dependencies {
     "org.scalatest" %% "scalatest" % "3.2.2" % "it,test",
     "org.scalamock" %% "scalamock" % "5.0.0" % "test",
     "org.scalatestplus" %% "scalacheck-1-14" % "3.2.2.0" % "test",
-    "org.scalacheck" %% "scalacheck" % "1.14.3" % "it,test"
+    "org.scalacheck" %% "scalacheck" % "1.14.3" % "it,test",
+    "com.softwaremill.diffx" %% "diffx-core" % "0.3.30" % "test",
+    "com.softwaremill.diffx" %% "diffx-scalatest" % "0.3.30" % "test"
   )
 
   val cats: Seq[ModuleID] = {
--- a/src/main/scala/io/iohk/ethereum/domain/Address.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/Address.scala
@@ -2,11 +2,9 @@ package io.iohk.ethereum.domain
 
 import akka.util.ByteString
 import io.iohk.ethereum.crypto
-import io.iohk.ethereum.domain.SignedTransaction.FirstByteOfAddress
 import io.iohk.ethereum.mpt.ByteArrayEncoder
 import io.iohk.ethereum.utils.ByteUtils.padLeft
 import org.bouncycastle.crypto.AsymmetricCipherKeyPair
-import org.bouncycastle.crypto.params.ECPublicKeyParameters
 import org.bouncycastle.util.encoders.Hex
 
 object Address {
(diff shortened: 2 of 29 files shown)


project/Dependencies.scala
```diff
@@ -54,7 +54,9 @@ object Dependencies {
     "org.scalatest" %% "scalatest" % "3.2.2" % "it,test",
     "org.scalamock" %% "scalamock" % "5.0.0" % "test",
     "org.scalatestplus" %% "scalacheck-1-14" % "3.2.2.0" % "test",
-    "org.scalacheck" %% "scalacheck" % "1.14.3" % "it,test"
+    "org.scalacheck" %% "scalacheck" % "1.14.3" % "it,test",
+    "com.softwaremill.diffx" %% "diffx-core" % "0.3.30" % "test",
+    "com.softwaremill.diffx" %% "diffx-scalatest" % "0.3.30" % "test"
```
mmrozek, line 59: Don't we need to update nix dependencies?

src/main/scala/io/iohk/ethereum/domain/Address.scala
```diff
@@ -33,6 +36,12 @@ object Address {
     require(bytes.length <= Length, s"Invalid address: $hexString")
     Address(bytes)
   }
+
+  def apply(keyPair: AsymmetricCipherKeyPair): Address = {
+    //byte 0 of encoded ECC point indicates that it is uncompressed point, it is part of bouncycastle encoding
+    val pub = keyPair.getPublic.asInstanceOf[ECPublicKeyParameters].getQ.getEncoded(false).tail
```
mmrozek, line 42: Couldn't we use `crypto.pubKeyFromKeyPair`?

project/Dependencies.scala
kapke: We do, I wanted buildkite to generate proper updates for me

src/main/scala/io/iohk/ethereum/domain/Address.scala
kapke: Good point!

mmrozek (approved):
LGTM!

src/main/scala/io/iohk/ethereum/domain/Address.scala
```diff
@@ -33,6 +35,11 @@ object Address {
     require(bytes.length <= Length, s"Invalid address: $hexString")
     Address(bytes)
   }
+
+  def apply(keyPair: AsymmetricCipherKeyPair): Address = {
+    val pub = crypto.pubKeyFromKeyPair(keyPair)
+    Address(crypto.kec256(pub).drop(FirstByteOfAddress))
+  }
```
lemastero, line 42: 👍

src/main/scala/io/iohk/ethereum/domain/Address.scala
```diff
@@ -33,6 +35,11 @@ object Address {
     require(bytes.length <= Length, s"Invalid address: $hexString")
     Address(bytes)
   }
+
+  def apply(keyPair: AsymmetricCipherKeyPair): Address = {
+    val pub = crypto.pubKeyFromKeyPair(keyPair)
+    Address(crypto.kec256(pub).drop(FirstByteOfAddress))
```
KonradStaniec, line 41: minor:  `.drop(FirstByteOfAddress)` is probably not necessary , as `Address` bytesting constructor takes only last 20byte also `FirstByteOfAddress` is weird name for something which drops 12 bytes

KonradStaniec (approved):
Only 1 minor comment

lemastero (approved):

PR #828, opened 2020-12-03
[FIX] Moved block metrics to the place after validation and execution of the block

pslaski:
# Description

We were measuring block metrics after the block execution but before validation, which mean that we were measuring blocks which could be removed later.

# Proposed Solution

Moved measuring block metrics to the place after the block execution and validation.


ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

PR #830, opened 2020-12-03
ETCM-451: Shorten logs in PeerActor 

aakoshh:
# Description

In DEBUG log level the `PeerActor` [logs](https://github.com/input-output-hk/mantis/blob/develop/src/main/scala/io/iohk/ethereum/network/PeerActor.scala#L261) received messages which includes transaction payloads, bloom filters etc, quickly filling up the logs, which then log rotation overwrites. This hampers our ability to troubleshoot issues with syncing since logs disappear very quickly after the syncing starts.

The PR adds a `toShortString` method to the `Message` type, following the convention of `toString` overrides, to keep the on-screen content to a minimum.


src/main/scala/io/iohk/ethereum/network/p2p/messages/CommonMessages.scala
```diff
@@ -121,6 +122,13 @@ object CommonMessages {
         s"totalDifficulty: $totalDifficulty" +
         s"}"
 
+    override def toShortString: String =
+      s"NewBlock { " +
+        s"code: $code, " +
+        s"block: ${Hex.toHexString(block.hash.toArray[Byte])}, " +
```
ntallar: This log proved really useful to check full non main chain blocks on the testnet, so I don't agree with this change

With all the other changes logged in this PR I'm not sure if there's any way of getting them

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV64.scala
```diff
@@ -127,6 +129,12 @@ object PV64 {
         s"chainWeight: $chainWeight" +
         s"}"
 
+    override def toShortString: String =
+      s"NewBlock { " +
+        s"block: ${Hex.toHexString(block.hash.toArray[Byte])}, " +
```
ntallar: Same for this change

ntallar (dismissed):
Who is the target beneficiary of this change? Because even though this lowers the amount of logs, there are probably many others that the beneficiary won't like to have, should we maybe have a different logback for this case?

Either way, as we are close to the release, I'm only blocking this PR for the marked changes, the above is not a blocker

ntallar:
This log also seems to a bit of a spammer:
```
2020-12-03 15:04:02,669 [i.i.e.network.PeerManagerActor] - Asking for another random node
```

src/main/scala/io/iohk/ethereum/network/p2p/messages/CommonMessages.scala
aakoshh: Sorry can you elaborate on what you mean by "full non main chain blocks"? Are there specific parts of the header you'd like to see? The default included the [header and the body](https://github.com/input-output-hk/mantis/blob/20369d331969884c474eb15c11900744ec1e4082/src/main/scala/io/iohk/ethereum/domain/Block.scala#L15), did you really benefit from seeing for example the [transaction payload](https://github.com/input-output-hk/mantis/blob/20369d331969884c474eb15c11900744ec1e4082/src/main/scala/io/iohk/ethereum/domain/Transaction.scala#L45) in the log? Is there a middle ground? Perhaps we can log the header instead of the full block?

aakoshh:
I don't want to destroy any useful information, but the amount of logs by default is overhelming, full binary block content flushed to the screen. Another option would be to use the `TRACE` level. 

The negative effect of the amount of blocks we saw yesterday was that it generated more than 10MB of blocks withing 30 seconds during fast sync, and the log ration as I understand keeps as fixed number of logs, so when I asked questions about how long it took to find the pivot block and how many peers the node had in the meantime, the logs were no longer available. The beneficiary would have been me who wanted to debug some issues using the logs. 

But if you think it's a bad change then just close it.

aakoshh:
> This log also seems to a bit of a spammer:
> 
> ```
> 2020-12-03 15:04:02,669 [i.i.e.network.PeerManagerActor] - Asking for another random node
> ```

It was useful to see what was happening, otherwise it would have been difficult to catch the situation you pointed out with the fixed small testnet size. https://github.com/input-output-hk/mantis/pull/831/commits/a183bc8cc11a0b918401a0c6a2bef64bd3912e8a should alleviate the issue in those cases, but let me know if it's also a problem with discovery enabled and we can put it back to TRACE.

aakoshh:
Oh, there's no `TRACE` in `ActorLogging` :(

kapke:
What about:
1. adding a property to logback.xml in similar fashion to ones controlling logs location, but for log level? In this way we can use `INFO` by default, but enable `DEBUG` easily with just system property/env variable/cli arg? 
2. introducing an object/class just for logging peer messages, but extending our `Logger` trait so `TRACE` can be used?

aakoshh:
> adding a property to logback.xml in similar fashion to ones controlling logs location, but for log level? In this way we can use INFO by default, but enable DEBUG easily with just system property/env variable/cli arg?

I didn't even check, just assumed this is how it worked. Definitely I can add a flag. It would also be possible to send different levels to stdout and the file, but then configuring is a bit more confusing than just one env var.

src/main/scala/io/iohk/ethereum/network/p2p/messages/CommonMessages.scala
aakoshh: Changed it so the header is included.

src/main/scala/io/iohk/ethereum/network/p2p/messages/PV64.scala
aakoshh: Included the header.

kapke:
IMO the `if` approach makes perfect sense here

kapke (dismissed):
LGTM!

aakoshh:
Applied a hack for tracing: the level is detected using TypeSafe Logging but it just influences what content goes into `ActorLogging.debug`.

src/main/scala/io/iohk/ethereum/network/PeerActor.scala
```diff
@@ -51,6 +52,18 @@ class PeerActor[R <: HandshakeResult](
 
   val peerId: PeerId = PeerId.fromRef(self)
 
+  // The actor logs incoming messages, which can be quite verbose even for DEBUG mode.
+  // ActorLogging doesn't support TRACE, but we can push more details if trace is enabled.
+  val isTraceEnabled = {
```
kapke: I thought about something like to workaround Akka limitations
```scala
class PeerActorLogger extends io.iohk.ethereum.utils.Logger {}
class PeerActor {
	val logger = new PeerActorLogger

	//...

	logger.log.debug(foo)
	logger.log.trace(bar)
}

```

aakoshh: Can do that too, I just didn't want to mix it that half of the messages are logged with the `ActorLogging` machinery which has a different configuration than the `Logger` stuff, except for this one. Best would be to get rid of `ActorLogging` altogether.

aakoshh: Pushed a version of that too, but just for messages.

kapke (approved):
LGTM!

aakoshh:
@ntallar are you happy with the changes?

PR #831, opened 2020-12-03
FIX: Update the bootstrap config

aakoshh:
# Description

Updated the `etc` and `mordor` bootstrap node configuration based on `core-geth`. Also added @KonradStaniec's [node](https://etcnodes.org/node/18.159.208.198#).

It turns out that what we had configured for `etc` were the Ethereum [mainnet nodes](https://github.com/etclabscore/core-geth/blob/53f90a413bd5a1b7c068d51c9843ec8958486e6c/params/bootnodes.go#L23). Replacing with the ones dedicated for Ethereum Classic may speed up discovery as well.


src/main/resources/chains/etc-chain.conf
```diff
@@ -158,15 +158,36 @@
...
+	"enode://7a7768a1603b6714acd16f646b84a5aff418869b5715fa606172d19e4e7d719699448b7707e30c7b59470b4fb8b38a020135641304861483a7dc3ba988e98490@47.108.52.30:30303",   //ETC Labs
+	"enode://903fceb08534c13fe1114b0c753a89c6c2ec50f973a85d308456c46721f8904b1cd47df9231114299be84e1e954db06552f791a0facbd86359aa9fd321d2ef50@47.240.106.205:30303", //ETC Labs
+	"enode://c7ec057ad9450d2d5c4002e49e53e1d90142acd54128c89e794d11401de026b91aa0e84e3f679fb6b47f7940e08b5e7d21d8178c5fa6ba0f36971098cb566ea6@101.133.229.66:30303", //ETC Labs
+	"enode://70b74fef51aa4f36330cc52ac04f16d38e1838f531f58bbc88365ca5fd4a3da6e8ec32316c43f79b157d0575faf53064fd925644d0f620b2b201b028c2b410d0@47.115.150.90:30303",  //ETC Labs
+	"enode://fa64d1fcb2d4cd1d1606cb940ea2b69fee7dc6c7a85ac4ad05154df1e9ae9616a6a0fa67a59cb15f79346408efa5a4efeba1e5993ddbf4b5cedbda27644a61cf@47.91.30.48:30303",    //ETC Labs
+
+  "enode://a7219ff608bdca40174d005610f80a8d216aa9695e34542d27d69cfd6bf5478fd3b5c83c08ad537a06425cce9cda975571053dcb174d44f109b7362d26798256@18.159.208.198:30303", // IOHK ec2-18-159-208-198.eu-central-1.compute.amazonaws.com
```
aakoshh, line 190: @KonradStaniec I haven't even realised so far that all nodes only specify their discovery port, not their TCP, so even though the `PeerDiscoveryManager` will treat them as already discovered, the `PeerManagerActor` will not be able to connect to them via TCP.

aakoshh: Added a filter to not treat them as discovered, so they don't get blacklisted.

aakoshh: Is this node using the default port?

aakoshh: Ah, okay, one can use the same port for both. Can you check that this connection string is correct? I haven't been able to connect to it.

aakoshh:
I don't know if it might cause problems but if we get the port wrong, the peer manager blacklists the peer's IP for 30 minutes, so even if discovery figures out later what the correct port is, it may take a long time until the same node is tried again. This only affects the bootstraps/known-nodes, the others' ports come form the ENR.

src/main/resources/chains/etc-chain.conf
```diff
@@ -158,15 +158,36 @@
...
+	"enode://7a7768a1603b6714acd16f646b84a5aff418869b5715fa606172d19e4e7d719699448b7707e30c7b59470b4fb8b38a020135641304861483a7dc3ba988e98490@47.108.52.30:30303",   //ETC Labs
+	"enode://903fceb08534c13fe1114b0c753a89c6c2ec50f973a85d308456c46721f8904b1cd47df9231114299be84e1e954db06552f791a0facbd86359aa9fd321d2ef50@47.240.106.205:30303", //ETC Labs
+	"enode://c7ec057ad9450d2d5c4002e49e53e1d90142acd54128c89e794d11401de026b91aa0e84e3f679fb6b47f7940e08b5e7d21d8178c5fa6ba0f36971098cb566ea6@101.133.229.66:30303", //ETC Labs
+	"enode://70b74fef51aa4f36330cc52ac04f16d38e1838f531f58bbc88365ca5fd4a3da6e8ec32316c43f79b157d0575faf53064fd925644d0f620b2b201b028c2b410d0@47.115.150.90:30303",  //ETC Labs
+	"enode://fa64d1fcb2d4cd1d1606cb940ea2b69fee7dc6c7a85ac4ad05154df1e9ae9616a6a0fa67a59cb15f79346408efa5a4efeba1e5993ddbf4b5cedbda27644a61cf@47.91.30.48:30303",    //ETC Labs
+
+  "enode://a7219ff608bdca40174d005610f80a8d216aa9695e34542d27d69cfd6bf5478fd3b5c83c08ad537a06425cce9cda975571053dcb174d44f109b7362d26798256@18.159.208.198:53818", // IOHK ec2-18-159-208-198.eu-central-1.compute.amazonaws.com
```
ntallar, line 190: I don't seem to be able to connect to this instace

src/main/resources/chains/mordor-chain.conf
```diff
@@ -141,6 +141,7 @@
   eth-compatible-storage = true
 
   # Set of initial nodes
+  # https://github.com/etclabscore/mordor/blob/master/static-nodes.json
```
ntallar, line 144: The below list doesn't seem to be the same as the from this link :thinking:

ntallar (changes requested):
Fast sync to etc mainnet took 8 minutes to start!

src/main/resources/chains/etc-chain.conf
aakoshh: Hm, I tried with the default ports, `18.159.208.198:30303` then `18.159.208.198:9072?discport=30303` but those didn't work, I was hoping this might as it's from etcnodes.org

@KonradStaniec do you know what might be wrong?

src/main/resources/chains/mordor-chain.conf
aakoshh: Not exactly, I only added new nodes, didn't remove ones we had but they didn't. Don't know who is right, I thought taking the union is the conservative approach.

src/main/resources/chains/etc-chain.conf
ntallar: Konrad is off today... I'd maybe remove this and re-add it once we speak with him about how he has it configured

src/main/resources/chains/mordor-chain.conf
ntallar: Hmm but I couldn't find any of the ones you added on that link :thinking:

ntallar (commented):
Apart from my comments answers, LGTM!

src/main/resources/chains/etc-chain.conf
aakoshh: Okay. He said he asked devops about firewalls and port forwarding etc.

src/main/resources/chains/mordor-chain.conf
aakoshh: You're right, I don't know what I mixed up, sorry.

aakoshh: Fixed.

src/main/resources/chains/etc-chain.conf
aakoshh: Removed.

ntallar (approved):
LGTM!
```
 .\\            //.
. \ \          / /.
.\  ,\     /` /,.-
 -.   \  /'/ /  .
 ` -   `-'  \  -
   '.       /.\`
      -    .-
      :`//.'
      .`.'
      .' BP 
```

PR #832, opened 2020-12-04
Update `mantis.client-version` in configuration

loverdos:
Update the client version reported from JSON-RPC


src/main/resources/application.conf
```diff
@@ -3,7 +3,7 @@ mantis {
   client-identity = null
 
   # Version string (reported by an RPC method)
-  client-version = "mantis/v2.0"
+  client-version = "mantis/v3.0"
```
ntallar: Shouldn't we remove this configuration and use `BuildInfo.version` as [here](https://github.com/input-output-hk/mantis/blob/9ba220545040fa9ba206d4850cfafa4941b6ee9b/src/main/scala/io/iohk/ethereum/utils/VersionInfo.scala#L25)?

loverdos: Very good point. Actually I introduced `BuildInfo.version` a long time ago and yes this is the proper way to do it.
Should we deprecate `mantis.client-version` completely then and use `BuildInfo` in [Config](https://github.com/input-output-hk/mantis/blob/ae88e6fc4d2ad4b9472ba971c7e0916191f53726/src/main/scala/io/iohk/ethereum/utils/Config.scala#L29-L29) as well? (or something from `VersionInfo` directly)

loverdos: If you agree I can update the PR accordingly

jmendiola222: agree

loverdos: I have updated the PR.

lemastero (approved):

PR #833, opened 2020-12-04
ETCM-446: Connection limit ranges

aakoshh:
# Description

Currently the node will try to connect to nodes until their `max-outgoing-peers` is reached and limit incoming connection to `max-incoming-peers`. The problem is once incoming peers are saturated over the network it's difficult to find nodes to connect to, and this situation is likely to arise if `outgoing >= incoming` is configured. 

The PR changes the connection strategy to be based on acceptable ranges:
* `min-outgoing-peers`: The node is not seeking new connections if the outgoing _handshaked_ connection count is above this number. Otherwise it tries to open up to `max-outgoing-peers` connections.
* `prune-incoming-peers`: A number of incoming peers to try to prune if the incoming handshaked peer count hit the maximum and we are rejecting connections with `TooManyPeers`.
* `min-prune-age`: A duration that a peer is minimally allowed to be connected before being selected for random pruning, and also a minimum amount of time that needs to pass between pruning attempts, to avoid hostile takeovers.

# Proposed Solution

There are two ideas:
1. We can try to aggressively connect to a lot of nodes in the beginning, expecting that many of them won't accept our connection, but once we have enough peers, we can stop at a lower level, only opening more if we fall below that.
2. We can close down incoming connections if the overall number goes beyond a limit, so that new nodes can be accepted into the network instead. This also compensates to the aggressive over-connecting in the beginning. 

The new default settings target at a 40/60 ratio for `min-outgoing / max-incoming`, to protect against eclipse attacks but also maintain that `total(max-incoming) >= total(min-outgoing)` on the the network level. `max-outgoing` is high to allow quickly trying to open many connections after startup, anticipating that most nodes will not be compatible or receptive.

Currently the pruning only kicks in when a new incoming connection would bring the number above the maximum. We could also prune periodically whenever the number equals the max. That would allow new joiners to be accepted immediately rather than when they try to re-connect after a short blacklisting.

To decide which connections to prune we take the average number of received _responses_ over the lifetime of the peers, with higher being better. The statistics are collected for the last 12 hours, which is the same time as the bonding duration in discovery.


src/main/resources/application.conf
```diff
@@ -142,11 +142,20 @@ mantis {
       # Maximum MPT components in a single response message (as a blockchain host)
       max-mpt-components-per-message = 200
 
-      # Maximum number of peers this node can connect to
-      max-outgoing-peers = 50
+      # Minimum number of peers this node tries to connect to at all times
+      min-outgoing-peers = 20
+
+      # Maximum number of peers this node can connect to at any time
+      max-outgoing-peers = 30
```
KonradStaniec: Maybe we can increase this nubmer to 50? At the node startup it is really important to traverse the list of discovered peers as fast as possible to try to grab as many connections as possible.

src/main/scala/io/iohk/ethereum/network/ConnectedPeers.scala
```diff
@@ -65,11 +74,49 @@ case class ConnectedPeers(
...
+    } else {
+      val candidates = handshakedPeers.collect {
+        case (_, p)
+            if (p.incomingConnection || !incoming)
+              && p.createTimeMillis <= ageThreshold
+              && !pruningPeers.contains(p.id) =>
+          p
+      }.toSeq
+
+      val toPrune = Random.shuffle(candidates).take(numPeers)
```
KonradStaniec: I have a mixed feeling about disconnecting randomly form peers, especially those which are connected to the node for a pretty long time as those usually provide most data and are most stable
Imo there are two forces at play:

-From network point of view this makes sense, as it makes space for new joiners.
-From node point of view it makes no sense, as node will randomly disconnect from peers to which it was connected pretty long time (so most probably most stable ones)

Maybe we should couple this strategy with some kinda tracking when was the last communication to given peer, and select only peers which do not provide us any data or do not communicate with us in any way?

src/main/scala/io/iohk/ethereum/network/ConnectedPeers.scala
```diff
@@ -65,11 +74,49 @@ case class ConnectedPeers(
...
+      currentTimeMillis: Long = System.currentTimeMillis
+  ): (Seq[Peer], ConnectedPeers) = {
+    val ageThreshold = currentTimeMillis - minAge.toMillis
+    if (lastPruneTimestamp > ageThreshold || numPeers == 0) {
+      // Protect against hostile takeovers by limiting the frequency of pruning.
+      (Seq.empty, this)
+    } else {
+      val candidates = handshakedPeers.collect {
+        case (_, p)
+            if (p.incomingConnection || !incoming)
```
KonradStaniec: imo this condition is quitle long to deserve its own function

src/main/scala/io/iohk/ethereum/network/ConnectedPeers.scala
```diff
@@ -65,11 +74,49 @@ case class ConnectedPeers(
...
+        outgoingPendingPeers -- peersId,
+        handshakedPeers -- peersId,
+        pruningPeers -- peersId,
+        lastPruneTimestamp = lastPruneTimestamp
+      )
     )
   }
+
+  def prunePeers(
+      incoming: Boolean,
```
KonradStaniec: why is this flag is necessary ? Maybe i am missing some thing, but in call sites we are passing here `true`

src/main/scala/io/iohk/ethereum/network/ConnectedPeers.scala
```diff
@@ -4,11 +4,15 @@ import java.net.InetSocketAddress
...
 import akka.util.ByteString
+import scala.concurrent.duration.FiniteDuration
+import scala.util.Random
 
 case class ConnectedPeers(
     private val incomingPendingPeers: Map[PeerId, Peer],
     private val outgoingPendingPeers: Map[PeerId, Peer],
-    private val handshakedPeers: Map[PeerId, Peer]
+    private val handshakedPeers: Map[PeerId, Peer],
+    private val pruningPeers: Map[PeerId, Peer],
```
KonradStaniec, line 13: just to make sure, this is necessary to not prune peer twice while waiting for its proper disconnect ?

src/main/resources/application.conf
aakoshh: Yes I was wondering whether the `max-outgoing-peers` should have a relation with the `max-incoming-peers`, but I guess in our case we know that we'll not achieve 50 handshaked outgoing ones. And even if we did, and `max-incoming-peers` is 30 and may get pruned back, if the `min-outgoing-peers` vs `max-incoming-peers` has an optimal (like 40/60) ratio it should over time still allow a stable network where everyone's demand is just satisfied.

src/main/scala/io/iohk/ethereum/network/ConnectedPeers.scala
aakoshh: Definitely that would be a great improvement, I just didn't want to touch even more components in this PR and introduce even more communication with `PeerActor` instances where I assume this information would be stored. 

The protection from rapidly throwing out old peers is that it will not do pruning more often than once per 30 minutes, so it would take some time until all the old guard gets randomly cleared out.

src/main/scala/io/iohk/ethereum/network/ConnectedPeers.scala
aakoshh: In theory we could prune outgoing connections as well, the communication is bi-directional on them, maybe some of them would be better replaced with other nodes, perhaps if they are slow. But in this PR it's just about the incoming ones, to free incoming slots.

src/main/scala/io/iohk/ethereum/network/ConnectedPeers.scala
aakoshh: Yes, not to prune the same peer twice and also to remember the overall number of peers we are trying to prune, so that if the routine is ran again, it will know that it doesn't have to prune yet another 10 random nodes.

src/main/scala/io/iohk/ethereum/network/ConnectedPeers.scala
aakoshh: Done.

src/main/resources/application.conf
aakoshh: Raised to 50.

src/main/scala/io/iohk/ethereum/network/ConnectedPeers.scala
aakoshh: Do you want this added to this PR, or a separate one where we can hash out the details?

aakoshh: Created a ticket ETCM-463 to maintain statistics.

aakoshh: See https://github.com/input-output-hk/mantis/pull/849 for a mechanism we could use for prioritisation.

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -445,4 +505,49 @@ object PeerManagerActor {
...
+    * with lower priorities being the ones to prune first.
+    */
+  def prunePriority(stats: Map[PeerId, PeerStatisticsActor.Stat], currentTimeMillis: Long)(peerId: PeerId): Double = {
+    stats
+      .get(peerId)
+      .flatMap { stat =>
+        val maybeAgeSeconds = stat.firstSeenTimeMillis
+          .map(currentTimeMillis - _)
+          .map(_ * 1000)
+          .filterNot(_ <= 0)
```
enriquerodbe:
```suggestion
          .filter(_ > 0)
```

src/main/scala/io/iohk/ethereum/network/TimeSlotStats.scala
```diff
@@ -0,0 +1,138 @@
...
+  /** Aggregate all stats in all slots within the duration. */
+  def getAll(window: Option[Duration] = None): Map[K, V] =
+    fold(Map.empty[K, V], window getOrElse duration) { case (acc, stats) =>
+      acc |+| stats
+    }
+
+  private def fold[A](init: A, window: Duration)(f: (A, Map[K, V]) => A) = {
+    val (start, end) = slotRange(currentTimeMillis, window)
+
+    def loop(idx: Int, acc: List[Map[K, V]]): List[Map[K, V]] = {
```
enriquerodbe: Can this method be `@tailrec`?

aakoshh: It should be. I can add it in the other PR just in case.

aakoshh: Done in https://github.com/input-output-hk/mantis/pull/849/commits/c4a61072663af066841db03edbd847b2e0c62ac8

enriquerodbe (approved):

PR #835, opened 2020-12-04
[FIX] Trigger block execution metrics on reorg as well

ntallar:
# Description

Measure block execution metrics on reorg as well.

The current measuring point is only part of the possible paths of blocks execution, meaning that some imported blocks are currently not measured


SoseMagarian (approved):
LGTM!

pslaski (approved):

PR #836, opened 2020-12-04
[ETCM-448] json rpc - status code

biandratti:
# Description

Today for every error in json rpc we return status code 200.
As per the 2.0 spec, there is no reason for using anything but 200 http status code whenever we are doing JSON-RPC over HTTP.
Also, returning 400 for valid JSON-RPC requests doesn't help when integrating with JSON-RPC clients.

# Proposed Solution

Any valid JSON-RPC request should generate a JSON-RPC response, the HTTP status code should be 200. No matter if the response is a success or error response.
Calling a method that hasn't been enabled or a method that doesn't exist are valid JSON-RPC requests, therefore the expected response is a JSON-RPC error and the status code 200.
Any invalid JSON-RPC request (malformed json, etc) should generate a 400 status code.
So in this way we have status code 400 for json rpc error: [-32600, -32602, -32700]. In other cases we have status code 200.

Possibly requires change of integration with faucet ui and wallet.


lemastero (approved):
LGTM 👍

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -83,6 +83,15 @@ trait JsonRpcHttpServer extends Json4sSupport with RateLimit with Logger {
...
 
+  private def handleResponse(f: Task[JsonRpcResponse]): Task[(StatusCode, JsonRpcResponse)] = f map { jsonRpcResponse =>
+    jsonRpcResponse.error match {
+      case Some(JsonRpcError(-32600, _, _)) => (StatusCodes.BadRequest, jsonRpcResponse)
+      case Some(JsonRpcError(-32601, _, _)) => (StatusCodes.NotFound, jsonRpcResponse)
+      case Some(JsonRpcError(_, _, _)) => (StatusCodes.InternalServerError, jsonRpcResponse)
+      case None => (StatusCodes.OK, jsonRpcResponse)
+    }
+  }
+
```
lemastero, line 97: can we have a test case for every code specified in [3.6.2   Errors](https://www.jsonrpc.org/historical/json-rpc-over-http.html#id19)

lemastero (changes requested):

lemastero:
Please update the description of the issue :)

kapke:
@conqeror Please verify it's working with wallet in case of errors (it should)
@biandratti Does faucet handle these errors properly?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -84,6 +84,16 @@ trait JsonRpcHttpServer extends Json4sSupport with RateLimit with Logger {
     }
   }
 
+  private def handleResponse(f: Task[JsonRpcResponse]): Task[(StatusCode, JsonRpcResponse)] = f map { jsonRpcResponse =>
+    jsonRpcResponse.error match {
+      case Some(JsonRpcError(error, _, _))
+          if List(JsonRpcError.InvalidRequest.code, JsonRpcError.ParseError.code, JsonRpcError.InvalidParams().code)
```
enriquerodbe: For readability, this List of errors could be extracted to a constant at the class level or companion object

src/test/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServerSpec.scala
```diff
@@ -242,6 +244,116 @@ class JsonRpcHttpServerSpec extends AnyFlatSpec with Matchers with ScalatestRout
...
+        )
+      )
+
+    val jsonRequest = ByteString("""{"jsonrpc":"2.0", "method": "asd", "id": "1"}""")
+    val postRequest =
+      HttpRequest(HttpMethods.POST, uri = "/", entity = HttpEntity(MediaTypes.`application/json`, jsonRequest))
+
+    postRequest ~> Route.seal(mockJsonRpcHttpServer.route) ~> check {
+      status shouldEqual StatusCodes.OK
+      responseAs[String] shouldEqual
```
enriquerodbe: What do you think if instead of comparing Strings (which in case of an error forces the developer to compare the whole string to find the difference), we parse the response as Json and assert each attribute individually?

biandratti:
Yes @kapke, the node and faucet using the same class, both have the same behavior. So, with this, we need to change faucet ui and wallet ui.

enriquerodbe (commented):
Looks good to me, only comments for readability

kapke:
@biandratti I meant - if faucet's rpc client handles errors from mantis node properly? IIRC faucet is being run as a separate process communicating with Mantis over RPC.

conqeror:
@kapke @biandratti
Regarding Faucet UI - it uses rpc client which already expects proper format. If I recall correctly, I was requesting proper rpc errors for `faucet_sendFunds` :)

Regarding mantis wallet - let me know when this gets merged and I will test it. I don't expect that it will require immediate changes, but it should enable us to handle errors more easily (...no more Internal JSON-RPC error everywhere :)

biandratti:
> @biandratti I meant - if faucet's rpc client handles errors from mantis node properly? IIRC faucet is being run as a separate process communicating with Mantis over RPC.

Good question @kapke 
Any invalid JSON-RPC request (malformed json, etc) should generate a 400 status code. Any way, if we have this error between faucet and node, we probably will have an integration error. Because any invalid json rpc request is validated first from faucet.
So in this way if we have an error with status code 200 or 400  (json rpc error: [-32600, -32602, -32700]) between the faucet and node always we will receive a json rpc error -32000 (status code 200).
Here we can see how it is resolved.
https://github.com/input-output-hk/mantis/blob/800d889ae619f883eb97e6db5cdd1c6e8b4ab996/src/main/scala/io/iohk/ethereum/faucet/jsonrpc/FaucetRpcService.scala#L50

kapke (approved):
LGTM!

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
biandratti: Done

src/test/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServerSpec.scala
biandratti: Done

lemastero (approved):
LGTM 👍  Nice work.

PR #838, opened 2020-12-07
[Config] Add bootstrap{1,2,3}.mantis.pw to mainnet bootstrap list

aakoshh:
# Description

Adds one of the IOHK static node IPs to the ETC bootstrap list.

Tried it with the `bootstrap3.mantis.pw` hostname as well instead of the IP; it works but since all the other entries are by IP, it looks like a DNS server one more centralised component that we don't have to depend on if the IP is added directly.


KonradStaniec:
could you also add as a part of this pr:
```
"enode://fbcd6fc04fa7ea897558c3f5edf1cd192e3b2c3b5b9b3d00be179b2e9d04e623e017ed6ce6a1369fff126661afa1c5caa12febce92dcb70ff1352b86e9ebb44f@18.193.251.235:9076?discport=30303", // bootstrap1.mantis.pw
"enode://1619217a01fb87745bb104872aa84314a2d42d99c7b915cd187245bfd898d679cbf78b3ea950c32051db860e2c4e3fe7d6329107587be33ab37541ca65046f91@18.198.165.189:9076?discport=30303", // bootstrap2.mantis.pw
```

?

KonradStaniec (approved):

PR #840, opened 2020-12-08
[FIX] blacklist for longer in case of header validation failure

KonradStaniec:
# Description

Increases duration for critical offences in fast sync.

Imagine scenario with 3 remote peers - two of them are on wrong fork for us, and only one of them is on correct for, and those two on wrong fork are slightly beter than this on good fork.

Such scenario is possible:
1. we sync up with the first peer on wrong fork, ultimately blacklisting him due to validation error. We rollback N blocks and start from behind
2.we sync up with the second peer on wrong fork, ultimately blacklisting him due to validation error. We rollback N blocks and start from behind
3. during the time we have been syncing with second peer, first peer has been unblacklisted, so we start syncing from him. And the same situation happens.

By increasing blacklist time for validation errors we make sure we will have time to try other peers.

btw. I did not introduce it to regular sync as it not easy to grab header validation error there and i would like to avoid big code changes before release


src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -431,14 +431,20 @@ class FastSync(
...
           case ImportedPivotBlock =>
             updatePivotBlock(ImportedLastBlock)
           case ValidationFailed(header, peerToBlackList) =>
             log.warning(s"validation fo header ${header.idTag} failed")
-            handleRewind(header, peerToBlackList, syncConfig.fastSyncBlockValidationN)
+            // pow validation failure indicate that either peer is malicious or it is on wrong fork
+            handleRewind(
+              header,
+              peerToBlackList,
+              syncConfig.fastSyncBlockValidationN,
```
mmrozek, line 445: I am not sure if we shouldn't rewind more blocks in case of critical failure. Some times ago I stuck in the wrong fork and only the increasing number of blocks helped me. WDYT?

mmrozek (approved):
One doubt only. Apart from that code LGTM

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
KonradStaniec: to be honest all those number were derived from old geth pr which also proided some security analysis so i am not eager the change those just before release, also until we will have proper branch resolving during fast sync all we do here i a little hacky either way. So i would leave it as it is to be honest

aakoshh:
> Imagine scenario with 3 remote peers - two of them are on wrong fork for us, and only one of them is on correct for, and those two on wrong fork are slightly beter than this on good fork.

What does it mean that 2 are on wrong fork "for us"? Which criteria tells that it's wrong? Is it a hard fork we don't want to follow?

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -431,14 +431,20 @@ class FastSync(
...
             // to genesis
             log.warning("Parent chain weight not found for block {}, not processing rest of headers", header.idTag)
-            handleRewind(header, peer, syncConfig.fastSyncBlockValidationN)
+            handleRewind(header, peer, syncConfig.fastSyncBlockValidationN, syncConfig.blacklistDuration)
           case HeadersProcessingFinished =>
             processSyncing()
           case ImportedPivotBlock =>
             updatePivotBlock(ImportedLastBlock)
           case ValidationFailed(header, peerToBlackList) =>
             log.warning(s"validation fo header ${header.idTag} failed")
```
aakoshh, line 440:
```suggestion
            log.warning(s"validation of header ${header.idTag} failed")
```

KonradStaniec:
Is it a hard fork we don't want to follow. i.e take Thanos, hard fork, it changes how proof of work validation work and we do not want to follow nodes which do not activated it.

aakoshh (approved):
Looks reasonable. I don't fully understand what the default case is, the "missing parent weight".

KonradStaniec:
this can happen when lets say there is minor fork in the chain. Let say we are on block `10` with hash `x`, and we ask peer for block from block `11`. Then we receive chain starting from block `11` which states that its parent is block `10` with hash `y`, it may happen that we may not validate this block as we do not validate every header pow, but we need is parent total diffiuctly to update it in db, she then when we try to reach for this block parent we will have `header parent not found error`.

In this case we also rewind blockchain a little bit, and hope fork will be resolved the next time we will get to this place.

PR #844, opened 2020-12-08
[FIX] Empty address handling

mmrozek:
# Description

Empty string should not be validated as the correct address

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
@@ -94,15 +94,14 @@ trait JsonMethodsImplicits {
 
     for {
       from <- input.get("from") match {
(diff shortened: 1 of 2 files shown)


src/main/scala/io/iohk/ethereum/jsonrpc/JsonMethodsImplicits.scala
```diff
@@ -94,15 +94,15 @@ trait JsonMethodsImplicits {
 
     for {
       from <- input.get("from") match {
-        case Some(JString(s)) => extractAddress(s)
+        case Some(JString(s)) if s.nonEmpty => extractAddress(s)
+        case Some(JString(_)) => extractAddress("0x0")
```
kapke, line 98: that line is invalid I think, no one should be able to send tx from "0x0" address

kapke (approved):
One minor thing and LGTM!

biandratti (approved):

biandratti:
LGTM!

PR #846, opened 2020-12-09
[FIX] enable discovery on sagano testnet

KonradStaniec:
# Description

Enables discovery for users node on sagano testnet


aakoshh:
So https://github.com/input-output-hk/mantis/blob/develop/src/universal/conf/base-testnet.conf#L27 doesn't need to be overridden anywhere in the repo?

ghost (approved):
LGTM

KonradStaniec:
So as i understand i could either:
- override it in `"testnet-internal-nomad`
- or modify in `base-testnet`

the other places where `base-testnet` is used is `gac-testnet`, but from what i known this won't be used, so i can directly change it in `base-testnet.conf`

aakoshh:
It seems logical to override it in `testnet-internal-nomad.conf`, at least that's what I thought you did based on the title.

KonradStaniec:
sure, i can go that way if it seems more logical at first sight 👍

aakoshh (approved):

PR #849, opened 2020-12-09
ETCM-463: Add PeerStatisticsActor to track message counts

aakoshh:
# Description

Allow https://github.com/input-output-hk/mantis/pull/833 to prioritise which peers to prune by counting the number of messages coming from them in a fixed time window.

# Proposed Solution

Adds a `TimeSlotStats` counter to the `PeerManagerActor`, which now subscribes to `MessageFromPeer` events by the event bus and adds 1 to the peer every time we receive a response for a request we sent. Doesn't increment the counter for requests sent by the other side, although that could make sense too, depends on what we consider a valuable connection: one that serves us data, or one that makes its presence known by requesting it. 

Some of this can be abused though, like `NewBlockHashes`, if it sends the same notification repeatedly, or the notifications about non-existing hashes; these conditions should be monitored on a different level and penalised. I'm guessing that would mean another component would have to publish events to the message bus about such faults, which would be a new event type.

The `TimeSlotStats` can be used with any statistic that can be merged, so we can always extend it to include faults, data size; less obvious would be the response time, at least not with this subscription, but perhaps the `PeerActor` can send messages to its parent directly instead. 

The idea is that we can call `peerMessageStats.getAll()` when we want to prune and use it to provide an ordering for which peers to prune first: the ones which we haven't pulled any data from in the last 30 minutes for example. 

(I added the counter as a `var` to the actor, seemed like there would be quite some noise if I introduce some kind of state wrapper for all the places `context become ...` occurs, but let me know if that's preferred).


src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -71,6 +70,15 @@ class PeerManagerActor(
   // Subscribe to the handshake event of any peer
   peerEventBus ! Subscribe(SubscriptionClassifier.PeerHandshaked)
 
+  // Subscribe to messages received from peers to maintain stats.
+  peerEventBus ! Subscribe(MessageSubscriptionClassifier)
```
KonradStaniec: My initial tought is that we should probably have separate actor that would gather those stats, to not interfere with `PeerManagerActor` other tasks. (there can be a lot of messages from a lot of peers) my plan would be:
- to move publication of the `PeerHandshaked` event to `PeerManagerActor`. `PeerActor` is child of `PeerManager` so it can inform `PeerManager` via `context.parent`
- then `PeerManagerActor` would be sole publisher of `PeerHandshaked` and `PeerDisconnected` events.
- then create actor `PeerStatiticsActor` which would subscirbe to those two events + all messages. This actor would maintain stats for all handshaked connections.
- `PeerStatiticsActor` could respons two different messages `getStatsForAll` and `getStatsForPeer(peerId)`

Having `PeerHandshaked` and `PeerDisconneted` events published from the same actor is important as this guarantees that `PeerEventBus` always receive `PeerHandshaked` first and `PeerDisconneted` second.

aakoshh: Hm, yes, although it sounds a bit more complicated change, and also make pruning more complicated since now it would involve an extra message exchange. Do you want all this in this PR?

KonradStaniec: I think this change in tracking should be done in this pr, as in case of many peers i would really like to avoid overloading PeerManagerActor with all those messages.

Also I think that this pr should be dealt with first and only after merging this one we should focus on merging pruning connections. (as i really do not like idea of dropping random incoming connections without any stats)

aakoshh: Yeah that was the plan, to merge this and update the other.

aakoshh: Moving the publishing of `PeerHandshakeSuccessful` from `PeerActor` to `PeerManagerActor` could result in subscribers seeing a `MessageFromPeer` first and a the successful handshake later, because the `PeerActor` keeps publishing messages to the event bus, while it has no control over when its parent will publish the handshake on its behalf. 

How about just introducing the `PeerStatisticsActor` in this PR and not try to modify who publishes handshakes and disconnects? It seems unrelated to this PR since I only gather messages, which only ever get published from handshaked peers. It's not a problem if the stats actor misses a disconnection event because eventually the statistics will be overwritten.

aakoshh: Done, the statistics are gathered in the `PeerStatisticsActor`.

src/main/scala/io/iohk/ethereum/network/PeerStatisticsActor.scala
```diff
@@ -0,0 +1,75 @@
...
+
+import akka.actor._
+import io.iohk.ethereum.network.PeerEventBusActor._
+import io.iohk.ethereum.network.p2p.messages.Codes
+import scala.concurrent.duration.FiniteDuration
+import cats.kernel.Monoid
+
+class PeerStatisticsActor(
+    peerEventBus: ActorRef,
+    var maybeStats: Option[TimeSlotStats[PeerId, PeerStatisticsActor.Stat]]
```
KonradStaniec: i would prefer to have here correctly constructed `TimeSlotStats` instance, and fail during actor initlisation if someone provides wrong param otherwise the whole system may think that stats work even if its not true.

src/main/scala/io/iohk/ethereum/network/PeerStatisticsActor.scala
```diff
@@ -0,0 +1,75 @@
...
+    )
+
+  case object GetStatsForAll
+  case class StatsForAll(stats: Map[PeerId, Stat])
+  case class GetStatsForPeer(peerId: PeerId)
+  case class StatsForPeer(peerId: PeerId, stat: Stat)
+
+  val MessageSubscriptionClassifier =
+    SubscriptionClassifier.MessageClassifier(
+      // Subscribe to response types, which indidate that we are getting data from that peer.
```
KonradStaniec: ocd mode: typo in `indidate`

src/main/scala/io/iohk/ethereum/network/TimeSlotStats.scala
```diff
@@ -0,0 +1,132 @@
...
+    val buffer: IndexedSeq[TimeSlotStats.Entry[K, V]]
+) {
+  import TimeSlotStats._
+
+  /** Overall length of the timewindow. */
+  def duration: FiniteDuration = slotDuration * slotCount
+  def slotCount: Int = buffer.size
+
+  /** Merge new stats for a given key in the current timestamp. */
+  def add(key: K, stat: V, timestamp: Timestamp = System.currentTimeMillis): TimeSlotStats[K, V] = {
```
KonradStaniec: as i understand `timestamp` is mainly for testing purposes ? (we do not use it in `PeerStatisticsActor`) Maybe 
`TimeSlotStats` should have dependency on `java.time.Clock` which so it could be mocked in tests, and not expose timing detail to the user.

src/main/scala/io/iohk/ethereum/network/PeerStatisticsActor.scala
```diff
@@ -0,0 +1,75 @@
...
+      sender ! StatsForAll(maybeStats.map(_.getAll()).getOrElse(Map.empty))
+
+    case GetStatsForPeer(peerId) =>
+      sender ! StatsForPeer(peerId, maybeStats.map(_.get(peerId)).getOrElse(Stat.empty))
+  }
+}
+
+object PeerStatisticsActor {
+  case class Stat(
+      responsesReceived: Int
```
KonradStaniec: to consider, maybe it is worth having also the timestamp when we heard from the peer last time ?

KonradStaniec:
@lemastero @biandratti @enriquerodbe as you will taking over project soon, I have added you as a reviewers so you will be aware what is happening here

src/main/scala/io/iohk/ethereum/network/PeerStatisticsActor.scala
aakoshh: Should we add separate settings for it then? I'm maintaining the possibility of not having it because I thought it could be based on the `min-prune-age` which seems like fair play for everyone, old connections and new as well. Otherwise an older peer can have more requests because it had been in the stats longer, and we'll have to resort to taking averages when doing prioritisation. Such settings might be set to zero, which shouldn't mean the system refuses to start. This is why I thought having stats in the `PeerManagerActor` is good because then it's its own jurisdiction of whether it uses them or not. We can potentially extend the interface in a way that `GetStatsFor...` sends a desired duration as well, so different clients can ask for different length windows. Getting complicated though.

Have you given any thought to how you want to use the stats for pruning?

src/main/scala/io/iohk/ethereum/network/PeerStatisticsActor.scala
aakoshh: Cheers :+1:

src/main/scala/io/iohk/ethereum/network/TimeSlotStats.scala
aakoshh: Yes it's a way to control time in the tests, although it could be used as a way to pass in exact observation times. I started out with `cats.effect.Clock[Id]` but it needs `Sync[Id]` which doesn't exist and then resorted to this, it's easier to pass in tests than winding a clock forward and backward. Do you have a strong preference?

src/main/scala/io/iohk/ethereum/network/PeerStatisticsActor.scala
aakoshh: Okay I'll add first and last observation times.

src/main/scala/io/iohk/ethereum/network/PeerStatisticsActor.scala
aakoshh: In general the whole system will have no idea how long the stats are for, only the programmer knows whether what goes on the `PeerStatisticsActor` makes any sense to what they are doing.

src/main/scala/io/iohk/ethereum/network/PeerStatisticsActor.scala
aakoshh: Done, added `lastSeenTimeMillis` and `firstSeenTimeMillis` to `Stat`.

src/main/scala/io/iohk/ethereum/network/TimeSlotStats.scala
aakoshh: Done, using `java.time.Clock`.

src/main/scala/io/iohk/ethereum/network/PeerStatisticsActor.scala
aakoshh: Also because someone may decide that we shouldn't waste the time collecting stats at all, and just set `slot-count` to 0, in which case all the other code that rely on it wouldn't have to change, they just have to deal with the possibility of empty stats.

aakoshh: Added separate configuration for it and the query messages need to send an explicit window they are interested in.

KonradStaniec (approved):

src/test/scala/io/iohk/ethereum/network/TimeSlotStatsSpec.scala
```diff
@@ -0,0 +1,286 @@
...
+object TimeSlotStatsSpec {
+
+  class MockClock extends Clock {
+    private var currentTimeMillis = System.currentTimeMillis
+    def windByMillis(by: Long): Unit =
+      currentTimeMillis = currentTimeMillis + by
+
+    override def instant(): Instant = Instant.ofEpochMilli(currentTimeMillis)
+    override def getZone(): ZoneId = ZoneId.of("UTC")
+    override def withZone(x: ZoneId): Clock = ???
```
biandratti: pending "???"

src/main/scala/io/iohk/ethereum/network/PeerStatisticsActor.scala
```diff
@@ -0,0 +1,113 @@
...
+import scala.concurrent.duration.FiniteDuration
+import cats.kernel.Monoid
+
+class PeerStatisticsActor(
+    peerEventBus: ActorRef,
+    var maybeStats: Option[TimeSlotStats[PeerId, PeerStatisticsActor.Stat]]
+) extends Actor {
+  import PeerStatisticsActor._
+
+  // Subscribe to messages received from handshaked peers to maintain stats.
```
biandratti: Maybe we can encapsulated calls to peerEventBus actor

override def preStart(): Unit = {
 self ! Initialization
}

private def handleInitilization: Receive = {
 // Subscribe to messages received from handshaked peers to maintain stats.
 peerEventBus ! Subscribe(MessageSubscriptionClassifier)
 // Removing peers is an optimisation to free space, but eventually the stats would be overwritten anyway.
 peerEventBus ! Subscribe(SubscriptionClassifier.PeerDisconnectedClassifier(PeerSelector.AllPeers))
}

What do you think?

src/test/scala/io/iohk/ethereum/network/TimeSlotStatsSpec.scala
aakoshh: It will not be called, the `MockClock` doesn't need it.

src/main/scala/io/iohk/ethereum/network/PeerStatisticsActor.scala
aakoshh: I just followed what's happening when the [PeerManagerActor subscribes](https://github.com/input-output-hk/mantis/blob/develop/src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala#L71-L72). 

The `preStart` looks cleaner than a side effecting constructor body, but why not send directly to the event bus from `preStart`, why the extra indirection?

aakoshh: I see [constructor](https://github.com/input-output-hk/mantis/blob/develop/src/main/scala/io/iohk/ethereum/transactions/PendingTransactionsManager.scala#L87) and [preStart](https://github.com/input-output-hk/mantis/blob/develop/src/main/scala/io/iohk/ethereum/blockchain/sync/PeerRequestHandler.scala#L37) both being used in the codebase for subscribing to the event bus. Strictly speaking the actor only needs to subscribe once, it wouldn't have to do it after restarts, which could be [achieved](https://stackoverflow.com/questions/51538451/constructor-vs-prestart-when-to-use-which) with `preStart` only. 

I'll move it into `preStart` :+1:

aakoshh: Done. Didn't want the extra `Receive` because then I'd have to think about whether to keep expecting it forever, or use `context.become` and it seems unnecessary.

src/main/scala/io/iohk/ethereum/network/PeerStatisticsActor.scala
```diff
@@ -0,0 +1,115 @@
...
+      sender ! StatsForAll(stats)
+
+    case GetStatsForPeer(window, peerId) =>
+      val stats = maybeStats.map(_.get(peerId, Some(window))).getOrElse(Stat.empty)
+      sender ! StatsForPeer(peerId, stats)
+  }
+}
+
+object PeerStatisticsActor {
+  case class Stat(
```
enriquerodbe: Maybe `Stat` is big enough to live in its own file?

aakoshh: Done.

src/test/scala/io/iohk/ethereum/network/TimeSlotStatsSpec.scala
biandratti: Yes, I know, it is the same case as the function “getZone()”.
Maybe in this case we can add only:
withZone(x: ZoneId): Clock = Clock.systemDefaultZone()

aakoshh: I didn't check if anything needs `getZone()` internally, so I thought it's easy enough to provide some default value.

If I implemented the method the way you suggest it would not conform to the definition of the method, which is "return a copy of this clock with a different time zone", i.e. if I used `windByMillis` to set the clock to something then used `withZone` I wouldn't see the same time. I think that's more dangerous than throwing an exception that flags up clearly that this method was not expected to be called and it's not implemented, at which point you can go an implement it correctly.

In general I don't think a test-local mock implementation needs to implement methods that the test doesn't rely on, because it's just a distraction. With `???` it's clear that method can be ignored. Even mockito throws if you call methods you never set up (I tried using it initially in this test but it didn't seem to work for some reason). 

Just checked that `getZone` isn't required either, so how about turning that into `???` as well?

aakoshh: Okay because it's so simple I implemented a working `withZone` to be consistent with `getZone`.

biandratti: Good! 
We have pending use of a tool from inspection of code quality (as sonarqube).
I don't sure if we can have any problems with syntax like “???”.
Anyway, good job.

aakoshh: Cheers, thanks for the review!

PR #850, opened 2020-12-09
[ETCM-269] Add missing runToFuture to healthcheck endpoint

enriquerodbe:
# Description

The health check endpoint was returning an unusable String. The reason was that a Monix task was being used in the `complete` method of Akka-http.

# Proposed Solution

Add `runToFuture` to the Monix task in the highest layer of the call stack.

# Testing

Unit test added.


lemastero (approved):
LGTM 👍

mmrozek (approved):
LGTM

biandratti (approved):

PR #856, opened 2020-12-11
[ETCM-472] Add missing field to receipts response

KonradStaniec:
# Description

According to spec https://eth.wiki/json-rpc/API we were missing few field:
`from`, `to`, `logsBloom`, `root`, `status`.


lemastero (approved):

src/main/scala/io/iohk/ethereum/jsonrpc/TransactionReceiptResponse.scala
```diff
@@ -1,16 +1,105 @@
...
+      blockNumber = blockHeader.number,
+      blockHash = blockHeader.hash,
+      from = signedTransactionSender,
+      to = stx.tx.receivingAddress,
+      cumulativeGasUsed = receipt.cumulativeGasUsed,
+      gasUsed = gasUsedByTransaction,
+      contractAddress = contractAddress,
+      logs = txLogs,
+      logsBloom = receipt.logsBloomFilter,
+      _: Option[ByteString],
```
enriquerodbe: I think it would be simpler to have:
```scala
val (root, status) = receipt.postTransactionStateHash match {
  case FailureOutcome => (None, Some(0))
  ...
}
```

and finally:

```scala
TransactionReceiptResponse(
  transactionHash = stx.hash,
  ...
  root,
  status
)
```

enriquerodbe (approved):

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -346,39 +345,22 @@ class EthService(
...
-              transactionIndex = txIndex,
-              transactionHash = stx.hash,
-              blockHash = header.hash,
-              blockNumber = header.number,
-              address = txLog.loggerAddress,
-              data = txLog.data,
-              topics = txLog.logTopics
-            )
-          }
+          receipt,
```
biandratti: Only a detail:
TransactionReceiptResponse(
          receipt= receipt,
          stx =stx,
          signedTransactionSender= sender,
          transactionIndex =txIndex,
          blockHeader= header,
          gasUsedByTransaction= gasUsed
        )

src/test/scala/io/iohk/ethereum/jsonrpc/EthServiceSpec.scala
```diff
@@ -1072,25 +1071,12 @@ class EthServiceSpec
...
-                transactionIndex = 1,
-                transactionHash = contractCreatingTransaction.hash,
-                blockHash = Fixtures.Blocks.Block3125369.header.hash,
-                blockNumber = Fixtures.Blocks.Block3125369.header.number,
-                address = fakeReceipt.logs.head.loggerAddress,
-                data = fakeReceipt.logs.head.data,
-                topics = fakeReceipt.logs.head.logTopics
-              )
-            )
+            fakeReceipt.copy(cumulativeGasUsed = fakeReceipt.cumulativeGasUsed + gasUsedByTx),
```
biandratti: ditto

biandratti:
In this case, do we need a label "BRAKES API"?

KonradStaniec:
imo we do not need this label as this pr only adds new fields which is not an api breaking operation

biandratti:
LGTM!

biandratti (approved):

PR #857, opened 2020-12-11
[ETCM-468] add getProof to MPT

lemastero:
# Description

One of the elements required for [EIP-1186](https://eips.ethereum.org/EIPS/eip-1186) is to be able to get proof from MerklePartitiaTrie. This PR adds this. Both `get` and `getProof` has some common parts encoded as `pathTraverse`


src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
```diff
@@ -594,4 +602,68 @@ class MerklePatriciaTrieSuite extends AnyFunSuite with ScalaCheckPropertyChecks
...
+      .put(key3, val3)
+
+    // when
+    val proof = trie.getProof(Hex.decode("00000001"))
+
+    // then
+    assert(proof.isEmpty)
+  }
+
+  test("getProof returns valid proof for existing key") {
```
enriquerodbe, line 553: It would be nice to have a generator for Merkle Patricia Trees so we could check properties instead of case by case.

pslaski (approved):

src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
```diff
@@ -11,12 +12,19 @@ import io.iohk.ethereum.mpt.MerklePatriciaTrie.{MPTException, defaultByteArraySe
...
 import scala.util.{Random, Try}
 import org.scalatest.funsuite.AnyFunSuite
 
 class MerklePatriciaTrieSuite extends AnyFunSuite with ScalaCheckPropertyChecks with ObjectGenerators {
 
-  val EmptyEphemNodeStorage = StateStorage.createTestStateStorage(EphemDataSource())._1.getBackingStorage(0)
+  val source: EphemDataSource = EphemDataSource()
+  val tuple: (StateStorage, NodeStorage, CachedNodeStorage) = StateStorage.createTestStateStorage(source)
+  val EmptyEphemNodeStorage = {
+    val value: StateStorage = tuple._1
```
biandratti: Maybe we can use this

val (stateStorage, nodeStorage, achedNodeStorage): (StateStorage, NodeStorage, CachedNodeStorage) = StateStorage.createTestStateStorage(source)
  val EmptyEphemNodeStorage = {
    val value: StateStorage = stateStorage
    value.getBackingStorage(0)
  }
  val EmptyNodeStorage = nodeStorage

biandratti (approved):

src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
lemastero: Nice idea, I end up with:
```scala
  val (stateStorage, emptyNodeStorage, _) = StateStorage.createTestStateStorage(dataSource)
  val emptyEphemNodeStorage = stateStorage.getBackingStorage(0)
```
good enough?

src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
lemastero: It is a bit tricky to make MPT generator, because MPT is a persistend data structure - this means it store in some persistend store like DB (in our case rocksdb) current state.

So we would need to generate MPT, storage, but then what about serializers, deserializers, MPT can have different types for keys and values (usually those are `Array[Byte]` but for storage (smart contract code) it is `BigInt`.

But there are already property tests. I changed this guy using the approach. (I needed to understand those tests and I noticed some duplications - so there are so helper methods extracted).

WDYT is it good enough?

enriquerodbe: Looks good yes, I wasn't aware about the MPT being coupled with the storage, but I'll have to learn more about that :)

enriquerodbe (approved):

PR #858, opened 2020-12-11
[CHORE] Add encrypt-key command to CLI

mmrozek:
# Description

`encrypt-key` command could be useful to set up a faucet


src/main/scala/io/iohk/ethereum/cli/CliCommands.scala
```diff
@@ -74,6 +78,23 @@ object CliCommands extends SecureRandomBuilder {
       }
     }
 
+  private val EncryptKey: Command[String] =
+    Command(name = encryptKeyCommand, header = "Encrypt private key") {
+
+      val privateKey = Opts
```
kapke: Shouldn't common options be defined outside commands? Or maybe even all options should replace current string constants in lines ~17-25?

src/test/scala/io/iohk/ethereum/cli/CliCommandsSpec.scala
```diff
@@ -92,6 +94,30 @@ class CliCommandsSpec extends AnyFlatSpec with Matchers with EitherValues {
...
+
+    val decrypted = (for {
+      encrypted <- EncryptedKeyJsonCodec.fromJson(json)
+      decrypted <- encrypted.decrypt("")
+    } yield decrypted).value
+
+    ByteStringUtils.hash2string(decrypted) shouldBe privateKey
+  }
+
+  it should "encrypt private key (with passphrase)" in {
```
kapke, line 110: minor: this test looks like a perfect candidate for a property-based one.

README.md
```diff
@@ -50,6 +50,11 @@ Possible networks: `etc`, `eth`, `mordor`, `testnet-internal`
 ./bin/mantis cli generate-key-pairs 5
 ```
 
+- encrypt private key (could be used to setup private faucet)
```
kapke: Probably bit more documentation (and explanation) would be helpful. Either here or in the help for that command

src/test/scala/io/iohk/ethereum/cli/CliCommandsSpec.scala
mmrozek: Not really. I don't want to test decryption/encryption here. It should be done in a different place. I only want to check if I get the correct JSON when I specify the correct arguments

src/main/scala/io/iohk/ethereum/cli/CliCommands.scala
mmrozek: I prefer to leave string constants because it is convenient to use them in tests

biandratti (approved):

lemastero (approved):
LGTM 👍

kapke (approved):
LGTM!

PR #859, opened 2020-12-14
[CHORE] Make election of pivot block more robust

mmrozek:
# Description

If Mantis connects with the malicious peer with the best block number bigger than other nodes (the difference bigger than `pivotBlockOffset`) it will never choose pivot block and will never start to sync.

# Proposed Solution

Blacklist the best peer if there are enough peers to retry the election process without it.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/fast/PivotBlockSelector.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/fast/PivotBlockSelector.scala
@@ -35,16 +35,16 @@ class PivotBlockSelector(
   override def receive: Receive = idle
 
   def idle: Receive = handleCommonMessages orElse { case SelectPivotBlock =>
-    val election @ ElectionDetails(correctPeers, expectedPivotBlock) = collectVoters
+    val election @ ElectionDetails(eligiblePeers, allPeers, expectedPivotBlock) = collectVoters
 
-    if (election.isEnoughVoters(minPeersToChoosePivotBlock)) {
+    if (election.isEnoughEligibleVoters(minPeersToChoosePivotBlock)) {
 
-      val (peersToAsk, waitingPeers) = correctPeers.splitAt(minPeersToChoosePivotBlock + peersToChoosePivotBlockMargin)
+      val (peersToAsk, waitingPeers) = eligiblePeers.splitAt(minPeersToChoosePivotBlock + peersToChoosePivotBlockMargin)
 
       log.info(
-        "Trying to choose fast sync pivot block using {} peers ({} correct ones). Ask {} peers for block nr {}",
-        peersToDownloadFrom.size,
-        correctPeers.size,
+        "Trying to choose fast sync pivot block using {} peers ({} eligible ones). Ask {} peers for block nr {}",
+        allPeers.size,
+        eligiblePeers.size,
         peersToAsk.size,
         expectedPivotBlock
       )
@@ -63,10 +63,13 @@ class PivotBlockSelector(
       log.info(
         "Cannot pick pivot block. Need at least {} peers, but there are only {} which meet the criteria ({} all available at the moment). Retrying in {}",
         minPeersToChoosePivotBlock,
-        correctPeers.size,
-        peersToDownloadFrom.size,
+        eligiblePeers.size,
+        allPeers.size,
         startRetryInterval
       )
+      if (election.isEnoughVoters(minPeersToChoosePivotBlock + 1)) { // TODO when we should start blacklisting best peers?
+        blacklist(eligiblePeers.head.id, blacklistDuration, "The best peer has too big best block number, blacklisting")
+      }
       scheduleRetry(startRetryInterval)
     }
   }
@@ -185,11 +188,11 @@ class PivotBlockSelector(
       .map { case (_, bestPeerBestBlockNumber) => bestPeerBestBlockNumber }
       .getOrElse(BigInt(0))
     val expectedPivotBlock = (bestPeerBestBlockNumber - syncConfig.pivotBlockOffset).max(0)
-    val correctPeers = peersSortedByBestNumber
+    val expectedEligiblePeers = peersSortedByBestNumber
       .takeWhile { case (_, number) => number >= expectedPivotBlock }
       .map { case (peer, _) => peer }
 
-    ElectionDetails(correctPeers, expectedPivotBlock)
+    ElectionDetails(expectedEligiblePeers, peersSortedByBestNumber.map(_._1), expectedPivotBlock)
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/fast/PivotBlockSelector.scala
```diff
@@ -63,10 +63,13 @@ class PivotBlockSelector(
...
         "Cannot pick pivot block. Need at least {} peers, but there are only {} which meet the criteria ({} all available at the moment). Retrying in {}",
         minPeersToChoosePivotBlock,
-        correctPeers.size,
-        peersToDownloadFrom.size,
+        eligiblePeers.size,
+        allPeers.size,
         startRetryInterval
       )
+      if (election.isEnoughVoters(minPeersToChoosePivotBlock + 1)) { // TODO when we should start blacklisting best peers?
+        blacklist(eligiblePeers.head.id, blacklistDuration, "The best peer has too big best block number, blacklisting")
```
biandratti: Maybe you can add bestPeer in the case class ElectionDetails. Or you can change the name eligiblePeers by eligiblePeersSortedByBestNumber. I think that by this way, it is more understandable that the list “eligiblepeers” was sorted, then you try to get the best peer “eligiblepeers.head.id”.

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/PivotBlockSelector.scala
```diff
@@ -185,11 +188,11 @@ class PivotBlockSelector(
       .map { case (_, bestPeerBestBlockNumber) => bestPeerBestBlockNumber }
```
biandratti, line 190: In the future maybe we can decouple the function “collectVoters” of this class, and reduce the complexity. By this way the function is moved to another class and we can add more unit tests. What do you think?

mmrozek: I am not sure about that. For me, this function doesn't have much sense in other contexts. It is collecting voters for pivot block election and only for that

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/PivotBlockSelector.scala
```diff
@@ -63,10 +63,15 @@ class PivotBlockSelector(
       log.info(
         "Cannot pick pivot block. Need at least {} peers, but there are only {} which meet the criteria ({} all available at the moment). Retrying in {}",
         minPeersToChoosePivotBlock,
-        correctPeers.size,
-        peersToDownloadFrom.size,
+        eligiblePeers.size,
+        allPeers.size,
         startRetryInterval
       )
+      if (election.isEnoughVoters(minPeersToChoosePivotBlock + 1)) { // TODO when we should start blacklisting best peers?
```
KonradStaniec, line 70: To be honest i am not sure about it. I am worried about the case like:
- we start our node, and we connect to one bootstrap node (which is deemed as truthful) 
- we find 3 just starting peers on block 1000
- we blacklist our bootstrap as it is to far a long
- we sync up to one of low block nodes on block 950

Maybe we should have similiar mechanism as besu that has few levels of checking  like: at first check if there are voters in range `best - 32` ,then in `best - 64` etc. up to lets say `128` ? It is not perferct but a little more robust to differnt network condition than our current solution.

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/PivotBlockSelector.scala
```diff
@@ -35,16 +35,16 @@ class PivotBlockSelector(
   override def receive: Receive = idle
 
   def idle: Receive = handleCommonMessages orElse { case SelectPivotBlock =>
```
lemastero, line 37: WDYT about extracting method from code used in orElse?
I would go even further and extract methods for both branches in `if` inside.

src/test/scala/io/iohk/ethereum/blockchain/sync/PivotBlockSelectorSpec.scala
```diff
@@ -110,6 +110,52 @@ class PivotBlockSelectorSpec
     )
   }
 
+  it should "retry if there are no enough eligible peers" in new TestSetup {
```
lemastero, line 113: can we somehow show that there is no enough eligible peers (assert, name etc)? it is hard for me to figure out that :)

src/test/scala/io/iohk/ethereum/blockchain/sync/PivotBlockSelectorSpec.scala
```diff
@@ -110,6 +110,52 @@ class PivotBlockSelectorSpec
...
+    updateHandshakedPeers(HandshakedPeers(threeAcceptedPeers))
+
+    time.advance(syncConfig.startRetryInterval)
+
+    peerMessageBus.expectMsgAllOf(
+      Subscribe(MessageClassifier(Set(Codes.BlockHeadersCode), PeerSelector.WithId(peer1.id))),
+      Subscribe(MessageClassifier(Set(Codes.BlockHeadersCode), PeerSelector.WithId(peer2.id))),
+      Subscribe(MessageClassifier(Set(Codes.BlockHeadersCode), PeerSelector.WithId(peer3.id)))
+    )
+  }
```
lemastero, line 134: I could use
// given
// when
// then
and possibly comments on two assertions that are done in this test, it is unclear for me how retry manifest

src/test/scala/io/iohk/ethereum/blockchain/sync/PivotBlockSelectorSpec.scala
```diff
@@ -110,6 +110,52 @@ class PivotBlockSelectorSpec
...
+
+    time.advance(syncConfig.startRetryInterval)
+
+    peerMessageBus.expectMsgAllOf(
+      Subscribe(MessageClassifier(Set(Codes.BlockHeadersCode), PeerSelector.WithId(peer2.id))),
+      Subscribe(MessageClassifier(Set(Codes.BlockHeadersCode), PeerSelector.WithId(peer3.id))),
+      Subscribe(MessageClassifier(Set(Codes.BlockHeadersCode), PeerSelector.WithId(peer4.id)))
+    )
+    peerMessageBus.expectNoMessage()
+
```
lemastero, line 156: how this blacklisting manifest? Is it possible to assert that?

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/PivotBlockSelector.scala
aakoshh: :heavy_plus_sign: on this.

Besu implements this strategy in [PivotBlockRetriever](https://github.com/hyperledger/besu/blob/8b6f8b6e2f8f498af71523580eb0d6e6779d06c7/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/PivotBlockRetriever.java), [FastSyncActions](https://github.com/hyperledger/besu/blob/8b6f8b6e2f8f498af71523580eb0d6e6779d06c7/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/FastSyncActions.java) and perhaps [PivotBlockConfirmer](https://github.com/hyperledger/besu/blob/8b6f8b6e2f8f498af71523580eb0d6e6779d06c7/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/fastsync/PivotBlockConfirmer.java), with the [default distance](https://github.com/hyperledger/besu/blob/8b6f8b6e2f8f498af71523580eb0d6e6779d06c7/ethereum/eth/src/main/java/org/hyperledger/besu/ethereum/eth/sync/SynchronizerConfiguration.java#L29) being 50.

enriquerodbe:
After discussion with the author and the team, we're not going to merge this as is because it needs more work. We can take a similar approach to what Besu or geth do.

PR #860, opened 2020-12-15
ETCM-480 Add buildinfo endpoint

enriquerodbe:
# Description

Add a new API endpoint under `/buildinfo` which returns a json object containing project name, version, Scala version, sbt version, git head commit, git current branch, git current tags, git described version, git uncommitted changes and library dependencies.

# Proposed Solution

The sbt-buildinfo plugin was already added so it was easily solved by just adding the required configs to the `build.sbt` and using it from the server endpoints definitions.

No big changes introduced.

# Testing

Unit test added.


lemastero (approved):

aakoshh (approved):
:+1:

biandratti (approved):

PR #861, opened 2020-12-16
Add sagano testnet configuration

lemastero:
# Description

Currently, build mantis client, do not have a config for sagano network out of the box. You need to rename current conf file.

# Proposed Solution

Copy config testnet-internal-nomad.conf to sagano.conf

# Testing

After building the client you should be able to run:
```bash
./bin/mantis-launcher sagano
```


src/universal/conf/sagano.conf
```diff
@@ -0,0 +1,12 @@
+include "base-testnet.conf"
+
+mantis {
```
enriquerodbe: Instead of copying everything, wouldn't it be better to `include "testnet-internal-nomad.conf"`?

faraggi:
How about an empty file and importing the previous conf until devops is updated?

src/universal/conf/sagano.conf
lemastero: Very good idea. Done. Tested locally.

enriquerodbe (approved):
I saw it later, but I think @faraggi comment was in the same direction

faraggi (approved):
LGTM

PR #863, opened 2020-12-17
Made some tests able to succeed more than once

dmitry-worker:
# Description

Some of the tests, where the `ActorSystem` was involved could be run successfully only once:
-`MantisServiceSpec`
-`TransactionHistoryServiceSpec`
-`RegularSyncSpec`

On the second run a `ClassCastException` was thrown

```
[info] MantisServiceSpec:                                                                                                                                                                                                                          
[info] Mantis Service                                                                                                                                                                                                                              
[info] - should get account's transaction history (1 second, 351 milliseconds)                                                                                                                                                                     
[info] - should validate range size against configuration *** FAILED *** (83 milliseconds)                                                                                                                                                         
[info]   java.lang.ClassCastException: interface akka.event.LoggingFilter is not assignable from class akka.event.slf4j.Slf4jLoggingFilter
```

Initial research told that something is spoiling the environment, because clean run always succeeded.
But the main source of the problem lied deeper - in a classloader! 

This seems not to be very suspicious:
```java.lang.ClassCastException: interface akka.event.LoggingFilter is not assignable from class akka.event.slf4j.Slf4jLoggingFilter```

But this situation:
```java.lang.ClassCastException: interface akka.event.LoggingFilter is not assignable from class akka.event.DefaultLoggingFilter```

...in a single jvm can be caused only when:
- LoggingFilter was loaded with ClassLoader 1
- DefaultLoggingFilter was loaded using ClassLoader 2

And here is a proof that it is a very annoying bug in akka itself:
https://samwrx.surge.sh/akka/2016/01/11/akkaloggingfilters.html


# Proposed Solution

**Always** specify a classloader when creating an ActorSystem.
This might seem ugly, but it works... And there's not much one can do with an upstream issue.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
+++ b/src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
@@ -74,7 +74,7 @@ trait AsyncConfigBuilder {
 }
 
 trait ActorSystemBuilder {
-  implicit lazy val system: ActorSystem = ActorSystem("mantis_system")
+  implicit lazy val system: ActorSystem = ActorSystem("mantis_system", classLoader = Some(getClass.getClassLoader))
 }
 
 trait PruningConfigBuilder extends PruningModeComponent {
--- a/src/test/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSyncSpec.scala
+++ b/src/test/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSyncSpec.scala
@@ -47,13 +47,13 @@ class RegularSyncSpec
   type Fixture = RegularSyncFixture
 
   val actorSystemResource =
-    Resource.make(Task { ActorSystem() })(system => Task { TestKit.shutdownActorSystem(system) })
+    Resource.make(Task { ActorSystem("default", classLoader = Some(getClass.getClassLoader)) })(system => Task { TestKit.shutdownActorSystem(system) })
   val fixtureResource = actorSystemResource.map(new Fixture(_))
 
   // Used only in sync tests
   var testSystem: ActorSystem = _
   override def beforeEach: Unit =
-    testSystem = ActorSystem()
+    testSystem = ActorSystem("default", classLoader = Some(getClass.getClassLoader))
   override def afterEach: Unit =
     TestKit.shutdownActorSystem(testSystem)
 

### CODE REVIEW

**Comment by kapke** [comment]
It may actually caused by sbt, which does quite a log of mangling of class loaders if tests are not forked to separate JVM.

**Comment by dmitry-worker** [comment]
> It may actually caused by sbt, which does quite a log of mangling of class loaders if tests are not forked to separate JVM.

@kapke Exactly! `currentThread` context supplied by sbt is not guaranteed to be the same during the session. 

As the documentation on akka says:
```
apply() | Creates a new ActorSystem with the name "default",  
obtains the current ClassLoader by first inspecting the current threads' getContextClassLoader,  
then tries to walk the stack to find the callers class loader, 
then falls back to the ClassLoader  associated with the ActorSystem class.
```

We can fork the tests, yes, that would be an alternative fix, that I didn't try because it will alter all the testing process, but I wouldn't allow a system to search and pick any first `classLoader` it finds.

File: src/test/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSyncSpec.scala [hunk 1]
```diff
@@ -47,13 +47,13 @@ class RegularSyncSpec
...
   val actorSystemResource =
-    Resource.make(Task { ActorSystem() })(system => Task { TestKit.shutdownActorSystem(system) })
+    Resource.make(Task { ActorSystem("default", classLoader = Some(getClass.getClassLoader)) })(system => Task { TestKit.shutdownActorSystem(system) })
   val fixtureResource = actorSystemResource.map(new Fixture(_))
 
   // Used only in sync tests
   var testSystem: ActorSystem = _
   override def beforeEach: Unit =
-    testSystem = ActorSystem()
+    testSystem = ActorSystem("default", classLoader = Some(getClass.getClassLoader))
```
lemastero, line 56: It looks a bit strange that we need 2 actor systems for one test and we use a non-standard way to close them. Normally we sth like:

```scala
class DebugServiceSpec
    extends TestKit(ActorSystem("ActorSystem_DebugServiceSpec"))
    with AnyFlatSpecLike
    with WithActorSystemShutDown
```

Maybe the issue arise, because we do not close properly ActorSystem and there are issues because of that?

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
```diff
@@ -74,7 +74,7 @@ trait AsyncConfigBuilder {
 }
 
 trait ActorSystemBuilder {
-  implicit lazy val system: ActorSystem = ActorSystem("mantis_system")
+  implicit lazy val system: ActorSystem = ActorSystem("mantis_system", classLoader = Some(getClass.getClassLoader))
```
lemastero, line 77: I think manual management of class loader for production code, to solve an issue with tests that fail when a run a couple of times might not be a good idea.

dmitry-worker: Right, I hate this one too, but the `Fixture` used in tests also brings on the Mantis Node cake, which isn't good either.
I have two ideas around that: 
1) We could get rid of this by implementing a `Fixture` that will not trigger a whole `NodeBuilder`
    This can be serious refactoring for the system that is built all around that pattern
2) We can lean to a particular `ClassLoader` to prevent `ActorSystem` from voluntarily selecting one. 
    I'm thankful to failed tests and also suppose that `it-will-not-fail-in-production` is a kind of survivorship bias. The production environment is clear and isolated, of course, but that particular bug may arise in another circumstances.
    
    Personally, I'd never rely on environment and will apply maximum effort to guarantee a robust behavior in any.

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSyncSpec.scala
dmitry-worker: >Maybe the issue arise, because we do not close properly ActorSystem and there are issues because of that?

Hmm, probably, if an ActorSystem wasn't properly closed `afterEach` test, we'd have errors on a first run, but I have no errors at all there.

lemastero:
> We can fork the tests, yes, that would be an alternative fix
IMHO this would be nice solution, it require only flag in build.sbt and no low level class loader management

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
kapke: You can also remove `classLoader` parameter in "production" part of code and overwrite it in tests only (and maybe even extract the override to a trait)

dmitry-worker:
Closed in favor of https://github.com/input-output-hk/mantis/pull/868

PR #841, opened 2020-12-08
Include PPoW info in BlockResponse

rtkaczyk:
This should be helpful when testing the Permissioned Proof of Work consensus.

**PLEASE DO NOT DELETE THE BRANCH WHEN MERGING**

PR #867, opened 2020-12-21
Include PPoW info in BlockResponse 

rtkaczyk:
Continuing from #841


enriquerodbe (approved):

src/main/scala/io/iohk/ethereum/jsonrpc/BlockResponse.scala
```diff
@@ -61,6 +68,17 @@ object BlockResponse {
 
     val (lcn, td) = weight.map(_.asTuple).separate
 
+    val signature =
+      if (block.header.extraData.length >= ECDSASignature.EncodedLength)
```
biandratti, line 72: Just a minor comment, when we have the condition (block.header.extraData.length =  ECDSASignature.EncodedLength) 
The following line:
block.header.extraData.takeRight(ECDSASignature.EncodedLength) is not required, right?
Anyway, this code is clear and understandable.

biandratti (approved):

lemastero (approved):
LGTM 👍

PR #870, opened 2020-12-21
[ETCM-423] remove mallet module

biandratti:
# Description

Today this module is not necessary. It was CLI minimal wallet app.
I doesn't seem to provide any more functionality than any web3 implementation, with the only exception being that it compile calls to IELE contracts.
Even if we eventually go back to using IELE or another VM I think this functionality should be implemented in a fork of web3.


src/main/scala/io/iohk/ethereum/utils/StringUtils.scala
```diff
@@ -1,9 +1,8 @@
-package io.iohk.ethereum.mallet.common
+package io.iohk.ethereum.utils
 
 import akka.util.ByteString
-import org.bouncycastle.util.encoders.Hex
 
-object StringUtil {
+object StringUtils {
 
   def unquote(s: String): String = {
```
lemastero: Can we get rid of this method, since we already touching this code?
It is unused and it throws exceptions without saying so in the type signature.

biandratti: sure, thank you for this detail!

enriquerodbe (approved):

src/main/scala/io/iohk/ethereum/utils/StringUtils.scala
lemastero: I see you deleted other unused stuff - nice catch! 👍

lemastero (approved):
LGTM 👍

PR #873, opened 2020-12-23
[ETCM-266]-replaced-rate-limiter-built-on-twitter

dmitry-worker:
# Description

Twitter-collections used for `LRU` cache for Scala 2.13 is unavailable.
Had to replace this `LRU` by something with similar functionality

# Proposed Solution

The simplest possible `LRU` is built on `LinkedHashMap`
Boilerplate code is replaced by `Directive0` implementation
Further improvement steps offered for JsonRPC server


src/main/scala/io/iohk/ethereum/db/cache/SimpleLRU.scala
```diff
@@ -0,0 +1,36 @@
...
+import java.util
+
+/**
+  * Simplest-possible implementation of insert-time based LRU
+  *
+  * @param maxElements - capacity of the inner map, between 0 and this value.
+  * @param ttlMillis   - entry invalidation interval
+  * @tparam K - stands for the type of the keys
+  */
+class SimpleLRU[K](maxElements: Int, ttlMillis: Long) {
```
enriquerodbe: Could we use Guava for this? I think we have that dependency anyway. I'm not sure we should have this kind of low-level implementations in this code base.

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/RateLimit.scala
```diff
@@ -1,25 +1,42 @@
...
+class RateLimit(config: RateLimitConfig) extends Directive0 with Json4sSupport {
+
+  private implicit val serialization: Serialization = native.Serialization
+  private implicit val formats: Formats = DefaultFormats + JsonSerializers.RpcErrorJsonSerializer
+
+  // It determines whether a request needs to be blocked
+  // Such algebras prevent if-elseif-else boilerplate in the JsonRPCServer code
+  val blockingAlgebra: (RemoteAddress => Boolean) = {
+    if (config.enabled) {
+      val lru = new SimpleLRU[RemoteAddress](
```
enriquerodbe: Can this `val lru` be declared and initialized in the outer `RateLimit` class body? I think that would make it clear that this is the same cache instance always.

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/RateLimit.scala
```diff
@@ -1,25 +1,42 @@
...
+
+
+class RateLimit(config: RateLimitConfig) extends Directive0 with Json4sSupport {
+
+  private implicit val serialization: Serialization = native.Serialization
+  private implicit val formats: Formats = DefaultFormats + JsonSerializers.RpcErrorJsonSerializer
+
+  // It determines whether a request needs to be blocked
+  // Such algebras prevent if-elseif-else boilerplate in the JsonRPCServer code
+  val blockingAlgebra: (RemoteAddress => Boolean) = {
```
enriquerodbe: This is a broader discussion we can have with the team, but in general I'm in favor of `def`s instead of `val` with function values.

src/test/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServerSpec.scala
```diff
@@ -215,7 +215,7 @@ class JsonRpcHttpServerSpec extends AnyFlatSpec with Matchers with ScalatestRout
       status shouldEqual StatusCodes.TooManyRequests
     }
 
-    fakeClock.advanceTime(2 * serverConfigWithRateLimit.rateLimit.minRequestInterval.toMillis)
+    Thread.sleep(30)
```
enriquerodbe: I'm not sure how this works, but it looks better to have a mock clock that we can move as we want rather than sleeping threads.

src/test/scala/io/iohk/ethereum/db/cache/SimpleLRUSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+      val results = (0 until 100).map { i => MockedLRU.checkAndRefreshEntry(i) }
+      assert( results.forall( _ == false ) )
+    }
+
+    "Drop the records according to maxElements" in {
+      val existing = (90 until 100).map { i => MockedLRU.checkAndRefreshEntry(i) }
+      assert( existing.forall( _ == true ) )
+
+      // maxElements guaranteed that we have no space for those
+      val absent = (0 until 10).map { i => MockedLRU.checkAndRefreshEntry(i) }
```
enriquerodbe: Why did the cache remove elements `0 until 10`? If elements `90 until 100` already existed there should still be space for those, right?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/RateLimit.scala
dmitry-worker: Sure, but the trick was to *not* create any LRUs in case when ratelimit is disabled in config.

src/main/scala/io/iohk/ethereum/db/cache/SimpleLRU.scala
dmitry-worker: Yes, I would certainly use guava when we have to store some values, not just the keys.

BUT even then I'd go caffeine based on the performance results: 
https://github.com/google/guava/issues/2063

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/RateLimit.scala
dmitry-worker: `blockingAlgebra` will return one function when ratelimit is on and another function when it's off.
As soon as it should be chosen only once, as config is read only on application start, it definitely should be the `val`

src/test/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServerSpec.scala
dmitry-worker: This is probably the weakest point in the PR, but I would remove that test anyway, because the ability to drop obsolete values from the cache is covered in another test (`SimpleLRUSpec`). So it might be an overkill to test it again and provide an external clock to test it.

src/test/scala/io/iohk/ethereum/db/cache/SimpleLRUSpec.scala
dmitry-worker: This test makes sure that when `90 until 100` are inside - and storage limit is 10 there's definitely no space for `0 until 10` - and they are absent

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/RateLimit.scala
dmitry-worker: But I've got an idea, thanks! Probably we don't even have to extract any IP when the feature is disabled...

src/main/scala/io/iohk/ethereum/db/cache/SimpleLRU.scala
enriquerodbe: Why not caffeine with dummy values?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/RateLimit.scala
```diff
@@ -1,25 +1,43 @@
...
+    config.latestTimestampCacheSize,
+    config.minRequestInterval.toMillis
+  )
+
+  // Such algebras prevent if-elseif-else boilerplate in the JsonRPCServer code
+  // It is also guaranteed that:
+  //   1) config.enabled is checked only once - on route init
+  //   2) no IP address is extracted unless config.enabled is true
+  //   3) no LRU is created unless config.enabled is true
+  val rateLimitAlgebra: (Unit => Route) => Route = {
```
enriquerodbe: This looks like premature optimization to me, I would leave a `def` that reads the `config.enabled` value each time. But if we're going to do this, we could have a `LimitStrategy` trait with two implementations: `RateLimit` (extract IP, have the LRU cache instantiated, limit if necessary) and `NoLimit` (just a no-op). The `JsonRpcHttpServer` would read the config once and instantiate the corresponding `LimitStrategy`. What do you think?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -54,37 +54,39 @@ trait JsonRpcHttpServer extends Json4sSupport with RateLimit with Logger {
...
     } ~ (path("buildinfo") & pathEndOrSingleSlash & get) {
       handleBuildInfo()
     } ~ (pathEndOrSingleSlash & post) {
-      (extractClientIP & entity(as[JsonRpcRequest])) { (clientAddress, request) =>
-        handleRequest(clientAddress, request)
-      } ~ entity(as[Seq[JsonRpcRequest]]) { request =>
-        handleBatchRequest(request)
+      // TODO: maybe rate-limit this one too?
+      entity(as[JsonRpcRequest]) {
+        case statusReq if statusReq.method == FaucetJsonRpcController.Status =>
```
biandratti, line 67: Here we can coordinate with devops. And change the endpoint used for the status health check. 
Today we are validating the status “healthcheck” through the endpoint json rpc “faucet_status”. Here we can change it for the request [GET] http “healthcheck”. This endpoint internally, works the same. 

In this way, we can remove the following condition:
if statusReq.method == FaucetJsonRpcController.Status

In this point we need coordinate with devops and change the endpoint that we have in the following line:
https://github.com/input-output-hk/mantis-ops/blob/784f14f0218ef9f383cd57d96367d6e6a07b5ea7/docker/mantis-faucet.nix#L8

What do you think?

src/main/scala/io/iohk/ethereum/db/cache/SimpleLRU.scala
biandratti: If we use the guava library we can make use of FakeTicker to advance the time at will. So in this way, we don't need to use  Thread.sleep in unit test

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
dmitry-worker: I'm looking forward to remove this workaround, but we should be aware if someone else is using this endpoint for health checking... If it is not the case, i'm sure I can remove it. WDYT, @dranik-blox?

src/main/scala/io/iohk/ethereum/db/cache/SimpleLRU.scala
dmitry-worker: @enriquerodbe @biandratti finished transition to guava :)
Upgraded that test as well.

src/test/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServerSpec.scala
dmitry-worker: Done

enriquerodbe (approved):

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/RateLimit.scala
```diff
@@ -1,25 +1,67 @@
...
+          if (exists) {
+            val err = JsonRpcError.RateLimitError(minInterval)
+            complete((StatusCodes.TooManyRequests, err))
+          } else {
+            f.apply(())
+          }
+        }
+      } else f.apply(())
   }
+
```
biandratti, line 70: Maybe we can encapsulate the validation behavior, for example:

private def isBelowRateLimit(ip: RemoteAddress): Boolean = {
    var exists = true
    lru.get(
      ip,
      () => {
        exists = false
        NotUsed
      }
    )
    exists
  }

And then call this method:
extractClientIP { ip =>
 if (isBelowRateLimit(ip)) {

src/test/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServerSpec.scala
```diff
@@ -215,7 +214,7 @@ class JsonRpcHttpServerSpec extends AnyFlatSpec with Matchers with ScalatestRout
       status shouldEqual StatusCodes.TooManyRequests
     }
 
-    fakeClock.advanceTime(2 * serverConfigWithRateLimit.rateLimit.minRequestInterval.toMillis)
+    mockJsonRpcHttpServerWithRateLimit.mockedTime = 50000000L
```
biandratti, line 217: I think that you can use this example to pass a FakeTicker to advance the time at will.
https://github.com/google/guava/blob/master/guava-tests/test/com/google/common/cache/CacheRefreshTest.java

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/RateLimit.scala
```diff
@@ -1,25 +1,67 @@
...
+          // But in this case we access our LRU twice.
+          lru.get(
+            ip,
+            () => {
+              exists = false
+              NotUsed
+            }
+          )
+          if (exists) {
+            val err = JsonRpcError.RateLimitError(minInterval)
```
lemastero: Can we separate LRU cache from RateLimit'ing?

I believe there are some rate-limiting capabilities built-in cats or Akka maybe we could use sth that is already build in?

dmitry-worker: I'm not sure I can get the idea of separating. 
Do you mean an ability to plug an opt-in rate limiting engine?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/RateLimit.scala
dmitry-worker: It's done

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/RateLimit.scala
dmitry-worker: Fixed that.  I think it's an overkill for a simple blocking mechanism - to have several classes for strategies.
Let's leave it for further improvement.

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
biandratti: Maybe we can create another task about it. 
What do you think?

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
```diff
@@ -160,12 +153,15 @@ object JsonRpcHttpServer extends Logger {
     }
 
   trait RateLimitConfig {
+    // TODO: Move the rateLimit.enabled setting upwards:
```
biandratti, line 156: ditto, maybe we can create another task about it.

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/RateLimit.scala
```diff
@@ -1,25 +1,71 @@
...
+  // It is also guaranteed that:
+  //   1) no IP address is extracted unless config.enabled is true
+  //   2) no LRU is created unless config.enabled is true
+  //   3) cache is accessed only once (using get)
+  override def tapply(f: Unit => Route): Route = {
+    if (config.enabled) {
+      extractClientIP { ip =>
+        if (isBelowRateLimit(ip)) {
+          val err = JsonRpcError.RateLimitError(minInterval)
+          complete((StatusCodes.TooManyRequests, err))
```
biandratti, line 63: Just a minor comment, we can reduce this line:
complete((StatusCodes.TooManyRequests, JsonRpcError.RateLimitError(minInterval)))

biandratti (approved):

biandratti:
Just a minor comment, Maybe you can create the task mentioned. So then we can continue working with improvements. 
Good job!

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
dmitry-worker: https://jira.iohk.io/browse/ETCM-509

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServer.scala
dmitry-worker: https://jira.iohk.io/browse/ETCM-510

src/main/scala/io/iohk/ethereum/jsonrpc/server/http/RateLimit.scala
dmitry-worker: Actually it was inline in the beginning, but I've decided that the code is quite unreadable :)
IMHO this will not affect performance and the compiler will inline it anyway because it is used once, immediately after creation. Maybe it would be better to construct a kind of builder for responses instead of serializable `Tuple`s. WDYT?

PR #876, opened 2020-12-30
[FIX] Not starting miner

pslaski:
# Description

Currently, askFor method is designed for sending a message to the actor and waiting for the response. Also, Task was used.
SendMiner prepared to communicate with Miner is using askFor, but not every usage should wait for the response and not every usage was executing task.

- divided sendMiner into sendMiner (used only for fire and forget message to the miner) and askMiner (previous sendMiner)

# Testing

run and check if miner is starting

### DIFF
--- a/src/main/scala/io/iohk/ethereum/consensus/ethash/EthashConsensus.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/ethash/EthashConsensus.scala
@@ -52,7 +52,13 @@ class EthashConsensus private (
 
   private implicit val timeout: Timeout = 5.seconds
 
-  override def sendMiner(msg: MinerProtocol): Task[MinerResponse] = {
+  override def sendMiner(msg: MinerProtocol): Unit = {
+    atomicMiner
+      .get()
+      .foreach(_ ! msg)
+  }
+
+  override def askMiner(msg: MinerProtocol): Task[MinerResponse] = {
     atomicMiner
       .get()
       .map(_.askFor[MinerResponse](msg))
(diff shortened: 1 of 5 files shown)


enriquerodbe (approved):

lemastero (approved):

biandratti (approved):

src/main/scala/io/iohk/ethereum/consensus/ethash/EthashConsensus.scala
```diff
@@ -52,7 +52,13 @@ class EthashConsensus private (
...
-  override def sendMiner(msg: MinerProtocol): Task[MinerResponse] = {
+  override def sendMiner(msg: MinerProtocol): Unit = {
+    atomicMiner
+      .get()
+      .foreach(_ ! msg)
+  }
+
+  override def askMiner(msg: MinerProtocol): Task[MinerResponse] = {
     atomicMiner
       .get()
```
dmitry-worker, line 63: I would fix the situation with `get()` and then...

src/main/scala/io/iohk/ethereum/consensus/ethash/EthashConsensus.scala
```diff
@@ -67,8 +73,6 @@ class EthashConsensus private (
           case MockedPow => MockedMiner(node)
         }
         atomicMiner.set(Some(miner))
```
dmitry-worker, line 75: ... and then set a miner. Because it is possible to set a miner in between of those.

pslaski: @dmitry-worker could you elaborate more?

lemastero: There is already PR opened that intend to fix that: https://github.com/input-output-hk/mantis/pull/777
There is no conclusion of the solution proposed there is the right one.

The change proposed here is an important fix. IMHO it is better to merge it and deal with potential concurrency issue in PR ☝️ .

PR #877, opened 2020-12-30
[ETCM-503] testing in CI takes a lot of time

dmitry-worker:
# Description

FastSyncSpec is failing one test repeatedly for some reason at CI... couldn't reproduce it locally..
Maybe it has something to do with a number of builds running simultaneously. They probably share the same hardware for testing.

# Proposed Solution

Increase the timeout for a test.


src/test/scala/io/iohk/ethereum/network/PeerStatisticsSpec.scala
```diff
@@ -48,7 +48,9 @@ class PeerStatisticsSpec
 
     val statA = stats.stats(alice)
     statA.responsesReceived shouldBe 2
-    statA.lastSeenTimeMillis shouldBe >(statA.firstSeenTimeMillis)
+    // @dv: the corner case when responses are received simultaneously
+    // can ruin the tests for no reason
+    statA.lastSeenTimeMillis shouldBe >=(statA.firstSeenTimeMillis)
```
aakoshh: Sorry for that. There may be better ways to test this via mocked clocks but I wanted to make sure that the first and last are treated separately. If they were always equal that would be an error.

src/test/scala/io/iohk/ethereum/blockchain/sync/FastSyncSpec.scala
```diff
@@ -22,7 +22,7 @@ class FastSyncSpec
     with FreeSpecBase
     with SpecFixtures
     with WithActorSystemShutDown { self =>
-  implicit val timeout: Timeout = Timeout(10.seconds)
+  implicit val timeout: Timeout = Timeout(30.seconds)
```
enriquerodbe, line 25: Does this mean that we have some tests that take more than 10 seconds to complete? We should improve those. If that's out of the scope of this PR maybe we can create a task to improve those tests.

enriquerodbe (approved):

src/test/scala/io/iohk/ethereum/blockchain/sync/FastSyncSpec.scala
dmitry-worker: Locally it runs for 2.6 seconds approximately. Maybe if I'll catch up with the logic of Fast/Regular sync it would be much easier to refactor these tests.

src/test/scala/io/iohk/ethereum/network/PeerStatisticsSpec.scala
dmitry-worker: I have provided a clock instance, please take a look into it, the pull requests has grown bigger though :(

src/main/scala/io/iohk/ethereum/network/TimeSlotStats.scala
```diff
@@ -13,8 +13,10 @@ class TimeSlotStats[K, V: Monoid] private (
     // The last written position in the buffer.
     val lastIdx: Int,
     // Ring buffer of slots statistics.
-    val buffer: TimeSlotStats.Buffer[K, V]
-)(implicit clock: Clock) {
+    val buffer: TimeSlotStats.Buffer[K, V],
+    // The time counter counter
+    clock: Clock
+) {
```
aakoshh: What was the problem with the way it was? I used an implicit just to make it a bit easier to return new instances of this otherwise immutable class.

src/test/scala/io/iohk/ethereum/utils/MockClock.scala
```diff
@@ -0,0 +1,17 @@
...
+                 zoneId: ZoneId = ZoneId.of("UTC")
+               ) extends Clock {
+  def windByMillis(by: Long): Unit =
+    currentTimeMillis = currentTimeMillis + by
+
+  override def instant(): Instant = Instant.ofEpochMilli(currentTimeMillis)
+  // The following are implemented for completness' sake but not used:
+  override def getZone(): ZoneId = zoneId
+  override def withZone(x: ZoneId): Clock = new MockClock(currentTimeMillis, zoneId)
+}
```
aakoshh: You may want to configure your editor to add a newline at the end of the file.

aakoshh (approved):
Thanks for the update, looks good!

As we discussed privately another option would have been to use `Instant` in the `PeerStat` type for `firstSeenTime` and `lastSeenTime` which has nanosecond precision, but this approach ultimately gives you full control.

src/main/scala/io/iohk/ethereum/network/TimeSlotStats.scala
aakoshh: It also seemed to make some sense that different parts of the app shouldn't have different clocks, so passing around an implicit one to whoever needs it might be better than threading it through. But ultimately it was only passed on in once place in this class so I get it if you don't like it.

dmitry-worker: Made clock implicit again

src/test/scala/io/iohk/ethereum/utils/MockClock.scala
dmitry-worker: Added newline

PR #871, opened 2020-12-22
[ETCM-491] Extend generated bash script with absolute path configs

enriquerodbe:
# Description

See https://github.com/input-output-hk/mantis/issues/851

# Proposed Solution

As @aakoshh commented in the issue, one solution is to extend the the generated bash script file so that we can set the config files paths correctly.

# Testing

These instructions should work for Unix and for Windows (just by changing `/` to `\`).

1. Build: `sbt dist`
2. Move to `target/universal`: `cd target/universal`
3. Unzip: `unzip mantis-3.2.1.zip`
4. Verify the following commands

### Should use etc by default (this is the test for the original issue)

- [x] `mantis-3.2.1/bin/mantis` - Check that it starts the client and connects to etc network. 

### Should use given chain

- [x] `mantis-3.2.1/bin/mantis-launcher sagano` - Check that it starts and connects to sagano testnet.

### Should not break other scripts
All of the following should be valid and work as usual
- [x] `mantis-3.2.1/bin/mantis-vm 8080`
- [x] `mantis-3.2.1/bin/eckeygen 8`
- [x] `mantis-3.2.1/bin/signatureValidator f0b91b55bdea6a6da2bcbd1270451e9ca014747b97e190a5a4275bad8eeda7436ab8081f0df5741a86a1bfb62798221103fff5898c4f768b707ff34b110329ed 6bf862422f0103d8ca6847019c0113c9b343a2489c2dc3fc67d853e87d98f68e5a5be1d021ec706197960912037d46bde44c3534a4ff562fd48e664f32c1392d1c 5fe7f977e71dba2ea1a68e21057beebb9be2ac30c6410aa38d4f3fbe41dcffd2`
- [x] `mantis-3.2.1/bin/faucet-server`

PR #880, opened 2021-01-04
[ETCM-491] Use etc.conf as default config file

enriquerodbe:
# Description

**Note:** This is a continuation of https://github.com/input-output-hk/mantis/pull/871. Here I'm addressing the last comment made by @aakoshh in that PR.

If the launcher is executed without arguments, then the default `conf/app.conf` file is used which uses etc network. This can lead to confusions because a user would expect to make changes to `conf/etc.conf` and run them using the launcher without arguments.

# Proposed Solution

If no argument is received, the launcher will select the `conf/etc.conf` file as the default configuration. This way the user will see the impacts of changing `conf/etc.conf` when running the launcher without arguments.

# Testing

All validations from https://github.com/input-output-hk/mantis/pull/871 should still work. And:

1. Change `conf/etc.conf` by setting the network to `testnet-internal-nomad`
2. Start the client with `bin/mantis-launcher`
3. Verify that the client starts and connects to the `testnet-internal-nomad` network

### DIFF
--- a/src/universal/bin/mantis-launcher
+++ b/src/universal/bin/mantis-launcher
@@ -7,6 +7,8 @@ CONFIG_FILE="./conf/$1.conf"
 if [ -f "$CONFIG_FILE" ]; then
   shift
   CHAIN_PARAM="-Dconfig.file=$CONFIG_FILE"
+elif [ -z "$1" ]; then
+  CHAIN_PARAM="-Dconfig.file=./conf/etc.conf"
 fi
 
 ./bin/mantis ${CHAIN_PARAM:+"$CHAIN_PARAM"} "$@"
--- a/src/universal/bin/mantis-launcher.bat
+++ b/src/universal/bin/mantis-launcher.bat
@@ -5,15 +5,20 @@ cd "%~dp0\.."
 set "CONFIG_FILE=conf\%1.conf"
 set "RESTVAR=%*"
 
-if not exist %CONFIG_FILE% goto :skip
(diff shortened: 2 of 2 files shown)


aakoshh (approved):
Thanks for taking action on my comment, defaulting to the config on empty seems to make sense, otherwise the user is in control and can do whatever they are trying without interference :+1: 

I took me a while to understand that this behaviour only applies to the `mantis-launcher` and not `mantis` itself. It should be possible to inject a whole script file into the `mantis` script so `./bin/mantis etc` would work the same way, but that's a different issue.

enriquerodbe:
> Thanks for taking action on my comment, defaulting to the config on empty seems to make sense, otherwise the user is in control and can do whatever they are trying without interference 👍
> 
> I took me a while to understand that this behaviour only applies to the `mantis-launcher` and not `mantis` itself. It should be possible to inject a whole script file into the `mantis` script so `./bin/mantis etc` would work the same way, but that's a different issue.

Thank you for the quick response 👍 

Regarding the launcher: I made that decision mostly because `mantis` launcher itself works perfectly fine and a user can configure anything by passing the correct parameters. I see the launcher and other bin scripts more as utilities, saving the users from typing `-Dconfig.file=conf/$network.conf` and letting them just use `mantis-launcher $network`.

That said, I also found a technical issue trying to extend the `mantis` script. The issue is that on Windows, the generated `mantis.bat` uses the arguments array `%*` directly, so shifting doesn't work, and I couldn't find any other solution to "bypass" a parameter like the `shift` in bash. So I also preferred consistency across platforms instead of a feature that only works on Unix.

PR #881, opened 2021-01-07
Fix sbtix build using nix-shell and scala2.13

jonringer:
# Description

Updated build.sbt to build with NixOS

updated sbtix in an attempt to fix the pure nix build, but was unsuccessful


jonringer:
no longer relevant with sbt-derivation

PR #882, opened 2021-01-09
Fix scala==2.13 nix build, cleanup

jonringer:
# Description

Fix build with scala 2.13

cc @input-output-hk/mantis-devs


dmitry-worker (approved):
Thanks @jonringer! We now have
 - no more repo.nix
 - no sbtix-related stuff in code

That's awesome! :+1: 

But there should be a fix for buildkite, cause it tries to follow the old build scenario.
Would you like me to update it or maybe we can manage it together using tmate? 
It would be great if one can test buildkite without commiting anything to the repo...

Regarding this PR - I believe I'm not an expert in Nix packaging, so we may require @manveru 's review.

nix/pkgs/mantis.nix
```diff
@@ -0,0 +1,89 @@
...
+  nativeBuildInputs = [ protoc-wrapper jdk makeWrapper ];
+
+  # read version from build.sbt
+  version = let
+    buildSbt = builtins.readFile ../../build.sbt;
+    captures = builtins.match ''.*version := "([^"]+)".*'' buildSbt;
+  in builtins.elemAt captures 0;
+
+  LD_LIBRARY_PATH = lib.makeLibraryPath [ stdenv.cc.cc.lib ];
+  PATH = lib.makeBinPath [ jre gawk ];
```
manveru: Please add coreutils to `PATH`. It's required by the launcher scripts

manveru (changes requested):
looks good otherwise, just needs one change.

nix/pkgs/mantis.nix
jonringer: it's why I'm appending to the previous path, but I think your way is probably more correct, thanks :)

jonringer: fixed

PR #884, opened 2021-01-11
ETCM-313 Improve blockchain download

enriquerodbe:
# Description

Refactor `FastSync` actor into smaller classes with less responsibilities each.

# Proposed Solution

- Extract a new actor `SyncingHandler` which does the actual downloading of blocks.
- Move all logic related to the pivot block into its own class.
- Extract persistence logic into its own `SyncingHandlerStorage` class.
- Extract validation logic into a `FastSyncValidator` class.

# Testing

Existing tests should all pass. New unit tests would be nice to add.


enriquerodbe:
Rebased this PR to develop in another branch. See https://github.com/input-output-hk/mantis/pull/888

PR #886, opened 2021-01-11
Caches implemented as ConcurrentMap

dmitry-worker:
# Description

Second life of my PR as a contributor before IOHK
```
val values = cache.values()
<do_something_with_values>
cache.clear()
```
Will lose everything put in cache between `values` and `clear`

# Proposed Solution

`drain()` method introduced in order to resolve concurrency issue
`ConcurrentMap` offers a fine-grained locking mode for effectively using Cache.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/db/cache/LruCache.scala
+++ b/src/main/scala/io/iohk/ethereum/db/cache/LruCache.scala
@@ -6,7 +6,7 @@ import io.iohk.ethereum.utils.Config.NodeCacheConfig
 import com.google.common.cache
 import com.google.common.cache.{CacheBuilder, RemovalNotification}
 
-import scala.collection.JavaConverters.asScalaIterator
+import scala.jdk.CollectionConverters._
 import scala.concurrent.duration.FiniteDuration
 
 class LruCache[K <: AnyRef, V <: AnyRef](
@@ -27,11 +27,11 @@ class LruCache[K <: AnyRef, V <: AnyRef](
       )
       .build()
 
-  override def drain(): Seq[(K, V)] = {
+  override def drain: Seq[(K, V)] = {
     this.lastClear = System.nanoTime()
     val mapView = lruCache.asMap
     val iter = mapView.entrySet.iterator
-    val result = asScalaIterator(iter).map { e =>
+    val result = iter.asScala.map { e =>
       val key = e.getKey
       val value = mapView.remove(key)
       key -> value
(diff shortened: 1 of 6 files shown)


src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
```diff
@@ -256,14 +256,24 @@ class BlockFetcherSpec
       val checkpointBlock = (new CheckpointBlockGenerator)
         .generate(
           firstBlocksBatch.last,
-          Checkpoint(CheckpointingTestHelpers.createCheckpointSignatures(Seq(crypto.generateKeyPair(secureRandom)), firstBlocksBatch.last.hash))
+          Checkpoint(
```
lemastero, line 259: If there are not changes in this file it would be good to revert changes, for better review. The PR is pretty complex.

src/main/scala/io/iohk/ethereum/db/cache/LruCache.scala
```diff
@@ -1,18 +1,20 @@
...
+import scala.jdk.CollectionConverters._
 import scala.concurrent.duration.FiniteDuration
 
 class LruCache[K <: AnyRef, V <: AnyRef](
     config: NodeCacheConfig,
     notificationHandler: Option[RemovalNotification[K, V] => Unit] = None
 ) extends Cache[K, V] {
 
-  private val lastClear = new AtomicLong(System.nanoTime())
+  @volatile private[this] var lastClear = System.nanoTime()
```
lemastero, line 17: Changing from:
`val lastClear = new AtomicLong`
to:
`@volatile private[this] var lastClear`
does not seems like good direction. Could you elaborate reason for this change?

lemastero (changes requested):

src/main/scala/io/iohk/ethereum/db/cache/LruCache.scala
dmitry-worker: `Volatile` guarantees that each individual read and write IS atomic per its definition.
And there's no `___And____` operations anywhere in code (`getAndSet`, `compareAndSet`, `getAndUpdate`, etc)

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
dmitry-worker: It's a change introduced by scalaFmt...

dzajkowski:
closing in favour of a split approach

PR #887, opened 2021-01-11
[ETCM-316] Fast-sync branch resolver

biandratti:
# Description

This task has as dependence the task [[ETCM-313]](https://github.com/input-output-hk/mantis/pull/892). When a invalid block is detected from the master peer. We need to search for a new peer potencial and validate the our chain saved with this new peer.
In this way, with this new peer we need to follow the next points.

1. First ask this peer for the last several blocks, starting from current mantis best + 1, in reverse. i.e let's say our current best block is 100, then we ask potential peer for blocks with numbers from `101` to `101 - X`, where x is some hyperparameter. Then we check the block one by one if they make a chain connection with our local chain. i.e first check if remote 101 is child of local 100, if not then check if remote 100 is child of local 99. If we find a match, rewind to a matched block and that starts syncing from there, and if not go to point 2.
2. If we do find not matched blocks in most fresh local blocks, it means that chains diverged earlier.  To find the first common block between local and remote blockchain we can use binary search as both blockchain are sorted lists. First we ask remote peer for a block in the middle of our local blockchain, if block from response is the same as local, then it means chains diverged later otherwise they diverged earlier. 

# Proposed Solution

We have an actor called FastSyncBranchResolver responsible for this behavior. When FastSync actor need to resolve it branch; FastSyncBranchResolver is called with the message “StartBranchResolver”. In this way, we have the following strategy:

1. Get a new master peer.
2. If we don't detect a master peer, we wait a second and retry again.
3. Get a set of the last BlockHeaders from this master peer.
4. Validate the set of blocks received from master peer andwe validate it with our chain. If deptect the last valid block, we discard the last invalid blocks in our chain and send a message “BranchResolvedSuccessful” to the fast sync to continue with the syncing. In case if we cant detect the last valid block it diverged earlier (5).
5. We continue with a binary search validation, block per block with the peer master until we detect the last valid block in our chain. When detecting this block, we discard the last invalid blocks in our chain and send a message “BranchResolvedSuccessful to the fast sync to continue with the syncing.

# Pending work

- ~We need to remove binary search logic from the actor, and put it into another normal class. Making it possible to test it completely in all the cases.~ :heavy_check_mark: 
- ~[task: ETCM-531] We need to change the behavior of traits BlacklistSupport and PeerListSupport. Both should be actors. Today when we switched between FastSync and FastSyncBranchResolver, we lost the list of peers in memory from blackList and handshake. So we need to uncouple this behavior and don't lose the changes. This new behavior should also work in the future to regular sync.~ :heavy_check_mark: 

# Testing

It needs a battery of integration test to cover these case.


src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSyncBranchResolver.scala
```diff
@@ -0,0 +1,163 @@
...
+      scheduler: Scheduler
+  ): Props =
+    Props(new FastSyncBranchResolver(peerEventBus, etcPeerManager, blockchain, syncConfig, appStateStorage, scheduler))
+
+  private val BlockHeadersHandlerName = "block-headers-request-handler"
+
+  case class SearchState(minBlockNumber: BigInt, maxBlockNumber: BigInt)
+
+  sealed trait BranchResolverRequest
+  case class StartBranchResolver() extends BranchResolverRequest
```
ten15bit: minor, but this should probably be a `case object`

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSyncBranchResolver.scala
```diff
@@ -0,0 +1,163 @@
...
+
+  sealed trait BranchResolverResponse
+  case class BranchResolvedSuccessful() extends BranchResolverResponse
```
ten15bit: should also be a `case object`

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSyncBranchResolver.scala
```diff
@@ -0,0 +1,182 @@
...
+        case Some(lastValidBlockHeader) =>
+          finalizeBranchResolver(bestBlockNumber, lastValidBlockHeader)
+        case None => requestBlockHeader(peer, SearchState(0, bestBlockNumber))
+      }
+    }
+    case RequestFailed(peer, reason) => handleRequestFailure(peer, sender(), reason)
+  }
+
+  private def binarySearchingLastValidBlock(searchState: SearchState): Receive = handleCommonMessages orElse {
+    case ResponseReceived(peer, BlockHeaders(blockHeaders), timeTaken) =>
```
KonradStaniec: during binary seach we only ask for one header from potential master peer, so we should validate if:
 - response has exactly one header
 - response is from master peer

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSyncBranchResolver.scala
```diff
@@ -0,0 +1,182 @@
...
+    val peerEventBus: ActorRef,
+    val etcPeerManager: ActorRef,
+    val blockchain: Blockchain,
+    val syncConfig: SyncConfig,
+    val appStateStorage: AppStateStorage,
+    implicit val scheduler: Scheduler
+) extends Actor
+    with ActorLogging
+    with PeerListSupport
+    with BlacklistSupport {
```
KonradStaniec: this will be a little problematic. 

As you can see with current encoding `BlacklistSupport` is per actor basis, this can lead to situation:

- In main fast sync actor, our master peer which is best peer failed (let say, sent us bad header), and we blacklisted him. We start `FastSyncBranchResolver` to chose new best peer and resolve branch
-  In `FastSyncBranchResolver` we get best peer and we get the same peer, as here he is not blacklisted
- we return to main fast sync actor and start sync with the same master.

Probably our blacklist, should be global for whole syncing component. to avoid it.

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSyncBranchResolver.scala
```diff
@@ -0,0 +1,182 @@
...
+        finalizeBranchResolver(blockchain.getBestBlockNumber(), childBlockHeader.number)
+      else requestBlockHeader(peer, searchState.copy(minBlockNumber = childBlockHeader.number + 1))
+    } else requestBlockHeader(peer, searchState.copy(maxBlockNumber = childBlockHeader.number))
+  }
+
+  private def finalizeBranchResolver(lastBlock: BigInt, lastValidBlock: BigInt): Unit = {
+    discardLastInvalidBlocks(lastBlock, lastValidBlock)
+    context become receive
+    log.info("branch resolution completed")
+    fastSync ! BranchResolvedSuccessful
```
KonradStaniec: we should probably return 
- new master peer
- first common block with master peer

KonradStaniec (commented):
In general idea looks good.

One thing i would change is to remove as much logic from actor, and move it to normal class to make it possible to test whole binary search logic without akka (and timeouts). So actor would be only to communicate with other actors

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSyncBranchResolver.scala
```diff
@@ -0,0 +1,193 @@
...
+
+  private def finalizeBranchResolver(lastBlock: BigInt, firstCommonBlockNumber: BigInt, masterPeer: Peer): Unit = {
+    discardLastInvalidBlocks(lastBlock, firstCommonBlockNumber)
+    context become receive
+    log.info(s"branch resolution completed - firstCommonBlockNumber $firstCommonBlockNumber")
+    fastSync ! BranchResolvedSuccessful(firstCommonBlockNumber = firstCommonBlockNumber, masterPeer = masterPeer)
+  }
+
+  private def getBestPeer: Option[(Peer, PeerInfo)] = {
+    handshakedPeers.toList.sortBy { case (_, peerInfo) => peerInfo.maxBlockNumber }(Ordering[BigInt].reverse).headOption
```
robinraju: I think we should perform an additional check here if this `bestPeer` is already blacklisted or not
I see this already addressed in https://jira.iohk.io/browse/ETCM-531

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSyncBranchResolver.scala
```diff
@@ -0,0 +1,193 @@
...
+      finalizeBranchResolver(blockchain.getBestBlockNumber(), parentBlockHeader.number - 1, searchState.masterPeer)
+    else
+      requestBlockHeader(searchState.copy(maxBlockNumber = parentBlockHeader.number))
+  }
+
+  private def finalizeBranchResolver(lastBlock: BigInt, firstCommonBlockNumber: BigInt, masterPeer: Peer): Unit = {
+    discardLastInvalidBlocks(lastBlock, firstCommonBlockNumber)
+    context become receive
+    log.info(s"branch resolution completed - firstCommonBlockNumber $firstCommonBlockNumber")
+    fastSync ! BranchResolvedSuccessful(firstCommonBlockNumber = firstCommonBlockNumber, masterPeer = masterPeer)
```
robinraju: @biandratti  Would there be any failures being reported back to `FastSync` actor? And does the branch resolving involves any timeouts?

ten15bit: Failures are now reported back via `BranchResolutionFailed` and there are request timeouts handled by `PeerRequestHandler`.

src/main/scala/io/iohk/ethereum/blockchain/sync/PeerListSupportNg.scala
```diff
@@ -44,6 +44,11 @@ trait PeerListSupportNg { self: Actor with ActorLogging =>
 
   def getPeerById(peerId: PeerId): Option[Peer] = handshakedPeers.get(peerId).map(_.peer)
 
+  def getPeerWithHighestBlock: Option[PeerWithInfo] =
+    peersToDownloadFrom.values.toList.sortBy { case PeerWithInfo(_, peerInfo) =>
+      peerInfo.maxBlockNumber
+    }(Ordering[BigInt].reverse).headOption
```
dzajkowski: nitpick: `Ordering[BigInt].reverse` could be an explicit constant value to save on `reverse` calls

src/test/scala/io/iohk/ethereum/blockchain/sync/fast/FastSyncBranchResolverSpec.scala
```diff
@@ -0,0 +1,309 @@
+package io.iohk.ethereum.blockchain.sync.fast
+
+import org.scalatest.wordspec.AnyWordSpec
+import org.scalatest.matchers.must.Matchers
+import org.scalamock.scalatest.MockFactory
+import io.iohk.ethereum.domain.Blockchain
+import io.iohk.ethereum.domain.BlockHeader
+import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
+import org.scalacheck.Gen
```
leo-bogastry:
```
import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
import org.scalacheck.Gen
```
are not used an can be removed

src/test/scala/io/iohk/ethereum/blockchain/sync/fast/FastSyncBranchResolverSpec.scala
```diff
@@ -0,0 +1,309 @@
...
+  }
+
+  private def headersMap(amount: Int, parent: Block): Map[BigInt, BlockHeader] = {
+    BlockHelpers.generateChain(amount, parent).map(b => (b.number, b.header)).toMap
+  }
+
+  private def headersList(blocksMap: Map[BigInt, Block]): List[BlockHeader] =
+    blocksMap.values.map(_.header).toList
+
+  private def headerNum(block: Block): BigInt = block.header.number
```
leo-bogastry: `headerNum` also not used

src/test/scala/io/iohk/ethereum/blockchain/sync/fast/FastSyncBranchResolverSpec.scala
```diff
@@ -0,0 +1,309 @@
...
+        (mockedBlockchain.getBlockHeaderByNumber _).expects(BigInt(99)).returning(headers.get(99))
+        (mockedBlockchain.removeBlock _).expects(headers(99).hash, false).returning(())
+        (mockedBlockchain.getBlockHeaderByNumber _).expects(BigInt(98)).returning(headers.get(98))
+        (mockedBlockchain.removeBlock _).expects(headers(98).hash, false).returning(())
+      }
+
+      val resolver = new FastSyncBranchResolver {
+        override val blockchain: Blockchain = mockedBlockchain
+      }
+      resolver.discardBlocksAfter(97)
```
leo-bogastry, line 59: This test seems to miss a final assertion

ten15bit: No, it does not...it's testing the expected calls to `Blockchain`. Whether `Blockchain` removes blocks correctly is not the responsibility of the branch resolver.

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSyncBranchResolverActor.scala
```diff
@@ -0,0 +1,273 @@
...
+
+  private val recentBlocksSearch: RecentBlocksSearch = new RecentBlocksSearch(blockchain)
+
+  override def receive: Receive = waitingForPeerWithHighestBlock
+
+  private def waitingForPeerWithHighestBlock: Receive = handlePeerListMessages orElse { case StartBranchResolver =>
+    getPeerWithHighestBlock match {
+      case Some(PeerWithInfo(peer, _)) => requestRecentBlockHeaders(peer, blockchain.getBestBlockNumber())
+      case None =>
+        log.info(s"Waiting for peers, rescheduling StartBranchResolver")
```
leo-bogastry: Could you please clean the string interpolations that are not necessary? I have seem so many in the code base in general

src/test/scala/io/iohk/ethereum/blockchain/sync/fast/FastSyncBranchResolverSpec.scala
leo-bogastry: 👍

leo-bogastry (approved):

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSyncBranchResolverActor.scala
ten15bit: removed this one...there's already a ticket to add a linter to the codebase to catch unused things

PR #889, opened 2021-01-12
[FIX] Block fetcher failing test fix

pslaski:
# Description

One of the block fetcher's test (`should process checkpoint blocks when checkpoint can fit into waiting headers queue`) was failing nondeterministically.

 - added setting `AwaitingBodiesToBeIgnored` as a fetching bodies status in case of changing waiting headers queue during bodies fetching (to avoid blacklisting peers which respond to pending block bodies requests)
 - fixed test structure

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
@@ -100,8 +100,12 @@ case class BlockFetcherState(
         .withKnownTopAt(block.number)
       Right(newState)
     } else if (isExistInWaitingHeaders(block.header.parentHash)) {
+      // ignore already requested bodies
+      val newFetchingBodiesState =
+        if (fetchingBodiesState == AwaitingBodies) AwaitingBodiesToBeIgnored else fetchingBodiesState
       val newState = copy(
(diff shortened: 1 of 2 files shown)


dmitry-worker:
The tests are no longer failing :+1: 
It looks like I don't have enough knowledge to get what's going on and why it is working now. 
@kapke, @ntallar if you have some time to take a look - please help us out.

lemastero (approved):
LGTM

kapke (approved):
LGTM!

pslaski:
@dmitry-worker 
This test was prepared for the scenario:
1. ask for the first part of headers -> ask for the next part of headers (result: 20 headers in the waiting headers queue)
2. ask for bodies for the first part of headers  (result: 10 ready blocks and 10 waiting headers)
3. send checkpoint which can fit into waiting headers queue (result: 10 ready blocks and 6 waiting headers - checkpoint in the end)
4. ask for bodies for the second part of headers  (result: 16 ready blocks)
5. Pick blocks from fetcher and check if there are valid and if there is a checkpoint 

there were 2 problems:
1. Bad test structure: Between 2) and 3) fetcher is also asking for bodies for the second part of headers (10 headers) which was missing in the test
2. bodies received from the previous point couldn't pass bodies validation (validation against waiting headers) as checkpoint were changing waiting headers queue in the meantime - that's a reason for adding setting AwaitingBodiesToBeIgnored as a fetching bodies status in case of changing waiting headers queue during bodies fetching

PR #893, opened 2021-01-14
[ETCM-514] Single array allocation for concatenation of ByteStrings

dmitry-worker:
# Description

As per @lemastero suggestion:
This PR replaces an ArrayBuffer with a preallocated array in order to achieve a constant time writing.


bytes/src/main/scala/io/iohk/ethereum/utils/ByteStringUtils.scala
```diff
@@ -61,9 +61,14 @@ object ByteStringUtils {
...
-    elements.foreach(el => builder.addAll(el.asByteArray))
-    ByteString(builder.result())
+    val elementList: List[ByteStringElement] = elements.toList
+    val len = elementList.foldLeft(0)(_ + _.len)
+    val array = Array.ofDim[Byte](len)
+    elementList.foldLeft(0) { (index, el) =>
+      System.arraycopy(el.asByteArray, 0, array, index, el.len)
+      index + el.len
+    }
+    ByteString(array)
```
dzajkowski: naive question: wouldn't a `ArrayBuilder.sizeHint` and `ArrayBuilder.addAll` achieve the same result?

dmitry-worker: Sure, but if we have known array size - perhaps we won't need a builder.

dzajkowski: true. 
my point is that it would be less code to maintain for the cost of one class allocation. observe that the underlying code is basically the same as what you wrote and quite similar to what `Array.concat` does.

thanks for the response!

KonradStaniec: Just a proposition, if we care about performance that much, and we know that underlying array will not be modified  we can use - ` ByteString.fromArrayUnsafe` which will reduce allocation even more

bytes/src/main/scala/io/iohk/ethereum/utils/ByteStringUtils.scala
```diff
@@ -61,9 +61,14 @@ object ByteStringUtils {
   }
 
   def concatByteStrings(elements: Iterator[ByteStringElement]): ByteString = {
-    val builder = new mutable.ArrayBuilder.ofByte
-    elements.foreach(el => builder.addAll(el.asByteArray))
-    ByteString(builder.result())
+    val elementList: List[ByteStringElement] = elements.toList
```
lemastero: I am sorry I was not precise enough when we discuss this.

When we migrated from 2.12 to 2.13 there was a problem that ByteString do not support all collections API like concatenation (because of removal of CanBuildFrom). So it was possible to write some utility that will concatenate `Array[byte]` that checks size of arrays and prepare final one with proper size.

I am afraid that when argument is iterator we need to convert `toList` so we loose perf benefits 😢

dmitry-worker: I have removed `toList`, hopefully a `Seq` from VarArg is fast enough prepending elements.

bytes/src/main/scala/io/iohk/ethereum/utils/ByteStringUtils.scala
```diff
@@ -55,14 +56,15 @@ object ByteStringUtils {
...
+    val it = head +: tail
     concatByteStrings(it)
   }
 
-  def concatByteStrings(elements: Iterator[ByteStringElement]): ByteString = {
-    val builder = new mutable.ArrayBuilder.ofByte
-    elements.foreach(el => builder.addAll(el.asByteArray))
-    ByteString(builder.result())
+  def concatByteStrings(elements: Iterable[ByteStringElement]): ByteString = {
+    val len = elements.foldLeft(0)(_ + _.len)
```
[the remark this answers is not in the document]
dmitry-worker: It's probably the first time I see that `fold` is not clear enough (and also it would be preferrable to traverse twice)

bytes/src/test/scala/io/iohk/ethereum/utils/ByteStringUtilsTest.scala
```diff
@@ -56,5 +57,44 @@ class ByteStringUtilsTest extends AnyWordSpec with Matchers {
       bsu.padTo(6, 1) shouldEqual longSeq
     }
 
+    "operate on bytebuf as the fastest tool for concatenation" in {
+      val buffer = arrayBufferTime
+      val builder = arrayBuilderTime
+      val bytebuf = byteBufferTime
+      assert(bytebuf < buffer)
```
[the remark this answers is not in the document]
dmitry-worker: Is it possible to allocate an off-heap `ArrayBuilder` to avoid GC as well as ByteBuffer?

bytes/src/test/scala/io/iohk/ethereum/utils/ByteStringUtilsTest.scala
```diff
@@ -56,5 +57,44 @@ class ByteStringUtilsTest extends AnyWordSpec with Matchers {
...
+      var aBuilder = new ArrayBuilder.ofByte
+      aBuilder.sizeHint(CAPACITY)
+      var time = System.nanoTime()
+      for (i <- 0 until CAPACITY) {
+          aBuilder += (i & 0xFF).toByte
+      }
+      System.nanoTime() - time
+  }
+
+  def byteBufferTime: Long = {
```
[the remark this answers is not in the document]
dmitry-worker: If it is okay for internal array to be re-allocated multiple times (without providing a hint) - then yes, ArrayBuilder can avoid pre-computing length.

bytes/src/test/scala/io/iohk/ethereum/utils/ByteStringUtilsTest.scala
```diff
@@ -56,5 +57,44 @@ class ByteStringUtilsTest extends AnyWordSpec with Matchers {
...
   }
+
+  def arrayBuilderTime: Long = {
+      var aBuilder = new ArrayBuilder.ofByte
+      aBuilder.sizeHint(CAPACITY)
+      var time = System.nanoTime()
+      for (i <- 0 until CAPACITY) {
+          aBuilder += (i & 0xFF).toByte
+      }
+      System.nanoTime() - time
```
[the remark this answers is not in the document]
dmitry-worker: Given that a direct ByteBuffer is much more effective in a language with GC - I'd replace `allocate` with `allocateDirect` and remove this microbench completely. For the current case - it's true that the results shown here cannot be used for estimating collection write performance in absolute numbers. But I can hardly tell if the relative result will not hold for some other JVM / GC.

dzajkowski:
closing due to inactivity

PR #897, opened 2021-01-18
[ETCM-518] refactor EthService into smaller, coherent classes

leo-bogastry:
# Description

EthService is quite a large service and it's due to handling many different types of endpoints at once. The proposed refactor is to split EthService into smaller chunks.

# Proposed Solution

Refactored EthService into:
EthBlocksService
EthFilterService
EthMiningService
EthTxService
EthUserService
EthInfoService

## Calls handled by Service:
**EthBlocksService:**
eth_blockNumber
eth_getBlockTransactionCountByHash
eth_getBlockByHash
eth_getBlockByNumber
eth_getBlockTransactionCountByNumber
eth_getUncleByBlockHashAndIndex
eth_getUncleByBlockNumberAndIndex
eth_getUncleCountByBlockNumber
eth_getUncleCountByBlockHash

**EthFilterService:**
eth_newFilter
eth_newBlockFilter
eth_newPendingTransactionFilter
eth_uninstallFilter
eth_getFilterChanges
eth_getFilterLogs
eth_getLogs

**EthMiningService:**
eth_mining
eth_getWork
eth_submitWork
eth_coinbase
eth_submitHashrate
eth_hashrate

**EthTxService:**
eth_getRawTransactionByHash
eth_getRawTransactionByBlockHashAndIndex
eth_getRawTransactionByBlockNumberAndIndex
eth_getTransactionByHash
eth_getTransactionReceipt
eth_gasPrice
eth_sendRawTransaction
eth_getTransactionByHash
eth_getTransactionReceipt
eth_pendingTransactions

**EthUserService:**
eth_getCode 
eth_getStorageRoot
eth_getStorageAt

**EthInfoService:**
eth_protocolVersion
eth_chainId
eth_syncing
iele_call
eth_call
eth_estimateGas


src/main/scala/io/iohk/ethereum/jsonrpc/EthBlocksService.scala
```diff
@@ -0,0 +1,198 @@
...
+import akka.util.ByteString
+import io.iohk.ethereum.domain.Blockchain
+import io.iohk.ethereum.jsonrpc.EthService.{BlockParam, ResolvedBlock}
+import io.iohk.ethereum.ledger.Ledger
+import io.iohk.ethereum.utils.{BlockchainConfig, Logger}
+import monix.eval.Task
+import org.bouncycastle.util.encoders.Hex
+
+object EthBlocksService {
+  case class ChainIdRequest()
```
dmitry-worker: Can be replaced with a `case object`

dmitry-worker (approved):
Great job refactoring that monster! 

The only thing I'd like to upgrade - replace `extractThis` and `extractThat` with the proper Decoders OR extractors such as JQuantity, JAddress etc. So the boilerplate like 
```
        case Some(JArray(blockStr :: uncleIndex :: Nil)) =>
          for {
            block <- extractBlockParam(blockStr)
            uncleBlockIndex <- extractQuantity(uncleIndex)
          } yield UncleByBlockNumberAndIndexRequest(block, uncleBlockIndex)
```
can be represented as: 
```
        case Some(JArray(JBlockString(block) :: JQuantity(uncleBlockIndex) :: Nil)) =>
          Right( UncleByBlockNumberAndIndexRequest(block, uncleBlockIndex) )
```

Not sure if it fits the PR though... WDYT?

leo-bogastry:
@dmitry-worker About replacing the extractors, as far as I see these extractors have some logic built inside them. Taking the example of `extractBlockParam` the way the extraction is done depends on the value of the String being `earliest`, `latest` or `pending`

src/main/scala/io/iohk/ethereum/jsonrpc/EthBlocksService.scala
leo-bogastry: Yes, it can. But such a change demands adding more boilerplate code, where `ChainIdRequest` is being referred as the type of an object we have to explicitly replace `ChainIdRequest` by `ChainIdRequest.type`.

dmitry-worker:
> @dmitry-worker About replacing the extractors, as far as I see these extractors have some logic built inside them. Taking the example of `extractBlockParam` the way the extraction is done depends on the value of the String being `earliest`, `latest` or `pending`

Sure, I've just tested an extractor for BlockParam: 
```
  sealed trait BlockParam

  object BlockParam {
    case class WithNumber(n: BigInt) extends BlockParam
    case object Latest extends BlockParam
    case object Pending extends BlockParam
    case object Earliest extends BlockParam


    def unapply(jv: JValue): Option[BlockParam] = jv match { 
      case JString("latest") => Some(Latest)
      case JString("earliest") => Some(Earliest)
      case JString("pending") => Some(Pending)
      case JLong(value) => Some(WithNumber(BigInt(value)))
      case _ => None
    }
  }
```

( Maybe it's not a perfect place to define an extractor though )

jvdp:
I think it would be nice to simplify the JSON encoding / decoding but it deserves a follow-up ticket IMHO.

Interesting proposal to use extractors for this; though it does lock the classes into one particular JSON-encoding. Is that a concern at all?

src/main/scala/io/iohk/ethereum/jsonrpc/EthBlocksService.scala
```diff
@@ -0,0 +1,198 @@
...
+  case class GetUncleCountByBlockHashResponse(result: BigInt)
+}
+
+class EthBlocksService(blockchain: Blockchain, ledger: Ledger, blockchainConfig: BlockchainConfig) {
+  import EthBlocksService._
+
+  private[jsonrpc] def consensus = ledger.consensus
+  private[jsonrpc] def blockGenerator = consensus.blockGenerator
+
+  def chainId(req: ChainIdRequest): ServiceResponse[ChainIdResponse] =
```
ntallar: Hmm I'm not sure if I get why this should be here.. isn't this closer to `eth_protocolVersion` than to blocks? Maybe it's better then for this to be on EthGeneralService?

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -146,446 +73,51 @@ object EthService {
...
+      case BlockParam.Pending =>
+        ledger.consensus.blockGenerator.getPendingBlockAndState
+          .map(pb => ResolvedBlock(pb.pendingBlock.block, pendingState = Some(pb.worldState)))
+          .map(Right.apply)
+          .getOrElse(resolveBlock(blockchain, ledger, BlockParam.Latest)) //Default behavior in other clients
+    }
+  }
 }
 
 class EthService(
```
ntallar: I find it a bit odd this being named EthService given that it doesn't implement all eth_ endpoints, maybe this should be renamed to EthGeneralService or something like that?

(I'd then move the resolveBlock function to another location)

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
```diff
@@ -146,446 +73,51 @@ object EthService {
...
-    jsonRpcConfig: JsonRpcConfig,
-    getTransactionFromPoolTimeout: FiniteDuration,
     askTimeout: Timeout
-) extends Logger {
+) {
 
   import EthService._
 
   val hashRate: ConcurrentMap[ByteString, (BigInt, Date)] = new TrieMap[ByteString, (BigInt, Date)]()
   val lastActive = new AtomicReference[Option[Date]](None)
```
ntallar: This are no longer needed here, right?

ntallar (commented):
Apart from the two above comment the design LGTM!

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
leo-bogastry: No, they can be removed

src/main/scala/io/iohk/ethereum/jsonrpc/EthService.scala
jvdp: How about we move `call` and `ieleCall` to `EthTxService` (they are related to transactions right?) so that only info-related methods are left and we call this `EthInfoService` or `EthStatusService`?

jvdp: Considering that `call` and `ieleCall` don't actually alter anything (please correct me if I'm wrong) perhaps they could also be part of the `EthInfoService`; they provide info about smart contracts (namely the output.)

robinraju (approved):
🚀

PR #899, opened 2021-01-22
Feature/etcm 533 non membership proof

### LINKED ISSUE #350: Implement (non-) membership proof for an MPT element with a given key
For client applications (like light clients) this would be very helpful. There's an implementation of this functionality in Geth.

bsuieric:
Proofs for non existent values(keys)

In case an address or storage-value does not exist, the proof needs to provide enough data to verify this fact. This means the client needs to follow the path from the root node and deliver until the last matching node. If the last matching node is a branch, the proof value in the node must be an empty one. In case of leaf-type, it must be pointing to a different relative-path in order to proof that the requested path does not exist.

Specification: https://eips.ethereum.org/EIPS/eip-1186
Jira ticket: https://jira.iohk.io/projects/ETCM/issues/ETCM-533

Proposed Solution
following similar implementation with geth client: in case of non existing account, an error should be returned. In case of non existing storage keys, the proof should be returned with default value 0

Manual tests based on other clients are welcomed

### DIFF
--- a/src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
+++ b/src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
@@ -2,7 +2,6 @@ package io.iohk.ethereum.txExecTest.util
 
 import java.time.Clock
 import java.util.concurrent.atomic.AtomicReference
-
 import akka.actor.ActorSystem
 import akka.util.ByteString
 import com.typesafe.config.ConfigFactory
@@ -15,6 +14,7 @@ import io.iohk.ethereum.db.storage.pruning.{ArchivePruning, PruningMode}
 import io.iohk.ethereum.db.storage.{AppStateStorage, StateStorage}
 import io.iohk.ethereum.domain.BlockHeader.HeaderExtraFields.HefEmpty
 import io.iohk.ethereum.domain.{Blockchain, UInt256, _}
+import io.iohk.ethereum.jsonrpc.ProofService.{StorageProof, StorageValueProof}
 import io.iohk.ethereum.ledger.{InMemoryWorldStateProxy, InMemoryWorldStateProxyStorage}
 import io.iohk.ethereum.mpt.MptNode
 import io.iohk.ethereum.network.EtcPeerManagerActor.PeerInfo
@@ -150,7 +150,7 @@ class BlockchainMock(genesisHash: ByteString) extends Blockchain {
       rootHash: NodeHash,
       position: BigInt,
       ethCompatibleStorage: Boolean
-  ): (BigInt, Seq[MptNode]) = (BigInt(0), Seq.empty)
+  ): StorageProof = StorageValueProof(position)
 
   override protected def getHashByBlockNumber(number: BigInt): Option[ByteString] = Some(genesisHash)
 
--- a/src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
+++ b/src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
@@ -150,7 +150,7 @@ class BlockchainMock(genesisHash: ByteString) extends Blockchain {
       rootHash: NodeHash,
       position: BigInt,
       ethCompatibleStorage: Boolean
-  ): (BigInt, Seq[MptNode]) = (1, Seq.empty)
+  ): (BigInt, Seq[MptNode]) = (BigInt(0), Seq.empty)
 
   override protected def getHashByBlockNumber(number: BigInt): Option[ByteString] = Some(genesisHash)
 
--- a/src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
+++ b/src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
@@ -150,7 +150,7 @@ class BlockchainMock(genesisHash: ByteString) extends Blockchain {
       rootHash: NodeHash,
       position: BigInt,
       ethCompatibleStorage: Boolean
-  ): Option[(BigInt, Seq[MptNode])] = None
+  ): (BigInt, Seq[MptNode]) = (1, Seq.empty)
 
   override protected def getHashByBlockNumber(number: BigInt): Option[ByteString] = Some(genesisHash)
 
--- a/src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
+++ b/src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
@@ -14,7 +14,7 @@ import io.iohk.ethereum.db.storage.pruning.{ArchivePruning, PruningMode}
 import io.iohk.ethereum.db.storage.{AppStateStorage, StateStorage}
 import io.iohk.ethereum.domain.BlockHeader.HeaderExtraFields.HefEmpty
 import io.iohk.ethereum.domain.{Blockchain, UInt256, _}
-import io.iohk.ethereum.jsonrpc.ProofService.{StorageProof, StorageValueProof}
+import io.iohk.ethereum.jsonrpc.ProofService.{EmptyStorageValueProof, StorageProof, StorageProofKey, StorageValueProof}
 import io.iohk.ethereum.ledger.{InMemoryWorldStateProxy, InMemoryWorldStateProxyStorage}
 import io.iohk.ethereum.mpt.MptNode
 import io.iohk.ethereum.network.EtcPeerManagerActor.PeerInfo
@@ -150,7 +150,7 @@ class BlockchainMock(genesisHash: ByteString) extends Blockchain {
       rootHash: NodeHash,
       position: BigInt,
       ethCompatibleStorage: Boolean
-  ): StorageProof = StorageValueProof(position)
+  ): StorageProof = EmptyStorageValueProof(StorageProofKey(position))
 
   override protected def getHashByBlockNumber(number: BigInt): Option[ByteString] = Some(genesisHash)
 
--- a/src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
@@ -1,7 +1,6 @@
 package io.iohk.ethereum.domain
 
 import java.util.concurrent.atomic.AtomicReference
-
 import akka.util.ByteString
 import cats.syntax.flatMap._
 import cats.instances.option._
@@ -13,6 +12,7 @@ import io.iohk.ethereum.db.storage._
 import io.iohk.ethereum.db.storage.pruning.PruningMode
 import io.iohk.ethereum.domain
 import io.iohk.ethereum.domain.BlockchainImpl.BestBlockLatestCheckpointNumbers
+import io.iohk.ethereum.jsonrpc.ProofService.{StorageProof, StorageValueProof}
 import io.iohk.ethereum.ledger.{InMemoryWorldStateProxy, InMemoryWorldStateProxyStorage}
 import io.iohk.ethereum.mpt.{MerklePatriciaTrie, MptNode}
 import io.iohk.ethereum.utils.{ByteStringUtils, Logger}
@@ -105,7 +105,7 @@ trait Blockchain {
       rootHash: ByteString,
       position: BigInt,
       ethCompatibleStorage: Boolean
-  ): (BigInt, Seq[MptNode])
+  ): StorageProof
 
   /**
     * Returns the receipts based on a block hash
@@ -313,16 +313,21 @@ class BlockchainImpl(
       rootHash: ByteString,
       position: BigInt,
       ethCompatibleStorage: Boolean
-  ): (BigInt, Seq[MptNode]) = {
-    val defaultValue = BigInt(0)
+  ): StorageProof = {
     val storage: MptStorage = stateStorage.getBackingStorage(0)
     val mpt: MerklePatriciaTrie[BigInt, BigInt] = {
       if (ethCompatibleStorage) domain.EthereumUInt256Mpt.storageMpt(rootHash, storage)
       else domain.ArbitraryIntegerMpt.storageMpt(rootHash, storage)
     }
-    val value = mpt.get(position).getOrElse(defaultValue)
-    val proof = mpt.getProof(position).getOrElse(Seq.empty)
-    (value, proof)
+    val value = mpt.get(position)
+    val proof = mpt.getProof(position)
+
+    (value, proof) match {
+      case (Some(value), Some(proof)) => StorageValueProof(position, value, proof)
+      case (None, Some(proof)) => StorageValueProof(position, proof = proof)
+      case (Some(value), None) => StorageValueProof(position, value)
+      case (None, None) => StorageValueProof(position)
+    }
   }
 
   private def persistBestBlocksData(): Unit = {
--- a/src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
@@ -12,7 +12,7 @@ import io.iohk.ethereum.db.storage._
 import io.iohk.ethereum.db.storage.pruning.PruningMode
 import io.iohk.ethereum.domain
 import io.iohk.ethereum.domain.BlockchainImpl.BestBlockLatestCheckpointNumbers
-import io.iohk.ethereum.jsonrpc.ProofService.{StorageProof, StorageValueProof}
+import io.iohk.ethereum.jsonrpc.ProofService.StorageProof
 import io.iohk.ethereum.ledger.{InMemoryWorldStateProxy, InMemoryWorldStateProxyStorage}
 import io.iohk.ethereum.mpt.{MerklePatriciaTrie, MptNode}
 import io.iohk.ethereum.utils.{ByteStringUtils, Logger}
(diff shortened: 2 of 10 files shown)


src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
```diff
@@ -150,7 +150,7 @@ class BlockchainMock(genesisHash: ByteString) extends Blockchain {
       rootHash: NodeHash,
       position: BigInt,
       ethCompatibleStorage: Boolean
-  ): Option[(BigInt, Seq[MptNode])] = None
+  ): (BigInt, Seq[MptNode]) = (BigInt(0), Seq.empty)
```
dmitry-worker: minor: 
The following tuple can be extracted as a trait, a class and an empty object, like: 
```
trait Proof { 
  value: BigInt,
  proof: Seq[ByteString]
}

class NonEmptyProof ...

object EmptyProof extends Proof { 
  override val value = BigInt(0)
  override val proof = Nil
}
```

src/main/scala/io/iohk/ethereum/jsonrpc/EthProofService.scala
```diff
@@ -143,37 +143,32 @@ class EthProofService(blockchain: Blockchain, blockGenerator: BlockGenerator, et
         blockchain.getAccountProof(address, blockNumber).map(_.map(asRlpSerializedNode)),
         noAccountProof(address, blockNumber)
       )
-      storageProof <- getStorageProof(account, storageKeys)
+      storageProof <- Either.right(getStorageProof(account, storageKeys))
```
dmitry-worker:
```suggestion
      storageProof = getStorageProof(account, storageKeys)
```

src/main/scala/io/iohk/ethereum/jsonrpc/EthProofService.scala
```diff
@@ -143,37 +143,32 @@ class EthProofService(blockchain: Blockchain, blockGenerator: BlockGenerator, et
...
         blockchain
           .getStorageProofAt(
             rootHash = account.storageRoot,
             position = storageKey.v,
             ethCompatibleStorage = ethCompatibleStorage
-          )
-          .map { case (value, proof) => StorageProof(storageKey, value, proof.map(asRlpSerializedNode)) }
-          .toRight(noStorageProof(account, storageKey))
+          ) match {
+        case (value, proof) => StorageProof(storageKey, value, proof.map(asRlpSerializedNode))
```
dmitry-worker: If a single pattern is exhaustive - then, probably, `match` clause is redundant.

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -307,16 +313,16 @@ class BlockchainImpl(
       rootHash: ByteString,
       position: BigInt,
       ethCompatibleStorage: Boolean
-  ): Option[(BigInt, Seq[MptNode])] = {
+  ): (BigInt, Seq[MptNode]) = {
+    val defaultValue = BigInt(0)
```
dmitry-worker: .. and default value together with empty seq can become an `EmptyProof` like in case here: https://github.com/input-output-hk/mantis/pull/899#discussion_r562542081

dmitry-worker (changes requested):
~~major:~~
~~Please add some tests with negative scenarios! (when the proof is not found for example)~~


minor:
As soon as we have 
```
  case class StorageProof(
      key: StorageProofKey,
      value: BigInt,
      proof: Seq[ByteString]
  )
```
we can probably replace the resulting tuples in some of the methods with a similar structure, that can be upgraded to a `StorageProof` when storage key is available. StorageProof can also inherit the `Proof` trait and represent it without copying, if needed.

dmitry-worker (approved):

src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
```diff
@@ -554,22 +554,25 @@ class MerklePatriciaTrieSuite extends AnyFunSuite with ScalaCheckPropertyChecks
...
+      val val2: Array[Byte] = Hex.decode("0102")
+      val val3: Array[Byte] = Hex.decode("0103")
+      val trie = EmptyTrie
+        .put(key1, val1)
+        .put(key2, val2)
+        .put(key3, val3)
+      // when
+      val proof: Option[Vector[MptNode]] = trie.getProof(key4)
+      // then
+      assert(proof.isDefined)
```
dmitry-worker, line 608: Got confused (this test actually IS testing a negative scenario)
Maybe we should also check that vector is empty?

src/test/scala/io/iohk/ethereum/domain/BlockchainSpec.scala
```diff
@@ -152,7 +152,10 @@ class BlockchainSpec extends AnyFlatSpec with Matchers with ScalaCheckPropertyCh
     //unhappy path
     val wrongAddress = Address(666)
     val retrievedAccountProofWrong = blockchain.getAccountProof(wrongAddress, headerWithAcc.number)
-    retrievedAccountProofWrong.isDefined shouldBe false
+    //the account doesn't exist, so we can't retrieve it, but we do receive a proof of non-existence with a full path of nodes that we iterated
+    retrievedAccountProofWrong.isDefined shouldBe true
+    retrievedAccountProofWrong.size shouldBe 1
+    mptWithAcc.get(wrongAddress) shouldBe None
```
lemastero, line 158: It is good practice to have separate test for different behaviours. Hence I would split this guy to `proof of existing Account` and `proof of non existing Account`

src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
AnastasiiaL: a proof for non-existing value is not necessarily empty, it can be a list of all nodes traversed while trying to find the value. So the only difference between empty and non-empty is the default value, but I like the idea of replacing a tuple with a class and having the default value for the `value` parameter then.

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
AnastasiiaL: @dmitry-worker so I made a change here, and replaced a tuple with a `StorageValueProof` class from EthProofService, but now I see that this way we are coupling 2 different modules and putting specifics of implementation into a Blockchain trait. What do you think?

src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
AnastasiiaL: the vector shouldn't be empty, it should contain the nodes it traversed while searching for the account. Butt good point to check not only that it's defined, but also that it contains something

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
dmitry-worker: I like the way it looks now, perfectly readable. As for the JsonRPC classes imported into a blockchain - it's a good point and I would ask @dzajkowski - does he prefer to move those guys from JsonRPC to the blockchain? Or maybe to create an internal image for these classes. This can be done afterwards, in a separate PR.

dzajkowski:
@bsuieric looks like it's not formatted

src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
dzajkowski: what does the case Some(Vector.empty) mean?

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -307,16 +313,21 @@ class BlockchainImpl(
...
-    } yield (value, proof)
+    val value = mpt.get(position)
+    val proof = mpt.getProof(position)
+
+    (value, proof) match {
+      case (Some(value), Some(proof)) => StorageValueProof(position, value, proof)
+      case (None, Some(proof)) => StorageValueProof(position, proof = proof)
+      case (Some(value), None) => StorageValueProof(position, value)
+      case (None, None) => StorageValueProof(position)
+    }
```
dzajkowski: If we decided to move to an ADT: I think this match should be moved to the companion object - essentially it's a construction detail. WDYT?

src/main/scala/io/iohk/ethereum/jsonrpc/EthProofService.scala
```diff
@@ -30,21 +24,33 @@ object ProofService {
 
   case class GetProofResponse(proofAccount: ProofAccount)
 
-  /** The key used to get the storage slot in its account tree */
-  case class StorageProofKey(v: BigInt) extends AnyVal
-
+  trait StorageProof {
```
dzajkowski: maybe `sealed trait`

src/main/scala/io/iohk/ethereum/jsonrpc/EthProofService.scala
```diff
@@ -30,21 +24,33 @@ object ProofService {
...
+  }
   /**
     * Object proving a relationship of a storage value to an account's storageHash
     *
     * @param key   storage proof key
     * @param value the value of the storage slot in its account tree
     * @param proof the set of node values needed to traverse a patricia merkle tree (from root to leaf) to retrieve a value
     */
-  case class StorageProof(
+  case class StorageValueProof(
```
dzajkowski: perhaps it would make sense to name the four distinct cases?

src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
AnastasiiaL: it means there was no root node in MPT, so no proof gathered

src/main/scala/io/iohk/ethereum/mpt/MerklePatriciaTrie.scala
```diff
@@ -84,10 +84,10 @@ class MerklePatriciaTrie[K, V] private (private[mpt] val rootNode: Option[MptNod
...
       node match {
-        case LeafNode(_, value, _, _, _) =>
+        case Some(LeafNode(_, value, _, _, _)) =>
           Some(vSerializer.fromBytes(value.toArray[Byte]))
 
-        case BranchNode(_, terminator, _, _, _) =>
+        case Some(BranchNode(_, terminator, _, _, _)) =>
           terminator.map(term => vSerializer.fromBytes(term.toArray[Byte]))
 
         case _ => None
```
dzajkowski: a little bit of code golf:
```scala
  def get(key: K): Option[V] = {
    pathTraverse[Option[V]](None, mkKeyNibbles(key)) {
      case (_, Some(LeafNode(_, value, _, _, _))) =>
        Some(vSerializer.fromBytes(value.toArray[Byte]))

      case (_, Some(BranchNode(_, terminator, _, _, _))) =>
        terminator.map(term => vSerializer.fromBytes(term.toArray[Byte]))

      case _ => None
    }.flatten
  }
```

src/main/scala/io/iohk/ethereum/mpt/MerklePatriciaTrie.scala
```diff
@@ -121,25 +121,25 @@ class MerklePatriciaTrie[K, V] private (private[mpt] val rootNode: Option[MptNod
     * @tparam T accumulator type
     * @return accumulated data or None if key doesn't exist
     */
-  private def pathTraverse[T](acc: T, searchKey: Array[Byte])(op: (T, MptNode) => T): Option[T] = {
+  private def pathTraverse[T](acc: T, searchKey: Array[Byte])(op: (T, Option[MptNode]) => T): Option[T] = {
```
dzajkowski, line 122: it's interesting that this was not the default flow.

src/test/scala/io/iohk/ethereum/domain/BlockchainSpec.scala
```diff
@@ -162,6 +165,24 @@ class BlockchainSpec extends AnyFlatSpec with Matchers with ScalaCheckPropertyCh
...
+
+    val headerWithAcc = Fixtures.Blocks.ValidBlock.header.copy(stateRoot = ByteString(mptWithAcc.getRootHash))
+
+    blockchain.storeBlockHeader(headerWithAcc).commit()
+
+    val wrongAddress = Address(666)
+    val retrievedAccountProofWrong = blockchain.getAccountProof(wrongAddress, headerWithAcc.number)
+    //the account doesn't exist, so we can't retrieve it, but we do receive a proof of non-existence with a full path of nodes that we iterated
+    retrievedAccountProofWrong.isDefined shouldBe true
+    retrievedAccountProofWrong.size shouldBe 1
```
dzajkowski: would it be possible to check the value of the first step of the proof?

src/test/scala/io/iohk/ethereum/jsonrpc/EthProofServiceSpec.scala
```diff
@@ -126,18 +78,113 @@ class EthProofServiceSpec
...
+    val result = retrievedAccountProofWrong.runSyncUnsafe()
+    result.isRight shouldBe true
+    result.fold(l => l, r => r.proofAccount.storageProof.map(v => {
+      v.proof.nonEmpty shouldBe true
+      v.value shouldBe BigInt(0)
+    }))
+  }
+
+  "EthProofService" should "return the proof and value for existing storage key" in new TestSetup {
+    val storageKey = Seq(StorageProofKey(key))
```
dzajkowski, line 98: looks like these can be extracted as a common helper (repeated in all tests). maybe:
```scala
val result = fetchProof(address, key, blockNo).runSyncUnsafe()
```

src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
dzajkowski: and what's the difference of meaning between None and Some(Vector.empty)?

AnastasiiaL: According to the spec and geth implementation - there always should be a proof in response, even if the proof is empty we should return the key (for which the search was requested), the value that defaults to zero and the proof, which defaults to empty Seq. That's why the getProof method on MerklePatriciaTrie can return None as a proof, but when we construct an object of StorageValueProof, we need an empty sequence instead of None

AnastasiiaL: ah, wait, I probably confused you with my answer :)
If you were asking about specifically MerkleParticiaTrie#getProof - case Some(Vector.empty) should never happen. If there's a root node - it will never be empty. If there's no root node - it's gonna return None. 

I was talking about the next level of transforming the results.

AnastasiiaL:
Nice addition with tests for MPT and making a proof always return a root node! Now we are following the spec, but with this we part ways with geth, which returns an empty proof instead of Seq(rootNode). Can you check if besu does the same as geth? We need to make a decision here - to follow the spec or other clients implementation.

src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
bsuieric: Can this be resolved?

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
bsuieric: Sounds great

src/main/scala/io/iohk/ethereum/jsonrpc/EthProofService.scala
bsuieric: Because, initially, was intended to get the proof only when the value was present

src/main/scala/io/iohk/ethereum/mpt/MerklePatriciaTrie.scala
bsuieric: Because, initially, was intended to get the proof only when the value was present

src/main/scala/io/iohk/ethereum/jsonrpc/EthProofService.scala
bsuieric: Not sure I follow, based on the pattern matching? I added the pattern matching in the apply method. Can we find a prettier way?

src/main/scala/io/iohk/ethereum/mpt/MerklePatriciaTrie.scala
bsuieric: applied your changes

src/test/scala/io/iohk/ethereum/domain/BlockchainSpec.scala
bsuieric: The first step will always be the root

src/test/scala/io/iohk/ethereum/mpt/MerklePatriciaTrieSuite.scala
dzajkowski: nitpick: the signature could be `Option[NonEmptyList]` to remove the option of having a Some(empty) result.

@bsuieric sure

src/test/scala/io/iohk/ethereum/domain/BlockchainSpec.scala
dzajkowski: exactly, so matching to the root gives us some confidence about the algorithm behaving in an expected way. it's not a big deal, just a note.

lemastero (approved):

dzajkowski (approved):

PR #903, opened 2021-01-26
[ETCM-540] Improve peer discovery algorithm

AnastasiiaL:
# Description

With each round of scanning for new nodes to connect, we only take N from that pool, trying to connect to them. This process takes a lot of time as not all peers are easy to connect to and if no nodes we suitable - we have to wait for another round of scanning. And then it's possible that we take N nodes from the new scan and they will contain nodes that were previously tried, but the connection was not successful. So instead of retrying those nodes, first we want to go through the whole list of discovered peers. 

# Proposed Solution

This PR introduces cashing of the tried nodes with a Map bounded by size to the size of the blacklist. So first we'll be trying to connect to the new nodes, that are not cashed in that map.

If the scanning round didn't find enough nodes for us to be picky - we try to connect to all of them regardless the fact if they were tried before.

Parameters that control blacklisting were reviewed - long blacklisting (in case wrong protocol, incompatible network, timeout during connection) was increased from 30min to 10h, as it's unlikely that nodes from another network/different protocol might suddenly change within 30min timeframe.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
+++ b/src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
@@ -1,6 +1,5 @@
 package io.iohk.ethereum.network
 
-import java.net.{InetSocketAddress, URI}
 import akka.actor.SupervisorStrategy.Stop
 import akka.actor._
 import akka.util.{ByteString, Timeout}
@@ -22,7 +21,12 @@ import io.iohk.ethereum.network.rlpx.RLPxConnectionHandler.RLPxConfiguration
 import monix.eval.Task
 import monix.execution.{Scheduler => MonixScheduler}
 import org.bouncycastle.util.encoders.Hex
+
+import java.net.{InetSocketAddress, URI}
+import java.util.Collections.newSetFromMap
+import scala.collection.mutable
 import scala.concurrent.duration._
+import scala.jdk.CollectionConverters._
 
 class PeerManagerActor(
     peerEventBus: ActorRef,
@@ -50,6 +54,8 @@ class PeerManagerActor(
 
   import PeerManagerActor._
   import akka.pattern.pipe
+
+  val triedNodes: mutable.Set[ByteString] = lruSet[ByteString](maxBlacklistedNodes)
 
   implicit class ConnectedPeersOps(connectedPeers: ConnectedPeers) {
 
@@ -126,6 +132,7 @@ class PeerManagerActor(
   private def maybeConnectToRandomNode(connectedPeers: ConnectedPeers, node: Node): Unit = {
     if (connectedPeers.outgoingConnectionDemand > 0) {
       if (connectedPeers.canConnectTo(node)) {
+        triedNodes.add(node.id)
         self ! ConnectToPeer(node.toUri)
       } else {
         peerDiscoveryManager ! PeerDiscoveryManager.GetRandomNodeInfo
@@ -134,9 +141,15 @@ class PeerManagerActor(
   }
 
   private def maybeConnectToDiscoveredNodes(connectedPeers: ConnectedPeers, nodes: Set[Node]): Unit = {
-    val nodesToConnect = nodes
+    val discoveredNodes = nodes
       .filter(connectedPeers.canConnectTo)
-      .take(connectedPeers.outgoingConnectionDemand)
+
+    val nodesToConnect = discoveredNodes
+      .filterNot(n => triedNodes.contains(n.id)) match {
+      case seq if seq.size >= connectedPeers.outgoingConnectionDemand =>
+        seq.take(connectedPeers.outgoingConnectionDemand)
+      case _ => discoveredNodes.take(connectedPeers.outgoingConnectionDemand)
+    }
 
     NetworkMetrics.DiscoveredPeersSize.set(nodes.size)
     NetworkMetrics.BlacklistedPeersSize.set(blacklistedPeers.size)
@@ -152,14 +165,17 @@ class PeerManagerActor(
 
     if (nodesToConnect.nonEmpty) {
       log.debug("Trying to connect to {} nodes", nodesToConnect.size)
-      nodesToConnect.foreach(n => self ! ConnectToPeer(n.toUri))
+      nodesToConnect.foreach(n => {
+        triedNodes.add(n.id)
+        self ! ConnectToPeer(n.toUri)
+      })
     } else {
       log.debug("The nodes list is empty, no new nodes to connect to")
     }
 
     // Make sure the background lookups keep going and we don't get stuck with 0
     // nodes to connect to until the next discovery scan loop. Only sending 1
-    // request so we don't rack up too many pending futures, just trigger a a
+    // request so we don't rack up too many pending futures, just trigger a
     // search if needed.
     if (connectedPeers.outgoingConnectionDemand > nodesToConnect.size) {
       peerDiscoveryManager ! PeerDiscoveryManager.GetRandomNodeInfo
@@ -184,7 +200,7 @@ class PeerManagerActor(
   private def getBlacklistDuration(reason: Long): FiniteDuration = {
     import Disconnect.Reasons._
     reason match {
-      case TooManyPeers => peerConfiguration.shortBlacklistDuration
+      case TooManyPeers | AlreadyConnected | ClientQuitting => peerConfiguration.shortBlacklistDuration
       case _ => peerConfiguration.longBlacklistDuration
     }
   }
@@ -552,4 +568,9 @@ object PeerManagerActor {
       }
       .getOrElse(0.0)
   }
+
+  def lruSet[A](maxEntries: Int): mutable.Set[A] =
+    newSetFromMap[A](new java.util.LinkedHashMap[A, java.lang.Boolean]() {
+      override def removeEldestEntry(eldest: java.util.Map.Entry[A, java.lang.Boolean]): Boolean = size > maxEntries
+    }).asScala
 }
--- a/src/test/scala/io/iohk/ethereum/network/PeerManagerSpec.scala
+++ b/src/test/scala/io/iohk/ethereum/network/PeerManagerSpec.scala
@@ -25,7 +25,7 @@ import org.scalatest.concurrent.Eventually
 import org.scalatest.flatspec.AnyFlatSpecLike
 import org.scalatest.matchers.should.Matchers
 import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
-import org.scalacheck.{Arbitrary, Gen, Shrink}, Arbitrary.arbitrary
+import org.scalacheck.{Arbitrary, Gen}, Arbitrary.arbitrary
 import scala.concurrent.duration._
 
 // scalastyle:off magic.number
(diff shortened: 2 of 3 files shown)


src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -134,9 +143,15 @@ class PeerManagerActor(
...
 
   private def maybeConnectToDiscoveredNodes(connectedPeers: ConnectedPeers, nodes: Set[Node]): Unit = {
-    val nodesToConnect = nodes
+    val discoveredNodes = nodes
       .filter(connectedPeers.canConnectTo)
-      .take(connectedPeers.outgoingConnectionDemand)
+
+    val nodesToConnect = discoveredNodes
+      .filterNot(n => triedNodes.contains(n.id)) match {
+      case seq if seq.size > connectedPeers.outgoingConnectionDemand =>
```
leo-bogastry: Can the `if` be `>=` instead?

src/test/scala/io/iohk/ethereum/network/PeerManagerSpec.scala
```diff
@@ -286,6 +286,45 @@ class PeerManagerSpec
     }
   }
 
+  it should "try to connect to discovered nodes if there's an outgoing demand: new nodes first, retried last" in new TestSetup {
```
leo-bogastry, line 289: Just for my understanding, what is the value of `connectedPeers.outgoingConnectionDemand` during this test?

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
AnastasiiaL: no, because seq is a set of nodes, and if condition is on it's size

src/test/scala/io/iohk/ethereum/network/PeerManagerSpec.scala
AnastasiiaL: `maxOutgoingPeers` is set to 30, `outgoingConnectionDemand` is calculated based on that number minus number connections established. So we start with 30

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
AnastasiiaL: or you meant not removing the if but changing the condition? :) yes, it makes sense

leo-bogastry: Haha sorry, I will be more verbose next time :)

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -152,14 +167,17 @@ class PeerManagerActor(
 
     if (nodesToConnect.nonEmpty) {
       log.debug("Trying to connect to {} nodes", nodesToConnect.size)
-      nodesToConnect.foreach(n => self ! ConnectToPeer(n.toUri))
+      nodesToConnect.foreach(n => {
+        triedNodes.put(n.id, n)
```
KonradStaniec: maybe we should mark node as tried also when we connect to it via random tries - https://github.com/input-output-hk/mantis/pull/903/files#diff-00a5e82132cd8fdaa408d88c385d12965f491c99a68c704c8a7db90a481dbff4R138 ?

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -48,6 +51,12 @@ class PeerManagerActor(
     */
   override val maxBlacklistedNodes: Int = 32 * 8 * discoveryConfig.kademliaBucketSize
 
+  def lruCache[A, B](maxEntries: Int): mutable.Map[A, B] = new java.util.LinkedHashMap[A, B]() {
```
KonradStaniec: probably it would be more space efficient to have something as :
```
final class LruSet[K <: AnyRef](maxSize: Long) {
  import LruSet._
  private val cache: LinkedHashMap[K, DummyObject.type] = lruCache[K, DummyObject.type](maxSize)

  def put(key: K): Unit = {
    cache.put(key, DummyObject)
  }

  def contains(key: K): Boolean = {
    cache.contains(k)
  }

  def size: Long = cache.size()
}
object LruSet {
  private case object DummyObject
}
```
That way we would not keep values unecessarily in memory, as we only call `put` and `contains`

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -134,9 +143,15 @@ class PeerManagerActor(
...
-    val nodesToConnect = nodes
+    val discoveredNodes = nodes
       .filter(connectedPeers.canConnectTo)
-      .take(connectedPeers.outgoingConnectionDemand)
+
+    val nodesToConnect = discoveredNodes
+      .filterNot(n => triedNodes.contains(n.id)) match {
+      case seq if seq.size >= connectedPeers.outgoingConnectionDemand =>
+        seq.take(connectedPeers.outgoingConnectionDemand)
+      case _ => discoveredNodes.take(connectedPeers.outgoingConnectionDemand)
```
KonradStaniec, line 151: I am no sure if this will be a problem, but this means that if network size is less than `triedNodes.maxSize`, then after first traversal we are back to our old way of checking/establishing connections, and if network size is much larger then we may still not try all nodes as they will be constantly changing in `triedNodes` cache and in `blacklist`. So in summary, I am not sure how much this will help, and the best way to find out is run mantis several times on etc mainnet and mordor to check it out.

KonradStaniec (commented):
did you try running mantis node with those changes to see how fast it traverses the peers it get from discovery ?

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
AnastasiiaL: the logic behind that is
- if the network size is small - then it's better to retry previously failed nodes than not do anything at all if we don't have any new nodes to try
- if the network size is large - we should try only new nodes until the moment the cache is full and we start replacing old values with new ones. 

Do you think it makes sense to review the max size of the cache?

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
AnastasiiaL: could you elaborate a bit more on that please? I don't quite grasp how keeping the cash in one place over the other helps with the space efficiency. Thanks

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
AnastasiiaL: yes, makes total sense. Was thinking about it also, but forgot to add

AnastasiiaL:
> did you try running mantis node with those changes to see how fast it traverses the peers it get from discovery ?

I ran to see that nothing got broken, but was not sure how to see if there's any improvement in the speed as it's working only in a specific situation when previously blacklisted modes are tried instead of new ones, for this the network should be large together with a large amount of blacklisted nodes.

Edit: I'm measuring the time to start syncing at the moment

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
KonradStaniec: currently we when adding nodes to cache we add always tuple `id -> node` which means that both `id` and `Node` case class are kept in memory. 

When using approach above, we would always add different `id` for each peer, but as value we would add only `DummyObject` which is singleton, which means all values in map are actually references to one object in memory, which ultimatly means that instead of multiple (with current default around 4k) values in map, we only have one.

AnastasiiaL: aaah I understand now. Thanks a lot for a detailed explanation!

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
KonradStaniec: I agree with this logic, and probably it is enough for small network. 

But in case of large network, we may still not try all peer provided by discovery. Lets say the network have 8k peers (which after some time is not impossible) and current max cache size is 4k. Then after cache gets full, our situation is basically the same as currently i.e we will have 4k peers to try and will have no way to be certain we tried them all.

as for reviewing cache size it may help, but i am not sure there is way to derive optimal value as network size can size dynamically.

In general I like this apprach becouse it solves problem in most of cases in simple way, but the question is this enough are what are benefits in comparission to baseline.

KonradStaniec:
Time to start syncing is pretty tricky measure which has also other factors in it.

 Maybe the best metric would be number of peers we tried to connect to after some time ( maybe as percentage of nodes provided by discovery) ? ( just thinking out loud here as i am also not sure 😅 )

AnastasiiaL:
> Time to start syncing is pretty tricky measure which has also other factors in it.
> 
> Maybe the best metric would be number of peers we tried to connect to after some time ( maybe as percentage of nodes provided by discovery) ? ( just thinking out loud here as i am also not sure 😅 )

Indeed, after running it multiple times I saw the time to sync drastically differs. Ended up looking at the ratio of connected nodes to discovered ones :)

So running on develop branch the percentage of connected nodes after 10min is 0.273%
                                                                                                          while on this branch is 0.316 %. That's a 16% improvement.
 That's an averaged out percentage out of 5 tries. Not sure it's very reliable but the change definitely didn't make things worse

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
AnastasiiaL: I found a nicer way to discard the value - using a set and keeping the map underhood
```
def lruSet[A](maxEntries: Int): mutable.Set[A] =
    newSetFromMap[A](new java.util.LinkedHashMap[A, java.lang.Boolean]() {
      override def removeEldestEntry(eldest: java.util.Map.Entry[A, java.lang.Boolean]): Boolean = size > maxEntries
    }).asScala
```
    
Will resolve this thread if it looks good to you

KonradStaniec: as for me it looks good 👍

KonradStaniec (approved):
It would nice to check what the ratio looks after 1h i.e when we had acquire more peers, but if it helps even a little bit  then LGTM! 

When i was thinking about this problem my inital solution was to change our interactions with discovery form:  ask discovery for all found peers every 1 minute, and try to connect to `X` peers
to something like: ask discovery for all found peers, try to connect to all of them (by calling to them in batches as it is now), and ask for discovery for all found peers again only after we check we are sure we checked all peers from previous batch. 

But maybe this something for other pr.

AnastasiiaL:
Just checked the ration after running for 1h:
develop: Discovered 5343 nodes, Blacklisted 1708 nodes, handshaked to 11/80  ----- 0.206%
etcm-540: Discovered 5336 nodes, Blacklisted 2824 nodes, handshaked to 17/80 ------ 0.319%

this difference could be caused by increased time for the blacklist as well, as seen above the number of blacklisted nodes with the new configuration is 65% higher

AnastasiiaL:
@KonradStaniec I like your idea of changing the interaction with discovery as well. No need asking for new nodes while we didn't process the previously discovered ones. The question in this case would be how fast the discovered batch is processed and if it doesn't leave us with an outdated info. I think the follow-up task could be created, WDYT @dzajkowski?

dzajkowski (approved):

PR #904, opened 2021-01-26
[ETCM-528] block creation metrics

jvdp:
This PR aims to help improve our understanding of the performance of `eth_getWork` and Mantis' usefulness for coordinating mining pools.

The JSON RPC Controller already had instrumentation but without a method label, so I've added this. In addition I added a `ConsensusMetrics` object with timer metrics for the different `generateBlock` implementations (with a class label.) I've updated the mantis-dashboard.json file with the new metrics.

(Recreated from #902 with a consistent branch name.)


leo-bogastry (approved):
It looks good to me!

robinraju (approved):
LGTM

bsuieric (approved):

PR #905, opened 2021-01-26
[ETCM-555] Add metrics on the time taken to download blocks (headers, bodies and receipts) and also MPT nodes. Expose in Grafana other already existing metrics.

leo-bogastry:
# Description
The goal of this PR is help measuring FastSync performance.

# Proposed Solution
Some metrics were already in place but were not available in the local docker-compose with Grafana setup. Following metrics added to Grafana:
```
fastsync.block.pivotBlock.number.gauge
fastsync.block.bestFullBlock.number.gauge
astsync.block.bestHeader.number.gauge
fastsync.state.totalNodes.gauge
fastsync.state.downloadedNodes.gauge
```

Other measurement were already being done, but were not available on a metric (time elapsed during the download of headers bodies and receipts). Following metrics added:
```
fastsync.block.downloadBlockHeaders.timer
fastsync.block.downloadBlockBodies.timer
fastsync.block.downloadBlockReceipts.timer
fastsync.state.downloadState.timer
```

The elapsed time is now being calculated using System.nanoTime() instead of System.currentTimeMillis(), because System.nanoTime() is more reliable when calculated elapsed time

### DIFF
--- a/docker/monitoring/grafana/provisioning/dashboards/mantis-dashboard.json
+++ b/docker/monitoring/grafana/provisioning/dashboards/mantis-dashboard.json
@@ -15,7 +15,7 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "iteration": 1611829944845,
+  "iteration": 1611838822810,
   "links": [],
   "panels": [
     {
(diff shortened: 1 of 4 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -171,6 +171,8 @@ class FastSync(
 
       case ResponseReceived(peer, BlockHeaders(blockHeaders), timeTaken) =>
         log.info("*** Received {} block headers in {} ms ***", blockHeaders.size, timeTaken)
+        SyncMetrics.setBlockHeadersDownloadedGauge(timeTaken)
```
jvdp: I'm kinda curious about using a gauge for this (and the other metrics) instead of a timer. A downside compared to the timer is that you lose some of the timings between samples (ie. Grafana pulling data from Prometheus.) With a timer you instead get an average and maximum for each interval, and can extend it with a histogram.

leo-bogastry: When you use a timer you can pass a clear method to it and it will calculate the time for you. This case is not so straightforward.
I forgot to add a screenshot of how the metrics were looking like in Grafana, now I am running mantis already on another branch...

leo-bogastry:
I have added one more metric about the total time it takes to download MPT nodes, the value was already available and being printed in logs. I have also adjusted current metrics to be histograms instead (and fixed the naming)
Screenshots of added panels in Grafana:
Screenshot 2021-01-27 at 17 07 15
Screenshot 2021-01-27 at 14 47 58
Screenshot 2021-01-27 at 14 55 57

I'm sorry that the last panel is not showing data!

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
leo-bogastry: So I have transformed the metrics into timers, but the values are calculated the same way. And indeed I am using rates now

leo-bogastry:
Some last adjustments done

leo-bogastry:
I have undone the modification of using `System.nanoTime` instead of `System.currentTimeMillis()`, given the discussion that took place in https://github.com/input-output-hk/mantis/pull/907#discussion_r565892302

robinraju (approved):
🚀

leo-bogastry:
Rebased with `develop`

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/SyncMetrics.scala
```diff
@@ -1,28 +1,39 @@
...
 import com.google.common.util.concurrent.AtomicDouble
 import io.iohk.ethereum.blockchain.sync.fast.FastSync.SyncState
 import io.iohk.ethereum.metrics.MetricsContainer
 
+import scala.concurrent.duration.MILLISECONDS
+
 object SyncMetrics extends MetricsContainer {
 
-  private[this] final val PivotBlockNumberGauge =
+  private final val PivotBlockNumberGauge =
```
AnastasiiaL, line 12: shall we keep private[this]? it imposes a bigger restriction

leo-bogastry: As far as I understand `private[this]` is not more restrictive. No place outside this object can access that val already, or tries to

AnastasiiaL: [private] makes it available to the current instance and other instances of the class it’s declared in. Since we are already using a singleton - there are no other instances that could access it. So there's no difference in this case. Some consider it a good practice to default to the highest restriction, I guess this was the reason private[this] was used here on the first place. We could make a decision as a team on which guideline to follow

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -440,7 +452,7 @@ class FastSync(
           case ImportedPivotBlock =>
             updatePivotBlock(ImportedLastBlock)
           case ValidationFailed(header, peerToBlackList) =>
-            log.warning(s"validation of header ${header.idTag} failed")
+            log.warning("validation of header {} failed", header.idTag)
```
dzajkowski, line 455: oh no.... we need to move to a proper logging wrapper. making ticket.

leo-bogastry:
Rebased

docker/monitoring/grafana/provisioning/dashboards/mantis-dashboard.json
```diff
@@ -91,14 +91,774 @@
...
+          "interval": "",
+          "legendFormat": "{{instance}}",
+          "refId": "A"
+        }
+      ],
+      "thresholds": [],
+      "timeFrom": null,
+      "timeRegions": [],
+      "timeShift": null,
+      "title": "MTP Nodes Download time",
```
ten15bit: minor...`MTP` should be `MPT`

leo-bogastry: Good catch! Thanks, it's fixed

ten15bit (approved):
:+1:

PR #906, opened 2021-01-27
[FIX] Align nix expression evaluation with recent changes

dzajkowski:
# Description

`nix eval '(1 + 1)'` used to evaluate but some recent changes resulted in this encoding being no longer supported.

This resulted in:
```
error: --- BadURL --- nix
'(import nix/sources.nix).nixpkgs' is not a valid URL
```

# Proposed Solution

@Infinisil proposed an alternative that should be stable:
replace the above with `nix-instantiate --eval --expr '1 + 1'` calls.

# Testing

The obvious test of buildkite working again.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,8 +1,8 @@
 steps:
   - label: ":nix::point_right::pipeline:"
     command: |
-      export NIX_PATH="nixpkgs=$(nix eval --raw '(import nix/sources.nix).nixpkgs')"
-      nix eval --json '(import ./.buildkite { pipeline = ./.buildkite/pipeline.nix; })' \
+      export NIX_PATH="nixpkgs=$(nix-instantiate --verbose ./nix/sources.nix)"
+      nix-instantiate --eval --strict --json --expr '(import ./.buildkite { pipeline = ./.buildkite/pipeline.nix; })' \
       | buildkite-agent pipeline upload --no-interpolation
     agents:
       queue: project42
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -1,8 +1,8 @@
 steps:
   - label: ":nix::point_right::pipeline:"
     command: |
-      export NIX_PATH="nixpkgs=$(nix eval --raw '(import nix/sources.nix).nixpkgs')"
-      nix eval --json '(import ./.buildkite { pipeline = ./.buildkite/pipeline.nix; })' \
+      export NIX_PATH="nixpkgs=$(nix-instantiate --verbose ./nix/sources.nix)"
```
infinisil:
```suggestion
      export NIX_PATH="nixpkgs=$(nix-instantiate --eval --strict --json --read-write-mode -E '(import nix/sources.nix).nixpkgs' | tr -d '"')"
```

A bit of a handful, but:
- `--eval` to only evaluate Nix, not attempt an instantiation
- `--strict` because we want to fully evaluate the result, needed for `--json`
- `--json` so that the result (which is a set) gets converted into a string automatically (the path to nixpkgs)
- `--read-write-mode` to ensure that the resulting path to nixpkgs actually exists. By default `--eval` can't write anything to the store
- `tr -d '"'` to strip the `"` on the left and right of the result. This will be an advantage of the new `nix eval` once that's stable, because it supports `--raw` which does that automatically.

infinisil (approved):

PR #907, opened 2021-01-27
[ETCM-556] Add a metric for the total time it takes for FastSync to complete and update docker-compose Grafana dashboard

leo-bogastry:
# Description

The goal of this PR is help measuring FastSync performance. 
A previous PR with more metrics was opened: https://github.com/input-output-hk/mantis/pull/905

# Proposed Solution
A new metric called `fastsync.totaltime.minutes.gauge` was added to record the amount of minutes elapsed since the start until the finish of StartSync.

# Remarks
Waiting for FastSync to complete takes a really long time with the current available networks, so this was not tested the way I would like to. 

Screenshot of a dummy value:
Screenshot 2021-01-27 at 16 40 08

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
@@ -1,7 +1,6 @@
 package io.iohk.ethereum.blockchain.sync.fast
 
 import java.time.Instant
-
 import akka.actor._
 import akka.util.ByteString
 import cats.data.NonEmptyList
@@ -29,6 +28,7 @@ import io.iohk.ethereum.utils.ByteStringUtils
 import io.iohk.ethereum.utils.Config.SyncConfig
 import org.bouncycastle.util.encoders.Hex
 
+import java.util.concurrent.TimeUnit
 import scala.annotation.tailrec
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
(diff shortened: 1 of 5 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -151,6 +151,10 @@ class FastSync(
     private val heartBeat =
       scheduler.scheduleWithFixedDelay(syncRetryInterval, syncRetryInterval * 2, self, ProcessSyncing)
 
+    private val startTime: Long = System.nanoTime()
+    private val toMinutes: Long = 6000000000L
+    private def totalMinutesTaken(): Long = (System.nanoTime() - startTime) / toMinutes
```
dzajkowski: not sure what the gain of using nano time here is. AFAIR nano time is considerably slower than millis so maybe it would be a good idea to switch to `currentTimeMillis`?

leo-bogastry: The reason for using `nanoTime()` is because when it come to recording elapsed time `currentTimeMillis()` is not so accurate and `nanoTime()` is recommended instead. About the performance I found older articles showing `currentTimeMillis()` being way faster, but more recent benchmarks show `nanoTime()` being faster actually :/

bsuieric: AFAIK currentTimeMillis is faster, but the question is: is this considered a high precision task? can we live with less precision? both options are fine with me

leo-bogastry: I don't have a strong reason to keep it in `nanoTime`, so I have updated the calculation to use `currentTimeMillis` instead

jvdp (approved):

bsuieric (approved):

PR #908, opened 2021-01-27
Attempt to fix pipeline

jonringer:
When upgrading the buildkite agents, I caused some regression when switching to nixUnstable.


jonringer:
fixed in a similar way in https://github.com/input-output-hk/mantis/commit/d886a96dbb1fc1c5d58cca9afdf80d509b31e215

PR #914, opened 2021-02-02
ETCM-530 Add a setup for Besu and Geth that allows to evaluate their performance

leo-bogastry:
# Description

Add a setup for Besu and Geth that allows to evaluate their performance.

# Proposed Solution

Similarly to what is already available for Mantis, a docker setup was added for Besu and Geth where metrics are exposed and a basic Grafana dashboard is available.
More metrics can be explored and added to Grafana dashboards later.

# Testing

Run each provided script


robinraju (approved):
LGTM, verified it locally.

jvdp (approved):
Haven't tried it locally but it looks good to my eyeballs 👍 

Have you considered having one `docker-compose.yml` so you can share grafana and prometheus and run subsets as required? Or everything when doing a comparative benchmark.

leo-bogastry:
> Haven't tried it locally but it looks good to my eyeballs 👍
> 
> Have you considered having one `docker-compose.yml` so you can share grafana and prometheus and run subsets as required? Or everything when doing a comparative benchmark.

I understand your point of view, but that sounds messy actually, having more than one ETC client running on our machines and competing for resources, it would be more far from a real scenario

leo-bogastry:
Rebased

jvdp:
> I understand your point of view, but that sounds messy actually, having more than one ETC client running on our machines and competing for resources, it would be more far from a real scenario

Ah yeah that makes sense, they'd only get in eachother's way.

PR #916, opened 2021-02-03
ETCM-147 use explicitly triggered scheduler for SyncControllerSpec

jvdp:
The actors used by `SyncController` use the scheduler to retry various actions and its tests rely on this behavior to trigger, which involves retrying over period of up to 10 or 30 seconds. By using an explicitly triggered scheduler we can instead advance the clock artificially and only a short time.

Because of the complexity of the actors and the different timeouts I had to find the time steps (100ms and 3000ms) by trial and error, unfortunately. If they are too high or too low the specs will fail. In addition, the speed up is only about 2-4x.

### DIFF
--- a/src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
+++ b/src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
@@ -48,17 +48,8 @@ class SyncControllerSpec
     with Eventually
     with NormalPatience {
 
-  implicit var system: ActorSystem = _
-
-  before {
-    system = ActorSystem("SyncControllerSpec_System", ConfigFactory.load("explicit-scheduler"))
-  }
-
-  after {
-    Await.result(system.terminate(), 10.seconds)
-  }
-
-  "SyncController" should "download pivot block and request block headers" in new TestSetup() {
+  "SyncController" should "download pivot block and request block headers" in withTestSetup() { testSetup =>
+    import testSetup._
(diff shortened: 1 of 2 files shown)


src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
```diff
@@ -35,14 +35,23 @@ import org.scalatest.time.{Seconds, Span}
...
-class SyncControllerSpec extends AnyFlatSpec with Matchers with BeforeAndAfter with MockFactory with Eventually {
+class SyncControllerSpec
+    extends AnyFlatSpec
+    with Matchers
+    with BeforeAndAfter
+    with MockFactory
+    with Eventually
+    with NormalPatience {
 
   implicit var system: ActorSystem = _
```
ten15bit: If we really need a fresh actor system for every test we could provide it via a fixture to get rid of this var.

ten15bit (approved):

src/test/scala/io/iohk/ethereum/blockchain/sync/SyncControllerSpec.scala
jvdp: I refactored it a bit, can you take a look? I couldn't find a way to keep the fixtures based on anonymous classes and add cleanup (actor system shutdown), but I think this is the next best thing. (See also https://www.scalatest.org/user_guide/sharing_fixtures )

ten15bit (approved):

PR #919, opened 2021-02-08
[ETCM-573] Add metrics on block imports

leo-bogastry:
# Description

Add metrics on block imports about importing new blocks, mined blocks, checkpoint blocks and other non specified blocks

# Proposed Solution
The import of these block is always done the same way (on BlockImporter) but the call to make the import comes from different origins.
I have added a trait called `BlockImportType` and specified the four different kinds of block imports and added this information to the methods making the block import calls, and then used that information to update the appropriate metric.

# Testing

I have updated the Grafana dashboard with the new metric

### DIFF
--- a/src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
+++ b/src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
@@ -1,8 +1,12 @@
 package io.iohk.ethereum.sync
 
+import com.typesafe.config.ConfigValueFactory
 import io.iohk.ethereum.FreeSpecBase
+import io.iohk.ethereum.metrics.{Metrics, MetricsConfig}
 import io.iohk.ethereum.sync.util.RegularSyncItSpecUtils.FakePeer
 import io.iohk.ethereum.sync.util.SyncCommonItSpec._
+import io.iohk.ethereum.utils.Config
+import io.prometheus.client.CollectorRegistry
 import monix.execution.Scheduler
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.matchers.should.Matchers
@@ -12,6 +16,12 @@ import scala.concurrent.duration._
 class RegularSyncItSpec extends FreeSpecBase with Matchers with BeforeAndAfterAll {
   implicit val testScheduler = Scheduler.fixedPool("test", 16)
 
+  override def beforeAll(): Unit = {
+    Metrics.configure(
+      MetricsConfig(Config.config.withValue("metrics.enabled", ConfigValueFactory.fromAnyRef(true)))
+    )
+  }
+
   override def afterAll(): Unit = {
     testScheduler.shutdown()
     testScheduler.awaitTermination(120.second)
@@ -20,12 +30,12 @@ class RegularSyncItSpec extends FreeSpecBase with Matchers with BeforeAndAfterAl
   "peer 2 should sync to the top of peer1 blockchain" - {
     "given a previously imported blockchain" in customTestCaseResourceM(FakePeer.start2FakePeersRes()) {
       case (peer1, peer2) =>
-        val blockNumer: Int = 2000
+        val blockNumber: Int = 2000
         for {
-          _ <- peer1.importBlocksUntil(blockNumer)(IdentityUpdate)
+          _ <- peer1.importBlocksUntil(blockNumber)(IdentityUpdate)
           _ <- peer2.startRegularSync()
           _ <- peer2.connectToPeers(Set(peer1.node))
-          _ <- peer2.waitForRegularSyncLoadLastBlock(blockNumer)
+          _ <- peer2.waitForRegularSyncLoadLastBlock(blockNumber)
         } yield {
           assert(peer1.bl.getBestBlock().hash == peer2.bl.getBestBlock().hash)
         }
@@ -96,4 +106,39 @@ class RegularSyncItSpec extends FreeSpecBase with Matchers with BeforeAndAfterAl
     }
   }
 
+  "A metric about mining a new block should be available" in customTestCaseResourceM(
+    FakePeer.start2FakePeersRes()
+  ) { case (peer1, peer2) =>
+    import MantisRegistries._
+
+    val minedMetricBefore = sampleMetric(TimerCountMetric, MinedBlockPropagation)
+    val defaultMetricBefore = sampleMetric(TimerCountMetric, DefaultBlockPropagation)
+
+    for {
+      _ <- peer1.startRegularSync()
+      _ <- peer1.mineNewBlocks(10.milliseconds, 1)(IdentityUpdate)
+      _ <- peer1.waitForRegularSyncLoadLastBlock(1)
+      _ <- peer2.startRegularSync()
+      _ <- peer2.connectToPeers(Set(peer1.node))
+      _ <- peer2.waitForRegularSyncLoadLastBlock(1)
+    } yield {
+
+      val minedMetricAfter = sampleMetric(TimerCountMetric, MinedBlockPropagation).doubleValue()
+      val defaultMetricAfter = sampleMetric(TimerCountMetric, DefaultBlockPropagation).doubleValue()
+
+      minedMetricAfter shouldBe minedMetricBefore + 1.0d
+      defaultMetricAfter shouldBe defaultMetricBefore + 1.0d
+    }
+  }
+
+  object MantisRegistries {
+    val TimerCountMetric = "app_regularsync_blocks_propagation_timer_seconds_count"
+    val DefaultBlockPropagation = "DefaultBlockPropagation"
+    val MinedBlockPropagation = "MinedBlockPropagation"
+    def sampleMetric(metricName: String, blockType: String): Double = CollectorRegistry.defaultRegistry.getSampleValue(
+      metricName,
+      Array("blocktype"),
+      Array(blockType)
+    )
+  }
 }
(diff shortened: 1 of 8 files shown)


src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
```diff
@@ -96,4 +124,27 @@ class RegularSyncItSpec extends FreeSpecBase with Matchers with BeforeAndAfterAl
     }
   }
 
+  "A metric about mining a new block should be available" in customTestCaseResourceM(
```
jvdp, line 109: Very nice that you tested the metric 👍 (Though this was perhaps also out of necessity?)

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -389,6 +400,26 @@ object BlockImporter {
   case class ResolvingMissingNode(blocksToRetry: NonEmptyList[Block]) extends NewBehavior
   case class ResolvingBranch(from: BigInt) extends NewBehavior
 
+  sealed trait BlockImportType {
```
jvdp, line 403: Do you think this explicit distinction between the types of block imports can be used to refactor other pieces of code? Perhaps stuff that is now implicit in all the message passing? (Could be a nice follow-up ticket if so.)

src/test/scala/io/iohk/ethereum/network/PeerManagerSpec.scala
```diff
@@ -310,8 +310,11 @@ class PeerManagerSpec
 
     time.advance(360000) // wait till the peer is out of the blacklist
 
-    val newRoundDiscoveredNodes = discoveredNodes + Node.fromUri(new java.net.URI(
-      "enode://a59e33ccd2b3e52d578f1fbd70c6f9babda2650f0760d6ff3b37742fdcdfdb3defba5d56d315b40c46b70198c7621e63ffa3f987389c7118634b0fefbbdfa7fd@51.158.191.43:38556?discport=38556"))
+    val newRoundDiscoveredNodes = discoveredNodes + Node.fromUri(
+      new java.net.URI(
+        "enode://a59e33ccd2b3e52d578f1fbd70c6f9babda2650f0760d6ff3b37742fdcdfdb3defba5d56d315b40c46b70198c7621e63ffa3f987389c7118634b0fefbbdfa7fd@51.158.191.43:38556?discport=38556"
+      )
+    )
```
jvdp: Random file getting formatted :P 

I'm curious: are you using the IntelliJ formatter or scalafmt, or both? Because CI should already confirm that every source file is formatted, and yet your PRs contain a bunch of these changes.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSyncMetrics.scala
```diff
@@ -0,0 +1,21 @@
...
+object RegularSyncMetrics extends MetricsContainer {
+  private final val blockPropagationTimer = "regularsync.blocks.propagation.timer"
+
+  final val MinedBlockPropagationTimer = metrics.timer(blockPropagationTimer, "class", "MinedBlockPropagation")
+  final val CheckpointBlockPropagationTimer =
+    metrics.timer(blockPropagationTimer, "class", "CheckpointBlockPropagation")
+  final val NewBlockPropagationTimer = metrics.timer(blockPropagationTimer, "class", "NewBlockPropagation")
+  final val DefaultBlockPropagationTimer = metrics.timer(blockPropagationTimer, "class", "DefaultBlockPropagation")
+
+  def recordMinedBlockPropagationTimer(time: Long): Unit = MinedBlockPropagationTimer.record(time, NANOSECONDS)
```
jvdp: Could call the argument something like `nanos` to be more explicit about the unit.

jvdp (approved):
LGTM but I always get an uneasy feeling when looking at actor code...

src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
leo-bogastry: Yes, it was. I need to understand better what was happening. I started by modifying an existing test and then created this one that is a bit more specific. I also found a bit weird that there are no tests on metrics at all

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
leo-bogastry: I have not seen enough code yet to be able to say yes or no. I do think the code in general is not very easy to reason about and I hope we can improve it in a nearby future

src/test/scala/io/iohk/ethereum/network/PeerManagerSpec.scala
leo-bogastry: I use both I think. I think that maybe different people have different settings? There should not be all this reformats indeeed

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSyncMetrics.scala
leo-bogastry: Fixed

src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
```diff
@@ -12,20 +16,44 @@ import scala.concurrent.duration._
...
+    _: String,
+    Array("class"),
+    Array("NewBlockPropagation")
+  )
+
+  val getDefaultBlockValue = CollectorRegistry.defaultRegistry.getSampleValue(
+    _: String,
+    Array("class"),
+    Array("DefaultBlockPropagation")
+  )
```
dzajkowski: maybe we could start adding helpers that make it easier to read? eg.
```scala
MantisRegistries.sampleClass("DefaultBlockPropagation")

object MantisRegistries {
  def sampleClass(className: String) = CollectorRegistry.defaultRegistry.getSampleValue(
    _: String,
    Array("class"),
    Array(className)
  )
}
```

src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
```diff
@@ -96,4 +124,27 @@ class RegularSyncItSpec extends FreeSpecBase with Matchers with BeforeAndAfterAl
     }
   }
 
+  "A metric about mining a new block should be available" in customTestCaseResourceM(
+    FakePeer.start2FakePeersRes()
+  ) { case (peer1, peer2) =>
+    val minedMetricBefore = getMinedBlockValue("app_regularsync_blocks_propagation_timer_seconds_count")
```
dzajkowski: nitpick: the strings (eg. "app_regularsync_blocks_propagation_timer_seconds_count") are repeated - maybe a Constants helper object would make sense?

src/it/scala/io/iohk/ethereum/sync/util/RegularSyncItSpecUtils.scala
```diff
@@ -116,9 +116,9 @@ object RegularSyncItSpecUtils {
       val currentWeight = bl
         .getChainWeightByHash(block.hash)
         .getOrElse(throw new RuntimeException(s"ChainWeight by hash: ${block.hash} doesn't exist"))
-      val currentWolrd = getMptForBlock(block)
+      val currentWorld = getMptForBlock(block)
```
dzajkowski, line 119: 🎉

src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
leo-bogastry: I have added the helper, the code is more clean indeed. I renamed the label name from `class` to `blocktype`

src/test/scala/io/iohk/ethereum/network/PeerManagerSpec.scala
dzajkowski: recommendation: always run `sbt pp` before pushing, that way you will get the code formatted. 
also: not sure if CI is verifying test code

dzajkowski (approved):

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -389,6 +400,26 @@ object BlockImporter {
...
+  case object CheckpointBlockImport extends BlockImportType {
+    override def recordMetric(time: Long): Unit = RegularSyncMetrics.recordImportCheckpointPropagationTimer(time)
+  }
+
+  case object NewBlockImport extends BlockImportType {
+    override def recordMetric(time: Long): Unit = RegularSyncMetrics.recordImportNewBlockPropagationTimer(time)
+  }
+
+  case object DefaultBlockImport extends BlockImportType {
+    override def recordMetric(time: Long): Unit = RegularSyncMetrics.recordDefaultBlockPropagationTimer(time)
```
ten15bit: `nanos` would also be better here (and two more just above)

ten15bit (approved):
:+1:

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
leo-bogastry: Fixed!

src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
leo-bogastry: You are right, I created vals for all the strings, they all repeated

PR #920, opened 2021-02-10
ETCM-607 skip most mining-related assertions in EthashUtilsSpec on CI for faster builds

jvdp:
This PR speeds up CI a little by changing the 3 slowest specs in a blunt way: skipping most of the assertions on CI.

Before:

Unit | Test | Time
-- | -- | --
io.iohk.ethereum.consensus.ethash.EthashUtilsSpec | Ethash should compute proof of work using cache | 323.385
io.iohk.ethereum.consensus.validators.BlockHeaderValidatorSpec | BlockHeaderValidator should validate DAO block (extra data) | 213.061
io.iohk.ethereum.blockchain.sync.SyncSchedulerSpec | SyncScheduler should sync whole trie when receiving all nodes from remote side | 160.821

After:

Unit | Test | Time
-- | -- | --
io.iohk.ethereum.consensus.validators.BlockHeaderValidatorSpec | BlockHeaderValidator should validate DAO block (extra data) | 35.233
io.iohk.ethereum.consensus.ethash.EthashUtilsSpec | Ethash should compute proof of work using cache | 27.244
io.iohk.ethereum.blockchain.sync.SyncSchedulerSpec | SyncScheduler should sync whole trie when receiving all nodes from remote side | (no longer in top 50; sub-5s)

Time taken by CI drops by about 10 minutes accordingly.

These tests were exceptionally slow on CI (as opposed to running locally) so I think there's something else going on. I'm making a follow-up ticket to investigate this further.


dzajkowski:
good find

leo-bogastry:
I feel a bit uneasy with this PR. Are we dropping tests that are not considered necessary (meaning that we over test) or just because they are being slow? I have noticed that the CI takes some 2h, so 10 min won't be very noticeable in the end and we are cutting tests? I agree that CI is slow and we should address it, but we shouldn't achieve it by cutting tests

jvdp:
Yes, I completely understand. There was some degree of over-testing that is reduced with this change. These parts of the code are not currently being worked on, but they should be to tackle the performance issues that have come to light. Once those are fixed we can probably fully enable the tests again.

dzajkowski:
@LeonorLunatech keep in mind that the skip is applied on ci only, locally the tests still run, right @jvdp ?

jvdp:
> @LeonorLunatech keep in mind that the skip is applied on ci only, locally the tests still run, right @jvdp ?

Correct

ten15bit (approved):

leo-bogastry (approved):
There are follow-up tickets on this issue right?

bsuieric (approved):

PR #921, opened 2021-02-10
[ETCM-531] Cache-based blacklist implementation [!pr]

ten15bit:
# Description

The current implementation doesn't allow the blacklist to be shared between multiple actors (e.g. fast-sync and upcoming branch resolver). This PR adds a new and thread-safe implementation that can be access and modified by multiple actors.

# Proposed Solution

Requirements of the blacklist include thread-safety, time-based expiration and size-based eviction of elements. Using a cache like [Caffeine](https://github.com/ben-manes/caffeine) provides all the required features and introduces less complexity than implementing an actor-based DIY solution.

In the future, regular sync will also be moved to this new approach. The current/old blacklist and peerlist implementations will then be removed.

* Added a cache-based blacklist implementation that is now used for fast-sync
* Added an updated support for peerlist that uses the new blacklist implementation

# Testing

* one new test for cache-based blacklist
* updated test for fast-sync


src/main/scala/io/iohk/ethereum/blockchain/sync/Blacklist.scala
```diff
@@ -0,0 +1,47 @@
...
+import io.iohk.ethereum.utils.Logger
+
+import scala.concurrent.duration.FiniteDuration
+
+import scala.jdk.CollectionConverters._
+import scala.jdk.OptionConverters._
+import scala.jdk.DurationConverters._
+
+trait Blacklist {
+  def isBlacklisted(id: BlackListId): Boolean
```
dzajkowski: nitpick: do you mind refactoring the name to `BlacklistId`?

ten15bit: sure, will do :+1:

build.sbt
```diff
@@ -111,24 +111,26 @@ lazy val node = {
     Seq(
       Dependencies.akka,
       Dependencies.akkaHttp,
-      Dependencies.json4s,
```
ten15bit, line 114: The only dependencies that were added are `scaffeine` and `guava-test`, but it's now alphabetically sorted...that's why there are so many changed lines ;)

src/main/scala/io/iohk/ethereum/blockchain/sync/PeerListSupportNg.scala
```diff
@@ -0,0 +1,71 @@
...
+import io.iohk.ethereum.network.PeerEventBusActor.PeerEvent.PeerDisconnected
+import io.iohk.ethereum.network.PeerEventBusActor.SubscriptionClassifier.PeerDisconnectedClassifier
+import io.iohk.ethereum.network.PeerEventBusActor.{PeerSelector, Subscribe, Unsubscribe}
+import io.iohk.ethereum.network.{EtcPeerManagerActor, Peer, PeerId}
+import io.iohk.ethereum.utils.Config.SyncConfig
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+
+trait PeerListSupportNg { self: Actor with ActorLogging =>
```
ten15bit, line 14: Mostly copied from `PeerListSupport` with a little bit of cleanup. `PeerListSupport` will go away eventually and then this will be renamed to `PeerListSupport`.

build.sbt
dzajkowski: thanks for the sort! 👍

src/main/scala/io/iohk/ethereum/blockchain/sync/Blacklist.scala
```diff
@@ -0,0 +1,67 @@
...
+      .expireVariably()
+      .toScala
+      .fold {
+        log.warn(
+          s"Unexpected error while adding peer [${id.value}] to blacklist using custom expiration time. Falling back to default expiration."
+        )
+        cache.put(id, reason)
+      } { varExpirationPolicy =>
+        varExpirationPolicy.put(id, reason, duration.toJava)
+      }
```
dzajkowski: a bit of codegolf:
```scala
} { _.put(id, reason, duration.toJava) }
```

src/main/scala/io/iohk/ethereum/blockchain/sync/Blacklist.scala
```diff
@@ -0,0 +1,67 @@
...
+
+object CacheBasedBlacklist {
+
+  def empty(maxSize: Int): CacheBasedBlacklist = {
+    val cache =
+      Scaffeine()
+        .expireAfter[BlacklistId, String](
+          create = (_, _) => 60.minutes,
+          update = (_, _, _) => 60.minutes,
+          read = (_, _, _) => 60.minutes
```
dzajkowski: would it make sense to have the duration here come from a config?

src/main/scala/io/iohk/ethereum/blockchain/sync/Blacklist.scala
```diff
@@ -0,0 +1,67 @@
...
+  def keys: Set[BlacklistId]
+}
+
+object Blacklist {
+  trait BlacklistId {
+    def value: String
+  }
+}
+
+final case class CacheBasedBlacklist(cache: Cache[BlacklistId, String]) extends Blacklist with Logger {
```
AnastasiiaL: Is string value (being a reason for blacklisting) ever used? Looks like we are not interested in it. Maybe for space efficiency we could use Boolean or a constant value? Or maybe there's a set implementation of cache available?

src/main/scala/io/iohk/ethereum/blockchain/sync/Blacklist.scala
ten15bit: In theory™ the duration should never be used (because we always set a custom duration), but we can move it to the config together maxElements to get rid of the magic number :+1: 

I'll create a ticket for it.

src/main/scala/io/iohk/ethereum/blockchain/sync/Blacklist.scala
ten15bit: At the moment it isn't used but I think it would be a very interesting metric to add to our dashboard. I turned the reason Strings into proper types and the different types of reasons are now represented as objects so it shouldn't have a negative impact on memory.

src/main/scala/io/iohk/ethereum/blockchain/sync/Blacklist.scala
ten15bit: Changed it to a pattern match because I found the fold not very readable.

src/main/scala/io/iohk/ethereum/blockchain/sync/Blacklist.scala
```diff
@@ -0,0 +1,161 @@
...
+  trait BlacklistId {
+    def value: String
+  }
+
+  sealed trait BlacklistReason {
+    def reasonType: BlacklistReasonType
+    def description: String
+  }
+  object BlacklistReason {
+    trait BlacklistReasonType {
```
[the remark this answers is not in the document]
ten15bit: Sealed traits with case objects are usually a better approach then enumerations which suffer from type erasure and no checks whether a match is exhaustive.

ten15bit:
> This pull request would normally be cancelled because the title contains "[!pr]"; however since the review was requested manually from the PullRequest dashboard, it was sent to our network to be reviewed. If you wish to cancel PullRequest review, you can do so manually from the [PullRequest dashboard](https://app.pullrequest.com/dash/projects/1/input-output-hk/mantis/reviews).

Which asks permissions to "act on your behalf" and view all your email addresses when all I wanna do is unsubscribe... :-1:

src/main/scala/io/iohk/ethereum/blockchain/sync/Blacklist.scala
AnastasiiaL: good point on adding a metric and seeing which reasons for blacklisting prevail. So thumbs up for introducing types for blacklisting reasons, but I think they are descriptive enough, so we don't have to keep the record for extra description. I'll leave another comment with proposed simplification

src/main/scala/io/iohk/ethereum/blockchain/sync/Blacklist.scala
```diff
@@ -0,0 +1,161 @@
...
+
+object Blacklist {
+  import BlacklistReason._
+  import BlacklistReasonType._
+
+  trait BlacklistId {
+    def value: String
+  }
+
+  sealed trait BlacklistReason {
```
AnastasiiaL, line 34: sealed trait BlacklistReason {
           def reasonType: BlacklistReasonType
           }
           
sealed trait BlacklistReasonType {
      def name: String
    }
    
 --> so then case objects of BlacklistReasonType are left almost the same, but without codes (unless there's a usage for them)
 -->  all BlacklistReasons are turned to case objects, description is discarded.
    
    
PS: just a suggestion for simplification, if you think it makes more sense to keep it like it is - I'll approve it like that

ten15bit: Thanks for the suggestion but I disagree about removing information that we currently log and that could be helpful for debugging.

src/main/scala/io/iohk/ethereum/blockchain/sync/Blacklist.scala
ten15bit: Replied on the other comment but I disagree. The name does not contain the additional information that is currently available for debugging.

AnastasiiaL (approved):

src/main/scala/io/iohk/ethereum/blockchain/sync/PeerListSupportNg.scala
```diff
@@ -0,0 +1,72 @@
...
+    handshakedPeers.filterNot { case (peerId, _) =>
+      blacklist.isBlacklisted(peerId)
+    }
+
+  def getPeerById(peerId: PeerId): Option[Peer] = handshakedPeers.get(peerId).map(_.peer)
+
+  def blacklistIfHandshaked(peerId: PeerId, duration: FiniteDuration, reason: BlacklistReason): Unit =
+    handshakedPeers.get(peerId).foreach(_ => blacklist.add(peerId, duration, reason))
+
+  private def updatePeers(peers: Map[Peer, PeerInfo]): Unit = {
```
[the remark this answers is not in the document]
ten15bit: The actor receives a message for disconnected peers and removes them.

src/test/scala/io/iohk/ethereum/blockchain/sync/CacheBasedBlacklistSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+import io.iohk.ethereum.network.PeerId
+import org.scalatest.matchers.must.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+
+import scala.concurrent.duration._
+import com.google.common.testing.FakeTicker
+import com.github.blemale.scaffeine.Scaffeine
+import java.util.concurrent.TimeUnit
+
+class CacheBasedBlacklistSpec extends AnyWordSpecLike with Matchers {
```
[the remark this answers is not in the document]
ten15bit: Thread-safety is provided by the cache implementation (i.e. caffeine). So it's like saying "let's write a test to check whether Akka is working correctly".

src/main/scala/io/iohk/ethereum/blockchain/sync/PeerListSupportNg.scala
```diff
@@ -0,0 +1,72 @@
...
+    val updated = peers.map { case (peer, peerInfo) =>
+      (peer.id, PeerWithInfo(peer, peerInfo))
+    }
+    updated.filterNot(p => handshakedPeers.keySet.contains(p._1)).foreach { case (peerId, _) =>
+      peerEventBus ! Subscribe(PeerDisconnectedClassifier(PeerSelector.WithId(peerId)))
+    }
+    handshakedPeers = updated
```
leo-bogastry, line 57: shouldn't this be `handshakedPeers = handshakedPeers + updated`?

src/test/scala/io/iohk/ethereum/blockchain/sync/CacheBasedBlacklistSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  "CacheBasedBlacklist" should {
+    "add elements and respect max number of elements" in withBlacklist(3) { blacklist =>
+      blacklist.add(peer1, 1.minute, reason)
+      blacklist.add(peer2, 1.minute, reason)
+      blacklist.add(peer3, 1.minute, anotherReason)
+      blacklist.add(peer4, 1.minute, anotherReason)
+      blacklist.add(peer5, 1.minute, reason)
+      blacklist.cache.cleanUp()
+      assert(blacklist.keys.size <= 3)
```
leo-bogastry: Any reason for a single assertion that will pass even if `blacklist.keys.size` is zero?

ten15bit: If you look at the eviction description of caffeine you'll see that it's not guaranteed to be exactly `maxElements` (so 3 in this case).

src/main/scala/io/iohk/ethereum/blockchain/sync/PeerListSupportNg.scala
ten15bit: No, `EtcPeerManagerActor.HandshakedPeers` contains all handshaked peers. So we're replacing our current list with it.

src/main/scala/io/iohk/ethereum/blockchain/sync/PeerListSupportNg.scala
```diff
@@ -0,0 +1,72 @@
...
+import io.iohk.ethereum.network.PeerEventBusActor.PeerEvent.PeerDisconnected
+import io.iohk.ethereum.network.PeerEventBusActor.SubscriptionClassifier.PeerDisconnectedClassifier
+import io.iohk.ethereum.network.PeerEventBusActor.{PeerSelector, Subscribe, Unsubscribe}
+import io.iohk.ethereum.network.{EtcPeerManagerActor, Peer, PeerId}
+import io.iohk.ethereum.utils.Config.SyncConfig
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+
+trait PeerListSupportNg { self: Actor with ActorLogging =>
```
pbvie, line 14: This trait is mostly a copy of the current `PeerListSupport`.

src/test/scala/io/iohk/ethereum/blockchain/sync/CacheBasedBlacklistSpec.scala
ten15bit: ...but I added a `> 0` :+1:

PR #924, opened 2021-02-15
ETCM-626 - remove Option.get when searching for the best block

bsuieric:
Unsafe use of Option.get causes node desync when getting new best block

### DIFF
--- a/src/main/scala/io/iohk/ethereum/consensus/ethash/EthashBlockCreator.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/ethash/EthashBlockCreator.scala
@@ -28,20 +28,24 @@ class EthashBlockCreator(
   lazy val blockchainConfig = consensus.blockchainConfig
 
   def getBlockForMining(
-      parentBlock: Block,
+      parentBlock: Option[Block],
       withTransactions: Boolean = true,
       initialWorldStateBeforeExecution: Option[InMemoryWorldStateProxy] = None
   ): Task[PendingBlockAndState] = {
-    val transactions =
-      if (withTransactions) getTransactionsFromPool else Task.now(PendingTransactionsResponse(Nil))
-    Task.parZip2(getOmmersFromPool(parentBlock.hash), transactions).map { case (ommers, pendingTxs) =>
-      blockGenerator.generateBlock(
-        parentBlock,
-        pendingTxs.pendingTransactions.map(_.stx.tx),
-        coinbase,
-        ommers.headers,
-        initialWorldStateBeforeExecution
-      )
+    val transactions = if (withTransactions) getTransactionsFromPool else Task.now(PendingTransactionsResponse(Nil))
+    parentBlock match {
+      case Some(parent) =>
+        Task.parZip2(getOmmersFromPool(parent.hash), transactions).map { case (ommers, pendingTxs) =>
+          blockGenerator.generateBlock(
+            parent,
+            pendingTxs.pendingTransactions.map(_.stx.tx),
+            coinbase,
+            ommers.headers,
+            initialWorldStateBeforeExecution
+          )
+        }
+      case None =>
+        Task.raiseError(new RuntimeException("error reading block"))
     }
   }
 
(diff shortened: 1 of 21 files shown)


src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -119,31 +119,29 @@ class LedgerImpl(
...
           }
-          importResult.foreach(measureBlockMetrics)
-          importResult
-
-        case None =>
-          Task.now(BlockImportFailed(s"Couldn't get total difficulty for current best block with hash: $hash"))
-
-      }
+        }
+      case _ => Task.now(BlockImportFailed(s"Couldn't find the current best block"))
```
dzajkowski: two notes:
- `case _ =>` can be explicit and I think it should, especially since the previous case is 18 lines higher. but basically if we can be explicit, we should.
- I think that putting a log.error here is crucial

src/main/scala/io/iohk/ethereum/jsonrpc/QAService.scala
```diff
@@ -44,11 +44,15 @@ class QAService(
   def generateCheckpoint(
       req: GenerateCheckpointRequest
   ): ServiceResponse[GenerateCheckpointResponse] = {
-    Task {
-      val hash = req.blockHash.getOrElse(blockchain.getBestBlock().hash)
-      val checkpoint = generateCheckpoint(hash, req.privateKeys)
-      syncController ! NewCheckpoint(hash, checkpoint.signatures)
-      Right(GenerateCheckpointResponse(checkpoint))
+    val hash = req.blockHash orElse blockchain.getBestBlock().map(_.hash)
```
dzajkowski: I know we don't have a style set in stone yet but can we agree that the infix notation is not good scala? :)

src/main/scala/io/iohk/ethereum/jsonrpc/EthMiningService.scala
```diff
@@ -78,26 +78,27 @@ class EthMiningService(
...
-      response
+            Right(
+              GetWorkResponse(
+                powHeaderHash = ByteString(kec256(BlockHeader.getEncodedWithoutNonce(pb.block.header))),
+                dagSeed = EthashUtils.seed(pb.block.header.number.toLong),
+                target = ByteString((BigInt(2).pow(256) / pb.block.header.difficulty).toByteArray)
+              )
+            )
+          }
+        case _ => Task.now(Left(JsonRpcError.BlockNotFound))
```
AnastasiiaL: responding "Block not found" to getWork request will be misleading as no one was requesting a block. Internal Server Error would be more appropriate IMO

src/main/scala/io/iohk/ethereum/consensus/ethash/EthashBlockCreator.scala
```diff
@@ -28,20 +28,24 @@ class EthashBlockCreator(
   lazy val blockchainConfig = consensus.blockchainConfig
 
   def getBlockForMining(
-      parentBlock: Block,
+      parentBlock: Option[Block],
```
AnastasiiaL: I don't think it makes sense to call this method with optional parentBlock as parentBlock is essential for execution. We could return an error one level higher, from the place this method is called

src/main/scala/io/iohk/ethereum/jsonrpc/EthMiningService.scala
ten15bit: Please also make sure the error is logged somewhere.

src/main/scala/io/iohk/ethereum/consensus/ethash/EthashMiner.scala
```diff
@@ -56,39 +56,46 @@ class EthashMiner(
...
+              self ! ProcessMining
+            }
+          }
+          .onErrorHandle { ex =>
+            log.error(ex, "Unable to get block for mining")
+            context.system.scheduler.scheduleOnce(10.seconds, self, ProcessMining)
+          }
+          .runAsyncAndForget
+      case None => {
+        log.error("Unable to get block for mining")
```
leo-bogastry: Can we make here explicit that `getBestBlock ` returned nothing, in this error message?

bsuieric: good point to differentiate the logs between a random exception and None

leo-bogastry (approved):
Nice to see all those `get`s removed 👍

AnastasiiaL (approved):

PR #925, opened 2021-02-15
Bug/etcm 636 [!pr]

AnastasiiaL:
# Description

None.get on bestBlock happened because of the inconsistency between cache and db. This PR introduces uni and integration tests that try to reproduce the issue. In case of it test - it's not possible to reproduce as probable cause is the thread of execution dying in the middle of updates.

Also updating one method with preferred order of execution - first saving to cache, after to db.

### DIFF
--- a/src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
+++ b/src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
@@ -1,0 +1,108 @@
+package io.iohk.ethereum.ledger
+
+import akka.testkit.TestProbe
+import akka.util.ByteString
+import cats.data.NonEmptyList
+import io.iohk.ethereum.blockchain.sync.regular.{BlockFetcher, BlockImporter}
+import io.iohk.ethereum.consensus.blocks.CheckpointBlockGenerator
+import io.iohk.ethereum.domain._
+import io.iohk.ethereum.mpt.MerklePatriciaTrie
+import io.iohk.ethereum.utils.Config.SyncConfig
+import io.iohk.ethereum.utils.Config
+import io.iohk.ethereum.Fixtures
+import monix.execution.Scheduler
+import org.scalamock.scalatest.MockFactory
+import org.scalatest.BeforeAndAfterAll
+import org.scalatest.flatspec.AsyncFlatSpecLike
+import org.scalatest.matchers.should.Matchers
+import scala.concurrent.duration._
+
+class BlockImporterItSpec extends MockFactory with TestSetupWithVmAndValidators with AsyncFlatSpecLike with Matchers with BeforeAndAfterAll {
+
+  implicit val testScheduler = Scheduler.fixedPool("test", 32)
+
+  override def afterAll(): Unit = {
+    testScheduler.shutdown()
+    testScheduler.awaitTermination(60.second)
+  }
+
+  val bl = BlockchainImpl(storagesInstance.storages)
+
+  val blockQueue = BlockQueue(bl, SyncConfig(Config.config))
+
+  val genesis = Block(
+    Fixtures.Blocks.Genesis.header.copy(stateRoot = ByteString(MerklePatriciaTrie.EmptyRootHash)),
+    Fixtures.Blocks.Genesis.body
+  )
+  val genesisWeight = ChainWeight.zero.increase(genesis.header)
+
+  bl.save(genesis, Seq(), genesisWeight, saveAsBestBlock = true)
+
+  lazy val checkpointBlockGenerator: CheckpointBlockGenerator = new CheckpointBlockGenerator
+
+  val fetcherProbe = TestProbe()
+  val ommersPoolProbe = TestProbe()
+  val broadcasterProbe = TestProbe()
+  val pendingTransactionsManagerProbe = TestProbe()
+  val supervisor = TestProbe()
+
+  override lazy val ledger: TestLedgerImpl =  new TestLedgerImpl(validators) {
+    override private[ledger] lazy val blockExecution = mock[BlockExecution]
+  }
+
+  val blockImporter = system.actorOf(
+    BlockImporter.props(
+      fetcherProbe.ref,
+      ledger,
+      bl,
+      syncConfig,
+      ommersPoolProbe.ref,
+      broadcasterProbe.ref,
+      pendingTransactionsManagerProbe.ref,
+      checkpointBlockGenerator,
+      supervisor.ref
+    ))
+
+  "BlockImporter" should "return a correct new best block after reorganising longer chain to a shorter one" in {
+
+    val genesis = bl.getBestBlock()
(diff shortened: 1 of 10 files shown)


leo-bogastry:
Is this PR dependant on https://github.com/input-output-hk/mantis/pull/925 being merged? Meaning that you will have to update the tests?

AnastasiiaL:
> Is this PR dependant on #925 being merged? Meaning that you will have to update the tests?

 mostly that's a separate effort here of investigating the underlying causes of that none.get situation. But the tests will be impacted a bit, given that the getBestBlock will return option after Bogdan's PR is merged.

src/test/scala/io/iohk/ethereum/blockchain/sync/ScenarioSetup.scala
```diff
@@ -21,8 +21,8 @@ import scala.concurrent.ExecutionContext
   * [[io.iohk.ethereum.nodebuilder.Node Node]].
   */
 trait ScenarioSetup extends StdTestConsensusBuilder with SyncConfigBuilder with StdLedgerBuilder {
-  protected lazy val executionContext = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(4))
-  protected lazy val monixScheduler = Scheduler(executionContext)
+  protected lazy val executionContextExecutor = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(4))
```
ten15bit, line 24: `ExecutionContext.fromExecutor` returns an `ExecutionContext` so I don't think this needs to be renamed.

src/main/scala/io/iohk/ethereum/ledger/BlockExecution.scala
```diff
@@ -59,7 +62,9 @@ class BlockExecution(
         .getBlockHeaderByHash(block.header.parentHash)
         .toRight(MissingParentError) // Should not never occur because validated earlier
       execResult <- executeBlockTransactions(block, parent)
-      worldToPersist = blockPreparator.payBlockReward(block, execResult.worldState)
+      worldToPersist <- Try {
+        blockPreparator.payBlockReward(block, execResult.worldState)
+      }.toEither.left.map(BlockExecutionError.MPTError(_))
```
ten15bit: with Cats this could be rewritten to
```
Either
  .catchOnly[Throwable](blockPreparator.payBlockReward(block, execResult.worldState))
  .leftMap(BlockExecutionError.MPTError.apply)
```

src/main/scala/io/iohk/ethereum/ledger/BlockExecution.scala
```diff
@@ -184,4 +189,6 @@ object BlockExecutionError {
   case object MissingParentError extends BlockExecutionError {
     override val reason: Any = "Cannot find parent"
   }
+
+  case class MPTError(reason: Throwable) extends BlockExecutionError
```
ten15bit: We should make all new case classes `final` (and also existing ones that you come across)...this will be enforced by code linting soon hopefully.

ten15bit (commented):
:+1: Only left some minor comments. Let's have a quick call today and then it's :heavy_check_mark: from my side.

src/test/scala/io/iohk/ethereum/blockchain/sync/ScenarioSetup.scala
AnastasiiaL: this was renamed, because the test suite extends 2 test setups, where each has an executionContext val resulting in "class BlockImporterItSpec inherits conflicting members"

src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
```diff
@@ -0,0 +1,196 @@
...
+
+    //returning discarded initial chain
+    blockchain.save(oldBlock2, Nil, oldWeight2, saveAsBestBlock = true)
+    blockchain.save(oldBlock3, Nil, oldWeight3, saveAsBestBlock = true)
+    blockchain.save(oldBlock4, Nil, oldWeight4, saveAsBestBlock = true)
+
+    blockImporter ! BlockFetcher.PickedBlocks(NonEmptyList.fromListUnsafe(newBranch))
+
+    Thread.sleep(200)
+    blockchain.getBestBlock().get shouldEqual newBlock3
```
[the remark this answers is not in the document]
AnastasiiaL: I fixed the descriptions of the tests so it should be easier to follow

src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
```diff
@@ -0,0 +1,196 @@
...
+  //saving initial main chain
+  blockchain.save(block1, Nil, weight1, saveAsBestBlock = true)
+  blockchain.save(oldBlock2, Nil, oldWeight2, saveAsBestBlock = true)
+  blockchain.save(oldBlock3, Nil, oldWeight3, saveAsBestBlock = true)
+  blockchain.save(oldBlock4, Nil, oldWeight4, saveAsBestBlock = true)
+
+  val oldBranch = List(oldBlock2, oldBlock3, oldBlock4)
+  val newBranch = List(newBlock2, newBlock3)
+
+  blockImporter ! BlockImporter.Start
```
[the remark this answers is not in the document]
AnastasiiaL: thanks for the suggestion

src/main/scala/io/iohk/ethereum/ledger/BlockExecution.scala
jvdp: Ooh nice!

src/main/scala/io/iohk/ethereum/ledger/BlockExecution.scala
```diff
@@ -59,7 +60,9 @@ class BlockExecution(
         .getBlockHeaderByHash(block.header.parentHash)
         .toRight(MissingParentError) // Should not never occur because validated earlier
       execResult <- executeBlockTransactions(block, parent)
-      worldToPersist = blockPreparator.payBlockReward(block, execResult.worldState)
+      worldToPersist <- Either
+        .catchOnly[Throwable](blockPreparator.payBlockReward(block, execResult.worldState))
```
jvdp: Now with cats making it easy to specify the error to catch, I'm thinking if we should specialize this to `MPTException`.

jvdp (approved):

leo-bogastry (approved):

src/main/scala/io/iohk/ethereum/ledger/BlockExecution.scala
AnastasiiaL: you tell me. Is it possible other exceptions are thrown while executing payBlockReward?

jvdp: Perhaps but not at first glance. I propose to be more conservative and only capture the ones we know about now (so `MPTException`.)

ten15bit (approved):

PR #929, opened 2021-02-22
ETCM-541 attempt UPnP port mapping to aid in peer discovery & connection

jvdp:
With this addition Mantis will try to use UPnP to open an UDP port for Kademlia / peer discovery and a TCP port for the Ethereum protocol. Can be disabled via `mantis.network.automatic-port-forwarding = false` (defaults to true.)


src/main/resources/application.conf
```diff
@@ -47,6 +47,9 @@ mantis {
       port = 9076
     }
 
+    # Try automatic port forwarding via UPnP
+    automatic-port-forwarding = true
```
leo-bogastry, line 51: Should this be on by default?

scalastyle-config.xml
```diff
@@ -1,7 +1,7 @@
 <scalastyle>
     <name>Scalastyle standard configuration</name>
     <check level="error" class="org.scalastyle.file.FileTabChecker" enabled="true"></check>
-    <check level="error" class="org.scalastyle.file.FileLengthChecker" enabled="true">
+    <check level="error" class="org.scalastyle.file.FileLengthChecker" enabled="false">
```
aakoshh, line 4: Does this need to be part of this PR? Which file is too long all of a sudden?

src/main/scala/io/iohk/ethereum/network/discovery/DiscoveryServiceBuilder.scala
```diff
@@ -17,6 +17,7 @@ import monix.eval.Task
 import monix.execution.Scheduler
 import scodec.bits.BitVector
 import scodec.Codec
+import monix.eval.instances.CatsConcurrentForTask
```
aakoshh: Looks like this isn't needed any more.

src/main/scala/io/iohk/ethereum/nodebuilder/StdNode.scala
```diff
@@ -62,6 +62,7 @@ abstract class BaseNode extends Node {
 
     startPeerManager()
 
+    startPortForwarding
```
aakoshh: Since this is side effecting, it should have `()` appended to it. I'd also recommend adding some logging somewhere so allocation can't just silently fail.

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
```diff
@@ -676,6 +681,27 @@ trait SyncControllerBuilder {
...
+      )
+      .whenA(Config.Network.automaticPortForwarding)
+      .allocated
+      .runToFuture
+
+  def stopPortForwarding(): Future[Unit] =
+    startPortForwarding.flatMap { case (_, release) =>
+      release.runToFuture
+    }
+}
```
aakoshh: It would be cool if starting and stopping could be done repeatedly, at least in theory. The lazy val can only be used once, but you could use an `AtomicReference` to set/unset it when its started/stopped.

This is what I ended up with in a similar situation where I had to use mutable state with resource allocation:

```scala
  // Reference to an allocated miner and its release task, if it has been started.
  // The release cannot be ignored or it would keep the other resources from being released.
  private val maybeMinerPromiseRef =
    new AtomicReference[Option[Promise[(ActorRef, Task[Unit])]]](None)

  private def startMiningProcess(node: Node): Unit = {
    val minerPromise = Promise[(ActorRef, Task[Unit])]()
    if (maybeMinerPromiseRef.compareAndSet(None, Some(minerPromise))) {

      val miner: Resource[Task, ActorRef] = ???

      miner.allocated.runToFuture
        .flatMap {
          case (actor, release) =>
            minerPromise.success(actor -> release)
            sendMiner(MinerProtocol.StartMining)
        }
        .recover {
          case NonFatal(ex) =>
            minerPromise.tryFailure(ex)
        }
    }
  }

  private def stopMiningProcess(): Unit = {
    sendMiner(MinerProtocol.StopMining).flatMap { _ =>
      maybeMinerPromiseRef
        .get()
        .map { minerPromise =>
          minerPromise.future.flatMap {
            case (_, release) =>
              maybeMinerPromiseRef.set(None)
              release.runToFuture
          }
        }
        .getOrElse(Future.successful(()))
    }
  }
```

aakoshh (commented):
Nice work :+1: 

Couple of comments about logging and repeatability.

src/main/resources/application.conf
jvdp: For it to be useful, I think so. Someone who wouldn't map the ports themselves probably won't be inclined to enable this manually either, right?

scalastyle-config.xml
jvdp: NodeBuilder.scala, but the limit is 800 so I wouldn't call that "all of a sudden." IMHO that is close enough to disabling this rule, so that's why this is the next logical step. ;)

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
jvdp: Seems easy enough 👍

src/main/scala/io/iohk/ethereum/nodebuilder/StdNode.scala
jvdp: With the `AtomicReference` this is also fixable :+1:

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
```diff
@@ -676,6 +681,35 @@ trait SyncControllerBuilder {
...
+          .openPorts(
+            Seq(Config.Network.Server.port),
+            Seq(discoveryConfig.port).filter(_ => discoveryConfig.discoveryEnabled)
+          )
+          .whenA(Config.Network.automaticPortForwarding)
+          .allocated
+          .runToFuture
+          .map(_._2)
+      }
+    )
```
aakoshh: I think this isn't doing the same thing as my version with the `Promise` was: `compareAndSet` takes both arguments by value, so the `Some` part will be evaluated even if the value is already set to something, and as a side effect start another port forwarder if the method is called twice in a row. Whereas with the `Promise` based version, if there already was a value, then all that happened was that we discarded a never-used `Promise` instance. 

```scala
@ import java.util.concurrent.atomic.AtomicReference 
import java.util.concurrent.atomic.AtomicReference

@ import scala.concurrent.ExecutionContext.Implicits.global 

@ import scala.concurrent.Future 

@ val ref: AtomicReference[Option[Future[Unit]]] = new AtomicReference(None) 
ref: AtomicReference[Option[Future[Unit]]] = None

@ ref.compareAndSet(None, Some(Future{ println("Hello!") })) 
res4: Boolean = true

Hello!
@ ref.compareAndSet(None, Some(Future{ println("Hello!") })) 
Hello!
res5: Boolean = false
```

See the 2nd invocation of `compareAndSet` returned `false` but still the side effect was executed. What you want is _only_ carry out the action if the return value was `true`.

scalastyle-config.xml
aakoshh: Fair enough :)

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
jvdp: Ah yeah you're absolutely right. Going to improve this.

The atomic reference might as well contain a boolean btw, with the promise being a separate value. (I dislike how side-effecty all this stuff is...)

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
aakoshh: Actually I should probably have made the stopping safer as well like this:
```scala
  private def stopMiningProcess(): Unit = {
      maybeMinerPromiseRef
        .getAndSet(None) match {
          case None => Future.unit
          case Some(promise) => promise.future.flatMap {
            case (_, release) => release.runToFuture
          }
        }
    }
  }
```
So it definitely only executes once.

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
```diff
@@ -676,6 +681,35 @@ trait SyncControllerBuilder {
...
+
+  def stopPortForwarding(): Future[Unit] =
+    portForwardingRelease.get().fold(Future.unit) {
+      _.flatMap(_.runToFuture)
+    }
```
aakoshh: This should also set the reference back to `None`, so it can be started again. See [this](https://github.com/input-output-hk/mantis/pull/929#discussion_r582019832) comment for a better way of doing it than the original proposal.

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
aakoshh: Yeah it's unfortunate. I was wondering if it should be wrapped up in a helper; so far it didn't need reuse, but it's tricky to get right. 

The reason I thought the `Promise` to be in the reference was so that it can be replaced with a fresh one upon restart. Also because I could always tell that the system isn't running by the absence of it (which you don't need here).

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
jvdp: ~I think there's a potential race condition in your original snippet: if `stopMiningProcess` is called before the promise is updated with the `release` task then this will never run. I think there's not an easy way to prevent this without locking (with eg. `synchronize`), what do you think?~

Ah no I suppose it will wait till the promise is updated instead.

aakoshh: Yes once you have the promise instance, you can call `.future` to subscribe to its completion. Just have to make sure it will happen even if there was an error.

aakoshh (approved):
Awesome, thanks for persevering with me! :checkered_flag:

scalastyle-config.xml
aakoshh: I see people do this to suppress this in one file they don't want to deal with:`
```scala
// scalastyle:off number.of.methods method.length file.size.limit
trait Blockchain
```

PR #931, opened 2021-02-25
ETCM-645: Add it-test for checkpointing

AnastasiiaL:
# Description

Integration tests added to simulate peers exchanging the checkpoints

### DIFF
--- a/src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
+++ b/src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
@@ -3,6 +3,7 @@ package io.iohk.ethereum.sync
 import com.typesafe.config.ConfigValueFactory
 import io.iohk.ethereum.FreeSpecBase
 import io.iohk.ethereum.metrics.{Metrics, MetricsConfig}
+import io.iohk.ethereum.network.PeerId
 import io.iohk.ethereum.sync.util.RegularSyncItSpecUtils.FakePeer
 import io.iohk.ethereum.sync.util.SyncCommonItSpec._
 import io.iohk.ethereum.utils.Config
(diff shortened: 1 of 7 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -329,7 +329,7 @@ class BlockImporter(
         Right(Nil)
       case UnknownBranch =>
         val currentBlock = blocks.head.number.min(startingBlockNumber)
-        val goingBackTo = currentBlock - syncConfig.branchResolutionRequestSize
+        val goingBackTo = (currentBlock - syncConfig.branchResolutionRequestSize).max(0)
```
jvdp, line 332: What is this change for? Reverting this doesn't seem to break anything. (Also goes for the `branchResolutionRequestSize = 30` changes.)

jvdp (approved):
So the tests look OK to me, superficially. (This `FakePeer` interface seems convenient.) Have you found a way to break them that indicates they're testing the right thing?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
AnastasiiaL: before the change val goingBackTo could be set to negative numbers, this shouldn't be the case as the resolution of the branches can't start from before the genesis block.

I actually had this happening during tests -> hence the fix.

As for branchResolutionRequestSize values changes in tests, I just aligned them with the value we use in application.conf.

AnastasiiaL:
> So the tests look OK to me, superficially. (This `FakePeer` interface seems convenient.) Have you found a way to break them that indicates they're testing the right thing?

not sure what you mean with breaking the tests, but changing the test setup breaks them easily

bsuieric (approved):

jvdp:
> not sure what you mean with breaking the tests, but changing the test setup breaks them easily

By disabling / breaking the feature they're supposed to test, I meant.

AnastasiiaL:
> > not sure what you mean with breaking the tests, but changing the test setup breaks them easily
> 
> By disabling / breaking the feature they're supposed to test, I meant.

checkpointing is a huuuge piece of functionality, touching a lot of layers, it's not something that can be commented out. I just added the tests that were supposed to be there long time ago

src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
```diff
@@ -76,6 +95,91 @@ class RegularSyncItSpec extends FreeSpecBase with Matchers with BeforeAndAfterAl
...
+      _ <- peer1.connectToPeers(Set(peer2.node))
+      _ <- peer1.waitForRegularSyncLoadLastBlock(length)
+    } yield {
+      assert(peer1.bl.getBestBlock().get.hash == peer2.bl.getBestBlock().get.hash)
+      assert(peer1.bl.getBestBlock().get.number == peer2.bl.getBestBlock().get.number && peer1.bl.getBestBlock().get.number == length)
+      assert(peer1.bl.getLatestCheckpointBlockNumber() == peer2.bl.getLatestCheckpointBlockNumber())
+    }
+  }
+
+  //TODO: investigate why reorganisation is not triggered after 2 nodes with conflicting branches connect
```
leo-bogastry, line 160: Have you created a ticket for this?

src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
```diff
@@ -76,6 +95,91 @@ class RegularSyncItSpec extends FreeSpecBase with Matchers with BeforeAndAfterAl
...
+      _ <- peer1.mineNewBlocks(100.milliseconds, 1)(IdentityUpdate)
+      _ <- peer2.waitForRegularSyncLoadLastBlock(blockNumber + 5)
+    } yield {
+      assert(peer1.bl.getBestBlock().get.hash == peer2.bl.getBestBlock().get.hash)
+      assert(peer1.bl.getLatestCheckpointBlockNumber() == peer2.bl.getLatestCheckpointBlockNumber())
+    }
+  }
+
+  "peers should chose the branch with a checkpoint discarding blocks that come after the checkpoint" in customTestCaseResourceM(
+    FakePeer.start2FakePeersRes()
```
leo-bogastry, line 141: I'm missing something understanding this test. I was expecting some mined blocks after the checkpointing creation

src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
AnastasiiaL: yes, its ETCM-722

src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
AnastasiiaL: the test describes a scenario when a node with 30 blocks, receives a signal of generating a checkpoint at height 25. Since the chain can't be amended, it means that 25 blocks are kept as is, 5 blocks are discarded, and 1 checkpoint at height 26 is generated

leo-bogastry (approved):

PR #933, opened 2021-03-01
ETCM-[165, 166]: Publish the RLP and Crypto libraries to Sonatype.

aakoshh:
# Description

Add steps to the CI pipeline to publish the `mantis-rlp` and `mantis-crypto` artifacts to Sonatype so they can be reused in other projects.

On the `develop` branch, the version is expected to be `-SNAPSHOT`, so it can overwrite previous versions.
On the `master` branch the version is set to release by removing `-SNAPSHOT`, so it can only be released once.

After that it's expected that `develop` should have a bump in version, but that's currently not part of the automation.

# Proposed Solution

This is based of what we're currently [doing](https://github.com/input-output-hk/scalanet/pull/119) in `scalanet`: instead of a version file I added a `version.sbt` file which could be used by [sbt-release](https://github.com/sbt/sbt-release) to bump and commit a new version, as described [here](https://medium.com/rahasak/publish-scala-library-project-to-maven-central-with-sonatype-d7edaa67d275), but currently it's just updated from the `publish.sh` script for the release, it's not committed and pushed back to Github. See the SBT guide for the [Sonatype settings](https://www.scala-sbt.org/1.x/docs/Using-Sonatype.html).

Ended up using this plugin: https://github.com/olafurpg/sbt-ci-release
What needs to be published is set int the `publish.sh` script using environment variables [used by the plugin](https://github.com/olafurpg/sbt-ci-release/blob/main/plugin/src/main/scala/com/geirsson/CiReleasePlugin.scala#L162-L163).

Its default assumption is that it will do a _release_ if the build is tagged in git, and push a snapshot if it's just a normal commit. Tagging is faked in the publish script at the moment, but if the release strategy were to change, it would be easy to modify it that way.

To make it work we added the following environment variables to the build: 
* `GPG_KEY`: GPG key used to sign the published artifacts, in Base64 format (see the `sbt-ci-release` plugin for the command to produce it)
* `GPG_KEY_ID`: ID of the GPG key, to tell the SBT plugin which key to use
* `GPG_PASSPHRASE`: Pass phrase to import the GPG key
* `OSS_USERNAME`: Sonatype username
* `OSS_PASSWORD`: Sonatype password

Added `-SNAPSHOT` to the version, which is stored in `version.sbt`. 

# Testing

I temporarily left the name of this branch as a conditional, which resulted in the `SNAPSHOT` versions being [published](https://oss.sonatype.org/#nexus-search;gav~io.iohk~mantis-*~~~).

See the [logs](https://buildkite.com/input-output-hk/mantis/builds/2083#74b5456c-ceb6-4798-94be-437fc9aad7f8) on Buildkite.


aakoshh:
@KonradStaniec can you add the required credentials to the build? I don't know the values.

aakoshh:
It would be nice if the `pipeline.nix` could use `wait` steps so publishing only happens if everything worked fine. I should see if that works.

aakoshh:
Apparently we should try to get a Mantis specific Sonatype account and PGP key.

aakoshh:
@dzajkowski @KonradStaniec this should be ready for approval now.

aakoshh:
@KonradStaniec kindly reminded me that we'd need this published for Scala 2.13 as well as 2.12, which I completely forgot.

aakoshh:
I wonder how nix likes cross building. In any case I noticed that my PR has a `./nix/pkgs/mantis.nix` which looks like it's currently missing on `develop` (there's one at `./nix/mantis.nix` instead); I just did what the README told me to update the dependencies, not sure if it's still correct.

.buildkite/publish.sh
```diff
@@ -0,0 +1,81 @@
...
+gpg --version
+
+
+# The build agent might have this key from before.
+GPG_EXISTS=$(gpg --list-keys "$GPG_KEY_ID" >&2 && echo "yes" || echo "no")
+
+if [[ "$GPG_EXISTS" == "no" ]]; then
+	echo "$GPG_KEY" | base64 --decode | gpg --batch --import
+fi
+# Local testing showed that without this the SBT plugin got "Bad passphrase".
```
aakoshh, line 22: It turns out it's because without this the `--pinentry-mode loopback` doesn't work. When I removed that from the command the plugin printed, and tried it in the console, it worked with the `--passphrase` flag, so it's not an issue of surrounding with `'` or having a bad password.

bytes/src/main/scala/io/iohk/ethereum/utils/ByteStringUtils.scala
```diff
@@ -61,7 +61,7 @@ object ByteStringUtils {
 
   def concatByteStrings(elements: Iterator[ByteStringElement]): ByteString = {
     val builder = new mutable.ArrayBuilder.ofByte
-    elements.foreach(el => builder.addAll(el.asByteArray))
+    elements.foreach(el => builder ++= el.asByteArray)
```
dzajkowski, line 64: looks like an unrelated change. intentional?

version.sbt
```diff
@@ -0,0 +1,6 @@
+// NOTE: This is replaced with `sed` during release,
+// but it could also be removed, and be determined
+// based on `git` tags by https://github.com/dwijnand/sbt-dynver,
+// which is a dependency of `sbt-ci-release`.
+
+version in ThisBuild := "3.2.1-SNAPSHOT"
```
dzajkowski, line 6: won't this result in having the libraries locked with the version of mantis?

.buildkite/publish.sh
```diff
@@ -0,0 +1,81 @@
...
+function release {
+    SCALA_VERSION=$1
+    export SCAPEGOAT_VERSION=$2
+
+    sbt "++ $SCALA_VERSION ; ci-release"
+}
+
+function releaseAll {
+    release 2.12.10 1.4.5
+    release 2.13.4 1.4.7
```
dzajkowski, line 58: oh this duplication with build.sbt will come back to bite us....

nix/pkgs/mantis.nix
```diff
@@ -0,0 +1,89 @@
+{ src
```
dzajkowski: I think that this is a duplicated file and should be dropped. @manveru wdyt?

.buildkite/publish.sh
jvdp: It shouldn't be too hard to move more of this logic into sbt.

bytes/src/main/scala/io/iohk/ethereum/utils/ByteStringUtils.scala
aakoshh: This change was necessary because of cross publishing. We want to use these libraries from Scala 2.12 as well as Scala 2.13, and 2.12 doesn't have `addAll`.

version.sbt
aakoshh: It will, yeah. I think that at least is easy to understand in the context of a mono-repo, although it obscures semantic versioning of these libraries. If we wanted the version them separately then probably it would be better to move them out of Mantis into their own repository. Then we could strip them of their Akka dependency, enhance the API of the RLP type classes to not raise exceptions. Right now we just want to make a stop to copy pasting the RLP code between projects, and just ride the coattails of Mantis. Happy to hear your thoughts on versioning in the long term.

.buildkite/publish.sh
aakoshh: You'd think so, right! I spent hours trying to figure out with a code like this:

```
scapegoatVersion in ThisBuild := (scalaBinaryVersion.value match {
  case "2.12" => "1.4.5"
  case "2.13" => println("Still 2.13"); "1.4.7"
})
```
I start `sbt`, type `++ 2.12.10` and see "Still 2.13" printed every time, then I type `scalaBinaryVersion` and it says `2.12.10`, really frustrating.

nix/pkgs/mantis.nix
aakoshh: NB: the README and the `update-nix.sh` still create and refer to this.

.buildkite/publish.sh
aakoshh: If you want to give it a go, you should be able to try wit `sbt +compile` for example (or at least `sbt +publishLocal`), the `+` should trigger cross building, and then it fails because it tries to grab the 1.4.7 version for 2.12.10, even when I moved the `scapegoatVersion` under `node`. 

If I removed `in ThisBuild` then it wants to grab 1.0.0. No idea what I'm missing, this [should work](https://stackoverflow.com/questions/41403305/in-sbt-how-can-i-cross-build-with-dependencies-that-are-not-required-in-one-ver).

nix/pkgs/mantis.nix
manveru: I fixed a bunch of things in the `fix-update-nix` branch, so once that's merged you can rebase and run the script again.

aakoshh: @manveru I pulled the latest develop and tried to run the updated script, but get this error:
```console
aakoshh@af-dp:~/projects/iohk/mantis$ ./update-nix.sh 
Determining new sha for sbt build, this can take several minutes to do a 'sbt compile'
error: don't know what to do with argument '.#mantis.deps.outputHash'
Try 'nix --help' for more information.
aakoshh@af-dp:~/projects/iohk/mantis$ nix --version
nix (Nix) 2.3.10
```
Do you know what I'm doing wrong?

aakoshh: The build [failed](https://buildkite.com/input-output-hk/mantis/builds/2289#a8cc967c-bff2-4adc-9a5e-e678435158d1) as well in an unexpected way:
```

Determining new sha for sbt build, this can take several minutes to do a 'sbt compile'
--
  | error: value is null while a list was expected
  |  
  | at /nix/store/aaw01dbdp547458996y474ggivd4b7q6-source/nix/mantis.nix:24:6:
  |  
  | 23\|     captures = builtins.match ''.*version := "([^"]+)".*'' buildSbt;
  | 24\|   in builtins.elemAt captures 0;
  | \|      ^
  | 25\|
  | (use '--show-trace' to show detailed location information)
  | calculating hash failed!
```

aakoshh: Added comments to the README about the need to install the Nix Flakes feature and updated the script to point at `version.sbt`.

version.sbt
dzajkowski: I'll keep in mind that we should make this less dependent, thanks

dzajkowski (approved):
Thanks for aligning with build changes!

KonradStaniec (approved):

PR #934, opened 2021-03-04
[Kaizen] Add scalafmt some rewrite rules 

dzajkowski:
# Description

Scalafmt rewrite rules applied: AvoidInfix, RedundantBraces, SortImports, SortModifiers


jvdp:
What about `RedundantParens`?

Also, I am partial to `trailingCommas` to minimize diff sizes (but then the `align.preset=some` accomplishes the opposite...)

jb-lunatech:
- A follow-up/ related work would be to ensure that CICD enforces style checks: I noticed that _scalafmt_ has not been applied to all submodules. Ex: `bytes/scalafmtAll` results in a change of `Hex.scala`. 
- Semi-related suggestion: it would be good to create a `style-all` similiar to `compile-all` that can be used in `pp` command alias?

.scalafix.conf
```diff
@@ -0,0 +1,17 @@
...
+  NoAutoTupling
+  NoValInForComprehension
+  OrganizeImports
+  ProcedureSyntax
+  RemoveUnused
+]
+
+OrganizeImports {
+  # Align with IntelliJ IDEA so that they don't fight each other
+  groupedImports = Merge
```
jvdp, line 12: After dealing with a couple merge conflicts: can we _not_ do grouped imports? It's much easier on diffs to keep each on a separate line. If it becomes too much you can always use the wildcard.

PR #935, opened 2021-03-05
[ETCM-689] Update state sync and pivot block selector to use new blacklist

ten15bit:
# Description

Updates all remaining fast-sync components to use the new cache-based blacklist.

# Testing

* since the blacklist is now shared between fast-sync and pivot block selector, one test had to be updated because otherwise `peer1` wouldn't be considered for pivot block selection because it had previously sent us an invalid response

### DIFF
--- a/src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
+++ b/src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
@@ -21,7 +21,12 @@ import org.scalatest.matchers.should.Matchers
 
 import scala.concurrent.duration._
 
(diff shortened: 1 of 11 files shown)


src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
```diff
@@ -21,7 +21,12 @@ import org.scalatest.matchers.should.Matchers
 
 import scala.concurrent.duration._
 
-class BlockImporterItSpec extends MockFactory with TestSetupWithVmAndValidators with AsyncFlatSpecLike with Matchers with BeforeAndAfterAll {
```
ten15bit, line 24: nothing has changed here...reformat caused by running `sbt pp`

leo-bogastry (approved):

PR #937, opened 2021-03-11
[ETCM-594] Improve peer discovery algorithm

leo-bogastry:
# Description
Improves peer discovery algorithm by triggering the discovery of new nodes only after finishing processing the current list of new known nodes

# Proposed Solution
Instead of having a fixed and recurrent schedule triggering the discovery of new nodes, the discovery is triggered at specific moments:
- If no bootstrap / saved nodes are found
- If no new discovery nodes are found
- After processing all current new discovery nodes

### DIFF
--- a/src/main/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManager.scala
+++ b/src/main/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManager.scala
@@ -190,7 +190,6 @@ class PeerDiscoveryManager(
 
     maybeDiscoveredNodes
       .map(_ ++ alreadyDiscoveredNodes)
-      .map(_.filterNot(isLocalNode))
       .map(DiscoveredNodesInfo(_))
   }
 
--- a/src/it/scala/io/iohk/ethereum/sync/util/CommonFakePeer.scala
+++ b/src/it/scala/io/iohk/ethereum/sync/util/CommonFakePeer.scala
@@ -24,7 +24,15 @@ import io.iohk.ethereum.network.p2p.Message.Version
 import io.iohk.ethereum.network.p2p.messages.ProtocolNegotiator
 import io.iohk.ethereum.network.rlpx.AuthHandshaker
 import io.iohk.ethereum.network.rlpx.RLPxConnectionHandler.RLPxConfiguration
-import io.iohk.ethereum.network.{EtcPeerManagerActor, ForkResolver, KnownNodesManager, PeerEventBusActor, PeerManagerActor, PeerStatisticsActor, ServerActor}
+import io.iohk.ethereum.network.{
+  EtcPeerManagerActor,
+  ForkResolver,
+  KnownNodesManager,
+  PeerEventBusActor,
+  PeerManagerActor,
+  PeerStatisticsActor,
+  ServerActor
+}
 import io.iohk.ethereum.nodebuilder.PruningConfigBuilder
 import io.iohk.ethereum.security.SecureRandomBuilder
 import io.iohk.ethereum.sync.util.SyncCommonItSpec._
--- a/src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
+++ b/src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
@@ -55,7 +55,8 @@ class PeerManagerActor(
   import PeerManagerActor._
   import akka.pattern.pipe
 
-  val triedNodes: mutable.Set[ByteString] = lruSet[ByteString](maxBlacklistedNodes)
+  private[network] val triedNodes: mutable.Set[ByteString] = lruSet[ByteString](maxBlacklistedNodes)
+  private[network] val remainingNodesToConnect: mutable.Set[ByteString] = mutable.Set[ByteString]()
 
   implicit class ConnectedPeersOps(connectedPeers: ConnectedPeers) {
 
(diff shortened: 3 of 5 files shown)


src/main/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManager.scala
```diff
@@ -190,7 +190,6 @@ class PeerDiscoveryManager(
 
     maybeDiscoveredNodes
       .map(_ ++ alreadyDiscoveredNodes)
-      .map(_.filterNot(isLocalNode))
```
aakoshh, line 193: Why is it better to return the local node as a potential peer?

leo-bogastry: I removed this because that filtering was happening twice

aakoshh: Hm, I think it wasn't redundant, though, because [getting nodes from the discovery library](https://github.com/input-output-hk/mantis/blob/b81948c06cc1abaeaf6dc59c6eebfc8f56e8a22f/src/main/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManager.scala#L186-L188) will have it. 

You can remove it, just be aware to skip it in the `PeerManagerActor`; maybe it's cheaper to filter it there.

aakoshh: Although in that case it would make sense to remove it completely from this file.

leo-bogastry:
I am currently testing this on Mordor, with the current configuration the discovery of peers is way more intensive and completely pollutes the logs and seem to delay synching with the network.
This needs to be better thought of

aakoshh:
@LeonorLunatech I'm sorry about the log pollution. We realised there was a particularly bad case where there was no other node to discover and it went into a tight loop. @KonradStaniec added some logic in another program where we use discovery to limit the frequency of random lookups, perhaps that's what ails you as well. 

His update added this bit to the `PeerDiscoveryManager`:

```scala
      // Simple rate limiting approach so that:
      // 1. There is minimal time between lookups - minDelayBetweenRandomLookups
      // 2. In case of initially empty network (no bootstrap nodes) delay grows to maxDelayBetweenRandomLookups
      //    If for some reason other nodes in network find us, delay is set back to minDelay
      delayRef <- Resource.liftF(Ref.of[Task, FiniteDuration](discoveryConfig.minDelayBetweenRandomLookups))

      // Create an Iterant (like a pull-based Observable) that repeatedly performs a random lookup
      // (grabbing kademlia-bucket-size items at a time) and flattens the results. It will automatically
      // perform further lookups as the items are pulled from it.
      randomNodes = Iterant
        .repeatEvalF {
          for {
            currentDelay <- delayRef.get
            peers <- Task.defer(service.lookup(randomNodeId)).delayResult(currentDelay)
            _ <- if (networkIsEmpty(peers)) {
              delayRef.set(discoveryConfig.maxDelayBetweenRandomLookups.min(currentDelay * 2))
            } else if (currentDelay != discoveryConfig.minDelayBetweenRandomLookups) {
              delayRef.set(discoveryConfig.minDelayBetweenRandomLookups)
            } else {
              Task.unit
            }
          } yield peers
        }
        .flatMap(ns => Iterant.fromList(ns.toList))
        .map(toNode)
        .filter(!isLocalNode(_))
```

With this method to check whether the network is empty:
```scala
  private def networkIsEmpty(nodes: SortedSet[io.iohk.scalanet.discovery.ethereum.Node]): Boolean = {
    val setSize = nodes.size
    setSize == 0 || (setSize == 1 && nodes.head.id == PublicKey(BitVector(localNodeId)))
  }
```

leo-bogastry:
@aakoshh What you describe here seem to be quite opposite to what is described on the ticket I picked up.
The one I picked up mentions having new discovered peers starting to be processed before the processing of the previously discovered batch of peers has ended, and adjust this situation.

src/main/scala/io/iohk/ethereum/network/discovery/PeerDiscoveryManager.scala
leo-bogastry: `alreadyDiscoveredNodes` also filters the localNode, that is why I removed it

aakoshh: True, however `maybeDiscoveredNodes` can still have it. Perhaps you can modify that like so:
```scala
    val maybeDiscoveredNodes: Task[Set[Node]] =
      maybeDiscoveryService.fold(Task.pure(Set.empty[Node])) {
        _.getNodes.map { nodes =>
          nodes.map(toNode).filterNot(isLocalNode)
        }
      }
```

aakoshh:
Okay, I just wanted to mention this in relation to intense logging. We had to regulate the frequency of lookups, that's all.

I'm not sure about the batches, I thought it only takes what it currently can to top up to the maximum number of connections, e.g. if you want to connect to max 50, and currently you have 35 established or pending outgoing connections, then it will ask the discovery module, get, say, 1000 peers, pick 15 new from them to try to connect to, and discard the rest. So it doesn't process the whole previously discovered batch of peers, it always gets all of them and selects a few. 

This is why I added the random lookups: once it was quickly done ruling out the 15 it chose (or the previous ones), it took a long time for another attempt. The random lookups replaced connections one by one as they were ruled out.

PR #939, opened 2021-03-16
[Kaizen] Add build githash to startup log

dzajkowski:
# Description

title says it all


dzajkowski:
it will fail on nix version, will update it after it runs

ten15bit (approved):
:+1:

leo-bogastry (approved):

PR #940, opened 2021-03-16
etcm-678 etcm-660 fix for removing chain after the node restart

bsuieric:
# Description
Adding check validating if the block exists in chain - Using any of getXXXByHash is not always accurate - after restart the best block is often lower than before restart. This change is trying to resolve the branch and checking the existence of the block when trying to import blocks

### DIFF
--- a/src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
+++ b/src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
@@ -6,12 +6,15 @@ import cats.data.NonEmptyList
 import io.iohk.ethereum.blockchain.sync.regular.BlockImporter.NewCheckpoint
 import io.iohk.ethereum.blockchain.sync.regular.{BlockFetcher, BlockImporter}
 import io.iohk.ethereum.checkpointing.CheckpointingTestHelpers
+import io.iohk.ethereum.consensus.{GetBlockHeaderByHash, GetNBlocksBack}
 import io.iohk.ethereum.consensus.blocks.CheckpointBlockGenerator
+import io.iohk.ethereum.consensus.ethash.validators.{OmmersValidator, StdOmmersValidator}
+import io.iohk.ethereum.consensus.validators.Validators
 import io.iohk.ethereum.domain._
 import io.iohk.ethereum.mpt.MerklePatriciaTrie
 import io.iohk.ethereum.utils.Config.SyncConfig
 import io.iohk.ethereum.utils.Config
-import io.iohk.ethereum.{Fixtures, ObjectGenerators, crypto}
+import io.iohk.ethereum.{Fixtures, Mocks, ObjectGenerators, crypto}
 import io.iohk.ethereum.ledger.Ledger.BlockResult
 import monix.execution.Scheduler
 import org.scalamock.scalatest.MockFactory
@@ -61,6 +64,18 @@ class BlockImporterItSpec
       noEmptyAccounts = false,
       ethCompatibleStorage = true
     )
+
+  override protected lazy val successValidators: Validators = new Mocks.MockValidatorsAlwaysSucceed {
+    override val ommersValidator: OmmersValidator = (
+        parentHash: ByteString,
+        blockNumber: BigInt,
+        ommers: Seq[BlockHeader],
+        getBlockHeaderByHash: GetBlockHeaderByHash,
+        getNBlocksBack: GetNBlocksBack
+    ) =>
+      new StdOmmersValidator(blockchainConfig, blockHeaderValidator)
+        .validate(parentHash, blockNumber, ommers, getBlockHeaderByHash, getNBlocksBack)
+  }
 
   override lazy val ledger = new TestLedgerImpl(successValidators) {
     override private[ledger] lazy val blockExecution =
@@ -153,14 +168,19 @@ class BlockImporterItSpec
     blockchain.getBestBlock().get shouldEqual newBlock3
   }
 
-  it should "return Unknown branch, don't start reorganisation in case of PickedBlocks with block that has a parent who is not present in the chain " in {
+  it should "return Unknown branch, don't start reorganisation(therefore no block/ommer validation) in case of PickedBlocks with block that has a parent who is not present in the chain(or ommer not in  chain)" in {
     val newcomerBlock4: Block =
       getBlock(genesisBlock.number + 4, difficulty = 104, parent = oldBlock3.header.hash)
     val newcomerWeight4Duplicate = oldWeight3.increase(newcomerBlock4.header)
 
     //Block n5 with oldBlock4 as parent
     val newComerBlock5WithOldBlock4Parent: Block =
-      getBlock(genesisBlock.number + 5, difficulty = 108, parent = oldBlock4.header.hash)
+      getBlock(
+        genesisBlock.number + 5,
+        difficulty = 108,
+        parent = oldBlock4.header.hash,
+        ommers = Seq(oldBlock4.header)
+      )
 
     blockchain.save(oldBlock2, Nil, oldWeight2, saveAsBestBlock = true)
     blockchain.save(oldBlock3, Nil, oldWeight3, saveAsBestBlock = true)
@@ -169,7 +189,7 @@ class BlockImporterItSpec
     blockchain.saveBestKnownBlocks(blockchain.getBestBlockNumber() - 1)
     blockchain.save(newcomerBlock4, Nil, newcomerWeight4Duplicate, saveAsBestBlock = true)
 
-    //this is not reorganising anymore until oldBlock4(not part of the chain anymore) // but resolveBranch is returning UnknownBranch
+    //not reorganising anymore until oldBlock4(not part of the chain anymore), no block/ommer validation when not part of the chain, resolveBranch is returning UnknownBranch
     blockImporter ! BlockFetcher.PickedBlocks(NonEmptyList.fromListUnsafe(List(newComerBlock5WithOldBlock4Parent)))
 
     Thread.sleep(200)
--- a/src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
+++ b/src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
@@ -153,6 +153,29 @@ class BlockImporterItSpec
     blockchain.getBestBlock().get shouldEqual newBlock3
   }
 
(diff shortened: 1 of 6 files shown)


src/test/scala/io/iohk/ethereum/ledger/LedgerTestSetup.scala
```diff
@@ -393,15 +393,15 @@ trait MockBlockchain extends MockFactory { self: TestSetupWithVmAndValidators =>
       .once()
   }
 
-  def setHeaderByHash(hash: ByteString, header: Option[BlockHeader]): CallHandler1[ByteString, Option[BlockHeader]] =
-    (blockchain.getBlockHeaderByHash _).expects(hash).returning(header)
+  def setHeaderInChain(hash: ByteString, result: Boolean = true): CallHandler1[ByteString, Boolean] = {
```
AnastasiiaL: brackets are not needed, and this method is not used anywhere

bsuieric: it is used in BranchResolutionSpec

AnastasiiaL: ah right, I was looking at the line 404 removal. do you know why it's no longer called there?

bsuieric: now it's being called within the specific tests, so the steps taken to solve it are visible

AnastasiiaL: clear, only brackets left

AnastasiiaL (approved):

src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
```diff
@@ -153,6 +168,34 @@ class BlockImporterItSpec
     blockchain.getBestBlock().get shouldEqual newBlock3
   }
 
+  it should "return Unknown branch, don't start reorganisation(therefore no block/ommer validation) in case of PickedBlocks with block that has a parent who is not present in the chain(or ommer not in  chain)" in {
```
leo-bogastry: can't this test description be simplified?

bsuieric: return Unknown branch, in case of PickedBlocks with block that has a parent that's not in the chain?

leo-bogastry: Sounds better yes :)

src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
```diff
@@ -62,6 +65,18 @@ class BlockImporterItSpec
...
+    override val ommersValidator: OmmersValidator = (
+        parentHash: ByteString,
+        blockNumber: BigInt,
+        ommers: Seq[BlockHeader],
+        getBlockHeaderByHash: GetBlockHeaderByHash,
+        getNBlocksBack: GetNBlocksBack
+    ) =>
+      new StdOmmersValidator(blockchainConfig, blockHeaderValidator)
+        .validate(parentHash, blockNumber, ommers, getBlockHeaderByHash, getNBlocksBack)
+  }
```
dzajkowski, line 79: I know it's a test but what's the gain of `val` instead of `def`?

src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
```diff
@@ -153,6 +168,34 @@ class BlockImporterItSpec
     blockchain.getBestBlock().get shouldEqual newBlock3
   }
 
+  it should "return Unknown branch, in case of PickedBlocks with block that has a parent that's not in the chain" in {
+    val newcomerBlock4: Block =
+      getBlock(genesisBlock.number + 4, difficulty = 104, parent = oldBlock3.header.hash)
+    val newcomerWeight4Duplicate = oldWeight3.increase(newcomerBlock4.header)
+
+    //Block n5 with oldBlock4 as parent
+    val newComerBlock5WithOldBlock4Parent: Block =
```
dzajkowski: maybe it would make sense to propose a shorthand for those and stick to it? eg.
`newBlock5ParentOldBlock4`?
`newcomerBlock4` would be `newBlock4ParentOldBlock3`

just a thought

src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
```diff
@@ -153,6 +168,34 @@ class BlockImporterItSpec
...
+    blockchain.save(oldBlock3, Nil, oldWeight3, saveAsBestBlock = true)
+    blockchain.save(oldBlock4, Nil, oldWeight4, saveAsBestBlock = true)
+    // simulation of node restart
+    blockchain.saveBestKnownBlocks(blockchain.getBestBlockNumber() - 1)
+    blockchain.save(newcomerBlock4, Nil, newcomerWeight4Duplicate, saveAsBestBlock = true)
+
+    //not reorganising anymore until oldBlock4(not part of the chain anymore), no block/ommer validation when not part of the chain, resolveBranch is returning UnknownBranch
+    blockImporter ! BlockFetcher.PickedBlocks(NonEmptyList.fromListUnsafe(List(newComerBlock5WithOldBlock4Parent)))
+
+    Thread.sleep(200)
```
dzajkowski: that hurts. is there a plan to get rid of `Thread.sleep`?

src/test/scala/io/iohk/ethereum/domain/BlockchainSpec.scala
```diff
@@ -44,6 +44,15 @@ class BlockchainSpec extends AnyFlatSpec with Matchers with ScalaCheckPropertyCh
     assert(validBlock == block.get)
   }
 
+  it should "be able to do strict check of block existence in the chain" in new EphemBlockchainTestSetup {
+    val validBlock = Fixtures.Blocks.ValidBlock.block
+    blockchain.save(validBlock, Seq.empty, ChainWeight(100, 100), saveAsBestBlock = true)
+    assert(blockchain.isInChain(validBlock.hash))
+    // simulation of node restart
+    blockchain.saveBestKnownBlocks(validBlock.header.number - 1)
+    assert(!blockchain.isInChain(validBlock.hash))
```
dzajkowski: curious: why not use scala test matchers?

src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
bsuieric: val is the only option because it's overriding a val from MockValidatorsAlwaysSucceed

src/test/scala/io/iohk/ethereum/ledger/BranchResolutionSpec.scala
```diff
@@ -78,6 +77,7 @@ class BranchResolutionSpec
     "correctly handle a branch that goes up to the genesis block" in new BranchResolutionTestSetup {
       val headers = genesisHeader :: getChainHeadersNel(1, 10, genesisHeader.hash)
 
+      setHeaderInChain(genesisHeader.parentHash, result = false)
```
dzajkowski, line 80: the genesis block has a parent hash? O.o

src/test/scala/io/iohk/ethereum/ledger/LedgerTestSetup.scala
```diff
@@ -400,16 +400,14 @@ trait MockBlockchain extends MockFactory { self: TestSetupWithVmAndValidators =>
       .once()
   }
 
-  def setHeaderByHash(hash: ByteString, header: Option[BlockHeader]): CallHandler1[ByteString, Option[BlockHeader]] =
-    (blockchain.getBlockHeaderByHash _).expects(hash).returning(header)
+  def setHeaderInChain(hash: ByteString, result: Boolean = true): CallHandler1[ByteString, Boolean] =
```
dzajkowski, line 403: looking at `setGenesisHeader` why is `setHeaderInChain` returning `CallHandler1[ByteString, Boolean]` and the former is returning `Unit`?

src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
```diff
@@ -153,6 +168,34 @@ class BlockImporterItSpec
...
+    blockchain.getBestBlock().get shouldEqual newcomerBlock4
```
dzajkowski: there is a contruct in scala test called `eventually` that takes a timeout and it polls a give method for an expected outcome - maybe that would be better than an explicit `sleep` step?

src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
bsuieric: this can be removed from this test, I will check with @AnastasiiaL for the possibility of removing it from all tests

src/test/scala/io/iohk/ethereum/domain/BlockchainSpec.scala
bsuieric: there are a bunch of tests like this, will change them

src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
dzajkowski: gotcha, thanks

src/test/scala/io/iohk/ethereum/ledger/BranchResolutionSpec.scala
bsuieric: it doesn't, when trying to resolve the branch, it goes up until the genesis header and returns false, meaning it reached the genesis block. The genesis block is just a .copy(number = 0) of a valid block

dzajkowski (approved):
added a few nitpicks but overall looks ok

src/test/scala/io/iohk/ethereum/ledger/LedgerTestSetup.scala
bsuieric: setHeaderInChain expects 1 parameter while setGenesisheader has no parameters

src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
bsuieric: eventually is pretty neat :)

PR #941, opened 2021-03-16
[bugfix] Fixes a bug that caused miners to assume new epoch too early

ten15bit:
# Description

A recent refactoring introduced a small but critical bug in `EthashMiner` that caused miners to assume a new epoch one block too early and as a consequence they were unable to mine the final block of the previous epoch and got stuck forever.

# Proposed Solution / Important Changes Introduced

Removed the `+ 1` added to the pending block. It was previously required because it was added to the parent block, but the header number in the pending block is already incremented.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/consensus/ethash/EthashMiner.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/ethash/EthashMiner.scala
@@ -62,7 +62,7 @@ class EthashMiner(
           .getBlockForMining(blockValue)
           .map {
             case PendingBlockAndState(PendingBlock(block, _), _) => {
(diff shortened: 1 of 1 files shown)


AnastasiiaL (approved):

dzajkowski (approved):

dzajkowski:
would it be possible to cover this by some unit test so that we don't repeat the same error in the future?

leo-bogastry (approved):

bsuieric (approved):

ten15bit:
> would it be possible to cover this by some unit test so that we don't repeat the same error in the future?

yes, will be added in a separate PR to get the fix deployed to testnet immediately

PR #942, opened 2021-03-16
Feature/etcm 655

AnastasiiaL:
# Description

Stable changes from checkpointing:
- delayed fetching of new blocks from peers after the checkpoint was received internally, to give it time to get propagated without starting reorganisations right away
- Extend `checkpointing_getLatestBlock` RPC call with the parentBlock hash to only return the descendants of the previous checkpoint, update response type to exclude the situation when the hash returned without bockNumber of vice versa
- Update the size of the block queue where fetcher keeps all known forks in case those forks become main chain
- Enable logging of dead letters

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
@@ -267,7 +267,7 @@ class BlockFetcher(
   }
 
   private def handlePossibleTopUpdate(state: BlockFetcherState): Receive = {
-    //by handling these type of messages, fetcher can received from network, fresh info about blocks on top
+    //by handling these type of messages, fetcher can receive from network, fresh info about blocks on top
     //ex. After a successful handshake, fetcher will receive the info about the header of the peer best block
     case MessageFromPeer(BlockHeaders(headers), _) =>
       headers.lastOption.map { bh =>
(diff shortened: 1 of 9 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -45,8 +48,24 @@ class BlockFetcher(
   import BlockFetcher._
 
   implicit val ec: MonixScheduler = MonixScheduler(context.dispatcher)
+  implicit val sys = context.system
   implicit val timeout: Timeout = syncConfig.peerResponseTimeout + 2.second // some margin for actor communication
 
+  val queue = {
```
ten15bit: minor: should be `private` and a type annotation would be useful

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -45,8 +48,24 @@ class BlockFetcher(
   import BlockFetcher._
 
   implicit val ec: MonixScheduler = MonixScheduler(context.dispatcher)
+  implicit val sys = context.system
```
ten15bit: implicits should always have type annotations

ten15bit (approved):
just some linter-like comments :+1:

leo-bogastry (approved):

PR #943, opened 2021-03-19
[ETCM-716] Readd block headers to work queue in case of errors

ten15bit:
# Description

* fixes a bug where block headers weren't readded to the work queue in case of errors
* also moves the block headers work queue to `syncState` to keep it together with the work queues for bodies and receipts

# Proposed Solution

If requests time out or the peer disconnects we need to readd the requested block headers to the work queue so they will be requested from another peer.

# Testing

Full sync on sagano testnet and successful syncing on mainnet for some time.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
@@ -131,7 +131,6 @@ class FastSync(
     private var currentSkeletonState: Option[HeaderSkeleton] = None
     private var skeletonHandler: Option[ActorRef] = None
     private var batchFailuresCount = 0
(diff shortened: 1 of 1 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -677,17 +680,16 @@ class FastSync(
     private def handleRequestFailure(peer: Peer, handler: ActorRef, reason: BlacklistReason): Unit = {
       removeRequestHandler(handler)
 
+      requestedHeaders.get(peer).foreach(requested => syncState = syncState.enqueueHeaderRange(requested))
```
ten15bit, line 683: This is the relevant line for fixing the bug.

robinraju (approved):

dzajkowski (approved):

PR #945, opened 2021-03-22
ETCM-286: fixed some rpc test issues

grale88:
# Description

Part of fixes introduced for Mantis RPC tests

Changed test assertions when requesting future/non-existing blocks. Updated patch script to bring config and keystore files to appropriate mantis client path. Lowered web3j lib to be able to run tests from Intellij (transient dependency issue with Kamon)

# Testing
Follow existing README guide

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -113,7 +113,7 @@ object Dependencies {
     "org.scala-lang.modules" %% "scala-parser-combinators" % "1.1.2",
     "org.scala-sbt.ipcsocket" % "ipcsocket" % "1.1.0",
     "org.xerial.snappy" % "snappy-java" % "1.1.7.7",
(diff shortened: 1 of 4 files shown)


bsuieric:
only some tests were fixed? I have some that are failing. Can we exclude the transitive dependency with sbt intransitive, just a thought

grale88:
@bsuieric 
The last one tagged with PrivNetNoMining is failing as the state gets modified by some of the previously executed test and when calling for transaction count with pending block an error is returned.

With intransitive I suppose I would have to specify all other dependencies explicitly. But what is also the issue is that the API of that particular transitive lib is changed.

bsuieric (approved):

leo-bogastry (approved):

PR #946, opened 2021-03-22
ETCM-657 refactor config files to facilitate running in development

jvdp:
This PR moves the config files around a bit to make them easier to use during development, for example to run the mantis node in test mode:

    $ sbt -Dconfig.file=./src/main/resources/conf/testmode.conf run mantis

Previously, this wouldn't be possible due to `base.conf` not existing in a development checkout (it was only generated during the `sbt dist` task.)


dzajkowski:
since there no longer is a default `application.conf` maybe it would make sense to setup a check that fails if no config file is passed (it will default to some weird akka file I think) and maybe add a note in Readme? wdyt?

jvdp:
> since there no longer is a default `application.conf` maybe it would make sense to setup a check that fails if no config file is passed (it will default to some weird akka file I think) and maybe add a note in Readme? wdyt?

Also discussed in Slack; there is still a default application.conf during development, and for the distribution it still defaults to `etc.conf` in `mantis_launcher`.

bsuieric (approved):

leo-bogastry (approved):

PR #947, opened 2021-03-23
3.1.0 flake

manveru:
This is only used for the Hydra build. Do not merge!


flake.nix
```diff
@@ -0,0 +1,81 @@
...
+        "github:input-output-hk/Sbtix?rev=7b969a5641fce10500ca51cbe88af4ea160d7064";
+      flake = false;
+    };
+  };
+
+  outputs = { self, nixpkgs, flake-utils, nixpkgs-sbt, ... }@inputs:
+    let
+      overlay = final: prev: {
+        inherit (import nixpkgs-sbt { inherit (final) system; }) sbt;
+        sbtix = final.callPackage ./sbtix.nix { };
```
jonringer: I would not use sbtix, but instead sbt-derivation. Mantis is already using it on the `develop` branch

PR #948, opened 2021-03-23
[ETCM-709] Improve ommers validations.

leo-bogastry:
# Description
In testnet we observed the following error message `Block import error BlockImportFailed(OmmersAncestorsError)`, but looking at the code it was not possible to know exactly why the validation failed.
The goal of this PR is to allow for better debugging in case we observe the same error message again.

# Proposed Solution

Replaced OmmersAncestorsError by more specific OmmerIsAncestorError and OmmerParentIsNotAncestorError.
Renamed OmmersNotValidError to OmmersHeaderError and added the underlying reason for failure 
Refactored validateOmmersAncestors method.
Refactored validateOmmersHeaders method to not hide the underlying failing reason.
Reorganized some packages because some classes were not in the correct package and/or the Spec was not in the correct package. Renamed some specs to better reflect the class being tested
Raised the log level of some logs in BlockFetcher because there are barely any logs this actor when debug logs are disabled

# Testing
Connected to Sagano with FastSync off and RegularSync is importing blocks properly

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
@@ -98,6 +98,7 @@ class BlockFetcher(
 
   private def handleCommands(state: BlockFetcherState): Receive = {
     case PickBlocks(amount) => state.pickBlocks(amount) |> handlePickedBlocks(state) |> fetchBlocks
+
     case StrictPickBlocks(from, atLeastWith) =>
       // FIXME: Consider having StrictPickBlocks calls guaranteeing this
       // from parameter could be negative or 0 so we should cap it to 1 if that's the case
@@ -113,6 +114,7 @@ class BlockFetcher(
       }
 
       fetchBlocks(newState)
+
     case InvalidateBlocksFrom(blockNr, reason, withBlacklist) =>
       val (blockProvider, newState) = state.invalidateBlocksFrom(blockNr, withBlacklist)
       log.debug("Invalidate blocks from {}", blockNr)
@@ -157,7 +159,7 @@ class BlockFetcher(
 
   private def handleBodiesMessages(state: BlockFetcherState): Receive = {
     case Response(peer, BlockBodies(bodies)) if state.isFetchingBodies =>
-      log.debug(s"Received ${bodies.size} block bodies")
+      log.debug("Received {} block bodies", bodies.size)
       if (state.fetchingBodiesState == AwaitingBodiesToBeIgnored) {
         log.debug("Block bodies will be ignored due to an invalidation was requested for them")
         fetchBlocks(state.withBodiesFetchReceived)
@@ -171,9 +173,10 @@ class BlockFetcher(
               state.withBodiesFetchReceived.handleRequestedBlocks(newBlocks, peer.id)
           }
         val waitingHeadersDequeued = state.waitingHeaders.size - newState.waitingHeaders.size
-        log.debug(s"Processed $waitingHeadersDequeued new blocks from received block bodies")
+        log.debug("Processed {} new blocks from received block bodies", waitingHeadersDequeued)
         fetchBlocks(newState)
       }
+
     case RetryBodiesRequest if state.isFetchingBodies =>
       log.debug("Something failed on a bodies request, cancelling the request and re-fetching")
       val newState = state.withBodiesFetchReceived
@@ -182,8 +185,10 @@ class BlockFetcher(
 
   private def handleStateNodeMessages(state: BlockFetcherState): Receive = {
     case FetchStateNode(hash) => fetchStateNode(hash, sender(), state)
+
     case RetryFetchStateNode if state.isFetchingStateNode =>
       state.stateNodeFetcher.foreach(fetcher => fetchStateNode(fetcher.hash, fetcher.replyTo, state))
+
     case Response(peer, NodeData(values)) if state.isFetchingStateNode =>
       log.debug("Received state node response from peer {}", peer)
       state.stateNodeFetcher.foreach(fetcher => {
@@ -215,10 +220,13 @@ class BlockFetcher(
       }
       supervisor ! ProgressProtocol.GotNewBlock(newState.knownTop)
       fetchBlocks(newState)
+
     case MessageFromPeer(CommonMessages.NewBlock(block, _), peerId) =>
       handleNewBlock(block, peerId, state)
+
     case MessageFromPeer(PV64.NewBlock(block, _), peerId) =>
       handleNewBlock(block, peerId, state)
+
     case BlockImportFailed(blockNr, reason) =>
       val (peerId, newState) = state.invalidateBlocksFrom(blockNr)
       peerId.foreach(id => peersClient ! BlacklistPeer(id, reason))
@@ -259,8 +267,9 @@ class BlockFetcher(
     state.tryInsertBlock(block, peerId) match {
       case Left(_) if block.number <= state.lastBlock =>
         log.debug(
-          s"Checkpoint block ${ByteStringUtils.hash2string(blockHash)} is older than current last block ${state.lastBlock}" +
-            s" - clearing the queues and putting checkpoint to ready blocks queue"
+          "Checkpoint block {} is older than current last block {} - clearing the queues and putting checkpoint to ready blocks queue",
+          ByteStringUtils.hash2string(blockHash),
+          state.lastBlock
         )
         val newState = state
           .clearQueues()
@@ -279,7 +288,7 @@ class BlockFetcher(
         log.debug(error)
         handleFutureBlock(block, state)
       case Right(state) =>
-        log.debug(s"Checkpoint block [${ByteStringUtils.hash2string(blockHash)}] fit into queues")
+        log.debug("Checkpoint block [{}] fit into queues", ByteStringUtils.hash2string(blockHash))
         fetchBlocks(state)
     }
   }
@@ -289,20 +298,20 @@ class BlockFetcher(
     //ex. After a successful handshake, fetcher will receive the info about the header of the peer best block
     case MessageFromPeer(BlockHeaders(headers), _) =>
       headers.lastOption.map { bh =>
-        log.debug(s"Candidate for new top at block ${bh.number}, current known top ${state.knownTop}")
+        log.debug("Candidate for new top at block {}, current known top {}", bh.number, state.knownTop)
         val newState = state.withPossibleNewTopAt(bh.number)
         fetchBlocks(newState)
       }
     //keep fetcher state updated in case new mined block was imported
     case InternalLastBlockImport(blockNr) =>
-      log.debug(s"New mined block $blockNr imported from the inside")
+      log.debug("New mined block {} imported from the inside", blockNr)
       val newState = state.withLastBlock(blockNr).withPossibleNewTopAt(blockNr)
 
       fetchBlocks(newState)
 
     //keep fetcher state updated in case new checkpoint block was imported
     case InternalCheckpointImport(blockNr) =>
-      log.debug(s"New checkpoint block $blockNr imported from the inside")
+      log.debug("New checkpoint block {} imported from the inside", blockNr)
 
       val newState = state
         .clearQueues()
@@ -414,13 +423,16 @@ class BlockFetcher(
 
   private def handleRequestResult(fallback: FetchMsg)(msg: Any): Task[Any] = msg match {
     case failed: RequestFailed =>
-      log.debug("Request failed due to {}", failed)
+      log.warning("Request failed due to {}", failed)
       Task.now(fallback)
+
     case NoSuitablePeer =>
       Task.now(fallback).delayExecution(syncConfig.syncRetryInterval)
+
     case Failure(cause) =>
       log.error(cause, "Unexpected error on the request result")
       Task.now(fallback)
+
     case m =>
       Task.now(m)
   }
--- a/src/main/scala/io/iohk/ethereum/consensus/ethash/validators/OmmersValidator.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/ethash/validators/OmmersValidator.scala
@@ -2,6 +2,7 @@ package io.iohk.ethereum.consensus.ethash.validators
 
 import akka.util.ByteString
 import io.iohk.ethereum.consensus.ethash.validators.OmmersValidator.{OmmersError, OmmersValid}
+import io.iohk.ethereum.consensus.validators.BlockHeaderError
 import io.iohk.ethereum.consensus.{GetBlockHeaderByHash, GetNBlocksBack}
 import io.iohk.ethereum.domain.{Block, BlockHeader, Blockchain}
 
@@ -36,9 +37,10 @@ object OmmersValidator {
 
   object OmmersError {
     case object OmmersLengthError extends OmmersError
-    case object OmmersNotValidError extends OmmersError
+    case class OmmersHeaderError(errors: List[BlockHeaderError]) extends OmmersError
     case object OmmersUsedBeforeError extends OmmersError
-    case object OmmersAncestorsError extends OmmersError
+    case object OmmerIsAncestorError extends OmmersError
+    case object OmmerParentIsNotAncestorError extends OmmersError
     case object OmmersDuplicatedError extends OmmersError
   }
 
--- a/src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
@@ -1,15 +1,14 @@
-package io.iohk.ethereum.consensus.ethash.validators
+package io.iohk.ethereum.consensus.validators.std
 
 import akka.util.ByteString
+import io.iohk.ethereum.consensus.ethash.validators.OmmersValidator
 import io.iohk.ethereum.consensus.ethash.validators.OmmersValidator.OmmersError._
 import io.iohk.ethereum.consensus.ethash.validators.OmmersValidator.{OmmersError, OmmersValid}
-import io.iohk.ethereum.consensus.validators.BlockHeaderValidator
+import io.iohk.ethereum.consensus.validators.{BlockHeaderError, BlockHeaderValid, BlockHeaderValidator}
 import io.iohk.ethereum.consensus.{GetBlockHeaderByHash, GetNBlocksBack}
 import io.iohk.ethereum.domain.BlockHeader
-import io.iohk.ethereum.utils.BlockchainConfig
 
-class StdOmmersValidator(blockchainConfig: BlockchainConfig, blockHeaderValidator: BlockHeaderValidator)
-    extends OmmersValidator {
+class StdOmmersValidator(blockHeaderValidator: BlockHeaderValidator) extends OmmersValidator {
 
   val OmmerGenerationLimit: Int = 6 // Stated on section 11.1, eq. (143) of the YP
   val OmmerSizeLimit: Int = 2
@@ -30,7 +29,8 @@ class StdOmmersValidator(blockchainConfig: BlockchainConfig, blockHeaderValidato
     * @param getBlockHeaderByHash function to obtain an ancestor block header by hash
     * @param getNBlocksBack       function to obtain N blocks including one given by hash and its N-1 ancestors
     *
-    * @return [[OmmersValidator.OmmersValid]] if valid, an [[OmmersValidator.OmmersError]] otherwise
+    * @return [[io.iohk.ethereum.consensus.ethash.validators.OmmersValidator.OmmersValid]] if valid,
+    *         an [[io.iohk.ethereum.consensus.ethash.validators.OmmersValidator.OmmersError]] otherwise
     */
   def validate(
       parentHash: ByteString,
@@ -66,16 +66,21 @@ class StdOmmersValidator(blockchainConfig: BlockchainConfig, blockHeaderValidato
   /** Validates that each ommer's header is valid based on validations stated in section 11.1 of the YP
     *
     * @param ommers               the list of ommers to validate
-    * @param getBlockHeaderByHash function to obtain ommers' parents
-    *
-    * @return [[OmmersValidator.OmmersValid]] if valid, an [[OmmersValidator.OmmersError.OmmersNotValidError]] otherwise
+    * @param getBlockParentsHeaderByHash function to obtain ommers' parents
+    * @return [[OmmersValidator.OmmersValid]] if valid, an [[OmmersValidator.OmmersError.OmmersHeaderError]] otherwise
     */
   private def validateOmmersHeaders(
       ommers: Seq[BlockHeader],
-      getBlockHeaderByHash: GetBlockHeaderByHash
+      getBlockParentsHeaderByHash: GetBlockHeaderByHash
   ): Either[OmmersError, OmmersValid] = {
-    if (ommers.forall(blockHeaderValidator.validate(_, getBlockHeaderByHash).isRight)) Right(OmmersValid)
-    else Left(OmmersNotValidError)
+    val validationsResult: Seq[Either[BlockHeaderError, BlockHeaderValid]] =
+      ommers.map(blockHeaderValidator.validate(_, getBlockParentsHeaderByHash))
+
+    if (validationsResult.forall(_.isRight)) Right(OmmersValid)
+    else {
+      val errors = validationsResult.collect { case Left(error) => error }.toList
+      Left(OmmersHeaderError(errors))
+    }
   }
 
   /** Validates that each ommer is not too old and that it is a sibling as one of the current block's ancestors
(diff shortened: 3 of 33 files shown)


src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
```diff
@@ -144,19 +152,17 @@ class StdOmmersValidator(blockchainConfig: BlockchainConfig, blockHeaderValidato
...
   private def collectOmmersFromAncestors(
       parentHash: ByteString,
       blockNumber: BigInt,
       getNBlocksBack: GetNBlocksBack
-  ): Option[Seq[BlockHeader]] = {
+  ): Seq[BlockHeader] = {
     val numberOfBlocks = blockNumber.min(OmmerGenerationLimit).toInt
-    val ancestors = getNBlocksBack(parentHash, numberOfBlocks).map(_.body.uncleNodesList)
-    Some(ancestors).filter(_.length == numberOfBlocks).map(_.flatten)
+    getNBlocksBack(parentHash, numberOfBlocks).flatMap(_.body.uncleNodesList)
```
leo-bogastry, line 166: I have simplified both `collectAncestors` and `collectOmmersFromAncestors` methods, by removing the final check.
It was checking that the `getNBlocksBack` did return the expected number of items, otherwise it would just return None.
Is there any reason to keep such a check?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -113,9 +114,10 @@ class BlockFetcher(
       }
 
       fetchBlocks(newState)
+
     case InvalidateBlocksFrom(blockNr, reason, withBlacklist) =>
       val (blockProvider, newState) = state.invalidateBlocksFrom(blockNr, withBlacklist)
-      log.debug("Invalidate blocks from {}", blockNr)
+      log.info("Invalidate blocks from {}", blockNr)
```
dzajkowski: I still think that these should not be changed to info. a better approach (imo) would be to setup the testnet with the logbacks BlockFetcher entry set to debug. this way we avoid burdening every other usage of mantis with additional logging.

leo-bogastry: That is indeed a valid approach. I do find weird that BlockFetcher has so many debug log entries, while the rest of the application also uses info. Note: I did not change all logs in BlockFetcher to be info

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -149,17 +151,17 @@ class BlockFetcher(
...
+      log.info("Something failed on a headers request, cancelling the request and re-fetching")
 
       val newState = state.withHeaderFetchReceived
       fetchBlocks(newState)
   }
 
   private def handleBodiesMessages(state: BlockFetcherState): Receive = {
     case Response(peer, BlockBodies(bodies)) if state.isFetchingBodies =>
-      log.debug(s"Received ${bodies.size} block bodies")
+      log.debug("Received {} block bodies", bodies.size)
```
jvdp, line 162: So this is a situation where this change improves performance, due to actor logging not being based on macro's, correct?

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
```diff
@@ -66,16 +66,22 @@ class StdOmmersValidator(blockchainConfig: BlockchainConfig, blockHeaderValidato
...
   ): Either[OmmersError, OmmersValid] = {
-    if (ommers.forall(blockHeaderValidator.validate(_, getBlockHeaderByHash).isRight)) Right(OmmersValid)
-    else Left(OmmersNotValidError)
+    val validationsResult = ommers.map(blockHeaderValidator.validate(_, getBlockParentsHeaderByHash))
+
+    if (validationsResult.forall(_.isRight)) Right(OmmersValid)
+    else {
+      validationsResult.filterNot(_.isRight) match {
+        case Left(headerError) :: _ => Left(OmmersHeaderError(Some(headerError)))
+        case _ => Left(OmmersHeaderError(None))
```
jvdp: What does this case mean? Is there any way to make this more informative?

src/main/scala/io/iohk/ethereum/consensus/ethash/validators/OmmersValidator.scala
```diff
@@ -36,9 +37,10 @@ object OmmersValidator {
 
   object OmmersError {
     case object OmmersLengthError extends OmmersError
-    case object OmmersNotValidError extends OmmersError
+    case class OmmersHeaderError(error: Option[BlockHeaderError]) extends OmmersError
```
jvdp: An optional argument to an error class seems a bit weird to me as it's so uninformative. (Also commented below about this.)

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
```diff
@@ -85,26 +91,28 @@ class StdOmmersValidator(blockchainConfig: BlockchainConfig, blockHeaderValidato
...
+  private[std] def validateOmmersAncestors(
       parentHash: ByteString,
       blockNumber: BigInt,
       ommers: Seq[BlockHeader],
       getNBlocksBack: GetNBlocksBack
   ): Either[OmmersError, OmmersValid] = {
 
-    val ancestorsOpt = collectAncestors(parentHash, blockNumber, getNBlocksBack)
+    val ancestors = collectAncestors(parentHash, blockNumber, getNBlocksBack)
+    def ommersHashes: Seq[ByteString] = ommers.map(_.hash)
```
jvdp: Would it make sense to use `val` instead of `def`, and `lazy val` for some bindings that are not always used?

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
jvdp: Ideally this would've been covered by tests if it was important. I can't comment on the check itself.

What leads you to believe it's not necessary to keep? Perhaps we can leave an assert in place, just in case?

src/test/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidatorSpec.scala
```diff
@@ -77,7 +92,29 @@ class OmmersValidatorSpec extends AnyFlatSpec with Matchers with ScalaCheckPrope
   // scalastyle:off magic.number
   trait BlockUtils extends EphemBlockchainTestSetup {
 
-    val ommersValidator = new StdOmmersValidator(blockchainConfig, new EthashBlockHeaderValidator(blockchainConfig))
+    val ommersValidator = new StdOmmersValidator(new EthashBlockHeaderValidator(blockchainConfig))
+
+    val ommerInvalidBranch = BlockHeader(
+      parentHash = ByteString(Hex.decode("fd07e36cfaf327801e5696134b12345f6a89fb1e8f017f2411a29d0ae810ab8b")),
```
jvdp, line 112: Why did all this have to be added? What do all the scary hex strings mean?

jvdp (approved):
Better errors and simpler code is better 👍 

I have no opinion about the log level stuff. I think we might want to bump default log level for end users to warning anyway? There's room for a story to really think about what we want out of logging, I think.

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
leo-bogastry: I cannot find a single reason to keep it, it adds noise to the code. The method was returning an `Option[List]` which is quite redundant and makes everything more hard to read

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
leo-bogastry: I had some colleagues that had a very strong opinion on using templates, but I researched about this last Friday and apparently does not make a difference anymore :/

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
leo-bogastry: I had something more specific here, but then I got the build to fail because it was matching this to an error

src/main/scala/io/iohk/ethereum/consensus/ethash/validators/OmmersValidator.scala
leo-bogastry: The internal BlockHeaderError was being completely discarded and replaced by a generic `OmmersNotValidError`. This error happens when during the ommers header validation and it can fail because of 18 different reasons, and that will be the specific BlockHeaderError.
It's an `Option` for code reasons, programmatically has to cover all possible cases.
This may indeed a code smell, why is there a test that returns a generic error running this code? I did not check to be honest.

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
leo-bogastry: `lazy val` can also be. Those pieces of code only run 0 or 1 time, def is just a more compact notation :)

src/test/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidatorSpec.scala
leo-bogastry: It's for test I added. The scary hex'es are hashes :)

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
jvdp: A future refactoring could inadvertently cause some of this to run 1+ times, though.

src/test/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidatorSpec.scala
jvdp: Could these hashes be computed on the fly so you can see what they're based on? Are these from existing tests?

leo-bogastry: These are not random hashes, they validate the block. If you notice I had to write a different test, than the rest of the tests in the spec, trying to just make one hash invalid was breaking the whole block and the validation was failing before validating the ommers.
I did copy this data from another test, yes, and it made me realize that getting this kind of data is not trivial. Maybe it's available in block explorers, I did not check, it just crossed my mind

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
leo-bogastry: I have updated it to `lazy val`. Future proof :)

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
```diff
@@ -66,16 +66,22 @@ class StdOmmersValidator(blockchainConfig: BlockchainConfig, blockHeaderValidato
...
+      getBlockParentsHeaderByHash: GetBlockHeaderByHash
   ): Either[OmmersError, OmmersValid] = {
-    if (ommers.forall(blockHeaderValidator.validate(_, getBlockHeaderByHash).isRight)) Right(OmmersValid)
-    else Left(OmmersNotValidError)
+    val validationsResult = ommers.map(blockHeaderValidator.validate(_, getBlockParentsHeaderByHash))
+
+    if (validationsResult.forall(_.isRight)) Right(OmmersValid)
+    else {
+      validationsResult.filterNot(_.isRight) match {
+        case Left(headerError) :: _ => Left(OmmersHeaderError(Some(headerError)))
```
ten15bit: This will never match because `validationsResult.filterNot(_.isRight)` returns a `Seq`.

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
```diff
@@ -85,26 +91,28 @@ class StdOmmersValidator(blockchainConfig: BlockchainConfig, blockHeaderValidato
...
-    else Left(OmmersAncestorsError)
+    lazy val ancestorsParents: Seq[ByteString] = ancestors.map(_.parentHash)
+    lazy val ommersParentsHashes: Seq[ByteString] = ommers.map(_.parentHash)
+
+    // parent not an ancestor or is too old (we only compare up to 6 previous ancestors)
+    lazy val ommersParentsAreAllAncestors: Boolean = ommersParentsHashes.forall(ancestorsParents.contains)
+
+    if (ommersThatAreAncestors.nonEmpty) Left(OmmerIsAncestorError)
+    else if (!ommersParentsAreAllAncestors) Left(OmmerParentIsNotAncestorError)
+    else Right(OmmersValid)
```
ten15bit, line 115: I wouldn't default to `Right` in the validation.

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
ten15bit: I wouldn't remove it...in the worst case if `getNBlocksBack` returns an empty list, the ommers validation will return `Right`.

The `validateOmmersAncestors` method used to return a `OmmersAncestorsError` to indicate there was a problem getting the ancestors...that behavior seems to be lost as well.

src/main/scala/io/iohk/ethereum/consensus/ethash/validators/OmmersValidator.scala
ten15bit: This will probably have to contain a list of header errors.

see https://github.com/input-output-hk/mantis/pull/948/files#diff-4ffb3286be2984d1679ee76bbb2e5e404f64147f1d77ab0eb95f8031a1379635R81

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -414,13 +423,16 @@ class BlockFetcher(
 
   private def handleRequestResult(fallback: FetchMsg)(msg: Any): Task[Any] = msg match {
     case failed: RequestFailed =>
-      log.debug("Request failed due to {}", failed)
+      log.error("Request failed due to {}", failed)
```
ten15bit: `error` is usually used for fatal errors...I think a `warn` would be enough here

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
ten15bit: I agree that not all of them should be changed to `info`...I'd change the ones indicating failures to `warn` and only important ones to info.

ten15bit (commented):
Looks good! Some comments regarding the refactoring and log levels.

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
leo-bogastry: Yes, and I have deconstructed that in a head and tail, and there's a test proving it works https://github.com/input-output-hk/mantis/blob/feature/ETCM-709-improve-ommers-validation/src/test/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidatorSpec.scala#L31-L39
There's never more than one error reported here

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
leo-bogastry: I have rolled back the modification in the logs, expect one. I think this can be reviewed better by someone that actually touches the code

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
leo-bogastry: It's because no error was found, what would you return there then?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
leo-bogastry: Updated

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
leo-bogastry: `validateOmmersAncestors` was returning `OmmersUsedBeforeError`, but since two different validations were happening in that method I made it clear by replacing `OmmersUsedBeforeError` by two others that are more precise.

ten15bit: `validateOmmersAncestors` was returning `OmmersAncestorsError` in case `getNBlocksBack` returned an invalid number of results: https://github.com/input-output-hk/mantis/pull/948/files#diff-4ffb3286be2984d1679ee76bbb2e5e404f64147f1d77ab0eb95f8031a1379635L107

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
ten15bit: I gave an example below why defaulting to `Right` is problematic.

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
ten15bit: > here's never more than one error reported here

It's calling `blockHeaderValidator.validate` for every ommer - so why can't there be more than one error?

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
leo-bogastry: Ah, ok, now I understand your point. The `getNBlocksBack` is a method that is written by Mantis, so I don't understand why we would have to validate it every single time it's called.

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
LeonorLunatech: For a couple of reason, the main one was to keep the method signature and the second is because as soon an error is found the block validation is stopped, in general, and I saw no reason to change that

leo-bogastry: Indeed, you are right, it can return more than one error and I can group them all in a list.
I think I was following the approach of stopping the validation at the first error found. I will adjust it to be a list, meaning `case class OmmersHeaderError(error: List[BlockHeaderError])`

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
ten15bit: Ideally `getNBlocksBack` would return a type that allows encoding errors but since it doesn't, it's a good idea to check whether it returns the expected number of blocks.

In theory we should always be able to get the expected numbers of blocks, so if the number doesn't match something is definitely broken. By removing the check and defaulting to `Right` that problem will be silently ignored.

leo-bogastry: So I am adding a check in `private def validateOmmersNotUsed`,making sure that if the list is empty the validation fails, in `private[std] def validateOmmersAncestors` the validation will already fail if the list of ancestors is empty

leo-bogastry: I have also added a test named `report failure if getNBlocksBack returns an empty list of ancestors`, involving `validateOmmersAncestors` method

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
leo-bogastry: Sorry, but where?

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
leo-bogastry: I had to remove the explicit check that checked that the list of ommers from ancestors was not empty, because it was causing 2 ETS tests to fails

leo-bogastry:
Tested FastSync and RegularSync once again in `etc`

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
ten15bit: Sorry, missed the comment. I was referring to https://github.com/input-output-hk/mantis/pull/948#discussion_r609043993

In general I'd say the problem with defaulting to `Right` is that it will also return Right for every unhandled/unexpected case while the other way around is more restrictive, making the errors more visible.

In this specific case I'd re-add at the very least the check that was removed in https://github.com/input-output-hk/mantis/pull/948/files#diff-4ffb3286be2984d1679ee76bbb2e5e404f64147f1d77ab0eb95f8031a1379635L160

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
```diff
@@ -66,16 +66,24 @@ class StdOmmersValidator(blockchainConfig: BlockchainConfig, blockHeaderValidato
...
-    if (ommers.forall(blockHeaderValidator.validate(_, getBlockHeaderByHash).isRight)) Right(OmmersValid)
-    else Left(OmmersNotValidError)
+    val validationsResult: Seq[Either[BlockHeaderError, BlockHeaderValid]] =
+      ommers.map(blockHeaderValidator.validate(_, getBlockParentsHeaderByHash))
+
+    if (validationsResult.forall(_.isRight)) Right(OmmersValid)
+    else {
+      val ommersHeaderError = validationsResult
+        .collect { case Left(error) => error }
+        .map(error => OmmersHeaderError(List(error)))
```
ten15bit: `.map` and `.reduce` are not needed...can be simplified to
```
val errors = validationsResult.collect { case Left(error) => error }.toList
Left(OmmersHeaderError(errors))
```

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
ten15bit: We shouldn't ignore the issue if blocks aren't found. I'd suggest one of the two approaches

- add a check for the correct size of the result and return an appropriate error otherwise
- merge as it is but create a follow-up ticket to refactor `GetNBlocksBack` to return an error in case blocks aren't found (better approach imho)

leo-bogastry: I created a follow-up ticket ETCM-776

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
leo-bogastry: That is the same check that will be added in  ETCM-776

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
leo-bogastry: You are right, I really overcomplicated that :)

ten15bit: :)

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
ten15bit: ETCM-776 was created after I wrote that comment.

...and I still think that defaulting to Right in validations is less secure.

ten15bit (approved):

src/main/scala/io/iohk/ethereum/consensus/validators/std/StdOmmersValidator.scala
leo-bogastry: I understand the philosophy of what you are saying, but I also think that in this case it would really overcomplicate the code. The whole reason why I was not able to understand why a block validation failed in testnet was because those two separate errors `OmmerIsAncestorError` and `OmmerParentIsNotAncestorError` did not exist.
This code is only making these two specific validations otherwise there is no error.

leo-bogastry: What I can propose is a bigger refactoring, than ETCM-776. `GetNBlocksBack` should be part of the `for` comprehension, inside the more generic `validate` method and `validateOmmersAncestors` could be divided into two method (each one returning only one error)

leo-bogastry: I have rewritten ETCM-776

PR #949, opened 2021-03-24
[ETCM-685] Improve pivot block selection

robinraju:
# Description

If Mantis connects with a peer with the best block number bigger than other nodes (the difference bigger than `pivotBlockOffset`) it will wait for a minimum number of voters infinitely and never choose pivot block and will never start to sync.

# Proposed Solution

If a particular pivot block cannot be confirmed,
 * the pivot block number is pushed back by a configured amount of blocks and the confirmation process repeats.
 * if a maximum number of retries is reached, the task fails and the whole process is restarted.

`PivotBlockSelector.scala`

# Testing

Unit tests were added.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/fast/PivotBlockSelector.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/fast/PivotBlockSelector.scala
@@ -35,13 +35,23 @@ class PivotBlockSelector(
   import PivotBlockSelector._
   import syncConfig._
 
+  private var pivotBlockRetryCount = 0
+
+  //TODO make these config params
+  private val pivotBlockNumberResetDelta: BigInt = 50
+  private val maxPivotBlockFailuresCount = 3
+
   override def receive: Receive = idle
 
   private def idle: Receive = handlePeerListMessages orElse { case SelectPivotBlock =>
-    val election @ ElectionDetails(correctPeers, expectedPivotBlock) = collectVoters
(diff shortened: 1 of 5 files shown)


dzajkowski (approved):

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/PivotBlockSelector.scala
```diff
@@ -188,12 +210,20 @@ class PivotBlockSelector(
...
+        .map(_ - pivotBlockNumberResetDelta.max(0))
+        .getOrElse(bestPeerBestBlockNumber)
+
+    val expectedPivotBlock = (currentBestBlockNumber - syncConfig.pivotBlockOffset).max(0)
     val correctPeers = peersSortedByBestNumber
       .takeWhile { case (_, number) => number >= expectedPivotBlock }
       .map { case (peer, _) => peer }
 
-    ElectionDetails(correctPeers, expectedPivotBlock)
+    ElectionDetails(correctPeers, currentBestBlockNumber, expectedPivotBlock)
```
jvdp, line 226: I think there's a nice opportunity to split some of the functionality in this file out into a couple pure functions / methods involving the `ElectionDetails` type. We could add some unit tests for it that way.

jvdp (approved):
It gets the job done and nothing is standing out but that `PivotBlockSelector` is pretty smelly and is getting just a bit worse with these additions.

PR #944, opened 2021-03-22
[Etcm 521] Fast sync integration tests

robinraju:
# Description

This PR adds integration tests for the newly introduced fast sync improvements (Concurrent download of headers)

# Important Changes Introduced

Minor fixes were applied to the original implementation of `FastSync`, which prevented the branch resolution from being completed properly.

# Testing

Run the tests and look through the logs for branch resolution when a malicious master peer is found.

PR #951, opened 2021-03-24
ETCM-719: Remove header work queue from persisted state and fix final header skeleton size

ten15bit:
# Description

- a previous refactoring moved the block header work queue to sync state to keep it together with the work queues for block bodies and receipts - this had the unintended side-effect that it also got persisted (which we don't want to happen)
- there was a bug that happened under certain conditions when downloading the final header skeleton
  - let's say we still need to download skeleton headers for the range `5100` to `5200` and our max batch size is `30` the request that would be made would have been to get 4 headers with a gap of 30 - so we're actually requesting up to `5220` which the peer might not have. In that case it would send us back only 3 headers which would fail our validation (because we expect 4).
  - improved the logging of outstanding requests: it would sometimes say `waiting for [0] requests` when it was actually waiting for the header skeleton
- improves the check whether the peer that we're requesting the skeleton headers from has a high enough best block

# Proposed Solution

- moved the header work queue out of sync state again
- changed the calculation of `limit` so it would only request 3 headers in the example above, so the header skeleton would contain headers `5100` to `5190`. After that a new skeleton would be requested with one batch of size 10.

# Testing

- added a unit test for calculations performed in `HeaderSkeleton`
- test case to be added to #944 (if not covered already)
- full sync with fast-sync enabled on sagano and mainnet


src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -359,13 +366,13 @@ class FastSync(
     }
 
     private def waitingForBranchResolution: Receive = handleStatus orElse {
-      case FastSyncBranchResolverActor.BranchResolvedSuccessful(firstCommonBlockNumber, resolvedPeer) =>
+      case FastSyncBranchResolverActor.BranchResolvedSuccessful(firstCommonBlockNumber, newMasterPeer) =>
         // Reset the batch failures count
         batchFailuresCount = 0
 
         // Restart syncing from the valid block available in state.
         syncState = syncState.copy(bestBlockHeaderNumber = firstCommonBlockNumber)
```
robinraju: I think we need to update `syncState.nextBlockToFullyValidate` as well here. 
maybe `nextBlockToFullyValidate = firstCommonBlockNumber + 1`

robinraju (approved):

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
ten15bit: yes, makes sense to update `nextBlockToFullyValidate` as well :+1:

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -874,25 +884,27 @@ class FastSync(
...
+          assignedHandlers.size + skeletonHandler.size
+        )
       }
     }
 
     private def shouldRequestNewSkeleton(peerInfo: PeerInfo): Boolean =
       currentSkeletonState.isEmpty &&
         skeletonHandler.isEmpty &&
-        syncState.bestBlockHeaderNumber < syncState.safeDownloadTarget &&
-        peerInfo.maxBlockNumber >= syncState.pivotBlock.number
```
AnastasiiaL, line 890: need help understanding removing this line :)

AnastasiiaL (approved):
the rest I understand and it makes sense 👍

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
ten15bit: sure :) the check here didn't make a lot of sense because `peer` usually wasn't used for the download anyway (we use the master peer if it's defined).

it has been replaced with the logic here: https://github.com/input-output-hk/mantis/pull/951/files#diff-4b491aec4f7ba3bc8091f81d49100e0ccc38a2f396bcca38434118637e6c0befR987

which basically says:
- if we have a master peer and its best block is high enough, use the master peer
- otherwise check if this peer's best block is high enough and if yes, use it to download the header skeleton and as new master peer
- if the peer also doesn't have a high enough best block log a warning (logic will be called again with another peer automatically by the scheduled `processSyncing()` call)

AnastasiiaL: clear, thanks!

PR #952, opened 2021-03-25
[ETCM-661]: adds pottery network for local nomad/docker setup

krzysztofpaliga-iohk:
# Description

For nomad/docker configuration we need a fresh local network, which runs on its own.

# Proposed Solution

A new network can called pottery has been introduced.


src/main/resources/conf/chains/pottery-chain.conf
```diff
@@ -0,0 +1,160 @@
+{
```
jvdp: Is there stuff you can use from an existing chain conf perhaps? Then you could just include that and keep only the changed params.

src/main/resources/conf/chains/pottery-genesis.json
```diff
@@ -0,0 +1,12 @@
+{
+  "difficulty": "0x20000",
```
jvdp: I think a lower difficulty is possible: https://github.com/input-output-hk/mantis/blob/develop/src/rpcTest/resources/privateNetConfig/conf/rpc-test-private-genesis.json#L5

src/main/resources/conf/pottery.conf
```diff
@@ -0,0 +1,43 @@
...
+mantis {
+  blockchains {
+    network = "pottery"
+  }
+
+  network {
+    protocol-version = 64
+
+    discovery {
+      discovery-enabled = true
```
jvdp, line 12: How does this work? Every node will need to be bootstrapped somewhere, right?

jvdp (approved):
Not much to say about config files (other than a few specific points in my comments.) Great to have dedicated configuration for a local testnet, though 👍  Also keeps https://github.com/input-output-hk/mantis-automation a bit simpler, I imagine?

src/main/resources/conf/pottery.conf
krzysztofpaliga-iohk: As far as I know, every node uses bootstrapped nodes. I had some issues without discovery enabled. I can't recall what it was. : /

src/main/resources/conf/chains/pottery-chain.conf
krzysztofpaliga-iohk: good idea...

src/main/resources/conf/chains/pottery-genesis.json
krzysztofpaliga-iohk: good idea...

leo-bogastry (approved):

PR #956, opened 2021-03-31
[ETCM-714] Increase the blacklisting duration for peers that result in critical errors

leo-bogastry:
# Description

A very big number of peers is blacklisted and not all reasons are clear

# Proposed Solution

Increased the blacklisting duration for peers that result in critical errors
Some code improvements

# Testing
No impacting changes where made

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/BlacklistSupport.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/BlacklistSupport.scala
@@ -34,9 +34,11 @@ trait BlacklistSupport {
   }
 
   def undoBlacklist(blacklistId: BlacklistId): Unit = {
-    val peer = blacklistedPeers.get(blacklistId)
-    peer.foreach(_.cancel())
-    blacklistedPeers.remove(blacklistId)
+    if (blacklistedPeers.contains(blacklistId)) {
+      val peer = blacklistedPeers.get(blacklistId)
+      peer.foreach(_.cancel())
+      blacklistedPeers.remove(blacklistId)
+    }
   }
 
   def isBlacklisted(blacklistId: BlacklistId): Boolean =
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/PeerListSupport.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/PeerListSupport.scala
@@ -50,12 +50,11 @@ trait PeerListSupport {
       removePeer(peerId)
   }
 
-  def peerById(peerId: PeerId): Option[Peer] = handshakedPeers collectFirst {
+  def getHandshakedPeerById(peerId: PeerId): Option[Peer] = handshakedPeers collectFirst {
     case (peer, _) if peer.id == peerId => peer
   }
 
-  def blacklistIfHandshaked(peer: Peer, reason: String): Unit =
-    handshakedPeers.get(peer).foreach(_ => blacklist(peer.id, syncConfig.blacklistDuration, reason))
+  def blacklistPeer(peer: Peer, reason: String): Unit = blacklist(peer.id, syncConfig.blacklistDuration, reason)
 }
 object PeerListSupport {
   type PeersMap = Map[Peer, PeerInfo]
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
@@ -2,7 +2,6 @@ package io.iohk.ethereum.blockchain.sync.fast
 
 import akka.actor._
 import akka.util.ByteString
-import cats.NonEmptyTraverse.ops.toAllNonEmptyTraverseOps
 import cats.data.NonEmptyList
 import cats.implicits._
 import io.iohk.ethereum.blockchain.sync.Blacklist.BlacklistReason._
@@ -20,7 +19,7 @@ import io.iohk.ethereum.blockchain.sync.fast.SyncStateSchedulerActor.{
   StateSyncFinished,
   WaitingForNewTargetBlock
 }
-import io.iohk.ethereum.consensus.validators.{BlockHeaderError, BlockHeaderValid, Validators}
+import io.iohk.ethereum.consensus.validators.Validators
 import io.iohk.ethereum.db.storage.{AppStateStorage, FastSyncStateStorage}
 import io.iohk.ethereum.domain._
 import io.iohk.ethereum.mpt.MerklePatriciaTrie
@@ -36,13 +35,10 @@ import org.bouncycastle.util.encoders.Hex
 import java.time.Instant
 import java.util.concurrent.TimeUnit
 import scala.annotation.tailrec
+import scala.collection.mutable
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
-import scala.util.Random
-import scala.collection.mutable
-import scala.util.Try
-import scala.util.Success
-import scala.util.Failure
+import scala.util.{Random, Success, Try}
 
 // scalastyle:off file.size.limit
 class FastSync(
@@ -342,7 +338,7 @@ class FastSync(
         batchFailuresCount += 1
         if (batchFailuresCount > fastSyncMaxBatchRetries) {
           log.info("Max number of allowed failures reached. Switching branch and master peer.")
-          handleRewind(header, masterPeer.get, fastSyncBlockValidationN, blacklistDuration)
+          handleRewind(header, masterPeer.get, fastSyncBlockValidationN, criticalBlacklistDuration)
 
           // Start branch resolution and wait for response from the FastSyncBranchResolver actor.
           context become waitingForBranchResolution
@@ -564,6 +560,7 @@ class FastSync(
 
     private def handleRewind(header: BlockHeader, peer: Peer, N: Int, duration: FiniteDuration): Unit = {
       blacklist.add(peer.id, duration, BlockHeaderValidationFailed)
+
       if (header.number <= syncState.safeDownloadTarget) {
         discardLastBlocks(header.number, N)
         syncState = syncState.updateDiscardedBlocks(header, N)
--- a/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala
+++ b/src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala
@@ -61,9 +61,9 @@ object WireProtocol {
 
   object Disconnect {
     object Reasons {
-      val DisconnectRequested = 0x00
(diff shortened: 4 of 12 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/PeerListSupport.scala
```diff
@@ -50,12 +50,11 @@ trait PeerListSupport {
...
   }
 
-  def peerById(peerId: PeerId): Option[Peer] = handshakedPeers collectFirst {
+  def getHandshakedPeerById(peerId: PeerId): Option[Peer] = handshakedPeers collectFirst {
     case (peer, _) if peer.id == peerId => peer
   }
 
-  def blacklistIfHandshaked(peer: Peer, reason: String): Unit =
-    handshakedPeers.get(peer).foreach(_ => blacklist(peer.id, syncConfig.blacklistDuration, reason))
+  def blacklistPeer(peer: Peer, reason: String): Unit = blacklist(peer.id, syncConfig.blacklistDuration, reason)
```
leo-bogastry, line 57: I have removed the double call to `handshakedPeers.get(peer)`

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -342,7 +338,7 @@ class FastSync(
         batchFailuresCount += 1
         if (batchFailuresCount > fastSyncMaxBatchRetries) {
           log.info("Max number of allowed failures reached. Switching branch and master peer.")
-          handleRewind(header, masterPeer.get, fastSyncBlockValidationN, blacklistDuration)
+          handleRewind(header, masterPeer.get, fastSyncBlockValidationN, criticalBlacklistDuration)
```
leo-bogastry, line 341: According to the comments in base.conf, the `critical-blacklist-duration` should be used in certain cases:
`# Duration for high offense blacklisting of a peer. Blacklisting reason include: header validation failure.`

And I was once being used once

src/main/scala/io/iohk/ethereum/blockchain/sync/BlacklistSupport.scala
```diff
@@ -34,9 +34,11 @@ trait BlacklistSupport {
   }
 
   def undoBlacklist(blacklistId: BlacklistId): Unit = {
-    val peer = blacklistedPeers.get(blacklistId)
-    peer.foreach(_.cancel())
-    blacklistedPeers.remove(blacklistId)
+    if (blacklistedPeers.contains(blacklistId)) {
```
ten15bit, line 37: This check is not needed since `blacklistedPeers.get` will return an `Option`.

The whole method could be rewritten to `blacklistedPeers.remove(blacklistId).foreach(_.cancel())` though...but since `BlacklistSupport` is deprecated and will be removed eventually, it's probably not worth to start cleaning up the code style.

src/main/scala/io/iohk/ethereum/blockchain/sync/PeerListSupport.scala
ten15bit: Could you please explain that comment? As above, the `get` here is fine because it will return an `Option`.

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
ten15bit: If it was used before and then changed, it could also mean that the blacklisting was too aggressive and it was changed to the shorter period on purpose.

In my opinion it makes sense though to blacklist a master peer that repeatedly sent us wrong data for a long time so I'm ok with that change.

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -564,6 +561,8 @@ class FastSync(
 
     private def handleRewind(header: BlockHeader, peer: Peer, N: Int, duration: FiniteDuration): Unit = {
       blacklist.add(peer.id, duration, BlockHeaderValidationFailed)
+      log.debug("Blacklisting peer with id {}, reason: {}", peer.id, BlockHeaderValidationFailed.toString)
```
ten15bit: same as https://github.com/input-output-hk/mantis/pull/956#discussion_r606090017

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -633,14 +632,17 @@ class FastSync(
     private def handleBlockBodies(peer: Peer, requestedHashes: Seq[ByteString], blockBodies: Seq[BlockBody]): Unit = {
       if (blockBodies.isEmpty) {
         val knownHashes = requestedHashes.map(ByteStringUtils.hash2string)
-        blacklist.add(peer.id, blacklistDuration, EmptyBlockBodies(knownHashes))
+        val reason = EmptyBlockBodies(knownHashes)
```
ten15bit: same as https://github.com/input-output-hk/mantis/pull/956#discussion_r606090017

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -653,7 +655,9 @@ class FastSync(
     private def handleReceipts(peer: Peer, requestedHashes: Seq[ByteString], receipts: Seq[Seq[Receipt]]): Unit = {
       if (receipts.isEmpty) {
         val knownHashes = requestedHashes.map(ByteStringUtils.hash2string)
-        blacklist.add(peer.id, blacklistDuration, EmptyReceipts(knownHashes))
+        val reason = EmptyReceipts(knownHashes)
```
ten15bit: same as https://github.com/input-output-hk/mantis/pull/956#discussion_r606090017

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -675,7 +679,9 @@ class FastSync(
 
           case ReceiptsValidationResult.Invalid(error) =>
             val knownHashes = requestedHashes.map(h => Hex.toHexString(h.toArray[Byte]))
-            blacklist.add(peer.id, blacklistDuration, InvalidReceipts(knownHashes, error))
+            val reason = InvalidReceipts(knownHashes, error)
```
ten15bit: same as https://github.com/input-output-hk/mantis/pull/956#discussion_r606090017

src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala
```diff
@@ -61,9 +61,9 @@ object WireProtocol {
 
   object Disconnect {
     object Reasons {
```
ten15bit, line 63: The reasons are based on https://github.com/ethereum/devp2p/blob/master/rlpx.md#disconnect-0x01 so we shouldn't rename them.

ten15bit (changes requested):
The names of the reasons for disconnects are based on the official protocol so we shouldn't rename them. Regarding the other changes I've left my comments inline.

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -384,6 +380,7 @@ class FastSync(
 
     private def blockHeadersError(peer: Peer, blacklistReason: BlacklistReason): Unit = {
       blacklist.add(peer.id, blacklistDuration, blacklistReason)
+      log.debug("Blacklisting peer with id {}, reason: {}", peer.id, blacklistReason.toString)
```
ten15bit: This should already be logged in https://github.com/input-output-hk/mantis/blob/04208b047c1f9c4917e8fb662709911ab59afe44/src/main/scala/io/iohk/ethereum/blockchain/sync/Blacklist.scala#L152

src/main/scala/io/iohk/ethereum/blockchain/sync/PeerListSupport.scala
leo-bogastry: I tend to see a lot of redundant code, the same checks done over and over again, and this is one of them

src/main/scala/io/iohk/ethereum/blockchain/sync/BlacklistSupport.scala
leo-bogastry: I was just an optimisation because both `get` and `remove` only do something if the blacklist contains the peer

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
leo-bogastry: 👍

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
leo-bogastry: You are right, I was having issues seeing the reasons for all blacklisting, by just adding certain classes to show the DEBUG log level, specifically the logs in `BlacklistSupport` were not showing, I had to change them to INFO during the investigation, I guess I ended up by adding these as well. I will revert

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
leo-bogastry: You don't need to put a new comment on each occurrence of a previous comment. will remove all log entries around `blacklist.add` in FastSync

src/main/scala/io/iohk/ethereum/network/p2p/messages/WireProtocol.scala
leo-bogastry: I can put it back, but "useless peer" does not add any specific info and "disconnect request" as well. I will add a comment in the code explaining when these are used at least

src/main/scala/io/iohk/ethereum/blockchain/sync/PeerListSupport.scala
ten15bit: Removing `handshakedPeers.get(peer).foreach` changes the behavior of the method though. The point of this method is to only blacklist when the peer is handshaked, otherwise we could use `blacklist` directly in the actor.

So we shouldn't rename the method or remove the check.

leo-bogastry: This method is called right after a call to `handshakedPeers.get(peer)`, that it what I meant by redundant code, and it's only called from a single place.

leo-bogastry:
Closing this PR, because the main initial modifications were rolled back, and there are other tickets in the backlog to address blacklisting issues like ETCM-760 and ETCM-761

PR #958, opened 2021-04-06
Return only the child of the checkpoint block

AnastasiiaL:
# Description

Fix a bug with returning the same candidate for a checkpoint as the previous one (the paren)


leo-bogastry:
Is there a ticket number for this?

AnastasiiaL:
> Is there a ticket number for this?

not for this bug. The feature was described here https://jira.iohk.io/browse/ETCM-655

leo-bogastry:
> > Is there a ticket number for this?
> 
> not for this bug. The feature was described here https://jira.iohk.io/browse/ETCM-655

Then why not use that ticket to "tag"  the commit and PR title?

leo-bogastry (approved):

ten15bit (approved):

PR #959, opened 2021-04-07
[ETCM-715] Add new network metric (tried.peers)

jb-lunatech:
# Description

- Detail info log in `PeerManagerActor`
- Create new network metric (tried.peers)
- Create a new Grafana panel computing and displaying the ratio between tried / discovered peers

# Proposed Solution

This let Grafana compute the ratio. It is cleaner from the metrics and Scala code point of view, however it takes time to make it work in the Grafana dashboard.

The alternative is to set a ratio metric in Scala (with the drawback of having convoluted metrics)

# Testing

- Use `build.sh` in `docker/mantis`
- Look for the new _network_ panel in Grafana

### DIFF
--- a/src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
+++ b/src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
@@ -154,6 +154,7 @@ class PeerManagerActor(
     NetworkMetrics.DiscoveredPeersSize.set(nodes.size)
     NetworkMetrics.BlacklistedPeersSize.set(blacklistedPeers.size)
     NetworkMetrics.PendingPeersSize.set(connectedPeers.pendingPeersCount)
+    NetworkMetrics.TriedPeersSize.set(triedNodes.size)
 
     log.info(
       s"Total number of discovered nodes ${nodes.size}. " +
--- a/src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
+++ b/src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
@@ -156,9 +156,9 @@ class PeerManagerActor(
     NetworkMetrics.PendingPeersSize.set(connectedPeers.pendingPeersCount)
 
     log.info(
-      s"Discovered ${nodes.size} nodes, " +
-        s"Blacklisted ${blacklistedPeers.size} nodes, " +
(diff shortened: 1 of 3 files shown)


bsuieric (approved):

src/main/scala/io/iohk/ethereum/network/PeerManagerActor.scala
```diff
@@ -154,13 +154,16 @@ class PeerManagerActor(
...
     log.info(
-      s"Discovered ${nodes.size} nodes, " +
+      s"Discovered ${nodes.size} (can connect to ${discoveredNodes.size}) nodes, " +
         s"Blacklisted ${blacklistedPeers.size} nodes, " +
         s"handshaked to ${connectedPeers.handshakedPeersCount}/${peerConfiguration.maxOutgoingPeers + peerConfiguration.maxIncomingPeers}, " +
         s"pending connection attempts ${connectedPeers.pendingPeersCount}. " +
-        s"Trying to connect to ${nodesToConnect.size} more nodes."
+        s"Trying to connect to ${nodesToConnect.size} more nodes, " +
+        s"Total tried nodes ${triedNodes.size}, " +
+        s"Connected nodes ${connectedPeers.peers.size}."
```
leo-bogastry: This log message becomes a bit confusing like this, for example this is a log message when connecting to `etc`: `Discovered 1154 (can connect to 790) nodes, Blacklisted 426 nodes, handshaked to 6/80, pending connection attempts 4. Trying to connect to 40 more nodes, Total tried nodes 435, Connected nodes 10.`

leo-bogastry: What about something like this? 
`Total number of discovered nodes 1154.
Total number of connection attempts 435, blacklisted 426 nodes, connected to 10 nodes.
Handshaked 6/80, pending connection attempts with 4 nodes. Trying to connect to 40 more nodes.`

AnastasiiaL: agree, especially handshaked and connected showing different numbers is confusing.
and generally this looks like too much information

leo-bogastry: Looking in more detail, the `Connected nodes 10` is also confusing it is actually 6 handshaked and 4 pending, so maybe I would remove the  `Connected nodes 10` actually

leo-bogastry (approved):

PR #960, opened 2021-04-07
[ETCM-746] Add support for PoW with Keccak-256 (ECIP-1049)

leo-bogastry:
Added support for PoW with Keccak-256 (ECIP-1049)
Renamed several classes from Ethash to PoW because they will be used in both Ethash Pow and Keccak PoW

# Description

Add support for PoW with Keccak-256 (ECIP-1049)
Renamed several classes from Ethash to PoW because they will be used in both Ethash Pow and Keccak PoW

# Proposed Solution
Added `KeccakCalculation` object that computes the new `mixHash` and validates it against the block difficulty.
In order to control if a block should be mined using the new Keccak-Pow or current Ethash-Pow a simple if clause was added to the `EthashMiner` class. In ETCM-759 this will be refactored.

Added `KeccakBlockHeaderValidator` object that validates a block header for a block mined with the Keccak-Pow.
Added `PoWBlockHeaderValidator` class to controls whether (new)`KeccakBlockHeaderValidator` or (current) `EthashBlockHeaderValidator` should be used for the validation

NOTE:
Class `EthashDAGManager` is not new code it was extracted from `EthashMiner`.
Because I renamed the package, `EthashMiner` is marked as a new file, but only the code related with `if (isKeccak(blockNumber))`

The total number of files modified is big because of renaming

# Testing

The branch was tested against `etc` network with both FastSync and RegularSync synchronization.


crypto/src/main/scala/io/iohk/ethereum/crypto/package.scala
```diff
@@ -41,6 +41,15 @@ package object crypto {
   def kec256(input: ByteString): ByteString =
     ByteString(kec256(input.toArray))
 
+  def kec256PoW(header: Array[Byte], once: Array[Byte]): Array[Byte] = {
```
milenkara: typo? Probably needs to be `nonce`

leo-bogastry: Oh yeah... I remember writing `once` here and there instead of `nonce` 😅

src/main/scala/io/iohk/ethereum/consensus/pow/blocks/RestrictedPoWBlockGeneratorImpl.scala
```diff
@@ -1,17 +1,17 @@
...
+import io.iohk.ethereum.consensus.pow.RestrictedPoWSigner
+import io.iohk.ethereum.consensus.pow.validators.ValidatorsExecutor
 import io.iohk.ethereum.domain.{Address, Block, Blockchain, SignedTransaction}
 import io.iohk.ethereum.ledger.{BlockPreparator, InMemoryWorldStateProxy}
 import io.iohk.ethereum.utils.BlockchainConfig
 import org.bouncycastle.crypto.AsymmetricCipherKeyPair
 import io.iohk.ethereum.consensus.ConsensusMetrics
 
-class RestrictedEthashBlockGeneratorImpl(
+class RestrictedPoWBlockGeneratorImpl(
```
bsuieric, line 14: Perhaps renaming in ConsensusMetrics makes sense now so we align the description

leo-bogastry: 👍

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
```diff
@@ -22,16 +22,16 @@ import scala.util.{Failure, Success, Try}
...
   val fullConfig = FullConsensusConfig(ConsensusConfig(Config.config), specificConfig)
 
   def loadEthashConsensus(
       vm: VMImpl,
       blockchain: BlockchainImpl,
       blockchainConfig: BlockchainConfig,
       validators: ValidatorsExecutor
-  ): ethash.EthashConsensus = {
-    val consensus = EthashConsensus(vm, blockchain, blockchainConfig, fullConfig, validators, NoAdditionalEthashData)
+  ): pow.PoWConsensus = {
```
jb-lunatech: `pow` and `pow.PoWConsensus` are both imported, I suggest to simply write `PoWConsensus`

src/main/scala/io/iohk/ethereum/consensus/pow/EthashMiner.scala
```diff
@@ -0,0 +1,205 @@
+package io.iohk.ethereum.consensus
+package pow
+
+import akka.actor.{Actor, ActorLogging, ActorRef, Props}
```
jb-lunatech, line 4: What is the sentiment of the team about Typed Actor?

leo-bogastry: In ETCM-759 I am adding 2 new actors, that will be typed, but for now the decision has been to do new things with Typed Actors and not migrate the Classic ones. Maybe in the end this one will also be migrated

crypto/src/main/scala/io/iohk/ethereum/crypto/package.scala
leo-bogastry: Fixed

src/main/scala/io/iohk/ethereum/consensus/pow/blocks/RestrictedPoWBlockGeneratorImpl.scala
leo-bogastry: Fixed

src/ets/scala/io/iohk/ethereum/ets/blockchain/ScenarioSetup.scala
leo-bogastry: Fixed

jb-lunatech (approved):

bsuieric:
So we don't use it for now, just for test purposes? in ECIP-1049 I see there is an activation block

leo-bogastry:
> So we don't use it for now, just for test purposes? in ECIP-1049 I see there is an activation block

Where did you see that? I have the idea that that it is still not decided, on which block to do the switch, for mainnet. Or even if there's a final decision to really go forward with this

I don't understand what do you mean with "is this just for testing purposes?". Everything has to be tested before going to live (there's a test network called Astor that only mines with keccak, we will add support for that on ETCM-749)

src/main/scala/io/iohk/ethereum/consensus/pow/EthashDAGManager.scala
```diff
@@ -0,0 +1,94 @@
...
+import io.iohk.ethereum.utils.{ByteUtils, Logger}
+import org.bouncycastle.util.encoders.Hex
+
+import java.io.{File, FileInputStream, FileOutputStream}
+import scala.util.{Failure, Success, Try}
+
+class EthashDAGManager(blockCreator: EthashBlockCreator) extends Logger {
+  var currentEpoch: Option[Long] = None
+  var currentEpochDagSize: Option[Long] = None
+  var currentEpochDag: Option[Array[Array[Int]]] = None
```
AnastasiiaL, line 14: are vars unavoidable here?

src/main/scala/io/iohk/ethereum/consensus/pow/EthashDAGManager.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  def calculateDagSize(blockNumber: Long, epoch: Long): (Array[Array[Int]], Long) = {
+    (currentEpoch, currentEpochDag, currentEpochDagSize) match {
+      case (Some(`epoch`), Some(dag), Some(dagSize)) => (dag, dagSize)
```
AnastasiiaL, line 18:
```suggestion
      case (Some(_), Some(dag), Some(dagSize)) => (dag, dagSize)
```

src/main/scala/io/iohk/ethereum/consensus/pow/EthashMiner.scala
```diff
@@ -0,0 +1,205 @@
...
+    val headerHash = crypto.kec256(BlockHeader.getEncodedWithoutNonce(block.header))
+    val startTime = System.nanoTime()
+    val mineResult =
+      mineEthah(headerHash, block.header.difficulty.toLong, dagSize, dag, blockCreator.miningConfig.mineRounds)
+    (startTime, mineResult)
+  }
+
+  private def doKeccakMining(block: Block, numRounds: Int): (Long, MiningResult) = {
+    val rlpEncodedHeader = BlockHeader.getEncodedWithoutNonce(block.header)
+    val initNonce = BigInt(64, new Random()) // scalastyle:ignore magic.number
```
AnastasiiaL, line 113: maybe turn 64 into a 'not magic' number instead of omitting the warning?

src/main/scala/io/iohk/ethereum/consensus/pow/EthashMiner.scala
```diff
@@ -0,0 +1,205 @@
...
+        node.system.actorOf(minerProps)
+      case consensus =>
+        wrongConsensusArgument[PoWConsensus](consensus)
+    }
+  }
+
+  private case object ProcessMining
+
+  // scalastyle:off magic.number
+  val MaxNonce: BigInt = BigInt(2).pow(64) - 1
```
AnastasiiaL, line 194: these numbers definitely need some explanation

AnastasiiaL (commented):
Looks good to me. Just minor comments about the code that was (probably) already there, but since we are touching these files - might as well improve it a bit

bsuieric:
I saw

bsuieric:
> > So we don't use it for now, just for test purposes? in ECIP-1049 I see there is an activation block
> 
> Where did you see that? I have the idea that that it is still not decided, on which block to do the switch, for mainnet. Or even if there's a final decision to really go forward with this
> 
> I don't understand what do you mean with "is this just for testing purposes?". Everything has to be tested before going to live (there's a test network called Astor that only mines with keccak, we will add support for that on ETCM-749)

I saw isKeccak is always false based on the configuration

leo-bogastry:
> > > So we don't use it for now, just for test purposes? in ECIP-1049 I see there is an activation block
> > 
> > 
> > Where did you see that? I have the idea that that it is still not decided, on which block to do the switch, for mainnet. Or even if there's a final decision to really go forward with this
> > I don't understand what do you mean with "is this just for testing purposes?". Everything has to be tested before going to live (there's a test network called Astor that only mines with keccak, we will add support for that on ETCM-749)
> 
> I saw isKeccak is always false based on the configuration

Yes, at the moment yes. It will only be tested in Astor for now

bsuieric (approved):

src/main/scala/io/iohk/ethereum/consensus/pow/EthashDAGManager.scala
leo-bogastry: This code is considered stable and working, it's complete out of the current scope doing any modification to it.
In the best case I can add val to replace the magic numbers you mentioned. The implementation of the ethash computation is based on this: https://eth.wiki/concepts/ethash/ethash

src/main/scala/io/iohk/ethereum/consensus/pow/EthashMiner.scala
leo-bogastry: I am placing 64 in a val, in ETCM-759, that touches this class

src/it/scala/io/iohk/ethereum/txExecTest/ECIP1017Test.scala
```diff
@@ -49,7 +49,8 @@ class ECIP1017Test extends AnyFlatSpec with Matchers {
       ecip1098BlockNumber = Long.MaxValue,
       treasuryAddress = Address(0),
       ecip1097BlockNumber = Long.MaxValue,
-      ecip1099BlockNumber = Long.MaxValue
+      ecip1099BlockNumber = Long.MaxValue,
+      ecip1049BlockNumber = None
```
dzajkowski, line 53: how does this translate to the configuration? if by default the "ecip1049-block-number" is set to 14796 (example) how would I remove it in a child config to get "None" as a result in the code? or is this only for the situation that the parameter was never defined?

src/main/scala/io/iohk/ethereum/consensus/pow/KeccakCalculation.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  /**
+    * Validates if mixHash <= 2^256 / difficulty
+    * @param mixHash
+    * @param difficulty
+    * @return boolean indicating whether PoW is valid or not
+    */
+  def isMixHashValid(mixHash: ByteString, difficulty: BigInt): Boolean = {
+    val mixHashInt = BigInt.apply(mixHash.toArray)
+    val threshold = BigInt(2).pow(256) / difficulty
```
dzajkowski: `BigInt(2).pow(256)` shouldn't this be a static value?

src/main/scala/io/iohk/ethereum/consensus/pow/EthashMiner.scala
```diff
@@ -0,0 +1,205 @@
+package io.iohk.ethereum.consensus
```
dzajkowski, line 1: a note on commit organisation: this would be way more readable if the package change was done in a previous commit and the code changes were done in this one. this way the actual code change would be highlighted and not the whole class.

src/main/scala/io/iohk/ethereum/consensus/pow/KeccakCalculation.scala
leo-bogastry: Yes, good point

src/main/scala/io/iohk/ethereum/consensus/pow/EthashMiner.scala
leo-bogastry: 👍

src/it/scala/io/iohk/ethereum/txExecTest/ECIP1017Test.scala
leo-bogastry: I think the answer is "this only for the situation that the parameter was never defined", but I am not sure that I understood the question.
At the moment in `etc-chain.conf` the configuration for this ECIP is added but all commented out:
https://github.com/input-output-hk/mantis/blob/feature/ETCM-746-support-keccak256-pow/src/main/resources/conf/chains/etc-chain.conf#L99-L102
That translates to `None`. 

For Astor it will be `ecip1049-block-number = "0"` because it uses Keccak from genesis block and for other networks it should be a specific block number (determining the switch)

dzajkowski: understood. my point is that the current flow is that if the flag is not set it defaults to Long.MaxValue. maybe changing this flow without addressing the other flags will make the code confusing for other devs? IOW maybe this change can be done as an overall improvement, and avoid this interim where there are mixed flows

leo-bogastry: This flag does not default to `Long.MaxValue`, It defaults to `None`.

dzajkowski: that's true. but every other flag of this type defaults to `Long.MaxValue`

leo-bogastry: Ok, I got what you mean. If you agree with me that defaulting to `None` is better than defaulting to `Long.MaxValue` I can create a ticket to address the issue. Or are you thinking an even more deep refactoring?

dzajkowski: please capture this in a ticket and we can come back to it. I have some ideas but I would also like to hear yours (IMO None isn't the solution but I'm open to a discussion about it).

leo-bogastry: I added ticket ETCM-794

leo-bogastry:
I realized that the tests for the EthashMiner never run by default. I forced then to run locally and all passed

PR #961, opened 2021-04-09
ETCM-732: Handle missing state node in regular sync after fast sync is done

milenkara:
# Description

Instead of handling failures of a `Task`, `BlockImporter` now handles appropriate error that indicates that block could not be imported due to a missing state node. 

# Proposed Solution

Since the error is now wrapped into instance of an `Either`, using pattern matching we can act on a newly introduced error type `BlockImportFailedDueToMissingNode`

# Testing

Issue cannot be reproduced on every run of fast sync. Once it is reproduced, observe that missing state node error is logged (during regular sync) and that `BlockFetcher` will try to fetch missing state node, after which regular sync continues as usual without any errors.

### DIFF
--- a/src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
+++ b/src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
@@ -14,13 +14,13 @@ import io.iohk.ethereum.domain._
 import io.iohk.ethereum.mpt.MerklePatriciaTrie
 import io.iohk.ethereum.utils.Config.SyncConfig
 import io.iohk.ethereum.utils.Config
-import io.iohk.ethereum.{Fixtures, Mocks, ObjectGenerators, crypto}
+import io.iohk.ethereum.{Fixtures, Mocks, NormalPatience, ObjectGenerators, Timeouts, crypto}
 import io.iohk.ethereum.ledger.Ledger.BlockResult
 import monix.execution.Scheduler
 import org.scalamock.scalatest.MockFactory
 import org.scalatest.BeforeAndAfterAll
-import org.scalatest.concurrent.Eventually.eventually
-import org.scalatest.flatspec.AsyncFlatSpecLike
+import org.scalatest.concurrent.Eventually
+import org.scalatest.flatspec.AnyFlatSpecLike
 import org.scalatest.matchers.should.Matchers
 
 import scala.concurrent.duration._
@@ -28,9 +28,11 @@ import scala.concurrent.duration._
 class BlockImporterItSpec
     extends MockFactory
     with TestSetupWithVmAndValidators
-    with AsyncFlatSpecLike
+    with AnyFlatSpecLike
     with Matchers
-    with BeforeAndAfterAll {
+    with BeforeAndAfterAll
+    with Eventually
+    with NormalPatience {
 
   implicit val testScheduler = Scheduler.fixedPool("test", 32)
 
@@ -162,7 +164,7 @@ class BlockImporterItSpec
 
     blockImporter ! BlockFetcher.PickedBlocks(NonEmptyList.fromListUnsafe(newBranch))
 
-    eventually { Thread.sleep(200); blockchain.getBestBlock().get shouldEqual newBlock3 }
+    eventually { blockchain.getBestBlock().get shouldEqual newBlock3 }
   }
 
   it should "return Unknown branch, in case of PickedBlocks with block that has a parent that's not in the chain" in {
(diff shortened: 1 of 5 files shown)


src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
```diff
@@ -240,4 +240,39 @@ class BlockImporterItSpec
...
+        pendingTransactionsManagerProbe.ref,
+        supervisor.ref
+      )
+    )
+
+    blockImporter ! BlockImporter.Start
+    blockImporter ! BlockFetcher.PickedBlocks(NonEmptyList.fromListUnsafe(List(invalidBlock)))
+
+    eventually {
+      Thread.sleep(1000)
```
leo-bogastry: Why the `Thread.sleep` if you are using `eventually`?

milenkara: I saw the `Thread.sleep` in some of the existing tests, so I reused it here. I will try to remove them and define a custom akka `Patience` for the `eventually`

milenkara: I have tried to use `eventually` without any `Thread.sleep`s and defining implicit/explicit `Patience` but for some reason it does not re-run the test if it fails. This is probably the reason that `Thread.sleep`s are introduced.

grale88 (approved):

src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
```diff
@@ -150,7 +152,7 @@ class BlockImporterItSpec
     blockImporter ! BlockFetcher.PickedBlocks(NonEmptyList.fromListUnsafe(newBranch))
 
     //because the blocks are not valid, we shouldn't reorganise, but at least stay with a current chain, and the best block of the current chain is oldBlock4
-    eventually { blockchain.getBestBlock().get shouldEqual oldBlock4 }
+    eventually {Thread.sleep(1000); blockchain.getBestBlock().get shouldEqual oldBlock4 }
```
dzajkowski: wouldn't it be cleaner to adjust the `implicit def patienceConfig: PatienceConfig = defaultPatienceConfig`?

src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
```diff
@@ -161,7 +165,11 @@ class BlockImport(
     reorgResult match {
       case Some(execResult) =>
         execResult.fold(
-          err => BlockImportFailed(s"Error while trying to reorganise chain: $err"),
+          {
+            case MPTError(reason) if reason.isInstanceOf[MissingNodeException] =>
+              BlockImportFailedDueToMissingNode(reason.asInstanceOf[MissingNodeException])
+            case err => BlockImportFailed(s"Error while trying to reorganise chain: $err")
+          },
```
dzajkowski, line 174: this could also be written as:
```scala
          {
            case MPTError(reason: MissingNodeException) => BlockImportFailedDueToMissingNode(reason)
            case err => BlockImportFailed(s"Error while trying to reorganise chain: $err")
          },
```

src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
milenkara: @dzajkowski I have tried both mixing our `LongPatience` trait into the test, and defining patience config directly before the `eventually` block. From the debug, I see that the my values are used, but for some reason method is not re-run.

src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
milenkara: Done.

src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
dzajkowski: interesting, let me give it a look

dzajkowski: `eventually` does not mix well with `AsyncFlatSpecLike`. either a `Future` and `AsyncFlatSpecLike` or `AnyFlatSpecLike` with `eventually`

milenkara: @dzajkowski  Done.

leo-bogastry (approved):

PR #962, opened 2021-04-09
WIP: WIP: WIP: experiment with a different way of saving the blockchain

bsuieric:
# Description

just an experiment that goes about saving all the blocks in the blockchain(having the tip always known - main chain) !!?? pretty nasty - work in progress

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
@@ -1069,8 +1069,8 @@ class FastSync(
         val bestReceivedBlock = fullBlocks.maxBy(_.number)
         val lastStoredBestBlockNumber = appStateStorage.getBestBlockNumber()
         if (lastStoredBestBlockNumber < bestReceivedBlock.number) {
-          blockchain.saveBestKnownBlocks(bestReceivedBlock.number)
           appStateStorage.putBestBlockNumber(bestReceivedBlock.number).commit()
+          appStateStorage.putBestBlockHash(bestReceivedBlock.hash)
         }
         syncState = syncState.copy(lastFullBlockNumber = bestReceivedBlock.number.max(lastStoredBestBlockNumber))
       }
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
@@ -66,7 +66,7 @@ class BlockImporter(
     case PrintStatus => log.info("Block: {}, is on top?: {}", blockchain.getBestBlock().get.number, state.isOnTop)
 
     case BlockFetcher.PickedBlocks(blocks) =>
-//      SignedTransaction.retrieveSendersInBackGround(blocks.toList.map(_.body))
+      SignedTransaction.retrieveSendersInBackGround(blocks.toList.map(_.body))
       importBlocks(blocks, DefaultBlockImport)(state)
 
     //TODO ETCM-389: Handle mined, checkpoint and new blocks uniformly
@@ -117,7 +117,7 @@ class BlockImporter(
     running(state.resolvingBranch(from))
 
   private def start(): Unit = {
-    log.debug("Starting Regular Sync, current best block is {}", startingBlockNumber)
+    log.info("Starting the incredible Regular Sync, current best block is {}", startingBlockNumber)
     fetcher ! BlockFetcher.Start(self, startingBlockNumber)
     supervisor ! ProgressProtocol.StartingFrom(startingBlockNumber)
     context become running(ImporterState.initial)
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
@@ -63,10 +63,10 @@ class BlockImporter(
   private def running(state: ImporterState): Receive = handleTopMessages(state, running) orElse {
     case ReceiveTimeout => self ! PickBlocks
 
-    case PrintStatus => log.info("Block: {}, is on top?: {}", blockchain.getBestBlockNumber(), state.isOnTop)
+    case PrintStatus => log.info("Block: {}, is on top?: {}", blockchain.getBestBlock().get.number, state.isOnTop)
 
     case BlockFetcher.PickedBlocks(blocks) =>
-      SignedTransaction.retrieveSendersInBackGround(blocks.toList.map(_.body))
+//      SignedTransaction.retrieveSendersInBackGround(blocks.toList.map(_.body))
       importBlocks(blocks, DefaultBlockImport)(state)
 
     //TODO ETCM-389: Handle mined, checkpoint and new blocks uniformly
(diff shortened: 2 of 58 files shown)


src/main/scala/io/iohk/ethereum/ledger/BlockExecution.scala
```diff
@@ -62,12 +62,13 @@ class BlockExecution(
         .getBlockHeaderByHash(block.header.parentHash)
         .toRight(MissingParentError) // Should not never occur because validated earlier
       execResult <- executeBlockTransactions(block, parent)
-      worldToPersist <- Either
-        .catchOnly[MPTException](blockPreparator.payBlockReward(block, execResult.worldState))
-        .leftMap(BlockExecutionError.MPTError.apply)
+//      worldToPersist <- Either
+//        .catchOnly[MPTException](blockPreparator.payBlockReward(block, execResult.worldState))
+//        .leftMap(BlockExecutionError.MPTError.apply)
```
dzajkowski: if we remove this step the chain will become not-sane. block rewards are a crucial part of validation and I'm a bit worried that the app will not build a chain because of this. why can't this be left in tact?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -121,7 +121,7 @@ class LedgerImpl(
       block: Block
   )(implicit blockExecutionScheduler: Scheduler): Task[BlockImportResult] = {
     val bestBlockHash = blockchain.getBestBlockHash()
-    blockchain.getChainWeightByHash(bestBlockHash) match {
+    blockchain.getChainWeightByHash(block.header.parentHash) match {
```
dzajkowski, line 124: is omitting the current block a desired step?

bsuieric: the current block is being processed, and the weight will be increased in BlockExecution based on what we pass as parentWeight

src/main/scala/io/iohk/ethereum/ledger/BlockExecution.scala
bsuieric: I removed it because the test was failing, worlds state starts with an empty state root hash. Will add it back to sync/test to etc and later fix the test

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -1069,8 +1069,8 @@ class FastSync(
         val bestReceivedBlock = fullBlocks.maxBy(_.number)
         val lastStoredBestBlockNumber = appStateStorage.getBestBlockNumber()
         if (lastStoredBestBlockNumber < bestReceivedBlock.number) {
-          blockchain.saveBestKnownBlocks(bestReceivedBlock.number)
           appStateStorage.putBestBlockNumber(bestReceivedBlock.number).commit()
+          appStateStorage.putBestBlockHash(bestReceivedBlock.hash)
```
dzajkowski, line 1073: wouldn't a `.commit()` step here align with the previous line?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -63,10 +63,10 @@ class BlockImporter(
   private def running(state: ImporterState): Receive = handleTopMessages(state, running) orElse {
     case ReceiveTimeout => self ! PickBlocks
 
-    case PrintStatus => log.info("Block: {}, is on top?: {}", blockchain.getBestBlockNumber(), state.isOnTop)
+    case PrintStatus => log.info("Block: {}, is on top?: {}", blockchain.getBestBlock().get.number, state.isOnTop)
 
     case BlockFetcher.PickedBlocks(blocks) =>
-      SignedTransaction.retrieveSendersInBackGround(blocks.toList.map(_.body))
+//      SignedTransaction.retrieveSendersInBackGround(blocks.toList.map(_.body))
```
dzajkowski: this is only for caching, right?

src/test/scala/io/iohk/ethereum/Fixtures.scala
```diff
@@ -130,7 +130,7 @@ object Fixtures {
         transactionsRoot = ByteString(Hex.decode("56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")),
         receiptsRoot = ByteString(Hex.decode("56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")),
         logsBloom = ByteString(Hex.decode("0" * 512)),
-        difficulty = BigInt("17179869184"),
+        difficulty = BigInt("1"),
```
dzajkowski, line 133: how does this affect the tests? seems arbitrary and not related to how best block is stored?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -119,32 +119,21 @@ class LedgerImpl(
...
-              importResult
-            case None =>
-              log.error(s"Getting total difficulty for current best block with hash: $hash failed")
-              Task.now(BlockImportFailed(s"Couldn't get total difficulty for current best block with hash: $hash"))
-          }
-        }
+  )(implicit blockExecutionScheduler: Scheduler): Task[BlockImportResult] = {
+    val bestBlockHash = blockchain.getBestBlockHash()
+    blockchain.getChainWeightByHash(bestBlockHash) match {
+      case Some(weight) =>
```
dzajkowski, line 125: wouldn't it make sense to still skip duplicate blocks?

dzajkowski:
it's interesting that the switch to block instead of number made a lot of code simpler

src/main/scala/io/iohk/ethereum/ledger/BlockExecution.scala
dzajkowski: it's not a prio to have units passing. I prefer to see a successful run with mainnet, and start poking at checkpointing.

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
dzajkowski: so this is a straight up bug?

bsuieric: no, in the previous case, importToTop means that currentBestBlock is the parent. And in case of reorganisation, we have to keep the weight of the bestBlock around, so we can check if another branch has a higher wieght

src/test/scala/io/iohk/ethereum/Fixtures.scala
bsuieric: it's not related, this is again a change that I had to do to run a test. The genesis block difficulty was higher than the block that were being added to the blockchain(could have had just increase the blocks difficulty instead). I would say it's not important at the moment, and can change it back

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
bsuieric: yes, but I didn't consider at all FastSync. For now we just work regular Sync

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
bsuieric: yes, it is. Would add it back when this go to staging

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
bsuieric: yes, it would make sense. Thanks

PR #965, opened 2021-04-14
ETCM-680 retesteth in Nix and on CI

jvdp:
Replaces the custom ETS implementation by retesteth, installed via Nix.

The new readme describes how to use it: https://github.com/input-output-hk/mantis/blob/feature/ETCM-680-retesteth-ci/ets/README.md (Also have a look at buildkite for annotations and output.)


nix-in-docker/run
```diff
@@ -0,0 +1,18 @@
...
+  --network="host" \
+  --rm \
+  -e BUILDKITE=true \
+  -v=$dir/..:/mantis \
+  -w=/mantis \
+  -v=mantis-root:/root \
+  -v=mantis-ops-nix:/nix \
+  -v=$dir/nix.conf:/etc/nix/nix.conf \
+  nixpkgs/nix-unstable:latest \
+  nix-shell "$@"
```
dzajkowski, line 18: any chance for a mention in the README?

nix/default.nix
```diff
@@ -1,27 +1,34 @@
...
-      src = cleanedSrc;
+      retesteth = final.callPackage ./retesteth.nix { };
+      lllc = final.callPackage ./lllc.nix { };
     };
-  };
+
   sbt-derivation-overlay = import sources.sbt-derivation;
-in import sources.nixpkgs {
+in
+import sources.nixpkgs {
```
dzajkowski, line 31: are the reformats coming from a linter? maybe those could be done in a different PR? maybe applied consistently to the whole codebase?

jvdp: Yeah it's coming from this: https://github.com/nix-community/nixpkgs-fmt

Which is baked into the Nix IDE package for VS Code. I'll make a separate PR.

nix/lllc.nix
```diff
@@ -0,0 +1,47 @@
...
+  postPatch = ''
+    substituteInPlace cmake/jsoncpp.cmake \
+      --replace "${jsoncppURL}" ${jsoncpp}
+  '';
+
+  preConfigure = ''
+    echo ${version} > commit_hash.txt
+  '';
+
+  # nativeBuildInputs = [ cacert cmake gcc gnumake perl psmisc ];
```
dzajkowski: is there a point to this comment?

src/ets/resources/config/mantis/genesis/Berlin.json
```diff
@@ -0,0 +1,15 @@
...
+       "byzantiumForkBlock" : "0x00",
+       "constantinopleForkBlock" : "0x00",
+       "constantinopleFixForkBlock" : "0x00",
+       "istanbulForkBlock" : "0x00",
+       "berlinForkBlock" : "0x00",
+       "chainID" : "0x01"
+    },
+    "accounts" : {
+    }
+}
```
dzajkowski: just a note: mantis does not support Berlin

test-ets.sh
```diff
@@ -1,32 +1,52 @@
...
-    else
-        echo "running a subset of ETS"
-        $SBT "ets:testOnly *VMSuite -- -Dexg=vmPerf*" &&
-        $SBT "ets:testOnly *BlockchainSuite -- -DuseLocalVM=true -Ding=bcForkStress*,bcMulti*,bcState*,bcTotalDiff*,bcValidBlock*,Transition*";
-    fi
-else
-    echo "SKIP_ETS_TESTS variable is set - skipping the tests";
-fi
+git submodule init
+git submodule update
```
dzajkowski, line 4: I think we need to automate this for the repository

test-ets.sh
```diff
@@ -1,32 +1,52 @@
...
+
+echo "booting Mantis and waiting for RPC API to be up"
+$SBT -Dconfig.file=./src/main/resources/conf/testmode.conf run &> mantis-log.txt &
+
+while ! nc -z localhost 8546; do   
+  sleep 0.1
```
dzajkowski, line 10: suggestion: we have a GET '/healthcheck' endpoint

dzajkowski: not for this PR but might prove more reliable

dzajkowski (approved):
LGTM. maybe @nrdxp could give this a look?

src/ets/resources/config/mantis/genesis/Berlin.json
jvdp: I just copied these from the `besu` config. At some point we are supposed to upstream the config to https://github.com/ethereum/retesteth/tree/master/retesteth/configs/clientconfigs I think.

test-ets.sh
jvdp: How?

test-ets.sh
jvdp: Not sure if that's a good fit. Giving it a try right now locally in testmode I get a status 500 response:

```json
{
  "checks":[
    {
      "description":"listening",
      "status":"OK"
    },
    {
      "description":"peerCount",
      "status":"OK"
    },
    {
      "description":"earliestBlock",
      "status":"OK"
    },
    {
      "description":"latestBlock",
      "status":"OK"
    },
    {
      "description":"pendingBlock",
      "status":"ERROR",
      "error":"java.util.NoSuchElementException: None.get"
    }
  ]
}
```

jvdp: To clarify: when something breaks this health check endpoint the result would be a stuck CI build.

nix/default.nix
jvdp: #973

nix/lllc.nix
```diff
@@ -0,0 +1,45 @@
...
+stdenv.mkDerivation rec {
+
+  pname = "lllc";
+  version = "14c9d5de6c7e58f1d1b18a04b57ec9f190d7cd2f";
+
+  src = builtins.fetchurl {
+    url = "https://github.com/winsvega/solidity/archive/${version}.tar.gz";
+    sha256 = "0s6nz0w8fr1347d05kx4z3z72b1k0kmcz6d4dwwc24rydjznkw6r";
+  };
+
```
jonringer, line 18: pretty sure this wasnt necessary for building retesteth

jvdp: The whole dependency you mean? It is a runtime dependency of retesteth (looked up via $PATH.)

https://github.com/ethereum/retesteth/blob/master/Dockerfile#L21

test-ets.sh
dzajkowski: good question. I think that easiest would be a task that runs every so often and checks if the modules are out of date and creates a PR is they are. I don't think this can be done now, just making a mental note.

btw I meant keeping the submodules in sync (`git submodule update --remote --merge`) and I see why this comment caused confusion. apologies

test-ets.sh
dzajkowski: isn't the same true for this while loop if the port does not become open?

jvdp: Yes, but the actual functionality (eg. the pending block) might well be broken, which the healthcheck will test. Only the HTTP server needs to be up, and after that it will fail quickly (or succeed testing, hopefully ;) )

test-ets.sh
jvdp: Ah, yeah, these commands only pull in the submodules in the checked out repo.

Updating those submodules might introduce breaking changes. Something like Scala Steward for this would be ideal...

test-ets.sh
dzajkowski: clear.

bsuieric (approved):

PR #966, opened 2021-04-15
ETS integration RPC endpoints

grale88:
# Description

Initial integration of ETS (Ethereum tests) executed through Retesteth test suite.

# Proposed Solution

Some of existing endpoints were updated according to newer changes in the runner and the tests. Additional methods were added, such as test_importRawBlock, test_getLogHash, debug_accountRangeAt and debug_storageRangeAt. Follow up tasks/bugs are created in order to ensure further progress and fixes for the tests that are failing.

Modifications were made to standard eth RPC responses because of Mantis custom fields that are avalable only when "testmode" is configured.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
@@ -90,13 +90,13 @@ class GenesisDataLoader(blockchain: Blockchain, blockchainConfig: BlockchainConf
     val initalRootHash = MerklePatriciaTrie.EmptyRootHash
 
     val stateMptRootHash = genesisData.alloc.zipWithIndex.foldLeft(initalRootHash) {
-      case (rootHash, (((address, genesisAccount), idx))) =>
+      case (rootHash, ((address, genesisAccount), _)) =>
         val mpt = MerklePatriciaTrie[Array[Byte], Account](rootHash, storage)
         val paddedAddress = address.reverse.padTo(addressLength, "0").reverse.mkString
         val stateRoot = mpt
           .put(
             crypto.kec256(Hex.decode(paddedAddress)),
-            Account(nonce = genesisAccount.nonce, balance = genesisAccount.balance, codeHash = genesisAccount.code)
+            Account(nonce = genesisAccount.nonce, balance = genesisAccount.balance)
           )
           .getRootHash
         stateRoot
--- a/src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
@@ -201,12 +201,7 @@ object GenesisDataLoader {
     }
 
     object UInt256JsonSerializer
-        extends CustomSerializer[UInt256](formats =>
-          (
-            { case jv => deserializeUint256String(jv) },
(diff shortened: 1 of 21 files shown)


src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
```diff
@@ -178,5 +191,23 @@ object GenesisDataLoader {
...
+      case other => throw new RuntimeException("Expected hex string, but got: " + other)
+    }
+
+    object UInt256JsonSerializer
+        extends CustomSerializer[UInt256](formats =>
+          (
+            { case jv => deserializeUint256String(jv) },
+            PartialFunction.empty
+          )
+        )
```
dzajkowski: formatting seems off here

src/main/scala/io/iohk/ethereum/jsonrpc/server/controllers/JsonRpcBaseController.scala
```diff
@@ -74,6 +76,7 @@ trait JsonRpcBaseController {
             JsonRpcControllerMetrics.recordMethodTime(request.method, time)
           }
       }
+      .flatTap { response => Task { log.debug(s"sending response ${response.inspect}") } }
```
dzajkowski, line 79: note: find out why there's no `tap`

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
```diff
@@ -152,16 +268,101 @@ class TestService(
...
+    )
+  }
+
+  def storageRangeAt(request: StorageRangeRequest): ServiceResponse[StorageRangeResponse] = {
+
+    val blockOpt = request.parameters.blockHashOrNumber
+      .fold(number => blockchain.getBlockByNumber(number), hash => blockchain.getBlockByHash(hash))
+
+    if (blockOpt.isEmpty) {
+      Task.now(Right(AccountsInRangeResponse(Map(), ByteString(0))))
```
dzajkowski: @jvdp do you know why this task is here? am I missing a side-effect?

jvdp: This has to be a bug and should be `return`ed, I'm pretty sure. It would make the `.get`s below safe.

dzajkowski (approved):
I don't like the intrusive paths taken (like in BlockPreaprator) but let's get this merged and move forward. Hopefully we can modularise the app and have such modules be `testmode` specific.

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
```diff
@@ -164,7 +164,7 @@ class TestService(
     accountAddresses = genesisData.alloc.keys.toList
     accountRangeOffset = 0
 
-    Task.now(Right(SetChainParamsResponse()))
+    SetChainParamsResponse().rightNow
```
jvdp, line 167: Being unfamiliar to this I find this worse, somehow... but it's purely subjective.

jvdp: Also, I'm kinda surprised that the core RPC methods aren't overloaded for 'bare' (non-`Task`) responses. Should be possible, right?

jvdp (approved):

PR #974, opened 2021-04-26
Bug/etcm 797

bsuieric:
# Description

https://jira.iohk.io/browse/ETCM-797

# Proposed Solution

calculate start epoch block, and calculate seed based on the epoch length and start epoch block

### DIFF
--- a/src/main/scala/io/iohk/ethereum/consensus/pow/EthashDAGManager.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/pow/EthashDAGManager.scala
@@ -13,11 +13,11 @@ class EthashDAGManager(blockCreator: EthashBlockCreator) extends Logger {
   var currentEpochDagSize: Option[Long] = None
   var currentEpochDag: Option[Array[Array[Int]]] = None
 
-  def calculateDagSize(blockNumber: Long, epoch: Long): (Array[Array[Int]], Long) = {
+  def calculateDagSize(blockNumber: Long, epoch: Long, ecip1099ActivationBlock: Long): (Array[Array[Int]], Long) = {
     (currentEpoch, currentEpochDag, currentEpochDagSize) match {
       case (Some(`epoch`), Some(dag), Some(dagSize)) => (dag, dagSize)
       case _ =>
-        val seed = EthashUtils.seed(blockNumber)
+        val seed = EthashUtils.seed(blockNumber, ecip1099ActivationBlock)
         val dagSize = EthashUtils.dagSize(epoch)
         val dagNumHashes = (dagSize / EthashUtils.HASH_BYTES).toInt
         val dag =
--- a/src/main/scala/io/iohk/ethereum/consensus/pow/EthashDAGManager.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/pow/EthashDAGManager.scala
@@ -13,11 +13,11 @@ class EthashDAGManager(blockCreator: EthashBlockCreator) extends Logger {
   var currentEpochDagSize: Option[Long] = None
   var currentEpochDag: Option[Array[Array[Int]]] = None
 
-  def calculateDagSize(blockNumber: Long, epoch: Long, ecip1099ActivationBlock: Long): (Array[Array[Int]], Long) = {
+  def calculateDagSize(blockNumber: Long, epoch: Long): (Array[Array[Int]], Long) = {
     (currentEpoch, currentEpochDag, currentEpochDagSize) match {
       case (Some(`epoch`), Some(dag), Some(dagSize)) => (dag, dagSize)
       case _ =>
-        val seed = EthashUtils.seed(blockNumber, ecip1099ActivationBlock)
+        val seed = EthashUtils.seed(blockNumber, blockCreator.blockchainConfig.ecip1099BlockNumber.toLong)
         val dagSize = EthashUtils.dagSize(epoch)
         val dagNumHashes = (dagSize / EthashUtils.HASH_BYTES).toInt
         val dag =
--- a/src/main/scala/io/iohk/ethereum/consensus/pow/EthashUtils.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/pow/EthashUtils.scala
@@ -63,11 +63,17 @@ object EthashUtils {
 
   private def epochBeforeEcip1099(blockNumber: Long): Long = blockNumber / EPOCH_LENGTH_BEFORE_ECIP_1099
 
-  def seed(blockNumber: Long): ByteString = {
-    val epoch = epochBeforeEcip1099(blockNumber)
+  def seed(blockNumber: Long, ecip1099ActivationBlock: Long): ByteString = {
+    val epoch: Long = epochBeforeEcip1099(startBlock(blockNumber, ecip1099ActivationBlock))
     (BigInt(0) until epoch)
       .foldLeft(ByteString(Hex.decode("00" * 32))) { case (b, _) => kec256(b) }
   }
+
+  def startBlock(blockNumber: Long, ecip1099ActivationBlock: Long): Long =
+    (blockNumber / calcEpochLength(blockNumber, ecip1099ActivationBlock)) * calcEpochLength(
+      blockNumber,
+      ecip1099ActivationBlock
+    ) + 1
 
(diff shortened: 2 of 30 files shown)


src/main/scala/io/iohk/ethereum/jsonrpc/DebugService.scala
```diff
@@ -41,4 +44,27 @@ class DebugService(peerManager: ActorRef, etcPeerManager: ActorRef) {
...
+              GetBlockHeaders(Left(req.number - 1), 2, 0, false),
+              peerId
+            ),
+            peerManager
+          )
+        }
+      }
+    } yield Right(GetSpecificBlockResponse())
+
+//    Task(Right(GetSpecificBlockResponse()))
```
bsuieric: remove comment

src/main/scala/io/iohk/ethereum/jsonrpc/DebugService.scala
```diff
@@ -41,4 +44,27 @@ class DebugService(peerManager: ActorRef, etcPeerManager: ActorRef) {
       .askFor[PeerInfoResponse](EtcPeerManagerActor.PeerInfoRequest(peer))
       .map(resp => resp.peerInfo)
   }
+
+  def getSpecificBlock(req: GetSpecificBlockRequest): ServiceResponse[GetSpecificBlockResponse] = {
```
bsuieric: 👌

jvdp (approved):

src/main/scala/io/iohk/ethereum/consensus/pow/EthashUtils.scala
```diff
@@ -63,12 +63,18 @@ object EthashUtils {
...
 
-  def seed(blockNumber: Long): ByteString = {
-    val epoch = epochBeforeEcip1099(blockNumber)
+  def seed(blockNumber: Long, ecip1099ActivationBlock: Long): ByteString = {
+    val epoch: Long = epochBeforeEcip1099(startBlock(blockNumber, ecip1099ActivationBlock))
     (BigInt(0) until epoch)
       .foldLeft(ByteString(Hex.decode("00" * 32))) { case (b, _) => kec256(b) }
   }
 
+  def startBlock(blockNumber: Long, ecip1099ActivationBlock: Long): Long =
```
dzajkowski: this is quite confusing - why such a convoluted computation?
can `calcEpochLength(blockNumber, ecip1099ActivationBlock)` change 
in this multiplication `calcEpochLength(blockNumber, ecip1099ActivationBlock)) * calcEpochLength(blockNumber, ecip1099ActivationBlock)`?

perhaps
```scala
val epochLength = calcEpochLength(blockNumber, ecip1099ActivationBlock)
((blockNumber / epochLength) * epochLength) + 1
```
or maybe even more explicitly:
```scala
val epochLength = calcEpochLength(blockNumber, ecip1099ActivationBlock)
val epoch: Long = blockNumber / epochLength
(epoch * epochLength + 1) / EPOCH_LENGTH_BEFORE_ECIP_1099
```
to avoid this indirection:
`epochBeforeEcip1099(startBlock(blockNumber, ecip1099ActivationBlock))`
?

jvdp: Deduplicating that `calcEpochLength(...)` call sounds good, but I'm not sure your other proposals are equivalent.

src/main/scala/io/iohk/ethereum/consensus/pow/EthashDAGManager.scala
```diff
@@ -13,11 +13,11 @@ class EthashDAGManager(blockCreator: EthashBlockCreator) extends Logger {
   var currentEpochDagSize: Option[Long] = None
   var currentEpochDag: Option[Array[Array[Int]]] = None
 
-  def calculateDagSize(blockNumber: Long, epoch: Long): (Array[Array[Int]], Long) = {
+  def calculateDagSize(blockNumber: Long, epoch: Long, ecip1099ActivationBlock: Long): (Array[Array[Int]], Long) = {
```
dzajkowski: wouldn't it make sense to make this parameter a characteristic of the manager itself? this cannot change during the run, and passing it all around seems tedious.

src/main/scala/io/iohk/ethereum/consensus/pow/EthashUtils.scala
bsuieric: I agree, maybe also the name should change to startEpochBlock, or add a comment, so we have more context when we look at it

src/main/scala/io/iohk/ethereum/jsonrpc/DebugJsonMethodsImplicits.scala
```diff
@@ -12,4 +19,19 @@ object DebugJsonMethodsImplicits extends JsonMethodsImplicits {
       def encodeJson(t: ListPeersInfoResponse): JValue =
         JArray(t.peers.map(a => JString(a.toString)))
     }
+
+  implicit val debug_getSpecificBlockRequest: JsonMethodCodec[GetSpecificBlockRequest, GetSpecificBlockResponse] =
```
dzajkowski: I think that this endpoint shouldn't be part of this PR.
1) it's not documented or tested
2) it's not part of the bugfix
3) will make it harder to backport to a release branch if the need arises

I'm fine with having this come from a ticket that's fleshed out and has its parameters properly identified, perhaps during a refinement session.

src/test/scala/io/iohk/ethereum/consensus/pow/EthashUtilsSpec.scala
```diff
@@ -18,8 +20,21 @@ class EthashUtilsSpec extends AnyFlatSpec with Matchers with ScalaCheckPropertyC
   val ecip1099forkBlockNumber: Long = 11460000
 
   "Ethash" should "generate correct hash" in {
-    forAll(Gen.choose[Long](0, 15000000L)) { blockNumber =>
-      seed(blockNumber, ecip1099forkBlockNumber) shouldBe seedForBlockReference(blockNumber)
+    val seedEpoch0 = ByteStringUtils.string2hash("0000000000000000000000000000000000000000000000000000000000000000")
+    val seedEpoch1 = ByteStringUtils.string2hash("290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563")
+    val seedEpoch382 = ByteStringUtils.string2hash("d3d0aa11197dcdcfcb3ad3c73d415af47299bddb47fda6081d31d9dd06462f6a")
+    val seedEpoch383 = ByteStringUtils.string2hash("bf532874eb434842e7a3e4acd113fe454541651872760d9b95d11d7f90ca25dc")
```
dzajkowski, line 26: where did these hashes come from? maybe this would be a good inspiration https://github.com/iquidus/ecip-1099-data ?

dzajkowski (changes requested):
Please remove the "debug_" endpoint. We can groom it and bring it in properly.

src/test/scala/io/iohk/ethereum/consensus/pow/EthashUtilsSpec.scala
jvdp: These are the seeds that correspond to the blocks, calculated with the Mantis code found to be working properly (valid blocks are valid.)

But the strings themselves are not so interesting for the purposes of this bug, just that they are the same across different blocks within a (ECIP-1099) epoch. (For the implementation itself the hashes are important of course.)

jvdp: (Also, the hash being all zeroes for epoch 0 is a good sign?)

src/main/scala/io/iohk/ethereum/jsonrpc/DebugJsonMethodsImplicits.scala
jvdp: Fair enough. I think this endpoint might be useful for state sync improvements, and we'll be able to grab it from this PR when that's the case.

src/main/scala/io/iohk/ethereum/consensus/pow/EthashUtils.scala
jvdp: I merged it all into the `seed` function, since it's only used there anyway.

dzajkowski (approved):

PR #979, opened 2021-05-05
[ETCM-759] refactor mining

leo-bogastry:
# Description
In ETCM-746 a temporary modification in EthashMiner was introduced to support the new PoW mining algorithm with Keccak mining
In order to keep a separation of concerns, this PR introduces the following refactoring:
- Introduced a PoWMiningCoordinator that will assert, per each block to be mined, if mining should to be done with ethash or keccak algorithm
- Simplified EthashMiner to only mine on demand (managed by PoWMiningCoordinator)
- Extracted from EthashMiner the Keccak specific code and added KeccakMiner (to only mine on demand, managed by PoWMiningCoordinator)

PoWMiningCoordinator protocol contains a placeholder to replace MockedMiner in the future
PoWMiningCoordinator is a typed actor. EthashMiner was modified to not be an actor anymore and just run a Monix task. The same to the KeccakMiner
Reorganized protocols (some is specific the MockedMiner for example)
Reorganized Spec setups for the miners

# Testing
Tested in pottery private network using mantis-automation (making sure ethash mining still works as expected)
Currently deployed in Staging (making sure ethash mining still works as expected)


bsuieric (approved):

bsuieric (approved):
LGTM, still unclear the use of MockedMiner, but I noticed you added a TODO to another ticket

src/main/scala/io/iohk/ethereum/consensus/pow/miners/EthashDAGManager.scala
```diff
@@ -0,0 +1,95 @@
...
+import scala.util.{Failure, Success, Try}
+
+class EthashDAGManager(blockCreator: PoWBlockCreator) extends Logger {
+  var currentEpoch: Option[Long] = None
+  var currentEpochDagSize: Option[Long] = None
+  var currentEpochDag: Option[Array[Array[Int]]] = None
+
+  def calculateDagSize(blockNumber: Long, epoch: Long): (Array[Array[Int]], Long) = {
+    (currentEpoch, currentEpochDag, currentEpochDagSize) match {
+      case (Some(`epoch`), Some(dag), Some(dagSize)) => (dag, dagSize)
```
dzajkowski: is `epoch` unused?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
```diff
@@ -0,0 +1,140 @@
...
+import io.iohk.ethereum.jsonrpc.EthMiningService
+
+import scala.concurrent.duration.DurationInt
+import scala.util.Random
+
+object PoWMiningCoordinator {
+  // TODO in ETCM-773 make trait sealed
+  trait CoordinatorProtocol
+  final case class StartMining(mode: MiningMode) extends CoordinatorProtocol
+  case object DoMining extends CoordinatorProtocol
```
dzajkowski: nitpick: `DoMining` => `StartMining`

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
```diff
@@ -0,0 +1,140 @@
...
+import io.iohk.ethereum.domain.{Block, Blockchain}
+import io.iohk.ethereum.jsonrpc.EthMiningService
+
+import scala.concurrent.duration.DurationInt
+import scala.util.Random
+
+object PoWMiningCoordinator {
+  // TODO in ETCM-773 make trait sealed
+  trait CoordinatorProtocol
+  final case class StartMining(mode: MiningMode) extends CoordinatorProtocol
```
dzajkowski: nitpick: `StartMining` => `SwitchMode`

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
```diff
@@ -0,0 +1,140 @@
...
+    case StartMining(mode) =>
+      log.info("Received message {}", StartMining(mode))
+      switchMiningMode(mode)
+
+    case DoMining =>
+      log.info("Received message ProcessMining")
+      blockchain
+        .getBestBlock()
+        .fold(
+          log.error("Unable to get block for mining: blockchain.getBestBlock() returned None")
```
dzajkowski, line 96: regarding the problems you're experiencing with pottery - might be that the startup process needs to be adjusted so that the client state is actually initialised to start mining.

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
dzajkowski: or maybe `TickMining`?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
```diff
@@ -0,0 +1,140 @@
...
+        ) { block =>
+          if (shouldMineWithKeccak(block.header.number)) mineWithKeccak(block) else mineWithEthash(block)
+        }
+      Behaviors.same
+
+    case MiningSuccessful | MiningUnsuccessful =>
+      log.info("Assigned miner has completed mining")
```
dzajkowski: nitpick: this assumes that some other process will log out the actual result of mining.

dzajkowski: IOW this might be a `debug` message.

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  private def switchMiningMode(mode: MiningMode): Behavior[CoordinatorProtocol] = mode match {
+    case RecurrentMining =>
+      context.self ! DoMining
+      handleMiningRecurrent()
+    case OnDemandMining => handleMiningOnDemand()
+  }
+
+  private def shouldMineWithKeccak(currentBlockNumber: BigInt): Boolean = {
```
dzajkowski: nitpick, not for this PR: wouldn't it make sense to future proof this step?
```scala
private def pickMiningAlogrithm(currentBlockNumber: BigInt): MiningAlgorithm
```

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
```diff
@@ -0,0 +1,140 @@
...
+    ecip1049BlockNumber match {
+      case None => false
+      case Some(blockNumber) => (currentBlockNumber + 1) >= blockNumber
+    }
+  }
+
+  private def mineWithEthash(bestBlock: Block): Unit = {
+    log.info("Spawning an EthashMiner")
```
dzajkowski: curious: this info will be spammed a lot on regular execution. do we really care that much?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
```diff
@@ -0,0 +1,140 @@
...
+        KeccakMiner(blockCreator, syncController, ethMiningService),
+        name = geMinerName("KeccakMiner", bestBlock.number),
+        DispatcherSelector.sameAsParent()
+      )
+    keccakMiner ! MinerProtocol.ProcessMining(bestBlock, context.self)
+  }
+
+  // The suffix is to make sure we never have two miners with the same name
+  // in case we spawn one while the previous is still shutting down
+  private def geMinerName(prefix: String, blockNumber: BigInt): String = {
```
dzajkowski: nitpick: so this is not deterministic, right? `generate` would be a better word.

src/test/scala/io/iohk/ethereum/consensus/pow/PoWMinerCoordinatorSpec.scala
```diff
@@ -0,0 +1,192 @@
...
+  it should "start on demand mining when receiving message StartMining(OnDemandMining)" in new TestSetup(
+    "OnDemandMining"
+  ) {
+    LoggingTestKit.info("Received message StartMining(OnDemandMining)").expect {
+      coordinator ! StartMining(OnDemandMining)
+    }
+    coordinator ! StopMining
+  }
+
+  it should "[Recurrent Mining] ProcessMining starts EthashMiner if mineWithKeccak is false" in new TestSetup(
```
dzajkowski: not sure these tags will scale. maybe we could use a spec that allows nesting (`WordSpec`, `FreeSpec`, `FunSpec`)?

src/test/scala/io/iohk/ethereum/consensus/pow/PoWMinerCoordinatorSpec.scala
```diff
@@ -0,0 +1,192 @@
...
+      coordinator ! StartMining(RecurrentMining)
+    }
+
+    coordinator ! StopMining
+  }
+
+  it should "[Recurrent Mining] ProcessMining starts KeccakMiner if mineWithKeccak is true" in new TestSetup(
+    "KeccakMining"
+  ) {
+    override val coordinator = system.spawn(
```
dzajkowski: wouldn't it make sense to extract the duplication for the `spawn` of the `PoWMiningCoordinator`? eg. move ti to `TestSetup`?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
```diff
@@ -44,17 +48,31 @@ class PoWConsensus private (
...
 
-  @volatile private[this] var minerRef: Option[ActorRef] = None
+  @volatile private[pow] var minerCoordinatorRef: Option[ActorRef[CoordinatorProtocol]] = None
+  // TODO in ETCM-773 remove MockedMiner
+  @volatile private[pow] var mockedMinerRef: Option[akka.actor.ActorRef] = None
 
+  final val BlockForgerDispatcherId = "mantis.async.dispatchers.block-forger"
   private implicit val timeout: Timeout = 5.seconds
 
   override def sendMiner(msg: MinerProtocol): Unit =
```
dzajkowski, line 58: would it be possible to have the param also typed to the MockedMiner protocol?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
```diff
@@ -44,17 +48,31 @@ class PoWConsensus private (
...
+        mockedMinerRef.foreach(_ ! MockedMiner.StopMining)
+        minerCoordinatorRef.foreach(_ ! PoWMiningCoordinator.StopMining)
+      case _ => log.warn("SendMiner method received unexpected message {}", msg)
+    }
 
-  override def askMiner(msg: MinerProtocol): Task[MinerResponse] = {
-    minerRef
-      .map(_.askFor[MinerResponse](msg))
+  // no interactions are done with minerCoordinatorRef using the ask pattern
+  override def askMiner(msg: MinerProtocol): Task[MockedMinerResponse] = {
```
dzajkowski: would it be possible to have the param also typed to the MockedMiner protocol?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
```diff
@@ -44,17 +48,31 @@ class PoWConsensus private (
...
+    mockedMinerRef
+      .map(_.askFor[MockedMinerResponse](msg))
       .getOrElse(Task.now(MinerNotExist))
+    Task.now(MinerNotExist)
```
dzajkowski: what is this code supposed to do? atm it passes a message to the mock and always returns `MinerNotExist` and the `      .getOrElse(Task.now(MinerNotExist))` only forces the ask to realise. is that the intent?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
```diff
@@ -65,15 +83,30 @@ class PoWConsensus private (
...
+        if (minerCoordinatorRef.isEmpty && mockedMinerRef.isEmpty) {
+          config.generic.protocol match {
+            case PoW | RestrictedPoW =>
+              log.info("Instantiating PoWMiningCoordinator")
+              minerCoordinatorRef = Some(
+                node.system.spawn(
+                  PoWMiningCoordinator(
+                    node.syncController,
+                    node.ethMiningService,
+                    blocKCreator,
```
dzajkowski: `K` typo?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
```diff
@@ -65,15 +83,30 @@ class PoWConsensus private (
    *
    * TODO further refactors should focus on extracting two types - one with a miner, one without - based on the config
    */
-  private[this] def startMiningProcess(node: Node): Unit = {
-    if (minerRef.isEmpty) {
+  private[this] def startMiningProcess(node: Node, blocKCreator: PoWBlockCreator): Unit = {
```
dzajkowski: nitpick: `startMiningCoordinator`?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
```diff
@@ -65,15 +83,30 @@ class PoWConsensus private (
...
+    if (minerCoordinatorRef.isEmpty && mockedMinerRef.isEmpty) {
```
dzajkowski: shouldn't this be hidden behind the mutex?

dzajkowski: as in: it's redundant

src/test/scala/io/iohk/ethereum/consensus/pow/PoWConsensusSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+    val powConsensus = PoWConsensus(
+      vm,
+      blockchain,
+      blockchainConfig,
+      ConsensusConfigs.fullConsensusConfig,
+      validator,
+      RestrictedPoWMinerData(key)
+    )
+
+    powConsensus.blockGenerator.isInstanceOf[RestrictedPoWBlockGeneratorImpl] shouldBe true
```
dzajkowski, line 50: looks like the test for "no miner" makes sure that the refs aren't set. wouldn't it make sense to also check here (send the msg "start" and check the expected ref)?

src/test/scala/io/iohk/ethereum/consensus/pow/PoWConsensusSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+      blockchain,
+      blockchainConfig,
+      fullConsensusConfig,
+      validator,
+      NoAdditionalPoWData
+    )
+
+    powConsensus.startProtocol(new TestConsensusNode())
+    powConsensus.minerCoordinatorRef shouldBe None
+    powConsensus.mockedMinerRef shouldBe None
```
dzajkowski, line 68: I'm confused. why have `use NoAdditionalPoWData block generator for PoWBlockGeneratorImpl` if this test does exactly the same plus verifies the `start`?

src/test/scala/io/iohk/ethereum/consensus/pow/PoWConsensusSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+      blockchainConfig,
+      fullConsensusConfig,
+      validator,
+      NoAdditionalPoWData
+    )
+
+    powConsensus.startProtocol(new TestConsensusNode())
+    powConsensus.minerCoordinatorRef shouldBe None
+    powConsensus.mockedMinerRef.isDefined shouldBe true
+  }
```
dzajkowski, line 87: I'm confused. why have use NoAdditionalPoWData block generator for PoWBlockGeneratorImpl if this test does exactly the same plus verifies the start?

src/test/scala/io/iohk/ethereum/consensus/pow/PoWConsensusSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  it should "start only one normal miner when miner protocol is PoW" in new TestSetup {
```
dzajkowski: this test does not verify what it says it does - there is no check for the amount of started miners in the body. there is only a verification that the refs are populated which means at least one.

project/Dependencies.scala
```diff
@@ -13,8 +13,10 @@ object Dependencies {
   val akka: Seq[ModuleID] = {
     Seq(
       "com.typesafe.akka" %% "akka-actor" % akkaVersion,
+      "com.typesafe.akka" %% "akka-actor-typed" % akkaVersion,
       "com.typesafe.akka" %% "akka-slf4j" % akkaVersion,
       "com.typesafe.akka" %% "akka-testkit" % akkaVersion,
+      "com.typesafe.akka" %% "akka-actor-testkit-typed" % akkaVersion,
```
AnastasiiaL: I believe you won't need these dependencies if you rebase on develop (it was already added there, together with the sha update)

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
```diff
@@ -44,17 +48,31 @@ class PoWConsensus private (
...
-    minerRef.foreach(_ ! msg)
+    msg match {
+      case mineBlocks: MockedMiner.MineBlocks => mockedMinerRef.foreach(_ ! mineBlocks)
+      case MinerProtocol.StartMining =>
+        mockedMinerRef.foreach(_ ! MockedMiner.StartMining)
+        minerCoordinatorRef.foreach(_ ! PoWMiningCoordinator.StartMining(PoWMiningCoordinator.RecurrentMining))
+      case MinerProtocol.StopMining =>
+        mockedMinerRef.foreach(_ ! MockedMiner.StopMining)
+        minerCoordinatorRef.foreach(_ ! PoWMiningCoordinator.StopMining)
+      case _ => log.warn("SendMiner method received unexpected message {}", msg)
```
AnastasiiaL, line 67: shall we use debug level for unhandled messages (in line with RegularSync)

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
```diff
@@ -65,15 +83,30 @@ class PoWConsensus private (
...
+                    node.ethMiningService,
+                    blocKCreator,
+                    blockchain,
+                    blockchainConfig.ecip1049BlockNumber
+                  ),
+                  "PoWMinerCoordinator",
+                  DispatcherSelector.fromConfig(BlockForgerDispatcherId)
+                )
+              )
+            case MockedPow =>
```
AnastasiiaL: just for my understanding: why do we need a mocked miner? Just not usual to see something mocked in the source code

src/main/scala/io/iohk/ethereum/consensus/pow/miners/EthashMiner.scala
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  private def mineEthah(
+      headerHash: Array[Byte],
+      difficulty: Long,
+      dagSize: Long,
+      dag: Array[Array[Int]],
+      numRounds: Int
+  ): MiningResult = {
+    val numBits = 64
```
dzajkowski: to make this an improvement: maybe move to a named constant?

src/main/scala/io/iohk/ethereum/consensus/pow/miners/EthashMiner.scala
```diff
@@ -0,0 +1,136 @@
...
+
+  override def receive: Receive = started
+
+  def started: Receive = { case ProcessMining(bestBlock, coordinatorRef) =>
+    log.debug("Received message ProcessMining with parent block {}", bestBlock.number)
+    processMining(bestBlock, coordinatorRef)
+  }
+
+  private def processMining(bestBlock: Block, coordinatorRef: akka.actor.typed.ActorRef[CoordinatorProtocol]): Unit = {
+    blockCreator
```
dzajkowski, line 38: since this is a one time `Task`, what's the need for an actor?

src/main/scala/io/iohk/ethereum/consensus/pow/miners/KeccakMiner.scala
```diff
@@ -0,0 +1,111 @@
...
+      log.debug("KeccakMiner received message {}", message)
+      processMining(bestBlock, coordinatorRef)
+      Behaviors.same
+  }
+
+  private def processMining(
+      bestBlock: Block,
+      coordinatorRef: ActorRef[CoordinatorProtocol]
+  ): Unit = {
+    blockCreator
```
dzajkowski, line 30: again - I don't see the point of having a dedicated actor

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  override def onMessage(msg: CoordinatorProtocol): Behavior[CoordinatorProtocol] = msg match {
+    case StartMining(mode) =>
+      log.info("Received message {}", StartMining(mode))
+      switchMiningMode(mode)
+  }
+
+  private def handleMiningRecurrent(): Behavior[CoordinatorProtocol] = Behaviors.receiveMessage {
+    case StartMining(mode) =>
+      log.info("Received message {}", StartMining(mode))
```
AnastasiiaL: I think all these messages should be debug level, otherwise will be too noisy

dzajkowski (changes requested):
overall looks good. please review my comments.

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
leo-bogastry: `MineNextBlock`?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
leo-bogastry: I don't think that makes the most sense. What about `SetMiningMode(mode: MiningMode)`?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
leo-bogastry: The error happens after is started mining, after is has mined a few blocks :(

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
leo-bogastry: > nitpick: this assumes that some other process will log out the actual result of mining.

Is does :)

> IOW this might be a debug message.

Indeed. I have adjusted it

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
leo-bogastry: At the moment I know I do. Until this actually goes "live" and the mining algo gets switched I think it does.

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
leo-bogastry: Good point. And there's a typo there 😬

src/test/scala/io/iohk/ethereum/consensus/pow/PoWMinerCoordinatorSpec.scala
leo-bogastry: There's coordinator instantiated there, but then I need to specify a specific `ecip1049BlockNumber` then it has to be overriden

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
leo-bogastry: I think it can.

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
leo-bogastry: I think it was just added by mistake 🙈

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
leo-bogastry: Given that the mockedMiner is supposed to be removed, yes. Otherwise no. What do you think? :)

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
leo-bogastry: I don't understand what you mean

src/test/scala/io/iohk/ethereum/consensus/pow/PoWConsensusSpec.scala
leo-bogastry: Afaik you can only do that with probes, and these are not probes

src/test/scala/io/iohk/ethereum/consensus/pow/PoWConsensusSpec.scala
leo-bogastry: Actually the instantiation of `PoWConsensus` is not equal, otherwise I also think that tests should document a class. So even if they had the same setup it could still make sense to have two tests

src/test/scala/io/iohk/ethereum/consensus/pow/PoWConsensusSpec.scala
leo-bogastry: I have adjusted the description

src/main/scala/io/iohk/ethereum/consensus/pow/miners/EthashMiner.scala
leo-bogastry: The EthashMiner and KeccakMiner don't need to run any state. So indeed a Task seems to be enough. I will do that one later, as the tests also need to be adjusted

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
leo-bogastry: This can mean an error, so if you make it a debug message it won't be visible

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
leo-bogastry: I hope the refinement meeting we had today was enough to answer the question :)
There's a ticket to remove the MockedMiner (ETCM-773)

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
leo-bogastry: This will only appear once in normal mining. If the mining mode changes (when doing tests) it is important to know

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
dzajkowski: naming hard. what else could it mine? maybe `MineNext` is enough?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
dzajkowski: oh, that's not a good sign.

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
dzajkowski: sure

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
dzajkowski: so maybe make it debug and control the message with `logback.xml`?

src/main/scala/io/iohk/ethereum/consensus/pow/miners/EthashDAGManager.scala
leo-bogastry: Yes, I removed it

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
dzajkowski: the `if` statement before the mutex is subject to context switch, and since it's two conditions it will return a false positive more often than a singular check - so IMO the mutex should be locked first and the check should be performed inside the mutex only as the external one provides little security.

src/main/scala/io/iohk/ethereum/consensus/pow/miners/EthashMiner.scala
leo-bogastry: I have refactored. Please check

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
leo-bogastry: So you mean, from this
```
if (minerCoordinatorRef.isEmpty && mockedMinerRef.isEmpty) {
      mutex.synchronized {
            if (minerCoordinatorRef.isEmpty && mockedMinerRef.isEmpty) {
``` 
to
```
 mutex.synchronized {
     if (minerCoordinatorRef.isEmpty && mockedMinerRef.isEmpty) {
```
?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
leo-bogastry: There are only two possible values that can be returned. Is it really worth the extra code?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
leo-bogastry: I changed these recurrent messages to debug

src/test/scala/io/iohk/ethereum/consensus/pow/PoWMinerCoordinatorSpec.scala
leo-bogastry: FreeSpec is a good fit

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
dzajkowski: I don't have a strong opinion here but I think that naming stuff properly is a good habit

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
dzajkowski: yep

dzajkowski (approved):
Please verify that the node can synch to sagano with these changes. Would it be possible to verify if ethash mining is still behaving as expected (compatible with sagano)?

src/main/scala/io/iohk/ethereum/consensus/pow/PoWMiningCoordinator.scala
leo-bogastry: I have refactored this

src/main/scala/io/iohk/ethereum/consensus/pow/PoWConsensus.scala
leo-bogastry: 👍

leo-bogastry:
> Please verify that the node can synch to sagano with these changes. Would it be possible to verify if ethash mining is still behaving as expected (compatible with sagano)?

A couple of day ago I deployed this branch in staging, and it is mining and synching. 
In Sagano I was able to synch, but in order to mine I have to update the miners there

leo-bogastry:
> Please verify that the node can synch to sagano with these changes. Would it be possible to verify if ethash mining is still behaving as expected (compatible with sagano)?

It's deployed in Sagano as well

PR #980, opened 2021-05-06
[ETCM-814] Don't burn part of the rewards that would go to treasury

bsuieric:
# Description

A miner should no longer be able to burn part of its reward that would be sent to the treasury. Change proposal ECIP 1098.

# Proposed Solution

Code responsible with burning the rewards is removed. 
Miner opting out of sending the rewards to treasury is not possible anymore.
Added block header validation that check treasuryOptOut for invalid blocks

Added block header validation that check treasuryOptOut for invalid blocks

# Testing

added unit tests, sync with sagano


leo-bogastry (approved):

dzajkowski:
looks like the fast sync validator is covered (`src/main/scala/io/iohk/ethereum/consensus/validators/BlockHeaderValidator.scala`) but regular sync seems to be missed

leo-bogastry:
👍

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
```diff
@@ -57,11 +51,8 @@ case class BlockHeader(
 
   override def toString: String = {
     val (treasuryOptOutString: String, checkpointString: String) = extraFields match {
```
AurelienRichez: shouldn't we also remove or rename `treasuryOptOutString` here ?

src/main/scala/io/iohk/ethereum/consensus/validators/BlockHeaderValidator.scala
```diff
@@ -49,6 +49,7 @@ object BlockHeaderError {
   case class HeaderFieldNotEmptyError(msg: String) extends BlockHeaderError
   case class HeaderNotMatchParentError(msg: String) extends BlockHeaderError
   case object CheckpointHeaderTreasuryOptOutError extends BlockHeaderError
+  case object BlockHeaderTreasuryOptOutError extends BlockHeaderError
```
AurelienRichez: It seems this is not used anymore so we can remove it.

src/main/scala/io/iohk/ethereum/domain/BlockHeader.scala
dzajkowski: agreed, initially I wanted to have a marker but it does not make sense in this context

AurelienRichez (approved):

PR #984, opened 2021-05-11
[ETCM-829] rewrite healthcheck

AurelienRichez:
# Description

This PR rewrite the healthcheck endpoint to have more precise info : 
  - it checks that we have at least one peer (before it only checked that the service responded without an error, but not the content of the response itself)
  - it returns block numbers for best stored block, best known block and best block we are fetching
  - it returns a 500 when the state did not change change for too long
  - it returns the sync status (STARTING, SYNCING, SYNCED)
  
 It can still be improved by returning the number of healthy peers instead of the raw number of peers.

The healthcheck endpoint is not backward compatible (for instance I renamed the field `description` to `name` because it seems more accurate)

two new parameters : 
 - `mantis.network.rpc.http.health.no-update-duration-threshold`
 - `mantis.network.rpc.http.health.syncing-status-threshold`
 
The values of some parameters are given as a generic parameter "info" which is a string. I think it is alright for a healthcheck endpoint but we might want to consider if we want something more strongly typed.

# Testing

You can call `<node_url>/healthcheck` to see the endpoint, and check that it moves correctly 

The result should look like this : 

```
{
  "checks": [
    {
      "name": "peerCount",
      "status": "OK",
      "info": "3"
    },
    {
      "name": "bestStoredBlock",
      "status": "OK",
      "info": "80000"
    },
    {
      "name": "bestKnownBlock",
      "status": "OK",
      "info": "12721559"
    },
    {
      "name": "bestFetchingBlock",
      "status": "OK",
      "info": "80000"
    },
    {
      "name": "updateStatus",
      "status": "OK"
    },
    {
      "name": "syncStatus",
      "status": "OK",
      "info": "SYNCING"
    }
  ]
}
```


src/main/scala/io/iohk/ethereum/jsonrpc/NodeJsonRpcHealthChecker.scala
```diff
@@ -37,45 +37,55 @@ class NodeJsonRpcHealthChecker(
 
   final val peerCountHC = JsonRpcHealthcheck
     .fromServiceResponse("peerCount", netService.peerCount(PeerCountRequest()))
-    .withInfo(_.value.toString)
-    .withPredicate("peer count is 0")(_.value > 0)
+    .map(healthcheck => healthcheck.withInfo(_.value.toString).withPredicate("peer count is 0")(_.value > 0))
```
dzajkowski: nitpick: `healthcheck => healthcheck.` => `_.`

src/main/scala/io/iohk/ethereum/jsonrpc/NodeJsonRpcHealthChecker.scala
```diff
@@ -1,43 +1,152 @@
...
+      )
 
   override def healthCheck(): Task[HealthcheckResponse] = {
-    val listeningF = listeningHC()
-    val peerCountF = peerCountHC()
-    val earliestBlockF = earliestBlockHC()
-    val latestBlockF = latestBlockHC()
-    val pendingBlockF = pendingBlockHC()
+    val responseTask = Task
+      .sequence(
```
dzajkowski: I think that `par` here was fine.

src/main/scala/io/iohk/ethereum/jsonrpc/NodeJsonRpcHealthChecker.scala
```diff
@@ -1,43 +1,152 @@
...
-    "latestBlock",
-    ethBlocksService.getBlockByNumber(BlockByNumberRequest(BlockParam.Latest, fullTxs = true))
-  )
-  final val pendingBlockHC = JsonRpcHealthcheck(
-    "pendingBlock",
-    ethBlocksService.getBlockByNumber(BlockByNumberRequest(BlockParam.Pending, fullTxs = true))
-  )
+  private var previousBestFetchingBlock: Option[(Instant, BigInt)] = None
+
+  final val peerCountHC = JsonRpcHealthcheck
```
dzajkowski: I think that all of the HCs can be `private`

dzajkowski (approved):
looks good, a few small comments

src/main/resources/conf/base.conf
```diff
@@ -253,6 +253,15 @@ mantis {
         peer-manager-timeout = 5.seconds
       }
 
+      health {
+        # If the client does see a change of the best known block number after this duration
+        # the healthcheck will consider it to be stuck and return an error
```
leo-bogastry: I guess you mean `otherwise the healthcheck will consider it to be stuck and return an error`

src/main/resources/conf/base.conf
```diff
@@ -253,6 +253,15 @@ mantis {
...
+        no-update-duration-threshold = 30.minutes
+        # If the client best known stored block is behind the best known block by more blocks
+        # than this parameter, we consider that it is syncing in the healthcheck
```
leo-bogastry: do you mean `we consider that it is NOT syncing in the healthcheck`?

src/test/scala/io/iohk/ethereum/jsonrpc/server/http/JsonRpcHttpServerSpec.scala
```diff
@@ -32,20 +33,44 @@ class JsonRpcHttpServerSpec extends AnyFlatSpec with Matchers with ScalatestRout
...
-                                       |      "status":"OK"
-                                       |    }
+                                       |    { "name": "peerCount", "status": "OK", "info": "2" }
                                        |  ]
-                                       |}""".stripMargin
+                                       |}""".stripMargin)
+    }
+  }
+
+  it should "respond to healthcheck with an error if one healtcheck fails" in new TestSetup {
```
leo-bogastry: Can you add tests for the other checks as well? Not just peerCount

leo-bogastry: And a typo in `if one healtcheck fails`

AurelienRichez: Actually, it does not really check `peerCount` itself because the checks come from a mock. The only logic that is tested here is the resulting json and the http code which should be 500 as soon as one check fails. I don't think that adding other tests with the `mockJsonRpcHealthChecker` really makes sense.

leo-bogastry: 👍

leo-bogastry (approved):

PR #986, opened 2021-05-14
[Kaizen] Introduce sbt-jmh bench module

dzajkowski:
Groundwork for some proper micro benchmarks


dzajkowski:
maybe this can wait, not really a priority now

PR #992, opened 2021-05-20
[ETCM-846] make ETS stBadOpcode tests pass

AurelienRichez:
# Description

This PR is part of the ogoing effort to make ETS pass all tests. The `stBadOpcode` suite did not pass because the accounts were not imported with the root hash for their storage. 

# Proposed Solution

The `getGenesisStateRoot` now create an in memory trie in order to compte the resulting hash. 

As this change is not specific to `stBadOpcode`, it also makes most the previously failing tests pass (9 failing test for `GeneralStateTests` while we had around 400 before).

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
@@ -86,14 +86,14 @@ class GenesisDataLoader(blockchain: Blockchain, blockchainConfig: BlockchainConf
     } yield ()
   }
 
-  def loadGenesisData(genesisData: GenesisData): Try[Unit] = {
+  def loadGenesisData(genesisData: GenesisData, storageRootHashes: Map[Address, ByteString] = Map.empty): Try[Unit] = {
     import MerklePatriciaTrie.defaultByteArraySerializable
 
     val stateStorage = blockchain.getStateStorage
     val storage = stateStorage.getReadOnlyStorage
     val initalRootHash = MerklePatriciaTrie.EmptyRootHash
 
-    val stateMptRootHash = getGenesisStateRoot(genesisData, initalRootHash, storage)
+    val stateMptRootHash = getGenesisStateRoot(genesisData, initalRootHash, storage, storageRootHashes)
     val header: BlockHeader = prepareHeader(genesisData, stateMptRootHash)
 
     log.debug(s"Prepared genesis header: $header")
@@ -122,12 +122,18 @@ class GenesisDataLoader(blockchain: Blockchain, blockchainConfig: BlockchainConf
     }
   }
 
-  private def getGenesisStateRoot(genesisData: GenesisData, initalRootHash: Array[Byte], storage: MptStorage) = {
+  private def getGenesisStateRoot(
+      genesisData: GenesisData,
+      initalRootHash: Array[Byte],
+      storage: MptStorage,
+      storageRootHashes: Map[Address, ByteString]
+  ) = {
     import MerklePatriciaTrie.defaultByteArraySerializable
 
     genesisData.alloc.zipWithIndex.foldLeft(initalRootHash) { case (rootHash, ((address, genesisAccount), _)) =>
       val mpt = MerklePatriciaTrie[Array[Byte], Account](rootHash, storage)
       val paddedAddress = address.reverse.padTo(addressLength, "0").reverse.mkString
+
       val stateRoot = mpt
         .put(
           crypto.kec256(Hex.decode(paddedAddress)),
@@ -135,7 +141,8 @@ class GenesisDataLoader(blockchain: Blockchain, blockchainConfig: BlockchainConf
             nonce = genesisAccount.nonce
               .getOrElse(blockchainConfig.accountStartNonce),
             balance = genesisAccount.balance,
-            codeHash = genesisAccount.code.map(codeValue => crypto.kec256(codeValue)).getOrElse(Account.EmptyCodeHash)
+            codeHash = genesisAccount.code.map(codeValue => crypto.kec256(codeValue)).getOrElse(Account.EmptyCodeHash),
+            storageRoot = storageRootHashes.getOrElse(Address(Hex.decode(paddedAddress)), Account.EmptyStorageRootHash)
           )
         )
         .getRootHash
--- a/src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
@@ -137,8 +137,8 @@ class GenesisDataLoader(blockchain: Blockchain, blockchainConfig: BlockchainConf
             nonce = genesisAccount.nonce
               .getOrElse(blockchainConfig.accountStartNonce),
             balance = genesisAccount.balance,
-            codeHash = genesisAccount.code.map(codeValue => crypto.kec256(codeValue)).getOrElse(Account.EmptyCodeHash),
-            storageRoot = genesisAccount.storage.map(computeStorageRootHash).getOrElse(Account.EmptyStorageRootHash)
+            codeHash = genesisAccount.code.fold(Account.EmptyCodeHash)(codeValue => crypto.kec256(codeValue)),
+            storageRoot = genesisAccount.storage.fold(Account.EmptyStorageRootHash)(computeStorageRootHash)
           )
         )
         .getRootHash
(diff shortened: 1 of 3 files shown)


KonradStaniec (approved):

dzajkowski:
please also use the ticket tag in commit messages ([ECTM-846]) this enables integration with jira

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
```diff
@@ -128,21 +129,40 @@ class GenesisDataLoader(blockchain: Blockchain, blockchainConfig: BlockchainConf
...
       val stateRoot = mpt
         .put(
           crypto.kec256(Hex.decode(paddedAddress)),
           Account(
             nonce = genesisAccount.nonce
               .getOrElse(blockchainConfig.accountStartNonce),
             balance = genesisAccount.balance,
-            codeHash = genesisAccount.code.map(codeValue => crypto.kec256(codeValue)).getOrElse(Account.EmptyCodeHash)
+            codeHash = genesisAccount.code.map(codeValue => crypto.kec256(codeValue)).getOrElse(Account.EmptyCodeHash),
+            storageRoot = genesisAccount.storage.map(computeStorageRootHash).getOrElse(Account.EmptyStorageRootHash)
```
jvdp: Nit: `fold` on an option is a bit more direct than `map(..).getOrElse`.

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
```diff
@@ -128,21 +129,40 @@ class GenesisDataLoader(blockchain: Blockchain, blockchainConfig: BlockchainConf
...
           )
         )
         .getRootHash
       stateRoot
     }
   }
 
+  private def computeStorageRootHash(storage: Map[UInt256, UInt256]): ByteString = {
```
jvdp, line 149: Is this based on some other code? It looks pretty involved for computing the root hash (which is probably done in other places as well?)

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
```diff
@@ -128,21 +129,40 @@ class GenesisDataLoader(blockchain: Blockchain, blockchainConfig: BlockchainConf
...
+    val emptyTrie = EthereumUInt256Mpt.storageMpt(
+      ByteString(MerklePatriciaTrie.EmptyRootHash),
+      new SerializingMptStorage(new ArchiveNodeStorage(new NodeStorage(EphemDataSource())))
+    )
+
+    val storageTrie = storage.foldLeft(emptyTrie) { (trie, tuple) =>
+      val (key, value) = tuple
```
jvdp: You can deconstruct this from the arguments directly. Could also put the conditional below as a separate pattern match.

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
```diff
@@ -167,39 +168,23 @@ class TestService(
...
-    val accountsWithCodes = accounts.filter(pair => pair._2.code.isDefined)
-
-    val worldToPersist = accountsWithCodes.foldLeft(world)((world, addressAccountPair) => {
-      world.saveCode(Address(addressAccountPair._1), addressAccountPair._2.code.get)
-    })
-
-    InMemoryWorldStateProxy.persistState(worldToPersist)
+  private def storeGenesisAccountCodes(accounts: Map[String, GenesisAccount]): Unit = {
+    for {
+      code <- accounts.map(pair => pair._2.code).collect { case Some(code) => code }
```
jvdp: Might as well use a `map` instead of a for-comprehension with only one line.

(Or a `foreach` actually, since it's a side-effect.)

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
```diff
@@ -167,39 +168,23 @@ class TestService(
...
-
-    InMemoryWorldStateProxy.persistState(worldToPersist)
+  private def storeGenesisAccountStorageData(accounts: Map[String, GenesisAccount]): Unit = {
+    val emptyStorage = domain.EthereumUInt256Mpt.storageMpt(
+      Account.EmptyStorageRootHash,
+      blockchain.getStateStorage.getBackingStorage(0)
+    )
+    val storagesToPersist = accounts.map(pair => pair._2.storage).collect { case Some(map) if map.nonEmpty => map }
+    for { storage <- storagesToPersist } {
+      storage.foldLeft(emptyStorage) { case (storage, (key, value)) => storage.put(key, value) }
```
jvdp: Why `foldLeft` and not another `for` loop or a `foreach`? Since it's only about the side-effect right.

jvdp (approved):
So it looks a lot simpler than the previous implementation and less dependent on `InMemoryWorldStateProxy` which sounds scary. (What's the impact of not using this anymore btw? Is it only for tests & test mode or also part of production code?)

Can you explain the main change a bit more? It's now more directly committing the genesis account to the blockchain instead of keeping it in the in-memory state?

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
AurelienRichez: It's about the side effect, but it's also important to get the return of `put`. Because it stores the new MPT in the database but does not remove the previous MPT, it just give back the new MPT corresponding to the modification (that's why I can reuse `emptyStorage` each time).

But this make me think that it's not really efficient as it creates intermediate MPT each time. I can use `update` which allows to add several keys at once. I will change this

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
AurelienRichez: From what I saw. Usually the hash is computed when actually storing the key values. For instance after running a transaction which modified the data we can extract the resulting root hash. The problem here is that there is no way to know the root hash unless we actually build it. So I create a MPT in memory in order to get the hash. I did not find any existing helper to do this.

I see two alternative to do actually get the root hash : 

 - injecting the root hash (new argument in `loadGenesisData` with precomputed hashed). We could do this by first storing the data and getting the resulting root hash.
 - building the hash inside the function

I choose the second because this is similar to what we already do with the contract code (we compute the hash of the code, but it is indeed less involved for the code because there is no MPT).

At the same time it seems weird that the Genesis data loader stores hashes but not the code and the contract storage, so a small change the responsibilities might be in order 🤔.

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
```diff
@@ -168,13 +168,11 @@ class TestService(
...
-    for {
-      code <- accounts.map(pair => pair._2.code).collect { case Some(code) => code }
-    } {
-      blockchain.storeEvmCode(kec256(code), code).commit()
-    }
-  }
+  private def storeGenesisAccountCodes(accounts: Map[String, GenesisAccount]): Unit =
+    accounts
+      .map(pair => pair._2.code)
+      .collect { case Some(code) => code }
```
jvdp: Sorry one more nit I missed earlier: you should be able to combine the `map` and `collect` to deconstruct the pair directly.

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
```diff
@@ -185,9 +183,9 @@ class TestService(
       .flatMap(pair => pair._2.storage)
       .map(accountStorage => accountStorage.filterNot { case (_, v) => v.isZero })
       .filter(_.nonEmpty)
-    for { storage <- storagesToPersist } {
-      storage.foldLeft(emptyStorage) { case (storage, (key, value)) => storage.put(key, value) }
-    }
+
+    //the identity map is needed to trigger an implicit conversion from UInt256 to BigInt
+    storagesToPersist.foreach(storage => emptyStorage.update(Nil, storage.toSeq.map { case (a, b) => (a, b) }))
```
jvdp: Could replace the comment and instead use: `identity: UIn256 => BigInt`.

src/main/scala/io/iohk/ethereum/blockchain/data/GenesisDataLoader.scala
jvdp: So many of the details go over my head but it sounds like there's room for improvement. Do you reckon it's worth creating a refactoring ticket?

(IMHO we could also improve some of this stuff with decent factory methods, eg. `MerklePatriciaTrie.empty`.)

AurelienRichez:
@jvdp 

The change is only for testmode so it should not affect normal production. We should probably at some point refactor the code so that the test specific parts are in a dedicated module though (there is a ticket for that)

The main problem here was that the rootHash for account storage was not set properly in `getGenesisStateRoot`. 

Regarding the ditching of `InMemoryWorldStateProxy` : 

From what I see `InMemoryWorldStateProxy` is more focused on building a transient state when running transactions, i.e, taking an initial state, modifying it and then commit the new resulting state. 

But when we load the genesis state we do not have an initial state to start with so we have a kind of a chicken and egg problem (no initial to set the storage, and the initial state needs a storage).

My first working solution was to create the genesis state with empty contract storage, call the function with `InMemoryWorldStateProxy`, get the root hashes from there and then reload the genesis this time with the real contract storage. 

I found it weird to have this logic (there was no reason to initialize the state in two passes), and with some guidance from Konrad I found how to directly store arbitrary MPTs. Calling directly the storage layer allows to create the desired state in the database directly without the double genesis initialization trick.

PR #993, opened 2021-05-20
ETCM-370: Refine responses validation on block fetcher side 

AnastasiiaL:
Currently `BlockFetcher` performs bunch of validations on received responses but then it doesn't blacklist peer even if the response is clearly invalid.

The goal of this PR is to:
  - refine validations performed in block fetcher
  - restructure validation errors into proper error hierarchy (currently they are just strings)
  - restore blacklisting, but make it depend on error returned (e.g. it's important to check whether received headers are forming chain with ones already enqueued but it should not result with blacklist, on the other hand if received headers are not forming chain themselves - it's clearly wrong response and it's worth blacklisting peer, who provided it)
- restore and extend test cases for blacklisting behaviour in Regular Sync

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
@@ -10,7 +10,6 @@ import io.iohk.ethereum.consensus.validators.BlockValidator
 import io.iohk.ethereum.domain.{Block, BlockBody, BlockHeader, HeadersSeq}
 import io.iohk.ethereum.network.PeerId
 import io.iohk.ethereum.network.p2p.messages.PV62.BlockHash
-import io.iohk.ethereum.utils.ByteStringUtils
 
 import scala.annotation.tailrec
 import scala.collection.immutable.Queue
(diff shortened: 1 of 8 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherState.scala
```diff
@@ -96,7 +94,7 @@ case class BlockFetcherState(
         .asRight[String]
         .ensure("Given headers should form a sequence without gaps")(HeadersSeq.areChain)
         .ensure("Given headers should form a sequence with ready blocks")(checkConsistencyWithReadyBlocks)
-        .ensure("Given headers do not form a chain with already stored ones")(headers =>
+        .ensure("Given headers should form a chain with waiting headers")(headers =>
```
AurelienRichez: If I understand correctly this error message must exactly match the on in `ValidationErrors`. This is not really obvious that there is a link between the two. And we could miss it if we want to add a new error type. 

I think it would be better to make this explicit by : 
  - make `ValidationErrors` a sealed trait
  - make `validatedHeaders` return an Either[ValidationErrors, Seq[BlockHeader]]
  - match by `ValidationErrors` in `processFetchCommands` instead of checking the error message
  
  It seems possible because we know every type of error that can occurs.
  What do you think ?

AnastasiiaL: you're completely right! I have no idea why I stopped with my changes half way, context switch is not always good for me :)

AurelienRichez (approved):

leo-bogastry (approved):
Nice PR description 👍

PR #995, opened 2021-05-25
[ETCM-856] make stSLoadTest pass

AurelienRichez:
Note : this PR is currently using `fix/ETCM-846-make-stBadOpcode-pass` as base (#992)  because it depends on it. I will change the target to develop when #992 is merged.

# Description

This PR makes stSLoadTest pass. The was with the way we condifured hard fork transition in the tests : 

  - We set the default block number of a given fork to 0 when not specified, so a hard fork was active by default
  - We did not recreate a new `blockPreparator` and `blockGenerator` so the new configuration was not applied in these.

I solved this by : 
  - By default, set the hard fork block to a high number if not specified (I chose Int.MaxNumber, but it is arbitrary)
  - For hard forks which are not specified by ETS (basically ETC hard fork), I aligned their block number to an ETH one : We might want to check that these are adequate (but at least the test is passing).
  - I changed some val to def to be sure the config
  
  As an aside, I also deactivated upnp when running in testmode.


src/main/resources/conf/testmode.conf
```diff
@@ -48,7 +48,11 @@ mantis {
     }
   }
 
-  network.rpc {
-    apis = "eth,web3,net,personal,mantis,test,iele,debug,qa,checkpointing"
+  network {
+    rpc {
+      apis = "eth,web3,net,personal,mantis,test,iele,debug,qa,checkpointing"
+    }
```
jvdp: I think this should be `rpc.apis` to prevent overriding (but I'm not 100% sure. but it's shorter anyways.)

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
```diff
@@ -161,13 +152,37 @@ class TestService(
...
     accountRangeOffset = 0
 
     SetChainParamsResponse().rightNow
   }
 
+  private def buildNewConfig(blockchainParams: BlockchainParams) = {
+    val byzantiumBlockNumber: BigInt = blockchainParams.byzantiumForkBlock.getOrElse(Int.MaxValue)
+    val istanbulForkBlockNumber: BigInt = blockchainParams.istanbulForkBlock.getOrElse(Int.MaxValue)
+
+    // For block number which are not specified by retesteth, we try to align the number to another fork
```
jvdp, line 167: What happened before this? And what should we do for future forks? What would happen if we'd also use `Int.MaxValue` here?

src/main/scala/io/iohk/ethereum/testmode/TestmodeConsensus.scala
```diff
@@ -20,7 +20,7 @@ import io.iohk.ethereum.consensus.pow.validators.ValidatorsExecutor
 class TestmodeConsensus(
     override val vm: VMImpl,
     blockchain: BlockchainImpl,
-    blockchainConfig: BlockchainConfig,
+    var blockchainConfig: BlockchainConfig,
```
jvdp: Is there any other way? I really feel we should avoid adding any more `var`s to the code base.

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
```diff
@@ -161,13 +152,37 @@ class TestService(
...
+    testLedgerWrapper.blockchainConfig.copy(
+      homesteadBlockNumber = blockchainParams.homesteadForkBlock.getOrElse(Int.MaxValue),
+      eip150BlockNumber = blockchainParams.EIP150ForkBlock.getOrElse(Int.MaxValue),
```
jvdp: Maybe you could define a constant `NEVER` or something for clarity?

jvdp (changes requested):
Requesting changes because of the `var`.

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
AurelienRichez: Before this I think the value defined in the configuration or the value defined in the previous run. So this would work on most of tests that just use the Istanbul fork, and tests that have not depend on something specific to a previous fork (actually I am surprised so many tests pass without this fix). 

However you are right. The way it is done, if we add a new fork then in the test the number will have whatever is defined in the configuration as a default, and if it is 0 then it will activate this fork everytime (but the compiler won't catch it). I think what we can do is a small refactoring of `BlockchainConfig` to extract the block numbers in another case class `ForkBlockNumbers` and rebuild completely `ForkBlockNumbers` when updating the config in the test. This way if we ever add a new fork the compiler will catch that we should also update the TestService instead of silently copying some value.

src/main/scala/io/iohk/ethereum/testmode/TestmodeConsensus.scala
AurelienRichez: Well, I think we could replace  `TestLedgerWrapper` with some utility class that provide a ledger and also a consensus implementation. We would still have a var somewhere but at least it would be more centralized. I will try something along those lines.

jvdp: Is thread safety a concern btw? This is essentially a shared resource right?

AurelienRichez: Not really. Clearly thread safety is not taken into account in the test service, but retesteth will run the test sequentially anway. There is a way to run tests in parallel but only with several instance of the client. But it wouldn't hurt to put the state in one place and put it behind an AtomicReference.

src/main/scala/io/iohk/ethereum/testmode/TestModeComponentsProvider.scala
```diff
@@ -0,0 +1,28 @@
...
+import io.iohk.ethereum.consensus.difficulty.DifficultyCalculator
+import io.iohk.ethereum.consensus.{Consensus, ConsensusConfig}
+import io.iohk.ethereum.domain.BlockchainImpl
+import io.iohk.ethereum.ledger.Ledger.VMImpl
+import io.iohk.ethereum.ledger.{Ledger, LedgerImpl, StxLedger}
+import io.iohk.ethereum.utils.BlockchainConfig
+import io.iohk.ethereum.utils.Config.SyncConfig
+import monix.execution.Scheduler
+
+/** Provides a ledger instance with modifiable blockchain config (used in test mode). */
```
jvdp: Is this comment still accurate?

jvdp (approved):
Nice 👍 ~so the only mutable thing is now the timestamp?~ nvm, there's still the blockchain config var of course.

dzajkowski:
the code is not passing scalastyle check

AurelienRichez:
There is still some mutable state, but at least it's not scattered in several classes. `TestService` holds all the state and `TestModeComponentsProvider` actually take the "mutable" parts as arguments.

src/main/scala/io/iohk/ethereum/testmode/TestModeComponentsProvider.scala
AurelienRichez: Indeed I should update it.

src/main/scala/io/iohk/ethereum/utils/BlockchainConfig.scala
```diff
@@ -44,12 +28,35 @@ case class BlockchainConfig(
...
-    ecip1099BlockNumber: BigInt
+    allowedMinersPublicKeys: Set[ByteString] = Set.empty
 ) {
   val minRequireSignatures: Int = (Math.floor(checkpointPubKeys.size / 2) + 1).toInt
+
+  def withUpdatedForkBlocks(update: (ForkBlockNumbers) => ForkBlockNumbers): BlockchainConfig =
+    copy(forkBlockNumbers = update(forkBlockNumbers))
 }
 
+case class ForkBlockNumbers(
```
leo-bogastry, line 39: 👍  Nice reorganization

leo-bogastry (approved):

PR #999, opened 2021-05-27
ETCM-845: Through rpc expose only blocks that are saved to storage

AnastasiiaL:
# Description

Several endpoints make use of the "best known block", which fails in case the best block wasn't yet saved to storage and only available in memory.

#Solution

fallback to best saved block in case best known block is not (yet) available

### DIFF
--- a/src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
@@ -289,10 +289,11 @@ class BlockchainImpl(
   override def getLatestCheckpointBlockNumber(): BigInt =
     bestKnownBlockAndLatestCheckpoint.get().latestCheckpointNumber
 
+  //returns the best known block if it's available in the storage, otherwise the best stored block
   override def getBestBlock(): Option[Block] = {
     val bestBlockNumber = getBestBlockNumber()
     log.debug("Trying to get best block with number {}", bestBlockNumber)
-    getBlockByNumber(bestBlockNumber)
+    getBlockByNumber(bestBlockNumber) orElse getBlockByNumber(appStateStorage.getBestBlockNumber())
   }
 
   override def getAccount(address: Address, blockNumber: BigInt): Option[Account] =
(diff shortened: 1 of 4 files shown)


AurelienRichez (approved):

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -142,6 +142,8 @@ trait Blockchain {
 
   def getBestBlockNumber(): BigInt
 
+  def getBestSavedBlockNumber(): BigInt
```
leo-bogastry: I'm wondering what are the repercussions of `getBestBlockNumber()` failing.
I have experienced an issue when calling `getBestBlock()` and getting `None` as a result. `getBestBlock()` calls `getBestBlockNumber()`.
Should `getBestBlockNumber()` be adjusted, instead of creating a new `getBestSavedBlockNumber()`?

src/main/scala/io/iohk/ethereum/jsonrpc/EthProofService.scala
```diff
@@ -210,7 +210,7 @@ class EthProofService(blockchain: Blockchain, blockGenerator: BlockGenerator, et
     blockParam match {
       case BlockParam.WithNumber(blockNumber) => getBlock(blockNumber).map(ResolvedBlock(_, pendingState = None))
       case BlockParam.Earliest => getBlock(0).map(ResolvedBlock(_, pendingState = None))
-      case BlockParam.Latest => getBlock(blockchain.getBestBlockNumber()).map(ResolvedBlock(_, pendingState = None))
+      case BlockParam.Latest => getBlock(blockchain.getBestSavedBlockNumber()).map(ResolvedBlock(_, pendingState = None))
```
jvdp: Would it perhaps make sense to add a method `getBestBlock` in `Blockchain` to first try to resolve the best block and fallback to best saved block? That removes the duplication of this logic and also the separate steps of "get block number" followed by "fetch block by number."

AnastasiiaL: did that already, fixing tests atm

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
AnastasiiaL: we need `getBestBlockNumber` as is, as some methods require to know best known number (even though it may be not stored yet. I adjusted `getBestBlock()` instead to fall back to the best stored block in case best known is not available. 

Will push in a bit after I'm done fixing tests

jvdp (approved):

PR #1000, opened 2021-05-28
Bugfix/remove block fetcher spec flakines

dzajkowski:
BlockFetcherSpec started failing randomly. Making the actor system test specific seems to fix random failures.

### DIFF
--- a/src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
+++ b/src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
@@ -36,8 +36,6 @@ import scala.concurrent.duration._
 
 class BlockFetcherSpec extends ScalaTestWithActorTestKit() with AnyFreeSpecLike with Matchers with SecureRandomBuilder {
 
-  val as: ActorSystem = ActorSystem("BlockFetcherSpec_System")
-
   "BlockFetcher" - {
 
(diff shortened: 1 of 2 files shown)


src/test/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcherSpec.scala
```diff
@@ -271,6 +269,8 @@ class BlockFetcherSpec extends ScalaTestWithActorTestKit() with AnyFreeSpecLike
   }
 
   trait TestSetup extends TestSyncConfig {
+    val as: ActorSystem = ActorSystem("BlockFetcherSpec_System")
```
jvdp, line 272: Will it shutdown?

dzajkowski: eventually for sure, but please note that it's not any worse compared to the previous approach (the difference is the amount of systems, ~10 instead of 1) but the fundamental behaviour did not change. to build a proper "start/stop" loop would mean a refactor instead of a quick-fix for flakiness

jvdp: Well not arguing against this for a quick fix for the flakiness.

Hypothetically, in an interactive sbt session (eg. IntelliJ or Metals with SBT enabled, or just sbt) this could leak memory.

PR #1001, opened 2021-05-28
[ETCM-763] [ETCM-819] improve fast sync stability

jvdp:
NB. this is based off of #977 

This PR includes:
* improved (debug) logging in FastSync to help diagnose issues
* add (unique) names for every actor (for the benefit of logging)
* fix edge cases where `Terminated` message wasn't handled and FastSync was stopped
* fix bug where skeleton handler actor failure wouldn't be recovered from, causing fast sync to get stuck (`currentSkeleton` and `skeletonHandler` are now cleared.)
* fix issue(s) caused by stale state / actors hanging around after branch resolution (every child actor is now killed and a new `SyncingHandler` is instantiated.)


src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -115,6 +115,12 @@ class FastSync(
       }
   }
 
+  private var actorCount = 0
+  private def countActor: Int = {
+    actorCount += 1
+    actorCount
+  }
```
AnastasiiaL: what about
```
private val actorCounter = new java.util.concurrent.atomic.AtomicInteger  
private def countActor: Int = actorCounter.incrementAndGet

```

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -362,19 +374,32 @@ class FastSync(
       }
     }
 
-    private def waitingForBranchResolution: Receive = handleStatus orElse {
+    private def waitingForBranchResolution: Receive = handleStatus orElse handleRequestFailure orElse {
       case FastSyncBranchResolverActor.BranchResolvedSuccessful(firstCommonBlockNumber, newMasterPeer) =>
+        log.debug(
+          s"resolved branch with first common block number $firstCommonBlockNumber for new master peer $newMasterPeer"
```
AnastasiiaL: nitpick: let's start logs messages with a capital letter

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -362,19 +374,32 @@ class FastSync(
...
-          nextBlockToFullyValidate = firstCommonBlockNumber + 1
+        log.debug("starting with fresh SyncingHandler")
+        val syncingHandler = new SyncingHandler(
+          syncState.copy(
+            bestBlockHeaderNumber = firstCommonBlockNumber,
+            nextBlockToFullyValidate = firstCommonBlockNumber + 1,
+            pivotBlockUpdateFailures = 0
+          )
         )
-        masterPeer = Some(newMasterPeer)
```
AnastasiiaL, line 375: masterPeer gets updated elsewhere now?

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -560,17 +588,23 @@ class FastSync(
       syncState = syncState.updateNextBlockToValidate(header, K, X)
     }
 
-    private def handleRewind(header: BlockHeader, peer: Peer, N: Int, duration: FiniteDuration): Unit = {
+    private def handleRewind(
+        header: BlockHeader,
+        peer: Peer,
+        N: Int,
+        duration: FiniteDuration,
+        process: Boolean = true
```
AnastasiiaL: `process` parameter name is a bit confusing. What would best describe the purpose of it? Maybe continueSync or concludeSync?

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSyncBranchResolverActor.scala
```diff
@@ -47,7 +47,11 @@ class FastSyncBranchResolverActor(
 
   private def waitingForPeerWithHighestBlock: Receive = handlePeerListMessages orElse { case StartBranchResolver =>
     getPeerWithHighestBlock match {
-      case Some(PeerWithInfo(peer, _)) => requestRecentBlockHeaders(peer, blockchain.getBestBlockNumber())
+      case Some(peerWithInfo @ PeerWithInfo(peer, _)) =>
+        log.debug(
+          s"starting branch resolution now with peer = $peerWithInfo and block number ${blockchain.getBestBlockNumber()}"
```
AnastasiiaL:
```suggestion
          "Starting branch resolution now with peer {} and block number {}", peerWithInfo, blockchain.getBestBlockNumber()
```

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
jvdp: Why not

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
jvdp: Yeah `continueSyncing` should cover it.

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
jvdp: Hmm actually no; it is selected again after branch resolution. I'll change this to be closer to the (old) intended behavior. 👍

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -340,12 +348,13 @@ class FastSync(
...
           log.info("Max number of allowed failures reached. Switching branch and master peer.")
-          handleRewind(header, masterPeer.get, fastSyncBlockValidationN, blacklistDuration)
+
+          blockHeadersQueue.dequeueAll(_ => true)
+
+          handleRewind(header, masterPeer.get, fastSyncBlockValidationN, blacklistDuration, continueSyncing = false)
 
           // Start branch resolution and wait for response from the FastSyncBranchResolver actor.
           context become waitingForBranchResolution
-          currentSkeletonState = None
```
AnastasiiaL, line 347: the skeleton state should be kept now?

jvdp: It will be reset by starting from a fresh `SyncingHandler`.

AnastasiiaL (approved):

leo-bogastry:
So one thing I have experienced when synching with Astor, was the fastSync failing without a proper error.
Here https://github.com/input-output-hk/mantis/pull/1001/files#diff-4b491aec4f7ba3bc8091f81d49100e0ccc38a2f396bcca38434118637e6c0befL637 when it calls `validateBlocks`, that method just returns `Valid` or `Invalid`.
So what do you think of also adding a log there?
This was my modification
```
def validateBlocks(
      requestedHashes: Seq[ByteString],
      blockBodies: Seq[BlockBody],
      log: LoggingAdapter
  ): BlockBodyValidationResult = {
    var result: BlockBodyValidationResult = Valid
    (requestedHashes zip blockBodies)
      .map { case (hash, body) => (blockchain.getBlockHeaderByHash(hash), body) }
      .forall {
        case (Some(header), body) =>
          val validationResult: Either[StdBlockValidator.BlockError, BlockValid] =
            validators.blockValidator.validateHeaderAndBody(header, body)
          result = validationResult.fold(
            error => {
              log.error("Block body validation failed with error {}", error)
              Invalid
            },
            _ => Valid
          )
          validationResult.isRight
        case _ =>
          result = DbError
          false
      }
    result
  }
```

AurelienRichez (approved):

jvdp:
@LeonorLunatech thanks for the suggestion, I added the log message and also refactored it a bit.

PR #1002, opened 2021-06-01
[ETCM-878] blockchain tests : BlockchainTests/vmArithmeticTest

AurelienRichez:
# Description

This PR is part of the ogoing effort to make ETS pass all tests. The vmArithmeticTest suite did not pass because of several problems : 

 - mining reward: We did not handle the `sealEngine`parameter and were always running with `NoReward` when mantis was running in test mode
 - The output of `getAccountStorage` sometime wrongly had leading zeros due to a sign bit added by BigInt.toByteAray.
 - `storageRangeAt` did not return a range of value and took a key instead of a hash as input. This kind of worked because most test used `0x00` as storage key
 - `getAccountsInRange` did not included the coinbase account. It also depended on some state instead of taking the hash given as a parameter to find the start of its range.

Fixing those problem also fixes lots of other tests

Most of the changes are external to the production code. The most impactful change is probably in `BlockchainImpl` where we now remove the leading zeros. I don't think this impact other parts of the application. 

Regarding `TestService.scala`, I removed the variable `accountRangeOffset`. I also added some state in testmode with a map `preimageCache`. This map is used to keep track of what storage address were used during one test run and be able to implement `storageRangeAt`. This is not really elegant so I will be happy for suggestion on how to make it better.

## Important note 
While this PR improves the implementation of `storageRangeAt` and `getAccountsInRange`, they are still not completely functional. They only work in test mode. There are also some edge case not handled : 

 - `getAccountsInRange` is only able to return account known in the genesis state and the account used for coinbase (So it won't return account created in a transaction)
 - `storageRangeAt` should be able to get the state not only on a given block but also after an arbitrary transaction inside the block. We currently only handle the first case. I'm not sure the tests actually need the second part so it might be alright for ETS.
 
 Also the `storageRangeAt` implementation is not really smart. It just iterate over known keys instead of actually reading the Merkle Patricia Trie. For tests with few values in storage It's good enough though.

### DIFF
--- a/ets/README.md
+++ b/ets/README.md
@@ -37,6 +37,7 @@ You can also run parts of the suite; refer to `ets/retesteth --help` for details
 You should run Mantis outside Nix as that is probably more convenient for your
 tooling (eg. attaching a debugger.)
 
+    rm -rf ~/.mantis/test # delete the storage to be sure no previous run is affecting the tests
     sbt -Dconfig.file=./src/main/resources/conf/testmode.conf -Dlogging.logs-level=WARN run
 
 Retesteth will need to be able to connect to Mantis, running on the host
@@ -44,8 +45,22 @@ system. First, find the IP it should use:
 
     nix-in-docker/run --command "getent hosts host.docker.internal"
 
-Edit `ets/config/mantis` and replace `0.0.0.0` with this IP.
-
 Finally, run retesteth in Nix in Docker:
 
-    nix-in-docker/run --command "ets/retesteth -t GeneralStateTests"
+    nix-in-docker/run --command "ets/retesteth -t GeneralStateTests -- --nodes <ip>:8546"
+
+## Useful options:
+
+You can run one test be selecting one suite and using `--singletest`, for instance: 
+
+    nix-in-docker/run -t BlockchainTests/ValidBlocks/VMTests/vmArithmeticTest -- --nodes <ip>:8546 --singletest add0"
+
+However it's not always clear in wich subfolder the suite is when looking at the output of retesteth.
+
+To get more insight about what is happening, you can use `--verbosity 6`. It will print every RPC call 
+made by retesteth and also print out the state by using our `debug_*` endpoints. Note however that 
+`debug_accountRange` and `debug_storageRangeAt` implementations are not complete at the moment :
+
+ - `debug_accountRange` will only list accounts known at the genesis state. 
+ - `debug_storageRangeAt` is not able to show the state after an arbitrary transaction inside a block.
+It will just return the state after all transaction in the block have run.
--- a/ets/README.md
+++ b/ets/README.md
@@ -37,7 +37,6 @@ You can also run parts of the suite; refer to `ets/retesteth --help` for details
 You should run Mantis outside Nix as that is probably more convenient for your
 tooling (eg. attaching a debugger.)
 
-    rm -rf ~/.mantis/test # delete the storage to be sure no previous run is affecting the tests
     sbt -Dconfig.file=./src/main/resources/conf/testmode.conf -Dlogging.logs-level=WARN run
 
 Retesteth will need to be able to connect to Mantis, running on the host
@@ -51,7 +50,7 @@ Finally, run retesteth in Nix in Docker:
 
 ## Useful options:
 
-You can run one test be selecting one suite and using `--singletest`, for instance: 
+You can run one test by selecting one suite and using `--singletest`, for instance: 
 
     nix-in-docker/run -t BlockchainTests/ValidBlocks/VMTests/vmArithmeticTest -- --nodes <ip>:8546 --singletest add0"
 
--- a/src/main/scala/io/iohk/ethereum/Mantis.scala
+++ b/src/main/scala/io/iohk/ethereum/Mantis.scala
@@ -2,6 +2,7 @@ package io.iohk.ethereum
 
 import io.iohk.ethereum.nodebuilder.{StdNode, TestNode}
 import io.iohk.ethereum.utils.{Config, Logger}
+import org.rocksdb
 
 import java.nio.file.{Files, Paths}
 import java.util.logging.LogManager
@@ -24,14 +25,7 @@ object Mantis extends Logger {
   }
 
   private def deleteRocksDBFiles(): Unit = {
-    val path = Paths.get(Config.Db.RocksDb.path)
-    if (path.toFile.exists()) {
-      log.info("Deleting previous database {}", Config.Db.RocksDb.path)
-      Files
-        .list(path)
-        .map(_.toFile)
-        .filter(_.isFile)
-        .forEach(f => f.delete())
-    }
+    log.warn("Deleting previous database {}", Config.Db.RocksDb.path)
+    rocksdb.RocksDB.destroyDB(Config.Db.RocksDb.path, new rocksdb.Options())
   }
 }
--- a/src/main/scala/io/iohk/ethereum/Mantis.scala
+++ b/src/main/scala/io/iohk/ethereum/Mantis.scala
@@ -3,8 +3,8 @@ package io.iohk.ethereum
 import io.iohk.ethereum.nodebuilder.{StdNode, TestNode}
 import io.iohk.ethereum.utils.{Config, Logger}
 
+import java.nio.file.{Files, Paths}
 import java.util.logging.LogManager
-import scala.reflect.io.Directory
 
 object Mantis extends Logger {
   def main(args: Array[String]): Unit = {
@@ -13,8 +13,7 @@ object Mantis extends Logger {
     val node =
       if (Config.testmode) {
         log.info("Starting Mantis in test mode")
-        log.info("Deleting previous database {}", Config.Db.RocksDb.path)
-        Directory(Config.Db.RocksDb.path).deleteRecursively()
+        deleteRocksDBFiles()
         new TestNode
       } else new StdNode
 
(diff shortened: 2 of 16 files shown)


ets/README.md
```diff
@@ -37,15 +37,30 @@ You can also run parts of the suite; refer to `ets/retesteth --help` for details
 You should run Mantis outside Nix as that is probably more convenient for your
 tooling (eg. attaching a debugger.)
 
+    rm -rf ~/.mantis/test # delete the storage to be sure no previous run is affecting the tests
```
jvdp: Can we do this cleanup while booting testmode?

ets/README.md
```diff
@@ -37,15 +37,30 @@ You can also run parts of the suite; refer to `ets/retesteth --help` for details
...
 system. First, find the IP it should use:
 
     nix-in-docker/run --command "getent hosts host.docker.internal"
 
-Edit `ets/config/mantis` and replace `0.0.0.0` with this IP.
-
 Finally, run retesteth in Nix in Docker:
 
-    nix-in-docker/run --command "ets/retesteth -t GeneralStateTests"
+    nix-in-docker/run --command "ets/retesteth -t GeneralStateTests -- --nodes <ip>:8546"
```
jvdp, line 49: Cool! 👍

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -320,7 +320,17 @@ class BlockchainImpl(
...
+    val bigIntValue = mpt.get(position).getOrElse(BigInt(0))
+    val byteArrayValue = bigIntValue.toByteArray
+
+    // BigInt.toArray actually might return one more byte than necessary because it add a sign bit, which in our case
+    // will always be 0. This would adds unwanted 0 bytes and might cause the value to be 33 byte long while an EVM
+    // word is 32 byte long.
+    if (bigIntValue != 0)
+      ByteString(byteArrayValue.dropWhile(_ == 0))
+    else
+      ByteString(byteArrayValue)
```
jvdp, line 333: Looks like a job for regexes to me!

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
```diff
@@ -290,57 +319,86 @@ class TestService(
...
           else UInt256(0).bytes
       ).rightNow
     }
   }
 
+  /** Get the list of storage values starting from _begin and up to _begin + _maxResults at given block.
+    * nexKey field is the next key hash if any key left in the state, or 0x00 otherwise.
+    *
+    * Normally, this RPC method is supposed to also be able to look up the state after after transaction
+    * _txIndex is executed. This is currently not supported in mantis.
```
jvdp, line 362: Do we have a ticket to track this lack of support? And are tests failing because of it?

src/main/scala/io/iohk/ethereum/testmode/TestModeWorldStateProxy.scala
```diff
@@ -0,0 +1,136 @@
...
+import io.iohk.ethereum.db.storage.EvmCodeStorage.Code
+import io.iohk.ethereum.domain.Account.accountSerializer
+import io.iohk.ethereum.domain.{Account, Address, UInt256}
+import io.iohk.ethereum.ledger.{InMemorySimpleMapProxy, InMemoryWorldStateProxy, InMemoryWorldStateProxyStorage}
+import io.iohk.ethereum.mpt.MerklePatriciaTrie
+
+/** This is a wrapper around InMemoryWorldStateProxy.
+  * Its only role is to store the storage key encountered during a run to store them for debugging purpose.
+  */
+class TestModeWorldStateProxy(
```
jvdp: Why not a case class?

jvdp (approved):
> While this PR improves the implementation of `storageRangeAt` and `getAccountsInRange`, they are still not completely functional. They only work in test mode.

With the knowledge you've gained fixing these tests so far: with so much of this stuff being test-mode specific, do you think there's a missed opportunity for covering more of the production code by ETS?

Also, is there an issue that the testservice now has very different operation compared to the other RPC services? (eg. the regular `eth_` methods listed here: https://github.com/ethereum/retesteth/wiki/RPC-Methods) Or am I misunderstanding how this works?

src/main/scala/io/iohk/ethereum/testmode/TestModeWorldStateProxy.scala
AurelienRichez: Yeah you're right. I blindly copied the logic from `InMemoryWorldStateProxy`. But a case class will give us the copy for free.

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
AurelienRichez: You mean we should change the output when transforming to a String in test service ? Here we are using BigInt and byte arrays. As It seems not really correct that `getAccountStorageAt` might return 33 bytes I fixed it here.

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
AurelienRichez: We have ETCM-784 and ETCM-758 which are related to this. I'll add a comment there. 

I did not find any test specifically failing because of this but I did not really look why other suites are failing. So I went with an iterative approach by implementing the minimum to make `vmArithmeticTest` pass.

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
AurelienRichez: Actually, we have a helper `bigIntToUnsignedByteArray` exactly for that in the project

ets/README.md
```diff
@@ -37,15 +37,30 @@ You can also run parts of the suite; refer to `ets/retesteth --help` for details
...
+
+## Useful options:
+
+You can run one test be selecting one suite and using `--singletest`, for instance:
```
AnastasiiaL: typo: be -> by

src/main/scala/io/iohk/ethereum/Mantis.scala
```diff
@@ -2,7 +2,10 @@ package io.iohk.ethereum
 
 import io.iohk.ethereum.nodebuilder.{StdNode, TestNode}
 import io.iohk.ethereum.utils.{Config, Logger}
+
+import java.nio.file.Path
```
AnastasiiaL: this import is not needed

src/main/scala/io/iohk/ethereum/Mantis.scala
```diff
@@ -11,6 +14,8 @@ object Mantis extends Logger {
     val node =
       if (Config.testmode) {
         log.info("Starting Mantis in test mode")
+        log.info(s"Deleting previous database ${Config.Db.RocksDb.path}")
```
AnastasiiaL:
```suggestion
        log.info("Deleting previous database {}", Config.Db.RocksDb.path)
```
the decision to use this format uniformly ^ was made a few months ago

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -320,7 +320,17 @@ class BlockchainImpl(
     val mpt =
       if (ethCompatibleStorage) domain.EthereumUInt256Mpt.storageMpt(rootHash, storage)
       else domain.ArbitraryIntegerMpt.storageMpt(rootHash, storage)
-    ByteString(mpt.get(position).getOrElse(BigInt(0)).toByteArray)
+
+    val bigIntValue = mpt.get(position).getOrElse(BigInt(0))
+    val byteArrayValue = bigIntValue.toByteArray
+
+    // BigInt.toArray actually might return one more byte than necessary because it add a sign bit, which in our case
```
AnastasiiaL:
```suggestion
    // BigInt.toArray actually might return one more byte than necessary because it adds a sign bit, which in our case
```

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -320,7 +320,17 @@ class BlockchainImpl(
...
+    // will always be 0. This would adds unwanted 0 bytes and might cause the value to be 33 byte long while an EVM
```
AnastasiiaL:
```suggestion
    // will always be 0. This would add unwanted 0 bytes and might cause the value to be 33 byte long while an EVM
```

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
```diff
@@ -245,10 +260,11 @@ class TestService(
 
   def importRawBlock(request: ImportRawBlockRequest): ServiceResponse[ImportRawBlockResponse] = {
     Try(decode(request.blockRlp).toBlock) match {
-      case Failure(_) => Task.now(Left(JsonRpcError(-1, "block validation failed!", None)))
+      case Failure(e) =>
```
AnastasiiaL: `e` is not used, why to name it?

AnastasiiaL (approved):
a couple non-blocking comments, otherwise looks good 👍🏻

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
```diff
@@ -259,7 +275,9 @@ class TestService(
       case BlockImportedToTop(blockImportData) =>
         val blockHash = s"0x${ByteStringUtils.hash2string(blockImportData.head.block.header.hash)}"
         ImportRawBlockResponse(blockHash).rightNow
-      case _ => Task.now(Left(JsonRpcError(-1, "block validation failed!", None)))
+      case e =>
+        log.warn(s"Block import failed with ${e}")
```
AnastasiiaL: log format also here and in couple more places

PR #1004, opened 2021-06-02
[ETCM-841] Rework protocol negotiation

dzajkowski:
Mantis implements a protocol version ETC/64 which is conflicting with ETH/64.
This PRs goal is to enable the coexistence of both and some future proofing (renames, simple protocol negotiation).

### DIFF
--- a/src/main/scala/io/iohk/ethereum/network/handshaker/EtcHelloExchangeState.scala
+++ b/src/main/scala/io/iohk/ethereum/network/handshaker/EtcHelloExchangeState.scala
@@ -30,7 +30,9 @@ case class EtcHelloExchangeState(handshakerConfiguration: EtcHandshakerConfigura
       case ProtocolVersions.ETC64 => EtcNodeStatus64ExchangeState(handshakerConfiguration)
       case ProtocolVersions.ETH63 => EtcNodeStatus63ExchangeState(handshakerConfiguration)
       case _ =>
-        log.debug(s"Connected peer does not support ${ProtocolVersions.ETH63} / ${ProtocolVersions.ETC64} protocol. Disconnecting.")
+        log.debug(
+          s"Connected peer does not support ${ProtocolVersions.ETH63} / ${ProtocolVersions.ETC64} protocol. Disconnecting."
+        )
         DisconnectedState(Disconnect.Reasons.IncompatibleP2pProtocolVersion)
     }
   }
--- a/src/main/scala/io/iohk/ethereum/network/handshaker/EtcHelloExchangeState.scala
+++ b/src/main/scala/io/iohk/ethereum/network/handshaker/EtcHelloExchangeState.scala
@@ -34,7 +34,7 @@ case class EtcHelloExchangeState(handshakerConfiguration: EtcHandshakerConfigura
       EtcNodeStatus63ExchangeState(handshakerConfiguration)
     else {
       log.debug(
-        s"Connected peer does not support eth ${ProtocolVersions.PV63.toByte} / ${ProtocolVersions.PV64.toByte} protocol. Disconnecting."
+        s"Connected peer does not support eth ${ProtocolVersions.PV63} / ${ProtocolVersions.PV64} protocol. Disconnecting."
       )
       DisconnectedState(Disconnect.Reasons.IncompatibleP2pProtocolVersion)
     }
--- a/src/main/scala/io/iohk/ethereum/network/handshaker/EtcHelloExchangeState.scala
+++ b/src/main/scala/io/iohk/ethereum/network/handshaker/EtcHelloExchangeState.scala
@@ -32,7 +32,9 @@ case class EtcHelloExchangeState(handshakerConfiguration: EtcHandshakerConfigura
       case Some(ProtocolVersions.ETH63) => EtcNodeStatus63ExchangeState(handshakerConfiguration)
       case _ =>
         log.debug(
-          s"Connected peer does not support ${ProtocolVersions.ETH63} / ${ProtocolVersions.ETC64} protocol. Disconnecting."
+          s"Connected peer does not support {} / {} protocol. Disconnecting.",
+          ProtocolVersions.ETH63,
+          ProtocolVersions.ETC64
         )
         DisconnectedState(Disconnect.Reasons.IncompatibleP2pProtocolVersion)
     }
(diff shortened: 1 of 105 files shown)


src/main/scala/io/iohk/ethereum/network/handshaker/EtcHelloExchangeState.scala
```diff
@@ -26,19 +26,15 @@ case class EtcHelloExchangeState(handshakerConfiguration: EtcHandshakerConfigura
...
-      log.debug(
-        s"Connected peer does not support eth ${ProtocolVersions.PV63.toByte} / ${ProtocolVersions.PV64.toByte} protocol. Disconnecting."
-      )
-      DisconnectedState(Disconnect.Reasons.IncompatibleP2pProtocolVersion)
+    capability match {
+      case ProtocolVersions.ETC64 => EtcNodeStatus64ExchangeState(handshakerConfiguration)
+      case ProtocolVersions.ETH63 => EtcNodeStatus63ExchangeState(handshakerConfiguration)
+      case _ =>
+        log.debug(
+          s"Connected peer does not support ${ProtocolVersions.ETH63} / ${ProtocolVersions.ETC64} protocol. Disconnecting."
```
AnastasiiaL: shall we align the log msg format?
```suggestion
          "Connected peer does not support {} / {} protocol. Disconnecting.", ProtocolVersions.ETH63, ProtocolVersions.ETC64
```

src/main/scala/io/iohk/ethereum/network/p2p/messages/Capability.scala
```diff
@@ -7,6 +7,22 @@ import io.iohk.ethereum.rlp.{RLPEncodeable, RLPException, RLPList, RLPSerializab
 case class Capability(name: String, version: Byte)
 
 object Capability {
+  def negotiate(c1: List[Capability], c2: List[Capability]): Option[Capability] = {
+    c1.intersect(c2).maxByOption(_.version) // FIXME ignores branches and other protocols
+  }
```
AnastasiiaL: curly braces can be omitted here

src/main/scala/io/iohk/ethereum/network/p2p/messages/Capability.scala
```diff
@@ -7,6 +7,22 @@ import io.iohk.ethereum.rlp.{RLPEncodeable, RLPException, RLPList, RLPSerializab
...
+
+  private val pattern = "(.*)/(\\d*)".r
+
+  def from(protocolVersion: String): Capability =
+    protocolVersion match {
+      case pattern(name, version) => Capability(name, version.toByte)
+      case _ => throw new RuntimeException(s"Unable to parse capability $protocolVersion")
```
AnastasiiaL, line 26:
```suggestion
      case _ => throw new RuntimeException("Unable to parse capability {}", protocolVersion)
```

dzajkowski: I'm not convinced that RuntimeException accepts template strings.

src/main/scala/io/iohk/ethereum/network/p2p/messages/Capability.scala
```diff
@@ -7,6 +7,21 @@ import io.iohk.ethereum.rlp.{RLPEncodeable, RLPException, RLPList, RLPSerializab
 case class Capability(name: String, version: Byte)
 
 object Capability {
+  def negotiate(c1: List[Capability], c2: List[Capability]): Option[Capability] =
+    c1.intersect(c2).maxByOption(_.version) // FIXME ignores branches and other protocols
+
+  private val pattern = "(.*)/(\\d*)".r
+
+  def from(protocolVersion: String): Capability =
```
AurelienRichez: Could we make it clear that this function throws ? `parseUnsafe` maybe ? 
Or make it return an option and let `BlockchainConfig` throw in case of none

src/main/scala/io/iohk/ethereum/nodebuilder/NodeBuilder.scala
```diff
@@ -246,7 +248,7 @@ trait PeerManagerActorBuilder {
       EthereumMessageDecoder,
       discoveryConfig,
       blacklist,
-      Config.Network.protocolVersion
+      blockchainConfig.capabilities // TODO replace with a list of capabilities
```
AurelienRichez: Isn't it already a list of capabilities ?

src/main/scala/io/iohk/ethereum/utils/BlockchainConfig.scala
```diff
@@ -122,6 +124,8 @@ object BlockchainConfig {
     val allowedMinersPublicKeys = readPubKeySet(blockchainConfig, "allowed-miners")
 
     val ecip1099BlockNumber: BigInt = BigInt(blockchainConfig.getString("ecip1099-block-number"))
+    val capabilities: List[Capability] =
+      blockchainConfig.getStringList("capabilities").asScala.toList.map(Capability.from)
```
AurelienRichez: I guess we should also check that the capability in the config is indeed a capability implemented by Mantis ?

src/main/scala/io/iohk/ethereum/network/p2p/messages/Capability.scala
```diff
@@ -7,6 +7,21 @@ import io.iohk.ethereum.rlp.{RLPEncodeable, RLPException, RLPList, RLPSerializab
 case class Capability(name: String, version: Byte)
 
 object Capability {
+  def negotiate(c1: List[Capability], c2: List[Capability]): Option[Capability] =
+    c1.intersect(c2).maxByOption(_.version) // FIXME ignores branches and other protocols
```
AurelienRichez: It seems that it duplicates the `best` function here.
We should probably have a single point where the best protocol is chosen, no ?
Something like :
```suggestion
   // FIXME ignores branches and other protocols
   c1.intersect(c2) match {
     case Nil => None
     case nonEmpty => best(nonEmpty)
   }
```

leo-bogastry:
The ETCM-841 ticket description mentions "The task is to rename the whole protocol to PV164 and mark @depracated". Are you thinking keeping it as ETC/64 instead? With or without deprecation?

dzajkowski:
@LeonorLunatech As I mentioned during refinement - I went through a few iterations on how to deal with this issue and by explicitly stating the namespace (prefix) mantis can have any number of protocols. The only limitation will be the negotiation mechanism.

TL;DR the deprecation is not needed with the prefix. This way mantis stays backwards compatible.

src/main/scala/io/iohk/ethereum/network/p2p/messages/Capability.scala
dzajkowski: good point. I'll stick to the exception. not the best solution but we need mantis to fail when configs are not properly defined.

AurelienRichez (approved):

leo-bogastry (approved):

PR #1006, opened 2021-06-03
[ETCM-893] Add support for chains with ids bigger than 127

leo-bogastry:
# Description

Currently Mantis only supports networks id a chaind id up to 127.
This brought problems syncing with Astor, that has a chainid of 212 

# Proposed Solution

Blockchain config, chainid field from Byte to BigInt
ECDSASignature, v field from Byte to BigInt

# Testing
Synchronization with ETC, Sagano and Astor were successfull
On a private network (network id = 2 / chain id = 61) a new tx was created, included in a block and successfully propagated

### DIFF
--- a/src/test/scala/io/iohk/ethereum/network/p2p/messages/NewBlockSpec.scala
+++ b/src/test/scala/io/iohk/ethereum/network/p2p/messages/NewBlockSpec.scala
@@ -12,10 +12,17 @@ import org.scalatest.funsuite.AnyFunSuite
 
 class NewBlockSpec extends AnyFunSuite with ScalaCheckPropertyChecks with ObjectGenerators with SecureRandomBuilder {
 
-  val chainId = Hex.decode("3d").head
+  val chainId_1 = 1
+  val chainId_212 = 212
 
   test("NewBlock v63 messages are encoded and decoded properly") {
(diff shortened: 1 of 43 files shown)


AurelienRichez (approved):

src/test/scala/io/iohk/ethereum/network/p2p/messages/NewBlockSpec.scala
```diff
@@ -12,10 +12,17 @@ import org.scalatest.funsuite.AnyFunSuite
 
 class NewBlockSpec extends AnyFunSuite with ScalaCheckPropertyChecks with ObjectGenerators with SecureRandomBuilder {
 
-  val chainId = Hex.decode("3d").head
+  val chainId_1 = 1
+  val chainId_212 = 212
```
AnastasiiaL: nitpick: mixing camelCase with snake_case. Shall we name them chainId1 and chainId212?

AnastasiiaL (approved):

src/test/scala/io/iohk/ethereum/network/p2p/messages/NewBlockSpec.scala
leo-bogastry: fixed

PR #1012, opened 2021-06-16
Etcm 926 miners

krzysztofpaliga-iohk:
Changes the mining algorithm from restricted-pow to pow


src/main/resources/conf/pottery.conf
```diff
@@ -14,7 +14,7 @@ mantis {
 	consensus {
     coinbase = "0011223344556677889900112233445566778899" # has to be changed for each node
     mining-enabled = true
-    protocol = "restricted-pow"
+    protocol = "pow"
```
dzajkowski, line 17: not that it matters much but I think only one change is mandatory.

dzajkowski (approved):

leo-bogastry (approved):

PR #1017, opened 2021-06-18
ETCM-938 Refactor Blockchain and extract methods getEvmCodeByHash and mptStateSavedKeys

leo-bogastry:
# Description

In order to simplify the Blockchain complexity, methods `getEvmCodeByHash` and `mptStateSavedKeys` where removed, since they are used in just a few locations
`getEvmCodeByHash` is just a `get` done directly to the EvmCodeStorage
`mptStateSavedKeys` is used only in `SyncStateScheduler` so it was moved there

# Testing
All should work as before

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/BlockchainHostActor.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/BlockchainHostActor.scala
@@ -2,6 +2,7 @@ package io.iohk.ethereum.blockchain.sync
 
 import akka.actor.{Actor, ActorLogging, ActorRef, Props}
 import akka.util.ByteString
+import io.iohk.ethereum.db.storage.EvmCodeStorage
 import io.iohk.ethereum.domain.{BlockHeader, Blockchain}
 import io.iohk.ethereum.network.PeerEventBusActor.PeerEvent.MessageFromPeer
 import io.iohk.ethereum.network.PeerEventBusActor.SubscriptionClassifier.MessageClassifier
@@ -20,6 +21,7 @@ import io.iohk.ethereum.network.p2p.messages.Codes
   */
 class BlockchainHostActor(
     blockchain: Blockchain,
+    evmCodeStorage: EvmCodeStorage,
     peerConfiguration: PeerConfiguration,
     peerEventBusActor: ActorRef,
     etcPeerManagerActor: ActorRef
@@ -54,7 +56,7 @@ class BlockchainHostActor(
         val maybeMptNodeData = blockchain.getMptNodeByHash(hash).map(e => e.toBytes: ByteString)
 
         //If no mpt node was found, fetch evm by hash
-        maybeMptNodeData.orElse(blockchain.getEvmCodeByHash(hash))
+        maybeMptNodeData.orElse(evmCodeStorage.get(hash))
       }
 
       Some(NodeData(nodeData))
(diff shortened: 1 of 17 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/BlockchainHostActor.scala
```diff
@@ -20,6 +21,7 @@ import io.iohk.ethereum.network.p2p.messages.Codes
   */
 class BlockchainHostActor(
     blockchain: Blockchain,
+    evmCodeStorage: EvmCodeStorage,
```
jvdp, line 24: Do you expect it will happen more during the refactoring that actors will have access to the storages directly? (To clarify: I don't really see a problem with it.)

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
```diff
@@ -120,6 +124,8 @@ object SyncController {
   def props(
       appStateStorage: AppStateStorage,
       blockchain: Blockchain,
+      evmCodeStorage: EvmCodeStorage,
+      nodeStorage: NodeStorage,
```
jvdp, line 128: I wonder at what point we'll create a case class `MantisEnv` or something that has all these instances in it.

jvdp (approved):

src/it/scala/io/iohk/ethereum/sync/FastSyncItSpec.scala
```diff
@@ -34,8 +34,14 @@ class FastSyncItSpec extends FlatSpecBase with Matchers with BeforeAndAfterAll {
...
     } yield {
-      assert(peer1.bl.getBestBlockNumber() == peer2.bl.getBestBlockNumber() - peer2.testSyncConfig.pivotBlockOffset)
-      assert(peer1.bl.getBestBlockNumber() == peer3.bl.getBestBlockNumber() - peer3.testSyncConfig.pivotBlockOffset)
+      assert(
+        peer1.bl
+          .getBestBlockNumber() == peer2.bl.getBestBlockNumber() - peer2.testSyncConfig.pivotBlockOffset
+      )
+      assert(
+        peer1.bl
+          .getBestBlockNumber() == peer3.bl.getBestBlockNumber() - peer3.testSyncConfig.pivotBlockOffset
```
AurelienRichez: nitpick: the line break in `peer1.bl.getBestBlockNumber()` makes it less readable imho

src/main/scala/io/iohk/ethereum/blockchain/sync/BlockchainHostActor.scala
AurelienRichez: I think might happen for state storage also as the Blockchain class does not really bring anything in term of read access (I think it's ok for read access).

We will probably want write access to be through a single service though, or redefined what we call a storage to be a class that manage several rocksdb namespace internally, because we have some writes that span several storages (for instance when adding a block we want to store the block itself but also update the transaction mapping and potentially some other metadata).

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
AurelienRichez: We already have something similar `BlockchainStorages` which is extended by `Storages` in the trait `StoragesComponent`, and with a concrete implementation in `DefaultStorages`. 

It is a big bag with every storage in it so we will probably want to separate them to have storages that work together in the same container, and identify storages that are a "view" of other storages.

AurelienRichez (approved):

src/main/scala/io/iohk/ethereum/blockchain/sync/BlockchainHostActor.scala
leo-bogastry: 👍

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
leo-bogastry: 👍

src/it/scala/io/iohk/ethereum/sync/FastSyncItSpec.scala
leo-bogastry: I've put it back!

AnastasiiaL (approved):

PR #1018, opened 2021-06-18
Feature/etcm 354/forkid

lukasz-golebiewski:
# Description

First part of https://eips.ethereum.org/EIPS/eip-2124
Adds the ForkId calculation and it's RLP encoding

### DIFF
--- a/src/main/scala/io/iohk/ethereum/forkid/ForkId.scala
+++ b/src/main/scala/io/iohk/ethereum/forkid/ForkId.scala
@@ -38,16 +38,7 @@ object ForkId {
       else None
     }
 
-    (maybeDaoBlock.toList ++
-      config.forkBlockNumbers.productIterator.toList.flatMap {
-        case i: BigInt => Some(i)
-        case i: Option[_] =>
-          i.flatMap {
-            case n if n.isInstanceOf[BigInt] => Some(n.asInstanceOf[BigInt])
-            case n => None
-          }
-        case default => None
-      })
+    (maybeDaoBlock.toList ++ config.forkBlockNumbers.all)
       .filterNot(v => v == 0 || v == noFork)
       .distinct
       .sorted
--- a/src/main/scala/io/iohk/ethereum/forkid/ForkId.scala
+++ b/src/main/scala/io/iohk/ethereum/forkid/ForkId.scala
@@ -7,6 +7,7 @@ import akka.util.ByteString
 import io.iohk.ethereum.utils.BlockchainConfig
 import io.iohk.ethereum.utils.BigIntExtensionMethods._
 import io.iohk.ethereum.utils.ByteUtils._
+import io.iohk.ethereum.rlp._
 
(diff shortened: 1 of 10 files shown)


dzajkowski:
were the geth test of any use? https://github.com/etclabscore/core-geth/blob/master/core/forkid/forkid_test.go#L132-L169

src/main/scala/io/iohk/ethereum/forkid/ForkId.scala
```diff
@@ -0,0 +1,69 @@
...
+      fork > head
+    }
+    new ForkId(crc.getValue(), next)
+  }
+
+  val noFork = BigInt("1000000000000000000")
+
+  def gatherForks(config: BlockchainConfig): List[BigInt] =
+    (config.daoForkConfig.map(_.forkBlockNumber).toList ++
+      config.forkBlockNumbers.productIterator.toList.flatMap {
```
dzajkowski: this assumes the order of fields in config case classes, right?

lukasz-golebiewski: The list is being sorted at the very end

src/main/scala/io/iohk/ethereum/forkid/ForkId.scala
```diff
@@ -0,0 +1,78 @@
...
+        case i: BigInt => Some(i)
+        case i: Option[_] =>
+          i.flatMap {
+            case n if n.isInstanceOf[BigInt] => Some(n.asInstanceOf[BigInt])
+            case n => None
+          }
+        case default => None
+      })
+      .filterNot(v => v == 0 || v == noFork)
```
AurelienRichez, line 42: I don't get why we filter out `noFork`. It does not seem to be used somewhere else

src/main/scala/io/iohk/ethereum/forkid/ForkId.scala
AurelienRichez: I would put that part with the `productIterator` inside `ForkBlockNumbers` directly. Even though it seem reasonable that every `BigInt` there will be a fork block number we never really know. If we put this function inside `ForkBlockNumbers` we make this assumption more explicit for devs modifying the case class.

If we want to avoid any logic in the config part, we could add a comment in ForkBlockNumbers which explain that every BigInt there must be a fork block number. 

What do you think ?

src/main/scala/io/iohk/ethereum/forkid/ForkId.scala
lukasz-golebiewski: The same configuration structure `BlockchainConfig` is re-used for multiple chains. Forks which shouldn't occur in a given chain are configured to happen at block height "1000000000000000000" which is equivalent to "never". We don't want those fork numbers on our fork id list

dzajkowski (approved):

AurelienRichez (approved):

PR #1019, opened 2021-06-18
[ETCM-841] Consume remaining data when extracting Hello in order

dzajkowski:
FrameCodec holds Secrets which is a mutable structure. The unfortunate reality is that it cannot identify that it is reading a message out of order. This goes exactly in opposite to how akka actors handle messages and can cause Frame deserialization to throw a "MAC mismatch" exception.

The fix is to cause the FrameCodec to read any available data immediately and not reschedule it to a mailbox.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
+++ b/src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
@@ -164,11 +164,12 @@ class RLPxConnectionHandler(
         case AuthHandshakeSuccess(secrets, remotePubKey) =>
           log.debug(s"Auth handshake succeeded for peer {}", peerId)
           context.parent ! ConnectionEstablished(remotePubKey)
-          if (remainingData.nonEmpty)
-            context.self ! Received(remainingData)
           // following the specification at https://github.com/ethereum/devp2p/blob/master/rlpx.md#initial-handshake
           // point 6 indicates that the next messages needs to be initial 'Hello'
-          context become awaitInitialHello(extractor(secrets))
+          // Unfortunately it is hard to figure out the proper order for messages to be handled in.
+          // FrameCodec assumes that bytes will arrive in the expected order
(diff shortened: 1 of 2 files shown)


src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
```diff
@@ -181,45 +182,37 @@ class RLPxConnectionHandler(
...
+          )
+        case Ack if cancellableAckTimeout.nonEmpty =>
+          //Cancel pending message timeout
+          cancellableAckTimeout.foreach(_.cancellable.cancel())
+          context become awaitInitialHello(extractor, None, seqNumber)
 
-    }
+        case AckTimeout(ackSeqNumber) if cancellableAckTimeout.exists(_.seqNumber == ackSeqNumber) =>
+          cancellableAckTimeout.foreach(_.cancellable.cancel())
+          log.error(s"[Stopping Connection] Sending 'Hello' to {} failed", peerId)
```
AnastasiiaL:
```suggestion
          log.error("[Stopping Connection] Sending 'Hello' to {} failed", peerId)
```

src/main/scala/io/iohk/ethereum/network/rlpx/RLPxConnectionHandler.scala
```diff
@@ -181,45 +182,37 @@ class RLPxConnectionHandler(
...
+          context stop self
+        case Received(data) =>
+          extractHello(extractor, data, cancellableAckTimeout, seqNumber)
+      }
 
-    private def handleReceiveHello(
+    private def extractHello(
         extractor: HelloCodec,
-        cancellableAckTimeout: Option[CancellableAckTimeout] = None,
-        seqNumber: Int = 0
```
AnastasiiaL, line 221: why not to keep the default values?

AnastasiiaL (approved):

lukasz-golebiewski (approved):

PR #1020, opened 2021-06-21
ETCM-941: Simplify Block Importer

AnastasiiaL:
# Description

- Top/NotOnTop state removed as it's a duplication from the BlockFetcher
- PrintStatus removed, as same info printed in the BlockFetcher already
- some code simplifications

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
@@ -24,7 +24,6 @@ import monix.execution.Scheduler
 
 import scala.concurrent.duration._
 
-// scalastyle:off cyclomatic.complexity
 class BlockImporter(
     fetcher: ActorRef,
     ledger: Ledger,
@@ -53,15 +52,8 @@ class BlockImporter(
     start()
   }
 
-  private def handleTopMessages(state: ImporterState, currentBehavior: Behavior): Receive = {
-    case OnTop => context become currentBehavior(state.onTop())
-    case NotOnTop => context become currentBehavior(state.notOnTop())
-  }
-
-  private def running(state: ImporterState): Receive = handleTopMessages(state, running) orElse {
+  private def running(state: ImporterState): Receive = {
     case ReceiveTimeout => self ! PickBlocks
-
-    case PrintStatus => log.info("Block: {}, is on top?: {}", blockchain.getBestBlockNumber(), state.isOnTop)
 
     case BlockFetcher.PickedBlocks(blocks) =>
       SignedTransaction.retrieveSendersInBackGround(blocks.toList.map(_.body))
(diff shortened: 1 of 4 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -244,27 +231,21 @@ class BlockImporter(
...
               Running
             case BlockImportFailedDueToMissingNode(missingNodeException) =>
               Task.raiseError(missingNodeException)
-            case BlockImportFailed(error) =>
-              if (informFetcherOnFail) {
-                fetcher ! BlockFetcher.BlockImportFailed(block.number, BlacklistReason.BlockImportError(error))
-              }
+            case BlockImportFailed(error) if informFetcherOnFail =>
+              fetcher ! BlockFetcher.BlockImportFailed(block.number, BlacklistReason.BlockImportError(error))
+            case _ => ()
```
AurelienRichez: I think the previous implementation was done this way intentionally to get the exhaustive check. It's probably not a big deal as if we ever change `BlockImportResult` we will probably check here also anyway. Otherwise we can compromise with `case BlockEnqueued | DuplicateBlock | UnknownParent | BlockImportFailed(_)`.

AurelienRichez (approved):

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -244,27 +231,21 @@ class BlockImporter(
               broadcastBlocks(blocks, weights)
               updateTxPool(importedBlocksData.map(_.block), Seq.empty)
               supervisor ! ProgressProtocol.ImportedBlock(block.number, internally)
-            case BlockEnqueued => ()
-            case DuplicateBlock => ()
-            case UnknownParent => () // This is normal when receiving broadcast blocks
```
dzajkowski, line 249: maybe `case _ => log.error`?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
dzajkowski: yeah, explicit cases are the way to go but I think we have to have the catch all in the end too

dzajkowski (approved):

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -24,7 +24,6 @@ import monix.execution.Scheduler
 
 import scala.concurrent.duration._
 
-// scalastyle:off cyclomatic.complexity
```
leo-bogastry, line 27: It's great that we don't need this anymore 🥳

PR #1021, opened 2021-06-21
[ETCM-927] enhance BlockchainTests/ValidBlocks/bcMultiChainTest/Chain…

strauss-m:
# Description

Enhance the test ChainAtoChainB_difficultyB.
Currently, the test can't completely pass due to a missing feature: transactions are not executed on blocks that are not part of the canonical blockchain.

# Proposed Solution

- a block insertion on an alternate chain is now valid and queryable by hash
- fields s, r and v in EthTransactionResponse are now pruned from leading 0.

# Testing

The test `BlockchainTests/ValidBlocks/bcMultiChainTest/ChainAtoChainB_difficultyB` fails later, on the storage check, instead at the first insertion of a block on the non-canonical (best) chain.

```
Request: {"jsonrpc":"2.0","method":"debug_storageRangeAt","params":["3", 2, "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b", "0x0000000000000000000000000000000000000000000000000000000000000000", 20],"id":26}
Reply: {"jsonrpc":"2.0","result":{"complete":true,"storage":{}},"id":26}
Error: Compare States: Missing expected address: '0x095e7baea6a6c7c4c2dfeb977efac326af552d87' (bcMultiChainTest/ChainAtoChainB_difficultyB_Istanbul, fork: Istanbul, block: 5)
Error: CompareStates failed with errors: MissingExpectedAccount (bcMultiChainTest/ChainAtoChainB_difficultyB_Istanbul, fork: Istanbul, block: 5)
```

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/EthBlocksService.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/EthBlocksService.scala
@@ -5,6 +5,7 @@ import io.iohk.ethereum.domain.{Blockchain, BlockchainReader}
 import monix.eval.Task
 import org.bouncycastle.util.encoders.Hex
 import io.iohk.ethereum.consensus.Consensus
+import io.iohk.ethereum.ledger.BlockQueue
 
 object EthBlocksService {
   case class BestBlockNumberRequest()
@@ -38,7 +39,8 @@ object EthBlocksService {
 class EthBlocksService(
     val blockchain: Blockchain,
     val blockchainReader: BlockchainReader,
-    val consensus: Consensus
+    val consensus: Consensus,
+    val blockQueue: BlockQueue
 ) extends ResolveBlock {
   import EthBlocksService._
 
@@ -71,8 +73,8 @@ class EthBlocksService(
     */
   def getByBlockHash(request: BlockByBlockHashRequest): ServiceResponse[BlockByBlockHashResponse] = Task {
     val BlockByBlockHashRequest(blockHash, fullTxs) = request
-    val blockOpt = blockchainReader.getBlockByHash(blockHash)
-    val weight = blockchain.getChainWeightByHash(blockHash)
+    val blockOpt = blockchainReader.getBlockByHash(blockHash) orElse blockQueue.getBlockByHash(blockHash)
+    val weight = blockchain.getChainWeightByHash(blockHash) orElse blockQueue.getChainWeightByHash(blockHash)
 
     val blockResponseOpt = blockOpt.map(block => BlockResponse(block, weight, fullTxs = fullTxs))
     Right(BlockByBlockHashResponse(blockResponseOpt))
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
@@ -162,6 +162,9 @@ class TestService(
 
     // remove current genesis (Try because it may not exist)
     Try(blockchain.removeBlock(blockchain.genesisHeader.hash, withState = false))
+    // TODO clear the storage ? When relaunching some tests on the same running test mantis client,
+    // we end up with duplicate blocks because they are still present in the storage layer
+    // for example: bcMultiChainTest/ChainAtoChainB_BlockHash_Istanbul
 
     // load the new genesis
     val genesisDataLoader = new GenesisDataLoader(blockchain, blockchainReader, stateStorage, currentConfig)
@@ -291,7 +294,7 @@ class TestService(
       case BlockImportedToTop(blockImportData) =>
         val blockHash = s"0x${ByteStringUtils.hash2string(blockImportData.head.block.header.hash)}"
         ImportRawBlockResponse(blockHash).rightNow
-      case BlockEnqueued =>
+      case BlockEnqueued | ChainReorganised(_, _, _) =>
         val blockHash = s"0x${ByteStringUtils.hash2string(block.hash)}"
         ImportRawBlockResponse(blockHash).rightNow
       case e =>
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
@@ -136,6 +136,10 @@ class TestService(
   def setChainParams(request: SetChainParamsRequest): ServiceResponse[SetChainParamsResponse] = {
     currentConfig = buildNewConfig(request.chainParams.blockchainParams)
 
+    // clear ledger's cache on test start
+    // setChainParams is expected to be the first remote call for each test
+    testModeComponentsProvider.clearState()
+
     val genesisData = GenesisData(
       nonce = request.chainParams.genesis.nonce,
       mixHash = Some(request.chainParams.genesis.mixHash),
@@ -276,14 +280,19 @@ class TestService(
         testModeComponentsProvider
           .blockImport(currentConfig, preimageCache, sealEngine)
           .importBlock(value)
-          .flatMap(handleResult)
+          .flatMap(handleResult(value))
     }
   }
 
-  private def handleResult(blockImportResult: BlockImportResult): ServiceResponse[ImportRawBlockResponse] = {
+  private def handleResult(
+      block: Block
+  )(blockImportResult: BlockImportResult): ServiceResponse[ImportRawBlockResponse] = {
     blockImportResult match {
       case BlockImportedToTop(blockImportData) =>
         val blockHash = s"0x${ByteStringUtils.hash2string(blockImportData.head.block.header.hash)}"
+        ImportRawBlockResponse(blockHash).rightNow
+      case BlockEnqueued =>
+        val blockHash = s"0x${ByteStringUtils.hash2string(block.hash)}"
         ImportRawBlockResponse(blockHash).rightNow
       case e =>
         log.warn("Block import failed with {}", e)
--- a/src/main/scala/io/iohk/ethereum/ledger/BlockQueue.scala
+++ b/src/main/scala/io/iohk/ethereum/ledger/BlockQueue.scala
@@ -83,6 +83,14 @@ class BlockQueue(blockchain: Blockchain, val maxQueuedBlockNumberAhead: Int, val
     blocks.contains(hash)
 
   /**
+    * Returns the weight of the block corresponding to the hash, or None if not found
+    * @param hash the block's hash to get the weight from
+    * @return the weight of the block corresponding to the hash, or None if not found
+    */
+  def getChainWeightByHash(hash: ByteString): Option[ChainWeight] =
+    blocks.get(hash).flatMap(_.weight)
+
+  /**
     * Takes a branch going from descendant block upwards to the oldest ancestor
     * @param descendant the youngest block to be removed
     * @param dequeue should the branch be removed from the queue. Shared part of branch won't be removed
@@ -123,6 +131,14 @@ class BlockQueue(blockchain: Blockchain, val maxQueuedBlockNumberAhead: Int, val
       blocks -= block.header.hash
       parentToChildren -= block.header.hash
     }
+
+  /**
+    * Clear the BlockQueue
+    */
+  def clear(): Unit = {
+    blocks.clear()
+    parentToChildren.clear()
+  }
 
   /**
     * Removes stale blocks - too old or too young in relation the current best block number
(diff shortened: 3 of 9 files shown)


src/main/scala/io/iohk/ethereum/ledger/BlockQueue.scala
```diff
@@ -76,12 +76,28 @@ class BlockQueue(blockchain: Blockchain, val maxQueuedBlockNumberAhead: Int, val
     }
   }
 
+  /**
+    * Remove all blocks currently queued
+    */
+  def clear(): Unit = {
+    blocks.clear()
+    parentToChildren.clear()
+  }
```
AurelienRichez: This seems to be unused as we rebuild a new BlockQueue when needed.

src/main/scala/io/iohk/ethereum/testmode/TestModeServiceBuilder.scala
```diff
@@ -42,9 +42,18 @@ trait TestModeServiceBuilder extends LedgerBuilder {
...
+      * Returns the chain weight if the block represented by its hash is either stored in the blockchain or enqueued
+      *
+      * @param hash the hash of the block to retrieve the chain weight
+      * @return
+      */
+    override def getChainWeightByHash(hash: ByteString): Option[ChainWeight] = testLedger.getChainWeightByHash(hash)
   }
 
   override lazy val ledger: Ledger = new TestLedgerProxy
+//  override lazy val ledger: Ledger = testModeComponentsProvider.ledger(blockchainConfig, SealEngineType.NoReward)
```
AurelienRichez: we can remove this :-)

src/main/scala/io/iohk/ethereum/testmode/TestModeServiceBuilder.scala
```diff
@@ -42,9 +42,18 @@ trait TestModeServiceBuilder extends LedgerBuilder {
     ): Task[BlockImportResult] = testLedger.importBlock(block)
     override def resolveBranch(headers: NonEmptyList[BlockHeader]): BranchResolutionResult =
       testLedger.resolveBranch(headers)
+
+    /**
+      * Returns the chain weight if the block represented by its hash is either stored in the blockchain or enqueued
+      *
+      * @param hash the hash of the block to retrieve the chain weight
+      * @return
+      */
```
AurelienRichez: no need to repeat the doc I think

src/main/scala/io/iohk/ethereum/testmode/TestModeComponentsProvider.scala
```diff
@@ -19,14 +21,42 @@ class TestModeComponentsProvider(
...
 
-  def ledger(blockchainConfig: BlockchainConfig, sealEngine: SealEngineType): Ledger =
-    new LedgerImpl(
-      blockchain,
-      blockchainConfig,
-      syncConfig,
-      consensus(blockchainConfig, sealEngine),
-      validationExecutionContext
+  private var cache = HashMap.empty[(BlockchainConfig, SealEngineType), Ledger]
+  private var blockQueue = BlockQueue(blockchain, syncConfig)
```
AurelienRichez: Why do we need a cache ? We would like to minimize state if possible. I think there is always only one ledger at a given time anyway during tests right ?

I guess we want to keep the same BlockQueue between the calls. Initially I tried to put all the state in the same place in `TestService`. It's not really a perfect place but it avoids scattering the state everywhere and be able to simplify it when possible. To be consistent we could inject `blockQueue` in `def ledger`, or put the current `Ledger`  in `TestService` ?

AurelienRichez (commented):
A few remarks. I don't know how it will play out with @jvdp effort to remove Ledger however.

Anyway we have more ETS tests passing so 👍 .

src/main/scala/io/iohk/ethereum/testmode/TestModeComponentsProvider.scala
strauss-m: In my test case, I got two different ledger instances due to different configs.
But maybe I should investigate deeper into this and make sure that for a given test, only one blockchain config is used, do you think it's a worthwhile effort ?
If there is only one config, one can drop the cache.

src/main/scala/io/iohk/ethereum/ledger/BlockQueue.scala
strauss-m: true.
It was belonging to a fix change that has been removed. I'll clean up.

src/main/scala/io/iohk/ethereum/testmode/TestModeServiceBuilder.scala
strauss-m: removed

src/main/scala/io/iohk/ethereum/testmode/TestModeServiceBuilder.scala
strauss-m: removed

jvdp:
> I don't know how it will play out with @jvdp effort to remove Ledger however.

It'll be interesting but I'm not afraid!

src/main/scala/io/iohk/ethereum/testmode/TestEthBlockServiceWrapper.scala
```diff
@@ -30,9 +31,18 @@ class TestEthBlockServiceWrapper(blockchain: Blockchain, ledger: Ledger, consens
     .getByBlockHash(request)
     .map(
       _.map(blockByBlockResponse => {
-        val fullBlock = blockchain.getBlockByNumber(blockByBlockResponse.blockResponse.get.number).get
-        BlockByBlockHashResponse(blockByBlockResponse.blockResponse.map(response => toEthResponse(fullBlock, response)))
-      })
+        blockByBlockResponse.blockResponse
+          .toRight("missing block response")
```
leo-bogastry: Can you add some more specific data to error?

strauss-m: I've added a little bit more context with regard to the error messages

src/main/scala/io/iohk/ethereum/testmode/TestEthBlockServiceWrapper.scala
```diff
@@ -32,9 +33,19 @@ class TestEthBlockServiceWrapper(
...
+          .flatMap(baseBlockResponse => baseBlockResponse.hash.toRight(s"missing hash for block $baseBlockResponse"))
+          .flatMap(hash => ledger.getBlockByHash(hash).toRight(s"Ledger: unable to find block for hash=${hash.toHex}"))
+          .map(fullBlock =>
+            BlockByBlockHashResponse(
+              blockByBlockResponse.blockResponse.map(response => toEthResponse(fullBlock, response))
+            )
+          )
+          .left
+          .map(errorMessage => JsonRpcError.LogicError(errorMessage))
+      }).flatten
```
leo-bogastry: You can make a few simplifications here:
- move the import to the top of the file, please
- remove the final `.flatten` and use `flatMap` instead
- you can remove `{` after `=>`, by replacing the `(` before `blockByBlockResponse` by `{`
In summary from:
`_.map(blockByBlockResponse => {`  to `_.flatMap { blockByBlockResponse =>`

src/main/scala/io/iohk/ethereum/testmode/TestModeComponentsProvider.scala
leo-bogastry: Were you able to investigate this?

src/main/scala/io/iohk/ethereum/jsonrpc/EthBlocksService.scala
```diff
@@ -74,8 +74,8 @@ class EthBlocksService(
     */
   def getByBlockHash(request: BlockByBlockHashRequest): ServiceResponse[BlockByBlockHashResponse] = Task {
     val BlockByBlockHashRequest(blockHash, fullTxs) = request
-    val blockOpt = blockchainReader.getBlockByHash(blockHash)
-    val weight = blockchain.getChainWeightByHash(blockHash)
```
jvdp, line 75: Are these two changes required by ETS? I wonder what would be a convenient way to refactor this with ledger gone...

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
```diff
@@ -157,6 +161,9 @@ class TestService(
 
     // remove current genesis (Try because it may not exist)
     Try(blockchain.removeBlock(blockchain.genesisHeader.hash, withState = false))
+    // TODO clear the storage ? When relaunching some tests on the same running test mantis client,
+    // we end up with duplicate blocks because they are still present in the storage layer
+    // for example: bcMultiChainTest/ChainAtoChainB_BlockHash_Istanbul
```
jvdp, line 167: How about when `setChainParams` is called?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
```diff
@@ -29,6 +29,13 @@ trait Ledger {
     */
   def getBlockByHash(hash: ByteString): Option[Block]
 
+  /**
+    * Returns the chain weight if the block represented by its hash is either stored in the blockchain or enqueued
+    * @param hash the hash of the block to retrieve the chain weight
+    * @return the chain weight if the block represented by its hash is either stored in the blockchain or enqueued
+    */
+  def getChainWeightByHash(hash: ByteString): Option[ChainWeight]
```
jvdp: Yeah please don't add more things to ledger 😅

src/main/scala/io/iohk/ethereum/testmode/TestEthBlockServiceWrapper.scala
```diff
@@ -31,10 +33,19 @@ class TestEthBlockServiceWrapper(
...
+          .flatMap(baseBlockResponse => baseBlockResponse.hash.toRight(s"missing hash for block $baseBlockResponse"))
+          .flatMap(hash => ledger.getBlockByHash(hash).toRight(s"Ledger: unable to find block for hash=${hash.toHex}"))
+          .map(fullBlock =>
+            BlockByBlockHashResponse(
+              blockByBlockResponse.blockResponse.map(response => toEthResponse(fullBlock, response))
+            )
+          )
+          .left
+          .map(errorMessage => JsonRpcError.LogicError(errorMessage))
+      }
```
jvdp, line 60: This is a bit hard to follow, might be nicer as `for`-comprehension?

src/main/scala/io/iohk/ethereum/testmode/TestModeComponentsProvider.scala
strauss-m: No, I didn't push further the idea. I don't know if the suite is actually using different config depending on tests, which would prevent a unique one (initial setup + each test config) convergence, and require the current multi-config cache.

src/main/scala/io/iohk/ethereum/testmode/TestEthBlockServiceWrapper.scala
strauss-m: thanks, update done

leo-bogastry (approved):

src/main/scala/io/iohk/ethereum/jsonrpc/EthBlocksService.scala
strauss-m: The underlying need was to be able to get a block by hash, no matter if it's on the canonical chain or an alternate one. The solution was to delegate the work to the ledger, which implements the getBlockByHash exactly as it (first look in the canonical blockchain, then into the BlockQueue). The getChainWeightByHash was added to match the getBlockByHash behaviour.

The feature (being able to query a block by its hash, no matter its location) is required by the ETS. After a successful insert, retesteth queries for the insert block by its hash to do some consistency check.
It is not restricted to the ledger, so if any other component provides this after the refactor, let's use the new one.

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
strauss-m: do you mean put the cleaning into the setChainParams call, or into the setChainParams caller ?

src/main/scala/io/iohk/ethereum/ledger/Ledger.scala
strauss-m: It is linked to the Ledger.getBlockByHash call.
According to their usage in EthBlocksService, getBlockByHash() and getChainWeightByHash() should both be available on whatever component provides them.
The refactor should be straightforward, wherever the getBlockByHash() method ends up, you can put the getChainWeightByHash() one too. The underlying behaviour is also similar to getBlockByHash: first query the (canonical) BlockChain, then the BlockQueue if not found.

AurelienRichez (approved):

PR #1022, opened 2021-06-22
[ETCM-943] partially split reading part of Blockchain

AurelienRichez:
# Description

The goal of this PR is to put the read only part of `Blockchain` in another class in order to reduce the dependency on this class. As a target I choose to make `BlockchainHostActor` independent of `Blockchain`. 

I moved a few function inside a new class named `BlockchainReader` for consistency, but we could come up with another name. 

It allowed to remove `Blockchain` at some other places but there are still read only methods that can be moved to go further. But as it is already a big change we should at least validate this before continuing.

# review

This is a big change, but mostly moving things around. Each commit can be viewed one by one and is mergeable, so we can split this PR if needed

One important thing to think about is : does the change help us for a future refactoring ? I don't necessarily see `BlockchainReader` as a final target for the structure of the code, but more as step to help us refactor step by step.

### DIFF
--- a/src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
+++ b/src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
@@ -8,14 +8,13 @@ import com.typesafe.config.ConfigFactory
 import io.iohk.ethereum.blockchain.sync.CacheBasedBlacklist
 import io.iohk.ethereum.db.components.Storages.PruningModeComponent
 import io.iohk.ethereum.db.components.{RocksDbDataSourceComponent, Storages}
-import io.iohk.ethereum.db.dataSource.{DataSourceBatchUpdate, RocksDbDataSource}
+import io.iohk.ethereum.db.dataSource.{DataSourceBatchUpdate}
 import io.iohk.ethereum.db.storage.NodeStorage.{NodeEncoded, NodeHash}
-import io.iohk.ethereum.db.storage.TransactionMappingStorage.TransactionLocation
 import io.iohk.ethereum.db.storage.pruning.{ArchivePruning, PruningMode}
-import io.iohk.ethereum.db.storage.{AppStateStorage, StateStorage}
+import io.iohk.ethereum.db.storage.AppStateStorage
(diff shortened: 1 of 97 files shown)


src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
```diff
@@ -32,10 +32,17 @@ import io.iohk.ethereum.security.SecureRandomBuilder
 import io.iohk.ethereum.utils.{Config, NodeStatus, ServerStatus}
 import monix.reactive.Observable
 import org.bouncycastle.util.encoders.Hex
+import org.scalamock.clazz.Mock
```
leo-bogastry: With `MockFactory` you shouldn't need this

src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
```diff
@@ -0,0 +1,105 @@
...
+import io.iohk.ethereum.db.storage.{
+  BlockBodiesStorage,
+  BlockHeadersStorage,
+  BlockNumberMappingStorage,
+  ReceiptStorage,
+  StateStorage
+}
+import io.iohk.ethereum.mpt.MptNode
+
+class BlockchainReader(
```
dzajkowski, line 13: would the next step be to create a trait `BlockchainReader` to allow for easier stubbing?

dzajkowski (approved):

src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
AurelienRichez: Yes I agree. In the end we might split it further and have several reader traits.

src/it/scala/io/iohk/ethereum/txExecTest/util/DumpChainApp.scala
AurelienRichez: indeed

leo-bogastry (approved):

AnastasiiaL (approved):

PR #1023, opened 2021-06-22
[ETCM-354] Bump cats version

lukasz-golebiewski:
# Description

Let's use newer versions of the cats libraries


dzajkowski (approved):
looks good. please make sure that all tests behave as expected (we the flakiness factor stays similar)

AnastasiiaL (approved):

PR #1024, opened 2021-06-23
[ETCM-354] Peer fork ID validation

lukasz-golebiewski:
# Description

Adds peer fork id validation as described in the section "validation rules" of https://eips.ethereum.org/EIPS/eip-2124

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -94,7 +94,9 @@ object Dependencies {
     "ch.qos.logback" % "logback-classic" % "1.2.3",
     "com.typesafe.scala-logging" %% "scala-logging" % "3.9.2",
     "net.logstash.logback" % "logstash-logback-encoder" % "6.4",
-    "org.codehaus.janino" % "janino" % "3.1.2"
+    "org.codehaus.janino" % "janino" % "3.1.2",
+    "org.typelevel" %% "log4cats-core" % "2.1.1",
+    "org.typelevel" %% "log4cats-slf4j" % "1.3.1"
   )
 
   val crypto = Seq("org.bouncycastle" % "bcprov-jdk15on" % "1.66")
--- a/src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
+++ b/src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
@@ -50,21 +50,21 @@ object ForkIdValidator {
     val checksums: Vector[BigInt] = calculateChecksums(genesisHash, forks)
 
     // find the first unpassed fork and it's index
-    val (unpassedFork, i) =
+    val (unpassedFork, unpassedForkIndex) =
       forks.zipWithIndex.find { case (fork, _) => currentHeight < fork }.getOrElse((maxUInt64, forks.length))
 
     // The checks are left biased -> whenever a result is found we need to short circuit
     val validate = (for {
       _ <- liftF(Logger[F].trace(s"Before checkMatchingHashes"))
       matching <- fromEither[F](
-        checkMatchingHashes(checksums, remoteId, currentHeight, i).toLeft("hashes didn't match")
+        checkMatchingHashes(checksums(unpassedForkIndex), remoteId, currentHeight).toLeft("hashes didn't match")
       )
       _ <- liftF(Logger[F].trace(s"checkMatchingHashes result: $matching"))
       _ <- liftF(Logger[F].trace(s"Before checkSubset"))
-      sub <- fromEither[F](checkSubset(checksums, forks, remoteId, i).toLeft("not in subset"))
+      sub <- fromEither[F](checkSubset(checksums, forks, remoteId, unpassedForkIndex).toLeft("not in subset"))
       _ <- liftF(Logger[F].trace(s"checkSubset result: $sub"))
       _ <- liftF(Logger[F].trace(s"Before checkSuperset"))
-      sup <- fromEither[F](checkSuperset(checksums, remoteId, i).toLeft("not in superset"))
+      sup <- fromEither[F](checkSuperset(checksums, remoteId, unpassedForkIndex).toLeft("not in superset"))
       _ <- liftF(Logger[F].trace(s"checkSuperset result: $sup"))
       _ <- liftF(Logger[F].trace(s"No check succeeded"))
       _ <- fromEither[F](Either.left[ForkIdValidationResult, Unit](ErrLocalIncompatibleOrStale))
@@ -73,7 +73,7 @@ object ForkIdValidator {
     for {
       _ <- Logger[F].debug(s"Validating $remoteId")
       _ <- Logger[F].trace(s" list: $forks")
-      _ <- Logger[F].trace(s"Unpassed fork $unpassedFork was found at index $i")
+      _ <- Logger[F].trace(s"Unpassed fork $unpassedFork was found at index $unpassedForkIndex")
       res <- validate.map(_.swap)
       _ <- Logger[F].debug(s"Validation result is: $res")
     } yield (res.getOrElse(Connect))
@@ -101,13 +101,12 @@ object ForkIdValidator {
     * 1b) No remotely announced fork; or not yet passed locally, connect.
     */
   private def checkMatchingHashes(
-      checksums: Vector[BigInt],
+      checksum: BigInt,
       remoteId: ForkId,
-      currentHeight: BigInt,
-      i: Int
+      currentHeight: BigInt
   ): Option[ForkIdValidationResult] =
     remoteId match {
-      case ForkId(hash, _) if checksums(i) != hash => None
+      case ForkId(hash, _) if checksum != hash => None
       case ForkId(_, Some(next)) if currentHeight >= next => Some(ErrLocalIncompatibleOrStale)
       case _ => Some(Connect)
     }
--- a/src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
+++ b/src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
@@ -1,0 +1,129 @@
+package io.iohk.ethereum.forkid
+
+import akka.util.ByteString
+import cats.Monad
+import cats.data.EitherT
+import cats.implicits._
+import io.iohk.ethereum.utils.BigIntExtensionMethods._
+import io.iohk.ethereum.utils.BlockchainConfig
+import io.iohk.ethereum.utils.ByteUtils._
+import monix.eval.Task
+import org.typelevel.log4cats.Logger
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+import java.util.zip.CRC32
+
+sealed trait ForkIdValidationResult
+case object Connect extends ForkIdValidationResult
+case object ErrRemoteStale extends ForkIdValidationResult
+case object ErrLocalIncompatibleOrStale extends ForkIdValidationResult
+
+object ForkIdValidator {
+
+  implicit val unsafeLogger = Slf4jLogger.getLogger[Task]
+
+  val maxUInt64 = (BigInt(0x7fffffffffffffffL) << 1) + 1 // scalastyle:ignore magic.number
+
+  def validatePeer[F[_]: Monad: Logger](
+      genesisHash: ByteString,
+      config: BlockchainConfig
+  )(currentHeight: BigInt, remoteForkId: ForkId): F[ForkIdValidationResult] = {
+    val forks = ForkId.gatherForks(config)
+    validatePeer[F](genesisHash, forks)(currentHeight, remoteForkId)
+  }
+
+  def validatePeer[F[_]: Monad: Logger](
+      genesisHash: ByteString,
+      forks: List[BigInt]
+  )(currentHeight: BigInt, remoteId: ForkId): F[ForkIdValidationResult] = {
+    val checksums: Vector[BigInt] = calculateCheckchecksums(genesisHash, forks)
+
+    // find the first unpassed fork and it's index
+    val (unpassedFork, i) = (forks :+ maxUInt64).zipWithIndex.find { case (fork, _) => currentHeight < fork }.get
+
+    // The checks are left biased -> whenever a result is found we need to short circuit
+    val validate: F[Option[ForkIdValidationResult]] = (for {
+      _ <- EitherT.liftF(Logger[F].debug(s"Before checkMatchingHashes"))
+      matching <- EitherT.fromEither[F](
+        checkMatchingHashes(checksums, remoteId, currentHeight, i).toLeft("hashes didn't match")
+      )
+      _ <- EitherT.liftF(Logger[F].debug(s"checkMatchingHashes result: $matching"))
+      _ <- EitherT.liftF(Logger[F].debug(s"Before checkSubset"))
+      sub <- EitherT.fromEither[F](checkSubset(checksums, forks, remoteId, i).toLeft("not in subset"))
+      _ <- EitherT.liftF(Logger[F].debug(s"checkSubset result: $sub"))
+      _ <- EitherT.liftF(Logger[F].debug(s"Before checkSuperset"))
+      sup <- EitherT.fromEither[F](checkSuperset(checksums, remoteId, i).toLeft("not in superset"))
+      _ <- EitherT.liftF(Logger[F].debug(s"checkSuperset result: $sup"))
+      _ <- EitherT.liftF(Logger[F].debug(s"No check succeeded"))
+      res <- EitherT.fromEither[F](Either.left[ForkIdValidationResult, Unit](ErrLocalIncompatibleOrStale))
+    } yield (res)).value
+      .map(_.swap)
+      .flatMap(res => Logger[F].debug(s"Validation result is: $res") >> Monad[F].pure(res.toOption))
+
+    Logger[F].debug(s"Validating $remoteId") >>
+      Logger[F].debug(s"Forks list: $forks") >>
+      Logger[F].debug(s"Unpassed fork $unpassedFork was found at index $i") >>
+      validate.map(_.getOrElse(Connect)) // Impossible to say if nodes are compatible, so we need to allow to connect
+  }
+
+  private def calculateCheckchecksums(
+      genesisHash: ByteString,
+      forks: List[BigInt]
+  ): Vector[BigInt] = {
+    val crc = new CRC32()
+    crc.update(genesisHash.asByteBuffer)
+    val genesisChecksum = BigInt(crc.getValue())
+
+    genesisChecksum +: (forks.map { fork =>
+      crc.update(bigIntToBytes(fork, 8))
+      BigInt(crc.getValue())
+    }).toVector
+  }
+
+  /**
+    * 1) If local and remote FORK_HASH matches, compare local head to FORK_NEXT.
+    * The two nodes are in the same fork state currently.
+    * They might know of differing future forks, but that’s not relevant until the fork triggers (might be postponed, nodes might be updated to match).
+    * 1a) A remotely announced but remotely not passed block is already passed locally, disconnect, since the chains are incompatible.
+    * 1b) No remotely announced fork; or not yet passed locally, connect.
+    */
+  private def checkMatchingHashes(
+      checksums: Vector[BigInt],
+      remoteId: ForkId,
+      currentHeight: BigInt,
+      i: Int
+  ): Option[ForkIdValidationResult] = {
+    if (checksums(i) == remoteId.hash) {
+      if (remoteId.next.isDefined && currentHeight >= remoteId.next.get) {
+        Some(ErrLocalIncompatibleOrStale)
+      } else {
+        Some(Connect)
+      }
+    } else { None }
+  }
+
+  /**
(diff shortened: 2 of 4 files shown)


src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,129 @@
...
+      )
+      _ <- EitherT.liftF(Logger[F].debug(s"checkMatchingHashes result: $matching"))
+      _ <- EitherT.liftF(Logger[F].debug(s"Before checkSubset"))
+      sub <- EitherT.fromEither[F](checkSubset(checksums, forks, remoteId, i).toLeft("not in subset"))
+      _ <- EitherT.liftF(Logger[F].debug(s"checkSubset result: $sub"))
+      _ <- EitherT.liftF(Logger[F].debug(s"Before checkSuperset"))
+      sup <- EitherT.fromEither[F](checkSuperset(checksums, remoteId, i).toLeft("not in superset"))
+      _ <- EitherT.liftF(Logger[F].debug(s"checkSuperset result: $sup"))
+      _ <- EitherT.liftF(Logger[F].debug(s"No check succeeded"))
+      res <- EitherT.fromEither[F](Either.left[ForkIdValidationResult, Unit](ErrLocalIncompatibleOrStale))
```
dzajkowski: there's a lot of boilerplate simply to get a lifted debug. perhaps a helper?
`private def liftedDebug[F[_]: Monad: Logger](s: => String) = EitherT.liftF(Logger[F].debug(s))`

```
    val validate: F[Option[ForkIdValidationResult]] = (for {
      _ <- liftedDebug[F](s"Before checkMatchingHashes")
      matching <- EitherT.fromEither[F](
        checkMatchingHashes(checksums, remoteId, currentHeight, i).toLeft("hashes didn't match")
      )
      _ <- liftedDebug[F](s"checkMatchingHashes result: $matching")
      _ <- liftedDebug[F](s"Before checkSubset")
      sub <- EitherT.fromEither[F](checkSubset(checksums, forks, remoteId, i).toLeft("not in subset"))
      _ <- liftedDebug[F](s"checkSubset result: $sub")
      _ <- liftedDebug[F](s"Before checkSuperset")
      sup <- EitherT.fromEither[F](checkSuperset(checksums, remoteId, i).toLeft("not in superset"))
      _ <- liftedDebug[F](s"checkSuperset result: $sup")
      _ <- liftedDebug[F](s"No check succeeded")
```

lukasz-golebiewski: I imported `EitherT._` and use the lifting functions without the explicit companion object

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,129 @@
...
+      genesisHash: ByteString,
+      forks: List[BigInt]
+  )(currentHeight: BigInt, remoteId: ForkId): F[ForkIdValidationResult] = {
+    val checksums: Vector[BigInt] = calculateCheckchecksums(genesisHash, forks)
+
+    // find the first unpassed fork and it's index
+    val (unpassedFork, i) = (forks :+ maxUInt64).zipWithIndex.find { case (fork, _) => currentHeight < fork }.get
+
+    // The checks are left biased -> whenever a result is found we need to short circuit
+    val validate: F[Option[ForkIdValidationResult]] = (for {
```
jvdp: This method looks a lot more complicated than it is. How about moving the tracing into the `check` methods, for a start? (Do we still need the tracing btw?)

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,129 @@
...
+      sub <- fromEither[F](checkSubset(checksums, forks, remoteId, i).toLeft("not in subset"))
+      _ <- liftF(Logger[F].trace(s"checkSubset result: $sub"))
+      _ <- liftF(Logger[F].trace(s"Before checkSuperset"))
+      sup <- fromEither[F](checkSuperset(checksums, remoteId, i).toLeft("not in superset"))
+      _ <- liftF(Logger[F].trace(s"checkSuperset result: $sup"))
+      _ <- liftF(Logger[F].trace(s"No check succeeded"))
+      res <- fromEither[F](Either.left[ForkIdValidationResult, Unit](ErrLocalIncompatibleOrStale))
+    } yield (res)).value
+      .map(_.swap)
+      .flatMap(res => Logger[F].debug(s"Validation result is: $res") >> Monad[F].pure(res.toOption))
```
jvdp: Could also use a for-comprehension for this part? (And bind the existing one to a name.)

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,129 @@
...
+case object ErrRemoteStale extends ForkIdValidationResult
+case object ErrLocalIncompatibleOrStale extends ForkIdValidationResult
+
+object ForkIdValidator {
+
+  implicit val unsafeLogger = Slf4jLogger.getLogger[Task]
+
+  val maxUInt64 = (BigInt(0x7fffffffffffffffL) << 1) + 1 // scalastyle:ignore magic.number
+
+  def validatePeer[F[_]: Monad: Logger](
```
jvdp, line 38: Could do with a scaladoc since it's 'external' API, right?

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,129 @@
...
+      forks: List[BigInt]
+  ): Vector[BigInt] = {
+    val crc = new CRC32()
+    crc.update(genesisHash.asByteBuffer)
+    val genesisChecksum = BigInt(crc.getValue())
+
+    genesisChecksum +: (forks.map { fork =>
+      crc.update(bigIntToBytes(fork, 8))
+      BigInt(crc.getValue())
+    }).toVector
```
jvdp, line 93: This code perhaps _looks_ functional but it's really relying on the mutable CRC instance. It's probably more performant this way, but it could be more descriptive using `List.unfold` or similar. (Eg. first produce a collection of collections to produce the CRC with, then flatten it.)

(But this is just a suggestion ofc.)

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,129 @@
...
+      i: Int
+  ): Option[ForkIdValidationResult] = {
+    if (checksums(i) == remoteId.hash) {
+      if (remoteId.next.isDefined && currentHeight >= remoteId.next.get) {
+        Some(ErrLocalIncompatibleOrStale)
+      } else {
+        Some(Connect)
+      }
+    } else { None }
+  }
```
jvdp: Could be nicer as a match on `remoteId`:

```scala
remoteId match {
  case ForkId(hash, _) if checksums(i) != hash => None
  case ForkId(_, Some(next)) if currentHeight >= next => Some(ErrLocalIncompatibleOrStale)
  case _ => Some(Connect)
}
```

(Gets rid of the nested `if` and the `Option.get`.)

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,129 @@
...
+      .take(i)
+      .find { case (sum, _) => sum == remoteId.hash }
+      .map { case (_, fork) => if (fork == remoteId.next.getOrElse(0)) Connect else ErrRemoteStale }
+
+  /**
+    * 3) If the remote FORK_HASH is a superset of the local past forks and can be completed with locally known future forks, connect.
+    * Local node is currently syncing. It might eventually diverge from the remote, but at this current point in time we don’t have enough information.
+    */
+  def checkSuperset(checksums: Vector[BigInt], remoteId: ForkId, i: Int): Option[ForkIdValidationResult] = {
+    checksums.drop(i).find(_ == remoteId.hash).map(_ => Connect)
```
jvdp: `collectFirst` could replace the `find` + `map`.

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,129 @@
...
+      checksums: Vector[BigInt],
+      forks: List[BigInt],
+      remoteId: ForkId,
+      i: Int
+  ): Option[ForkIdValidationResult] =
+    checksums
+      .zip(forks)
+      .take(i)
+      .find { case (sum, _) => sum == remoteId.hash }
+      .map { case (_, fork) => if (fork == remoteId.next.getOrElse(0)) Connect else ErrRemoteStale }
```
jvdp: `collectFirst` could be nice here.

src/test/scala/io/iohk/ethereum/forkid/ForkIdValidatorSpec.scala
```diff
@@ -0,0 +1,102 @@
...
+class ForkIdValidatorSpec extends AnyWordSpec with Matchers {
+
+  val config = blockchains
+
+  val ethGenesisHash = ByteString(Hex.decode("d4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3"))
+
+  "ForkIdValidator" must {
+    "correctly validate ETH peers" in {
+      // latest fork at the time of writing those assertions (in the spec) was Petersburg
+      val ethForksList: List[BigInt] = List(1150000, 1920000, 2463000, 2675000, 4370000, 7280000)
```
jvdp, line 26: Are all these different forks interesting for the purposes of testing? I mean, you are testing it with 6, would it make a material difference to the test quality if it was 5 or 7 or 9000?

Also, would this be a good opportunity for property based testing perhaps?

jvdp (approved):
Looks OK to my eyeballs but the code and tests could be simpler. And it will be more interesting once this is hooked into handshaking and running on a test net.

src/test/scala/io/iohk/ethereum/forkid/ForkIdValidatorSpec.scala
dzajkowski: now I'm curious - how does one property test an equation? by writing the same equation in a different way?

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,129 @@
...
+      genesisHash: ByteString,
+      config: BlockchainConfig
+  )(currentHeight: BigInt, remoteForkId: ForkId): F[ForkIdValidationResult] = {
+    val forks = ForkId.gatherForks(config)
+    validatePeer[F](genesisHash, forks)(currentHeight, remoteForkId)
+  }
+
+  def validatePeer[F[_]: Monad: Logger](
+      genesisHash: ByteString,
+      forks: List[BigInt]
```
leo-bogastry, line 48: Should `forks` be validated to not be empty?

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,129 @@
...
+  )(currentHeight: BigInt, remoteId: ForkId): F[ForkIdValidationResult] = {
+    val checksums: Vector[BigInt] = calculateCheckchecksums(genesisHash, forks)
+
+    // find the first unpassed fork and it's index
+    val (unpassedFork, i) = (forks :+ maxUInt64).zipWithIndex.find { case (fork, _) => currentHeight < fork }.get
```
leo-bogastry: Can we avoid the `.get`?

src/test/scala/io/iohk/ethereum/forkid/ForkIdValidatorSpec.scala
jvdp: Yes, but part of this 'equation' would be in how the arbitrary data is constructed. I agree that just re-implementing the same logic wouldn't make much sense.

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
lukasz-golebiewski: This code isn't functional because of a side effect in `map`. It is because of the nature of Java's `CRC` as you correctly noted. This however doesn't break the referential transparency of `calculateChecksum`, so the function itself remains `pure`.

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
lukasz-golebiewski: Good catch, added

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
lukasz-golebiewski: > This method looks a lot more complicated than it is.

This is vague, can you elaborate on that? 

>  How about moving the tracing into the check methods, for a start?

It could be done, but I can't see how moving logging instructions around is going to make the method less complicated.

> (Do we still need the tracing btw?)

Logging is part of our DoD and it might get useful when testing on testnet, so I wouldn't remove it.

src/test/scala/io/iohk/ethereum/forkid/ForkIdValidatorSpec.scala
dzajkowski: oh this isn't arbitrary data - it's hard fork heights.

lukasz-golebiewski: This test was provided as part of the specification in https://eips.ethereum.org/EIPS/eip-2124 so we would have to have a really good reason to modify or not include it here. 

Whether all the forks are interesting for the purposes of testing is probably a question to the original EIP's authors. 

@jvdp what are the properties you would like to test here?

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
lukasz-golebiewski: I don't think this is necessary

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
lukasz-golebiewski: Done

project/Dependencies.scala
```diff
@@ -94,7 +94,9 @@ object Dependencies {
     "ch.qos.logback" % "logback-classic" % "1.2.3",
     "com.typesafe.scala-logging" %% "scala-logging" % "3.9.2",
     "net.logstash.logback" % "logstash-logback-encoder" % "6.4",
-    "org.codehaus.janino" % "janino" % "3.1.2"
+    "org.codehaus.janino" % "janino" % "3.1.2",
+    "org.typelevel" %% "log4cats-core" % "2.1.1",
+    "org.typelevel" %% "log4cats-slf4j" % "1.3.1"
```
AurelienRichez, line 99: do we really need to add 3 deps for one class ?

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,140 @@
...
+        checkMatchingHashes(checksums, remoteId, currentHeight, i).toLeft("hashes didn't match")
+      )
+      _ <- liftF(Logger[F].trace(s"checkMatchingHashes result: $matching"))
+      _ <- liftF(Logger[F].trace(s"Before checkSubset"))
+      sub <- fromEither[F](checkSubset(checksums, forks, remoteId, i).toLeft("not in subset"))
+      _ <- liftF(Logger[F].trace(s"checkSubset result: $sub"))
+      _ <- liftF(Logger[F].trace(s"Before checkSuperset"))
+      sup <- fromEither[F](checkSuperset(checksums, remoteId, i).toLeft("not in superset"))
+      _ <- liftF(Logger[F].trace(s"checkSuperset result: $sup"))
+      _ <- liftF(Logger[F].trace(s"No check succeeded"))
```
AurelienRichez, line 69: I think we could do with just one log between each step right ?

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,140 @@
...
+    * The two nodes are in the same fork state currently.
+    * They might know of differing future forks, but that’s not relevant until the fork triggers (might be postponed, nodes might be updated to match).
+    * 1a) A remotely announced but remotely not passed block is already passed locally, disconnect, since the chains are incompatible.
+    * 1b) No remotely announced fork; or not yet passed locally, connect.
+    */
+  private def checkMatchingHashes(
+      checksums: Vector[BigInt],
+      remoteId: ForkId,
+      currentHeight: BigInt,
+      i: Int
```
AurelienRichez: we could directly pass the current fork block hash instead of `i` and a list. It's not really obvious that the index of the next fork is actually also the index of the current checksums.

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,140 @@
...
+      forks: List[BigInt]
+  )(currentHeight: BigInt, remoteId: ForkId): F[ForkIdValidationResult] = {
+    val checksums: Vector[BigInt] = calculateCheckchecksums(genesisHash, forks)
+
+    // find the first unpassed fork and it's index
+    val (unpassedFork, i) =
+      forks.zipWithIndex.find { case (fork, _) => currentHeight < fork }.getOrElse((maxUInt64, forks.length))
+
+    // The checks are left biased -> whenever a result is found we need to short circuit
+    val validate = (for {
```
AurelienRichez, line 57: can't we find the hash in checksums which match remote hash first ? So that we directly know if we should do the matching/subset/superset check directly ? I feel that it would simplify the function.

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  private[forkid] def validatePeer[F[_]: Monad: Logger](
+      genesisHash: ByteString,
+      forks: List[BigInt]
+  )(currentHeight: BigInt, remoteId: ForkId): F[ForkIdValidationResult] = {
+    val checksums: Vector[BigInt] = calculateCheckchecksums(genesisHash, forks)
+
+    // find the first unpassed fork and it's index
+    val (unpassedFork, i) =
```
AurelienRichez: `i` -> `nextForkIndex` or `unpassedForkIndex` ?

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
```diff
@@ -0,0 +1,140 @@
...
+    for {
+      _ <- Logger[F].debug(s"Validating $remoteId")
+      _ <- Logger[F].trace(s" list: $forks")
+      _ <- Logger[F].trace(s"Unpassed fork $unpassedFork was found at index $i")
+      res <- validate.map(_.swap)
+      _ <- Logger[F].debug(s"Validation result is: $res")
+    } yield (res.getOrElse(Connect))
+  }
+
+  private def calculateCheckchecksums(
```
AurelienRichez: typo `calculateChecksums`

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
AurelienRichez: Do we really need the monad ?
This is a synchronous function, so there is no need for an async abstraction. I get that logging is a kind of side effect, but from the caller point of view it does not really matter, and it adds some noise because everything needs to be inside a `liftF` now.

leo-bogastry (approved):

project/Dependencies.scala
AurelienRichez: sorry, it's only 2 my bad

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
lukasz-golebiewski: `Monad` has nothing to do with asynchronicity, it allows chaining effectful computations.
Logging is a side effect. Are you saying we should ignore that fact?

src/main/scala/io/iohk/ethereum/forkid/ForkIdValidator.scala
lukasz-golebiewski: I don't think I can see what you mean. Could you paste a code snippet with the suggested simplification?

project/Dependencies.scala
lukasz-golebiewski: Do we have a limit for that? :-)

dzajkowski (approved):
I created a set of topics in slack to continue the discussion but from my pov it LGTM

PR #1025, opened 2021-06-23
ETCM 944 Extract instantiation of InMemoryWorldStateProxy from Blockchain class

leo-bogastry:
# Description

On of the things Blockchain class is doing is providing an instantiation of InMemoryWorldStateProxy class. I think is because this class needs a few storages and they are available in Blockchain (because Blockchain is hoarding everything).

Two flavours of InMemoryWorldStateProxy were provided (a normal one and a readonly one). These were merged into one constructor added to InMemoryWorldStateProxy, and it's up to the class instantiating InMemoryWorldStateProxy to pass the normal or readonly `mptStorage`.

The biggest impact of this refactoring is in the tests.
Access to the RocksDB became a problem and was conflicting a lot.
I'm adding a few comments to the code as well


src/main/scala/io/iohk/ethereum/ledger/BlockExecution.scala
```diff
@@ -57,10 +59,19 @@ class BlockExecution(
   /** Executes a block (executes transactions and pays rewards) */
   private def executeBlock(block: Block): Either[BlockExecutionError, BlockResult] = {
     for {
-      parent <- blockchain
+      parentHeader <- blockchain
         .getBlockHeaderByHash(block.header.parentHash)
         .toRight(MissingParentError) // Should not never occur because validated earlier
-      execResult <- executeBlockTransactions(block, parent)
+      initialWorld = InMemoryWorldStateProxy(
```
leo-bogastry, line 65: I moved the creation of the `initialWorld` here because it made the method `executeBlockTransactions` more testable.

src/main/resources/conf/base.conf
```diff
@@ -386,7 +386,7 @@ mantis {
     nodes-per-request = 384
 
     # Minimum number of peers required to start fast-sync (by determining the pivot block)
-    min-peers-to-choose-pivot-block = 3
+    min-peers-to-choose-pivot-block = 1
```
AurelienRichez: Should we keep this change ? Or was it just for testing ?

AurelienRichez (approved):

src/main/resources/conf/base.conf
leo-bogastry: Ups... that was me testing. I will rollback

leo-bogastry: Fixed

src/it/scala/io/iohk/ethereum/txExecTest/ForksTest.scala
```diff
@@ -63,13 +57,17 @@ class ForksTest extends AnyFlatSpec with Matchers {
     val startBlock = 1
     val endBlock = 11
 
-    protected val testBlockchainStorages = FixtureProvider.prepareStorages(startBlock, fixtures)
+    protected val testBlockchainStorages = FixtureProvider.prepareStorages(endBlock, fixtures)
```
AnastasiiaL, line 60: this change begs some explanation :)

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -107,6 +107,19 @@ trait Blockchain {
       ethCompatibleStorage: Boolean
   ): StorageProof
 
+  /**
+    * Get the MptStorage
+    * @param blockNumber
+    * @return MptStorage
+    */
+  def getBackingStorage(blockNumber: BigInt): MptStorage
```
AnastasiiaL: why not to name it `getBackingMptStorage` and `getReadOnlyMptStorage`? otherwise it's to generic

AnastasiiaL (approved):

src/it/scala/io/iohk/ethereum/txExecTest/ForksTest.scala
leo-bogastry: Because the `prepareStorages` mocks the storages from the given block down to block zero. So instead of doing it 11 times, inside the foreach, I changed it to be done once for all blocks

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
leo-bogastry: Sure, I just added the original name, but this is better. Done

PR #1029, opened 2021-06-29
[ETCM-964] Fix preimage cache to make vmArithmeticTest pass

AurelienRichez:
# Description

We accidentally removed the storage key preimage cache in test mode during our refactoring, which broke lots of ETS tests. 

This PR fixes that to make those ETS test work again

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
@@ -117,8 +117,7 @@ class TestService(
     consensusConfig: ConsensusConfig,
     testModeComponentsProvider: TestModeComponentsProvider,
     initialConfig: BlockchainConfig,
-    transactionMappingStorage: TransactionMappingStorage,
-    preimageCache: collection.concurrent.Map[ByteString, UInt256]
+    transactionMappingStorage: TransactionMappingStorage
 )(implicit
     scheduler: Scheduler
 ) extends Logger {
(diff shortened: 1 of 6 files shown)


src/main/scala/io/iohk/ethereum/ledger/BlockExecution.scala
```diff
@@ -81,6 +73,18 @@ class BlockExecution(
     } yield execResult.copy(worldState = worldPersisted)
   }
 
+  protected def buildInitialWorld(block: Block, parentHeader: BlockHeader): InMemoryWorldStateProxy = {
```
AurelienRichez, line 76: ok this is weird. apparently this simple change causes this error : 
```
[error] /.../mantis/src/main/scala/io/iohk/ethereum/ledger/BlockExecution.scala:146:3: Could not find a
ny member to link for "BlockExecutionError".
[error]   /** Executes and validates a list of blocks, storing the results in the blockchain.
[error]   ^
[error] one error found
[error] Scaladoc generation failed
```
The class is mentioned in the scaladoc, and replacing it with a full name with the namespace solves the problem. But I don't really understand how this change causes a problem.

jvdp (approved):

src/main/scala/io/iohk/ethereum/jsonrpc/TestService.scala
```diff
@@ -131,6 +130,8 @@ class TestService(
   private var currentConfig: BlockchainConfig = initialConfig
   private var blockTimestamp: Long = 0
   private var sealEngine: SealEngineType = SealEngineType.NoReward
+  private val preimageCache: collection.concurrent.Map[ByteString, UInt256] =
```
lukasz-golebiewski, line 133: Looks like this `preimageCache` is used to store only storage data. If that's the case it would make sense to make this obvious from the type definition by using `AnyVal` wrappers for `ByteString` and `UInt256` like say `StorageHash`, `StorageKey`

lukasz-golebiewski (approved):

PR #1034, opened 2021-07-01
[ETCM-522][ETCM-523] Apply scalafix scalafmt rules

dzajkowski:
First commit introduces the new scala fix and fmt rules.
Second commit makes the checks on all modules mandatory on CI.
Fourth commit applies the rules to the codebase.


crypto/src/main/scala/io/iohk/ethereum/crypto/ECDSASignature.scala
```diff
@@ -63,18 +65,17 @@ object ECDSASignature {
     val pointSign = chainId match {
       case Some(id) if v == negativePointSign => (id * 2 + newNegativePointSign).toByte
       case Some(id) if v == positivePointSign => (id * 2 + newPositivePointSign).toByte
-      case None => v
-      case _ => throw new IllegalStateException(s"Unexpected pointSign. ChainId: ${chainId}, v: ${v}")
+      case None                               => v
```
AnastasiiaL, line 68: do we want that alignment? not sure it makes things prettier

PR #1037, opened 2021-07-06
ETCM-986 Remove method storeEvmCode from Blockchain

leo-bogastry:
# Description

Remove method storeEvmCode from Blockchain class and use EvmCodestorage direclty

# Proposed Solution
The method "storeEvmCode" just makes a direct call to the EvmStorage and is only used by a class and a couple of test classes

The code will be modified to have the EvmStorage called directly


jvdp (approved):
Awesome 👍

AurelienRichez (approved):

AnastasiiaL (approved):

PR #1038, opened 2021-07-06
Mutable BlockchainConfig in TestNode

jvdp:
This to fix some issues with test mode where (parts of) BlockchainConfig can change at run time, and also try out a refactoring to deal with issues stemming from the cake pattern:

- actors and RPC-service classes get an instance to the node (via the `BlockchainConfigBuilder` trait in this case) instead of the `BlockchainConfig` directly. This preserves the laziness in build initialization order (as provided by the cake) and configurability for tests. And also adds mutability, which is convenient for test mode.
- service (strategy pattern) classes such as BlockExecution are now not constructed with the `BlockchainConfig` dependency as a parameter but get it as implicit parameter instead. By using the implicit, impact on code is minimized and this refactoring is entirely mechanical in only a few places (the builder trait, the service class itself and test setup.)


src/main/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSync.scala
```diff
@@ -39,7 +40,8 @@ class RegularSync(
     syncConfig: SyncConfig,
     ommersPool: ActorRef,
     pendingTransactionsManager: ActorRef,
-    scheduler: Scheduler
+    scheduler: Scheduler,
+    node: BlockchainConfigBuilder
```
AurelienRichez: `node` is a bit surprising in that context. We can probably just call it `configBuilder`

src/main/scala/io/iohk/ethereum/testmode/TestModeComponentsProvider.scala
```diff
@@ -77,8 +77,8 @@ class TestModeComponentsProvider(
       blockchain,
       blockchainReader,
       evmCodeStorage,
-      blockchainConfig,
-      consensus(blockchainConfig, sealEngine).blockPreparator
+      consensus(blockchainConfig, sealEngine).blockPreparator,
```
AurelienRichez: I'm wondering if we should put the seal engine part inside the blockchain config.

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSyncSpec.scala
```diff
@@ -668,7 +669,7 @@ class RegularSyncSpec
 
         val parentBlock = testBlocks.last
         setImportResult(parentBlock, Task.eval(BlockImportedToTop(Nil)))
-        blockImport.importBlock(parentBlock)(Scheduler.global)
+        blockImport.importBlock(parentBlock)(Scheduler.global, blockchainConfig) // impl?
```
AurelienRichez: Is the comment still useful ?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSync.scala
jvdp: Valid remark. My reasoning was that there might be more traits tacked on later (eg. `node: BlockchainConfigBuilder with SyncConfig`.) Maybe this is only strictly needed for mutability. As for the cake pattern the impact will only become apparent later when these traits become abstact instead of responsible for building them. So the dependencies move from the 'cake traits' into these constraints in the constructor arguments.

Does that make sense? Any thoughts?

src/main/scala/io/iohk/ethereum/testmode/TestModeComponentsProvider.scala
jvdp: I think it's for test mode only.

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSyncSpec.scala
jvdp: Ah no, forgot to get back to that...

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporter.scala
```diff
@@ -331,7 +334,8 @@ object BlockImporter {
       ommersPool: ActorRef,
       broadcaster: ActorRef,
       pendingTransactionsManager: ActorRef,
-      supervisor: ActorRef
+      supervisor: ActorRef,
+      node: BlockchainConfigBuilder
```
strauss-m: _Why is this variable called "node" ?
On first read, I would have called it according to its function (like blockchainConfigBuilder) instead of its state (I recal lthat it's the node that provides the config retrieval feature)._

redacted since it relates to a later comment by @AurelienRichez

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSync.scala
strauss-m: I got the same reaction, and made exactly the same comment on a previous commit entry (removed now).

My gut feeling is that "node" should be called according to its function, and not its state, if i understand correctly the situtation: you expect the config building feature to be provided by a node instance.

src/main/scala/io/iohk/ethereum/jsonrpc/EthMiningService.scala
```diff
@@ -101,7 +101,7 @@ class EthMiningService(
                 dagSeed = EthashUtils
                   .seed(
                     pb.block.header.number.toLong,
-                    node.blockchainConfig.forkBlockNumbers.ecip1099BlockNumber.toLong
+                    blockchainConfig.forkBlockNumbers.ecip1099BlockNumber.toLong
```
strauss-m, line 103: In this edit, the construct "import node._," followed by directly accessing the inner def blockchainConfig is used here.
However, previous commit would directly refering to the complete node.blockchainConfig entry.

Is there a reason ?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSync.scala
jvdp: OK I'm outnumbered.

src/main/scala/io/iohk/ethereum/jsonrpc/EthMiningService.scala
jvdp: Yes, so this PR is to test refactoring of these class structures in a way that minimizes changes we have to make. Importing something that provides the implicit parameters helps accomplish this. And in this case we can use it to access `blockchainConfig` this way.

AurelienRichez (approved):

strauss-m (approved):

PR #1039, opened 2021-07-07
[ETCM-969] simple Branch implementation

AurelienRichez:
# Description

This PR creates a simple `BlockchainBranch` (name up for discussion). The goal is to create an immutable view of a specific branch. Currently it only works for the canonical branch, but the goal would be to have an unified representation for any branch we might want to use and allow to easily switch between branches.

For instance, instead of calling `blockchainReader.getBlockByNumber(nb)` we would call `blockchainReader.getBestBranch().getBlockByNumber(nb)`

the interface `BlockchainBranch` is the biggest change. At the end it will contain every methods which only make sense in the context of a specific branch (mostly those which need a block number). 

This implementation has a few drawback currently though : 
 - Initially I created `getBestBranch` in `BlockchainReader` with a `BlockchainBranch` because I figured that we should always at least have the genesis block (so at worst the best branch is just the genesis block). However this is not the case in the tests so I have to return an option which makes the interface cumbersome to use. I think it would be worthwhile to actually enforce this in order to have a more elegant interface.
  - The implementation might lead to a little more database access because `getBestBranch` is roughly equivalent to `getBestBlock` because a `BlockchainBranch` is defined by its tip and contains the block header. So for instance it would mean that to call `isInchain` we would get the tip and then check that a given block is in the right branch. This could probably be done without fetching the tip. I think this might be ok as the best block header can easily be cached if it becomes a problem

I had to change lots of tests for this to work. The problem is that lots of tests wrote a block in the storages but did not update the best block. As a consequence, the `getBestBranch` method does not return the actual best branch in the unit test. I 
changed the test to maintain consistency but I am not really sure that it is a good solution.

See #1044 for an improvement on the `Option` return value

### DIFF
--- a/src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
@@ -8,9 +8,9 @@ import io.iohk.ethereum.db.storage.BlockHeadersStorage
 import io.iohk.ethereum.db.storage.BlockNumberMappingStorage
 import io.iohk.ethereum.db.storage.ReceiptStorage
 import io.iohk.ethereum.db.storage.StateStorage
-import io.iohk.ethereum.domain.branch.BestBlockchainBranch
-import io.iohk.ethereum.domain.branch.BlockchainBranch
-import io.iohk.ethereum.domain.branch.EmptyBlockchainBranch
+import io.iohk.ethereum.domain.branch.BestBranch
+import io.iohk.ethereum.domain.branch.Branch
+import io.iohk.ethereum.domain.branch.EmptyBranch$
 import io.iohk.ethereum.mpt.MptNode
 import io.iohk.ethereum.utils.Logger
 
@@ -71,16 +71,16 @@ class BlockchainReader(
   def getReceiptsByHash(blockhash: ByteString): Option[Seq[Receipt]] = receiptStorage.get(blockhash)
 
   /** get the current best stored branch */
-  def getBestBranch(): BlockchainBranch =
+  def getBestBranch(): Branch =
     getBestBlock()
       .map { block =>
-        new BestBlockchainBranch(
+        new BestBranch(
           block.header,
           blockNumberMappingStorage,
           this
         )
       }
-      .getOrElse(EmptyBlockchainBranch)
+      .getOrElse(EmptyBranch$)
 
   def getBestBlockNumber(): BigInt = {
     val bestSavedBlockNumber = appStateStorage.getBestBlockNumber()
--- a/src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
@@ -8,6 +8,8 @@ import io.iohk.ethereum.db.storage.BlockHeadersStorage
 import io.iohk.ethereum.db.storage.BlockNumberMappingStorage
 import io.iohk.ethereum.db.storage.ReceiptStorage
 import io.iohk.ethereum.db.storage.StateStorage
+import io.iohk.ethereum.domain.branch.BestBlockchainBranch
+import io.iohk.ethereum.domain.branch.BlockchainBranch
 import io.iohk.ethereum.mpt.MptNode
 import io.iohk.ethereum.utils.Logger
 
@@ -85,6 +87,13 @@ class BlockchainReader(
     * @return Receipts if found
     */
   def getReceiptsByHash(blockhash: ByteString): Option[Seq[Receipt]] = receiptStorage.get(blockhash)
+
+  def getBestBranch(): BlockchainBranch =
+    new BestBlockchainBranch(
+      getBestBlock().map(_.header).getOrElse(genesisHeader),
+      blockNumberMappingStorage,
+      this
+    )
 
   def getBestBlockNumber(): BigInt = {
     val bestSavedBlockNumber = appStateStorage.getBestBlockNumber()
--- a/src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
@@ -10,6 +10,7 @@ import io.iohk.ethereum.db.storage.ReceiptStorage
 import io.iohk.ethereum.db.storage.StateStorage
 import io.iohk.ethereum.domain.branch.BestBlockchainBranch
 import io.iohk.ethereum.domain.branch.BlockchainBranch
+import io.iohk.ethereum.domain.branch.EmptyBlockchainBranch
 import io.iohk.ethereum.mpt.MptNode
 import io.iohk.ethereum.utils.Logger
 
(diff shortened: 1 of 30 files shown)


src/main/scala/io/iohk/ethereum/domain/branch/BestBlockchainBranch.scala
```diff
@@ -0,0 +1,37 @@
...
+    bestChainBlockNumberMappingStorage: BlockNumberMappingStorage,
+    blockchainReader: BlockchainReader
+) extends BlockchainBranch {
+
+  /* The following assumptions are made in this class :
+   *  - the whole branch exist in storage
+   *  - The various metadata and index are consistent
+   *
+   * If those assumptions are found to be false, then the application is in an inconsistent
+   * state and the class will throw.
```
leo-bogastry: Throw which error exactly?

AurelienRichez: In fact this comment might not be relevant anymore. It wrote it while trying to put too many function at once inside this class. You're right I should define a dedicated Exception if I econter the case though. I will remove the comment for now.

src/main/scala/io/iohk/ethereum/domain/branch/BestBlockchainBranch.scala
```diff
@@ -0,0 +1,37 @@
...
+   * */
+
+  override def getBlockByNumber(number: BigInt): Option[Block] =
+    if (tipBlockHeader.number >= number && number >= 0) {
```
AurelienRichez: This is some additionnal logic compared to what was existing. The rationale is that if we want a branch to be an immutable view, then I have to avoid fetching block "in the future" compared to the selected tip. 

This is still not perfect because if the `number -> block hash` index  is rewritten, then the interface is not really immutable. The problem should not be apparent unless the instance is used for too long.

src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
```diff
@@ -86,6 +69,18 @@ class BlockchainReader(
     */
   def getReceiptsByHash(blockhash: ByteString): Option[Seq[Receipt]] = receiptStorage.get(blockhash)
 
+  /** get the current best stored branch */
+  // FIXME this should not be an option as we should always have the genesis
+  // but some tests prevent it to simply be BlockchainBranch for now
```
leo-bogastry: `some tests prevent it to simply be BlockchainBranch for now` that sounds so wrong 🙈

src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
```diff
@@ -86,6 +69,18 @@ class BlockchainReader(
...
+  def getBestBranch(): Option[BlockchainBranch] =
```
leo-bogastry: As mentioned in our discusssion, I'm hoping that this method will be part of Consensus soon

leo-bogastry (approved):

src/main/scala/io/iohk/ethereum/domain/branch/BestBlockchainBranch.scala
```diff
@@ -0,0 +1,36 @@
...
+import akka.util.ByteString
+
+import io.iohk.ethereum.db.storage.BlockNumberMappingStorage
+import io.iohk.ethereum.domain.Block
+import io.iohk.ethereum.domain.BlockHeader
+import io.iohk.ethereum.domain.BlockchainReader
+
+/** A Branch instance which only works for the best canonical branch. As this branch
+  * currently has specific indexes (particularly regarding accessing blocks by number),
+  * it will uses thoses to provide better performance.
```
AnastasiiaL:
```suggestion
  * it will use thoses to provide better performance.
```
and what is `thoses`?

src/main/scala/io/iohk/ethereum/domain/branch/BestBlockchainBranch.scala
```diff
@@ -0,0 +1,36 @@
...
+  */
+class BestBlockchainBranch(
+    tipBlockHeader: BlockHeader,
+    bestChainBlockNumberMappingStorage: BlockNumberMappingStorage,
+    blockchainReader: BlockchainReader
+) extends BlockchainBranch {
+
+  /* The following assumptions are made in this class :
+   *  - The whole branch exist in storage
```
AnastasiiaL:
```suggestion
   *  - The whole branch exists in storage
```

src/main/scala/io/iohk/ethereum/domain/branch/BlockchainBranch.scala
```diff
@@ -0,0 +1,16 @@
+package io.iohk.ethereum.domain.branch
+
+import akka.util.ByteString
+
+import io.iohk.ethereum.domain.Block
+
+// TODO choose a name : ChainInstance, BlockchainBranch, Branch, Blockchain ?
+trait BlockchainBranch {
```
AnastasiiaL: is it gonna be used only for the main chain? or we could be interested in alternative branches?

src/main/scala/io/iohk/ethereum/domain/branch/EmptyBlockchainBranch.scala
```diff
@@ -0,0 +1,11 @@
+package io.iohk.ethereum.domain.branch
+import akka.util.ByteString
+
+import io.iohk.ethereum.domain.Block
+
+object EmptyBlockchainBranch extends BlockchainBranch {
```
AnastasiiaL: I wonder if the empty branch is possible, since we always have at least genesis block

AurelienRichez: I completely agree. The `EmptyBlockchainBranch` implementation is a compromise to allow refactoring without breaking everything. 

The current implementation does not guarantee that a genesis is set (it must be done by a side effect somewhere else), so lot of unit tests would fail when trying to get the best branch when no genesis was set. My previous implementation returned an `Option[BlockchainBranch]` to handle this, but this was tedious to use for no real reason. This implementation allows to keep a nicer interface without breaking the tests.

src/main/scala/io/iohk/ethereum/domain/branch/BlockchainBranch.scala
AurelienRichez: The intent is to be able to use it for other chains as well. So the interface itself does not make any assumption that we are indeed using the best chain.

src/main/scala/io/iohk/ethereum/domain/branch/BestBlockchainBranch.scala
AurelienRichez: Oops, Its from an old comment I removed.

AurelienRichez: I referred to the index, but the comment was initially longer. I'll rephrase that

src/main/scala/io/iohk/ethereum/domain/branch/BlockchainBranch.scala
AnastasiiaL: then maybe just Branch as a name will suit better, that'd be my choice

AnastasiiaL (approved):

src/main/scala/io/iohk/ethereum/domain/branch/BlockchainBranch.scala
AurelienRichez: OK. Let's go for it for now. This part is still heavily changing right now so we will see how this pans out.

PR #1040, opened 2021-07-07
[ETCM-921] Add optional access lists to transactions

dzajkowski:
This is the first PR to allow bodies to contain TypedTransactions and LegacyTransactions at the same time (Magneto HF).

It renames the base transaction type, introduces the TransactionWithAccessList class and fixes unit tests.

More to come.


AurelienRichez (approved):

src/main/scala/io/iohk/ethereum/domain/Transaction.scala
```diff
@@ -0,0 +1,92 @@
...
+    nonce: BigInt,
+    gasPrice: BigInt,
+    gasLimit: BigInt,
+    receivingAddress: Option[Address],
+    value: BigInt,
+    payload: ByteString
+) extends Transaction {
+  override def toString: String = {
+    val receivingAddressString =
+      receivingAddress.map(addr => Hex.toHexString(addr.toArray)).getOrElse("[Contract creation]")
```
lukasz-golebiewski: `Address` already has a good `toString` override you can use

src/main/scala/io/iohk/ethereum/domain/Transaction.scala
```diff
@@ -0,0 +1,92 @@
...
+    gasPrice: BigInt,
+    gasLimit: BigInt,
+    receivingAddress: Option[Address],
+    value: BigInt,
+    payload: ByteString,
+    accessList: List[AccessListItem]
+) extends TypedTransaction {
+  override def toString: String = {
+    val receivingAddressString =
+      receivingAddress.map(addr => Hex.toHexString(addr.toArray)).getOrElse("[Contract creation]")
```
lukasz-golebiewski: There's some duplication here, probably would make sense to move this to the `Transaction` trait

src/main/scala/io/iohk/ethereum/domain/Transaction.scala
```diff
@@ -0,0 +1,92 @@
...
+      s"gasPrice: $gasPrice " +
+      s"gasLimit: $gasLimit " +
+      s"receivingAddress: $receivingAddressString " +
+      s"value: $value wei " +
+      s"payload: ${if (isContractInit) "ContractInit: " else "TransactionData: "}${Hex.toHexString(payload.toArray[Byte])} " +
+      s"}"
+  }
+}
+
+case class TransactionWithAccessList(
```
lukasz-golebiewski, line 69: Could this by any chance simply be a `Transaction`? :-)

lukasz-golebiewski (approved):

src/main/scala/io/iohk/ethereum/domain/Transaction.scala
dzajkowski: it would collide with `trait Transaction`. or do you have a good name for the base one?

src/main/scala/io/iohk/ethereum/domain/Transaction.scala
dzajkowski: yep. I was hesitant about changing the original code but you are right, should be DRY-ed

src/main/scala/io/iohk/ethereum/domain/Transaction.scala
dzajkowski: good catch!

src/main/scala/io/iohk/ethereum/domain/Transaction.scala
lukasz-golebiewski: `TransactionBase` maybe?

dzajkowski: the deal with typed transactions is that there will be more of them. this one is just the first. I could name it `TypedTransaction01`.
02 has already been proposed and is being adapted afaik

PR #1042, opened 2021-07-07
ETCM-985 Remove method saveNode from Blockchain

leo-bogastry:
# Description

Remove method saveNode from Blockchain class and use StateStorage directly

# Proposed Solution
The method "saveNode" just makes a direct call to the StateStorage and is only in a few places

The code was modified to have the StateStorage called directly


jvdp (approved):
LGTM 👍

AnastasiiaL (approved):

PR #1043, opened 2021-07-08
[kaizen] Create EmptyBlochainBranch

AurelienRichez:
# Description

This is a proposal to make `getBestChain` usage more readable.

In my current PR #1039, `getBestChain` returns an `Option`. This does not really make sense as we at least should have a genesis so we always should be able to provide some chain. However the unit tests do not always set up a genesis so this assumption is not always true, and fixing this in the tests might be tricky. This means currently we have to put up  with an awkward `Option` and `flatMap`s in the code while in normal usage `None` would be a consistency error in the application.

This PR is a compromise where I introduce a new implementation with represent an empty chain (it always tell that a block is not in the chain). This way the interface is nicer, but we don't have to create a fully consistent test environment in unit tests. I'm not certain this is the best way, but It seems to be that it is an easy way to have the interface we should have, with no real downside compared to a plain `Option`. 

Any thought ?


jvdp:
Why'd you close it?

Also, the general idea makes sense to me. You (and the comment) note that there's always a genesis chain right; perhaps that could be the base case? (A branch with only the genesis block.)

PR #1044, opened 2021-07-08
[kaizen] Create EmptyBlochainBranch 

AurelienRichez:
# Description

This is a proposal to make `getBestChain` usage more readable.

In my current PR #1039, `getBestChain` returns an `Option`. This does not really make sense as we at least should have a genesis so we always should be able to provide some chain. However the unit tests do not always set up a genesis so this assumption is not always true, and fixing this in the tests might be tricky. This means currently we have to put up  with an awkward `Option` and `flatMap`s in the code while in normal usage `None` would be a consistency error in the application.

This PR is a compromise where I introduce a new implementation with represent an empty chain (it always tell that a block is not in the chain). This way the interface is nicer, but we don't have to create a fully consistent test environment in unit tests. I'm not certain this is the best way because it does not exactly solves the problem at hand (we still have a concept of empty chain which does not exactly exist in the "real world"). But it seems to be that it is an easy way to have the interface we should have, with no real downside compared to a plain `Option`. 

Any thought ?

### DIFF
--- a/src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
+++ b/src/it/scala/io/iohk/ethereum/sync/RegularSyncItSpec.scala
@@ -124,7 +124,7 @@ class RegularSyncItSpec extends FreeSpecBase with Matchers with BeforeAndAfterAl
       _ <- peer2.importBlocksUntil(30)(IdentityUpdate)
       _ <- peer1.startRegularSync()
       _ <- peer2.startRegularSync()
-      _ <- peer2.addCheckpointedBlock(peer2.blockchainReader.getBestBranch().get.getBlockByNumber(25).get)
+      _ <- peer2.addCheckpointedBlock(peer2.blockchainReader.getBestBranch().getBlockByNumber(25).get)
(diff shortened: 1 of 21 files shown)


jvdp:
(copied from https://github.com/input-output-hk/mantis/pull/1043#issuecomment-876417584 )

The general idea makes sense to me. You (and the comment) note that there's always a genesis chain right; perhaps that could be the base case? (A branch with only the genesis block.)

AurelienRichez:
@jvdp 
sorry. I had a bit of a misclick and could not reopen the original PR after a forced push

I fully agree that the base case should be a branch with one genesis block. I tried to make the unit test scenario always provide a genesis which was stored during test init, but ran into some NPEs in the initialization code. As I did not want to spend to much time rewriting the currently working tests, I thought this could be a decent compromise to have the nicer interface. 

It can still be worthwhile to guarantee that there is always a genesis and remove this implementation, but at least the signature of the function won't change.

leo-bogastry (approved):

jvdp (approved):

PR #1045, opened 2021-07-09
[Kaizen] Unflake RegularSynchSpec

dzajkowski:
the reason for the flakiness is a race condition between scalamock and akka testkit. by guaranteeing that actors will be shutdown when scalamock is closing down its stubs the possibility for the race is removed.

### DIFF
--- a/src/test/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSyncSpec.scala
+++ b/src/test/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSyncSpec.scala
@@ -87,6 +87,8 @@ class RegularSyncSpec
   def sync[T <: Fixture](test: => T): Future[Assertion] =
     Future {
       test
+      // this makes sure that actors are all done after the test (believe me, afterEach does not work with mocks)
+      TestKit.shutdownActorSystem(testSystem)
       succeed
     }
 
@@ -475,7 +477,7 @@ class RegularSyncSpec
         peersClient.setAutoPilot(new PeersClientAutoPilot)
         override lazy val branchResolution: BranchResolution = stub[BranchResolution]
         (blockchainReader.getBestBlockNumber _).when().returns(0)
-        (branchResolution.resolveBranch _).when(*).returns(NewBetterBranch(Nil)).repeat(10)
+        (branchResolution.resolveBranch _).when(*).returns(NewBetterBranch(Nil)).atLeastOnce()
         (blockImport
           .importBlock(_: Block)(_: Scheduler))
           .when(*, *)


src/test/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSyncSpec.scala
```diff
@@ -474,7 +476,7 @@ class RegularSyncSpec
         peersClient.setAutoPilot(new PeersClientAutoPilot)
         override lazy val branchResolution: BranchResolution = stub[BranchResolution]
         (blockchainReader.getBestBlockNumber _).when().returns(0)
-        (branchResolution.resolveBranch _).when(*).returns(NewBetterBranch(Nil)).repeat(10)
+        (branchResolution.resolveBranch _).when(*).returns(NewBetterBranch(Nil)).atLeastOnce()
```
dzajkowski, line 480: not convinced if this is required but the `repeat(10)` seems arbitrary.

leo-bogastry: Indeed, `atLeastOnce` or `anyNumberOf Times` seems more appropriate. I just always wonder why does a method needs to be called so many times. Is it a recursive one?

dzajkowski: the logic in RegularSync will run a retry on failure. the failure is the missing node (vide a bit lower) which should result in some logic and a retry. the 10 used to be 1 a long time ago.

AnastasiiaL (approved):
I wanna know the steps you took to arrive at this solution 🧐😄

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSyncSpec.scala
```diff
@@ -86,6 +86,8 @@ class RegularSyncSpec
   def sync[T <: Fixture](test: => T): Future[Assertion] =
     Future {
       test
+      // this makes sure that actors are all done after the test (believe me, afterEach does not work with mocks)
+      TestKit.shutdownActorSystem(testSystem)
```
jvdp, line 91: This 'fixture' pretty weird... are there any other tests with this?

jvdp (approved):

leo-bogastry (approved):

src/test/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSyncSpec.scala
dzajkowski: it's custom to this test, haven't seen it any other place.

AurelienRichez (approved):

PR #1051, opened 2021-07-12
[Fix] Simplify release mechanism by removing scapegoat selection

dzajkowski:
When bumping sbt a scapegoat selection mechanism was removed. It's still required.

### DIFF
--- a/.buildkite/publish.sh
+++ b/.buildkite/publish.sh
@@ -48,14 +48,13 @@ export CI_SONATYPE_RELEASE=$"; bytes/sonatypeBundleRelease; rlp/sonatypeBundleRe
 # so as a workaround the combos are called here explicitly.
 function release {
     SCALA_VERSION=$1
-    export SCAPEGOAT_VERSION=$2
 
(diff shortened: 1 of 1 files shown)


AurelienRichez (approved):

.buildkite/publish.sh
```diff
@@ -48,14 +48,13 @@ export CI_SONATYPE_RELEASE=$"; bytes/sonatypeBundleRelease; rlp/sonatypeBundleRe
 # so as a workaround the combos are called here explicitly.
```
aakoshh, line 48: Did you leave that comment for historic reason for why `release` was introduced in the first place? 

By the way it's possible now that `ci-release` could handle [everything](https://github.com/olafurpg/sbt-ci-release#how-do-i-publish-cross-built-projects). 

As per my other comment [here](https://github.com/input-output-hk/mantis/blob/23793740f67fde00ceb0c4ac00c963d7562b3739/.buildkite/publish.sh#L40-L44), inserting a `+` before each module would do the cross release, and then there wouldn't be a need to repeat the exact scala versions here.

PR #1052, opened 2021-07-12
[ETCM-995] move isInChain into BlockchainBranch

AurelienRichez:
# Description

Move isInChain function to `BlockchainBranch`

Depends on #1039

### DIFF
--- a/src/test/scala/io/iohk/ethereum/domain/BlockchainSpec.scala
+++ b/src/test/scala/io/iohk/ethereum/domain/BlockchainSpec.scala
@@ -52,11 +52,17 @@ class BlockchainSpec extends AnyFlatSpec with Matchers with ScalaCheckPropertyCh
 
   it should "be able to do strict check of block existence in the chain" in new EphemBlockchainTestSetup {
     val validBlock = Fixtures.Blocks.ValidBlock.block
+    blockchainWriter.save(
+      validBlock.copy(header = validBlock.header.copy(number = validBlock.number - 1)),
+      Seq.empty,
+      ChainWeight(100, 100),
+      saveAsBestBlock = true
+    )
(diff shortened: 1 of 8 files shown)


lukasz-golebiewski (approved):

leo-bogastry (approved):

src/test/scala/io/iohk/ethereum/domain/BlockchainSpec.scala
```diff
@@ -52,11 +52,17 @@ class BlockchainSpec extends AnyFlatSpec with Matchers with ScalaCheckPropertyCh
...
     val validBlock = Fixtures.Blocks.ValidBlock.block
+    blockchainWriter.save(
+      validBlock.copy(header = validBlock.header.copy(number = validBlock.number - 1)),
+      Seq.empty,
+      ChainWeight(100, 100),
+      saveAsBestBlock = true
+    )
     blockchainWriter.save(validBlock, Seq.empty, ChainWeight(100, 100), saveAsBestBlock = true)
-    blockchain.isInChain(validBlock.hash) === false
+    blockchainReader.getBestBranch().isInChain(validBlock.hash) should ===(true)
```
AnastasiiaL, line 62: I'm a bit concerned with condition changing from false to true😅

AurelienRichez: Good catch. I changed it consciously because previously this test did not assert anything (it just was the condition without `should`), but it failed when changing it. 

It seems to me that the real value should be true because the test : 
  - Writes a block the chain
  - Checks that after adding it the block is indeed in the chain (so `true` instead of `false`)
  - Removes that block artificially by saying that the best known block was at a reduce height
  - Checks that it is not in the chain anymore

AnastasiiaL (approved):

PR #1056, opened 2021-07-12
[Kaizen] Unflake RegularSyncSpec

dzajkowski:
Looks like a message was missing in the BlockImporter code.


src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockFetcher.scala
```diff
@@ -166,8 +167,9 @@ class BlockFetcher(
                 log.info("Dismissed received headers due to: {}", HeadersNotMatchingReadyBlocks.description)
                 peersClient ! BlacklistPeer(peer.id, BlacklistReason.UnrequestedHeaders)
                 state.withHeaderFetchReceived
-              case Left(err) =>
-                log.info("Dismissed received headers due to: {}", err)
+              case Left(HeadersNotMatchingWaitingHeaders) =>
+                log.info("Dismissed received headers due to: {}", HeadersNotMatchingWaitingHeaders.description)
+                peersClient ! BlacklistPeer(peer.id, BlacklistReason.UnrequestedHeaders)
```
AnastasiiaL, line 172: right, this one was missed.
but it shouldn't result in blacklisting.
I was following the description here - https://jira.iohk.io/browse/ETCM-370

dzajkowski: ok so the test runs into this case sometimes and that's why sometimes it fails. In this case I'll try a different condition.

PR #1057, opened 2021-07-12
[ETCM-980] encode empty "to" address in transaction RPC reply

strauss-m:
# Description

Transaction's field "to" was not encoded in contract creation since it was not set.
The facilitate ETS success, this field is expected to always be present.

# Proposed Solution

The fix allows the encoding of empty fields, and remove the duplicated (and not working) custom option serializer

### DIFF
--- a/src/main/scala/io/iohk/ethereum/jsonrpc/serialization/JsonSerializers.scala
+++ b/src/main/scala/io/iohk/ethereum/jsonrpc/serialization/JsonSerializers.scala
@@ -6,7 +6,6 @@ import org.bouncycastle.util.encoders.Hex
 import org.json4s.CustomSerializer
 import org.json4s.DefaultFormats
 import org.json4s.Formats
-import org.json4s.JNull
 import org.json4s.JString
 
 import io.iohk.ethereum.domain.Address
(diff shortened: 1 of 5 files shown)


jvdp:
Interesting, I actually did the same change way back but it was reverted quickly after: [`a8b92f5` (#965)](https://github.com/input-output-hk/mantis/pull/965/commits/a8b92f51d7d1ea6355ca6e268c67246f1b4b8cbb#diff-37f3a5ada6f3f4b82944cd1dbae68cdd4822629c9c70d21b2f984bb0984d168d)

I wonder if you'll run into anything. Does retesteth not complain about extra fields?

jvdp (approved):

AnastasiiaL (approved):

strauss-m:
Actually, unit-tests are not happy with this change.
eth_getTransactionReceipt is expecting either a root, or a status, and it's handled throught optional (see https://eth.wiki/json-rpc/API#eth_gettransactionreceipt).
I'm guess I should rollback the DefaultFormat.preserveEmptyValue part and do a hand-crafted custom json encoder for the EthTransactionResponse.

PR #1062, opened 2021-07-13
[ETCM-1014] retesteth: 0.1.0 → 0.1.1

nrdxp:
I force pushed a rebase on our retesteth fork with an update to the latest version, as well as some simple fixups to get it building. This updates the package definition to build this new version.

### DIFF
--- a/nix/retesteth.nix
+++ b/nix/retesteth.nix
@@ -79,13 +79,13 @@ let
 in
 stdenv.mkDerivation rec {
   pname = "retesteth";
-  version = "v0.1.0-accesslist_fix";
+  version = "v0.1.1-eip1559";
 
   src = fetchFromGitHub {
     owner = "input-output-hk";
     repo = "retesteth";
     rev = "remove-hunter";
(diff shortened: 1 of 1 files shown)


dzajkowski (approved):

jvdp (approved):

jvdp:
> I force pushed a rebase on our retesteth fork with an update to the latest version, as well as some simple fixups to get it building. This updates the package definition to build this new version.

Do you think it would be complicated to enable/disable hunter with a flag? Then we might upstream the changes.

nrdxp:
> Do you think it would be complicated to enable/disable hunter with a flag? Then we might upstream the changes.

The only change is to remove hunter and use pkg-config instead, so there would be nothing to upstream. We tried making hunter work originally by simply toggling it off, but that wasn't enough, and nothing we tried worked except removing it completely.

jvdp:
> The only change is to remove hunter and use pkg-config instead, so there would be nothing to upstream. We tried making hunter work originally by simply toggling it off, but that wasn't enough, and nothing we tried worked except removing it completely.

Ahh ok, so pretty complicated. Too bad.

PR #1065, opened 2021-07-14
Refresh retesteth genesis files

jvdp:


strauss-m:
I managed to get 'nix-in-docker run retesteth' running locally with the new retesteth version.
Basically I've added Berlin and London hardfork in the list of supported forks, updated the reward config file, and added the new London.json +  BerlinToLondonAt5.json genesis config file.

The Berlin + London supported forks added in config seems to be more a workaround, but I feel that retesteth is just not planned to work on non-london supporting clients.

For this to work, I need to add the '--singlenet Istanbul' parameter to force the Istanbul fork.

It's different from your solution, so I'd like to understand why the overall fork cleanup in most of the genersis/*.json file ? Maybe it's the missing part that forces me to specify the expected fork version in the command line.

As a reference, I've commited my change in a temporary branch: https://github.com/input-output-hk/mantis/commit/94d341a82135a2ef26c80946e7aceb727769949b

The full command line to launch locally the tests is:
`./nix-in-docker/run --command "ets/retesteth -t BlockchainTests/ValidBlocks/bcGasPricerTest -- --nodes 192.168.65.2:8546 --singletest RPC_API_Test --verbosity 6 --singlenet Istanbul"
`

Doing so gives me the same results than using the retesteth docker binary.

jvdp:
#1068 actually fixes the issue.

PR #1067, opened 2021-07-15
[ETCM-1015] Add Magneto block height to configs

lukasz-golebiewski:
# Description
This PR adds Magneto block height to the app's config files


src/test/scala/io/iohk/ethereum/consensus/blocks/BlockGeneratorSpec.scala
```diff
@@ -224,28 +224,12 @@ class BlockGeneratorSpec extends AnyFlatSpec with Matchers with ScalaCheckProper
       gasTieBreaker = false,
       ethCompatibleStorage = true,
       treasuryAddress = Address(0),
-      forkBlockNumbers = ForkBlockNumbers(
+      forkBlockNumbers = ForkBlockNumbers.Empty.copy(
```
AurelienRichez, line 227: 👍

AurelienRichez (approved):

strauss-m (approved):

src/main/resources/conf/chains/eth-chain.conf
```diff
@@ -98,6 +98,10 @@
   # https://ecips.ethereumclassic.org/ECIPs/ecip-1099
   ecip1099-block-number = "1000000000000000000"
 
+  # Magneto EVM and Protocol Upgrades
+  # https://ecips.ethereumclassic.org/ECIPs/ecip-1103
+  magneto-block-number = "1000000000000000000"
```
leo-bogastry, line 103: Instead of using a super high block number, why not define this config value as `Option[BigInt]`, and omit it in the files where it does not make sense?

lukasz-golebiewski: I was thinking about doing it like this, but in the end didn't to remain consistent. Perhaps it would make sense to refactor those configs with the approach above in a separate PR?

leo-bogastry: There is a configuration added by me, that is an `Option` already :) So total consistency is already gone 😅 
I would love to see this refactored

dzajkowski (approved):

PR #1068, opened 2021-07-15
[ETCM-1028] fix retesteth config after version upgrade to v0.1.1-eip1559

strauss-m:
# Description

retesteth doesn't launch anymore after version update

# Proposed Solution

retesteth seems to expect the London fork in the config file, which is currently not present since Mantis doesn't support Berlin neither London.

A workaround seems to declare London as a supported fork, but not Berlin.
That way, Berlin tests are skipped, and since there are no London tests in the repository, retesteth just runs all tests up to and including the Istanbul ones.

An issue has been opened to further investigate a potential retesteth bug or misunderstood documentation: https://github.com/ethereum/retesteth/issues/141

Please be aware that the London inclusion asks retesteth to run all the London-related test. Currently, they are not yet included, but they will once https://jira.iohk.io/browse/ETCM-1033 is done. The workaround won't work anymore at that point.


jvdp:
Can we leave a note for the person that will update the `ets/tests` git submodule that the new "london" tests are run (more or less) inadvertently? (Like a comment on the related PR, perhaps...)

jvdp (approved):

strauss-m:
> Can we leave a note for the person that will update the `ets/tests` git submodule that the new "london" tests are run (more or less) inadvertently? (Like a comment on the related PR, perhaps...)

Sure, where's the best place to do it ? I don't see any related jira ticket or in progress PR for such a task.

jvdp:
> Sure, where's the best place to do it ? I don't see any related jira ticket or in progress PR for such a task.

Sorry I was being a bit vague. I'll create a ticket and a note on it there.

AurelienRichez (approved):

AurelienRichez:
Nice 👍

PR #1069, opened 2021-07-16
Etcm 1030

AnastasiiaL:
# Description

ExecutionSync is gonna be orchestrating the whole regular sync process and is gonna replace current RegularSync class.

Note: tests will be added in a separate PR


leo-bogastry (approved):

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/ExecutionSync.scala
```diff
@@ -0,0 +1,142 @@
...
+      log.info("Starting regular sync")
+      peerEventBus ! Subscribe(
+        MessageClassifier(
+          Set(Codes.NewBlockCode, Codes.NewBlockHashesCode, Codes.BlockHeadersCode),
+          PeerSelector.AllPeers
+        )
+      )
+      val listOfPeersToFetchFromWithBestBlock: List[(Peer, Int)] = ???
+      val startFrom: BigInt = blockchainReader.getBestBlockNumber()
+      listOfPeersToFetchFromWithBestBlock.map(peerWithBlockNumber =>
```
strauss-m: style remark (which is of course personal): I would have gone with: 
```
listOfPeersToFetchFromWithBestBlock.map { case (peerId, peerBestBlockNumber) =>
  fetcherService.fetchBlocksUntil(peerId, Left(startFrom), Left(peerBestBlockNumber))
}
```

I tend to find it easier to read a case(a, b) than a tuple element acces like data._1, even though in this case the name is pretty straightforward

strauss-m (approved):

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/ExecutionSync.scala
```diff
@@ -0,0 +1,142 @@
...
+    BlockBroadcasterActor
+      .props(new BlockBroadcast(etcPeerManager), peerEventBus, etcPeerManager, blacklist, syncConfig, scheduler),
+    "block-broadcaster"
+  )
+
+  override def receive: Receive = running(
+    ProgressState(initialBlock = 0, currentBlock = 0, bestKnownNetworkBlock = 0)
+  )
+
+  def running(progressState: ProgressState): Receive = {
```
jvdp: So this class is aimed to be a drop-in replacement for `RegularSync`?

I have some questions / remarks:
1. do you have some ideas on how we will work towards being able to replace the `RegularSync` actor in the future. ie. what needs to be done before we "flip the switch"?
2. related: so you listed the incoming messages (with similar "context.become") but what about the outgoing messages that are required?
3. can we try to keep the actor-code minimal from the get go? Ie. put the actual implementation in separate object methods (or classes if we have to.) And the actor state could then be a case class / ADT with transition methods, unit-testable without involving an actor system.
4. related to that: do you think it'd be feasible to move code out of the `RegularSync` actor and replace them piece by piece? Since in the end the API of the actor should be largely the same as we refactor it. (Assuming that's the proposed approach.)

src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
```diff
@@ -39,6 +43,8 @@ class PeersClient(
   val statusSchedule: Cancellable =
     scheduler.scheduleWithFixedDelay(syncConfig.printStatusInterval, syncConfig.printStatusInterval, self, PrintStatus)
 
+  val activeFetchingNodes: mutable.Set[Peer] = lruSet[Peer](syncConfig.peersToFetchFrom)
```
jvdp: Is the LRU set the right behavior for this and would it be possible to split this out into something that is unit-testable?

Don't we want to reject peers based on something other than "least recently used"? What's the definition of "used" here btw?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporterService.scala
```diff
@@ -0,0 +1,14 @@
+package io.iohk.ethereum.blockchain.sync.regular
+
+import cats.data.NonEmptyList
+import io.iohk.ethereum.blockchain.sync.regular.BlockImporter.{BlockImportType, ImportFn, ResolvingBranch}
+import io.iohk.ethereum.domain.Block
+
+//not used atm, a part of the future ExecutionSync
+class BlockImporterService() {
```
jvdp: I'm not super convinced of the value of adding empty files like this.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/HeadersFetcher.scala
```diff
@@ -84,7 +82,7 @@ object HeadersFetcher {
     Behaviors.setup(context => new HeadersFetcher(peersClient, syncConfig, supervisor, context))
 
   sealed trait HeadersFetcherCommand
-  final case class FetchHeaders(blockNumber: BigInt, amount: BigInt) extends HeadersFetcherCommand
+  final case class FetchHeaders(block: Either[BigInt, ByteString], amount: BigInt) extends HeadersFetcherCommand
```
dzajkowski: wouldn't an explicit command be easier to reason about?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
```diff
@@ -0,0 +1,86 @@
...
+import monix.eval.Task
+import io.iohk.ethereum.blockchain.sync.PeerRequestHandler.RequestFailed
+import io.iohk.ethereum.consensus.validators.BlockValidator
+import io.iohk.ethereum.utils.Config.SyncConfig
+import cats.implicits._
+
+import scala.annotation.tailrec
+
+//not used atm, a part of the future ExecutionSync
+class FetcherService(validator: BlockValidator, blockchainReader: BlockchainReader, syncConfig: SyncConfig) {
```
dzajkowski, line 24: I think it would be very beneficial to get in the habit of adding code that is covered by unit tests.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
```diff
@@ -0,0 +1,86 @@
...
+  val batchSize = syncConfig.blockHeadersPerRequest
```
dzajkowski, line 25: I think that this is an unfortunate pattern. the class does not depend on the whole sysconfig class - it depends on one field from that class. how can we better express such dependencies?
if it's one field perhaps a tagged number would be good enough?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  // Checks that the received block bodies are an ordered subset of the ones requested
+  @tailrec
+  private def bodiesAreOrderedSubsetOfRequested(
+      requestedHeaders: Seq[BlockHeader],
+      respondedBodies: Seq[BlockBody],
+      matchedBlocks: Seq[Block] = Nil
+  ): Option[Seq[Block]] =
+    (requestedHeaders, respondedBodies) match {
+      case (Seq(), _ +: _) => None
```
dzajkowski, line 84: `Seq()` wouldn't this allocate an empty Seq on every match?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
```diff
@@ -0,0 +1,86 @@
...
+      case (_, Seq())      => Some(matchedBlocks)
+      case (header +: remainingHeaders, body +: remainingBodies) =>
+        val doMatch = validator.validateHeaderAndBody(header, body).isRight
+        if (doMatch)
```
dzajkowski: not sure what the value of `doMatch` is.
```
if(validator.validateHeaderAndBody(header, body).isRight)
```
is not readable?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
```diff
@@ -0,0 +1,86 @@
...
+      case None      => EitherT.leftT(RequestFailed(peer, "couldn't find blocks to fetch"))
+      case Some(seq) => seq.traverse(num => fetchBlocks(peer, batchSize, Left(num))).map(_ => ())
+    }
+  }
+
+  private def fetchBlocks(
+      peer: Peer,
+      amount: BigInt,
+      block: Either[BigInt, ByteString]
+  ): EitherT[Task, RequestFailed, Unit] =
```
dzajkowski: I would recommend making this function into a pure one. this way the dependency on the queue has to be explicit and it will aid testing.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  def fetchBlocksUntil(
+      peer: Peer,
+      startFromBlock: Either[BigInt, ByteString],
+      fetchUntilBlock: Either[BigInt, ByteString]
+  ): EitherT[Task, RequestFailed, Unit] = {
+    val endNumber: Option[BigInt] =
+      fetchUntilBlock.fold(Some(_), blockchainReader.getBlockHeaderByHash(_).map(_.number))
+    val startNumber: Option[BigInt] =
+      startFromBlock.fold(Some(_), blockchainReader.getBlockHeaderByHash(_).map(_.number))
```
dzajkowski, line 46: correct me if I'm wrong but wouldn't it be beneficial to request a block by hash from a peer?

dzajkowski (changes requested):
I think that this one merits some discussion. maybe we can address it on the synch meeting?

src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
```diff
@@ -129,6 +136,25 @@ class PeersClient(
 
     log.debug(s"Handshaked peers status (number of peers: ${handshakedPeersStatus.size}): $handshakedPeersStatus")
   }
+
+  //returns the next best peer after the one already returned previously
+  //TODO: make sure the next best peer has a different best block, so we don't fetch in parallel identical branches
+  private def nextBestPeer(peersToDownloadFrom: Map[PeerId, PeerWithInfo]): Option[Peer] = {
```
dzajkowski: this could (and IMO should) be a public function in an object. and it should be covered by a set of unit tests to document the expectation (`activeFetchingNodes.contains(peerAndWeight._1)).maxByOption(_._2).map(_._1)` in not obvious at first glance)

src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
```diff
@@ -129,6 +136,25 @@ class PeersClient(
...
+    val peersToUse = peersToDownloadFrom.values
+      .collect { case PeerWithInfo(peer, PeerInfo(_, chainWeight, true, _, _)) =>
+        (peer, chainWeight)
+      }
+
+    val peer =
+      peersToUse.filter(peerAndWeight => !activeFetchingNodes.contains(peerAndWeight._1)).maxByOption(_._2).map(_._1)
+    peer.foreach(activeFetchingNodes.add)
```
dzajkowski: I think that hiding side effects makes the code harder to read and reason about.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
jvdp: There is only one instance of `Nil` though right? https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/immutable/List.scala#L661

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/ExecutionSync.scala
AnastasiiaL: 1. Once the fetcher service is complete and block importer is replaced with a chain manager
2. what outgoing messages do you mean? those that were previously going to BlockImporter will now go to the ChainManager, that's mentioned in the comments, unless you mean something else
3. of course we can try :) what logic from this actor would you extract?
4.  I'm not sure I follow here: move the code out where? the API won't change much indeed, it's just gonna be a bit simpler and message propagating will follow another path, but the function of this actor will stay largely the same (unless we decide to change that)

src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
AnastasiiaL: we are rejecting peers for multiple reasons, but this piece of code has nothing to do to rejection, it's needed for picking best N peers out of our pool of discovered and connected peers, and we need to make sure that when we fetch from N chosen peers in parallel, all the peers are actually different, hence lru set. "used" here in the meaning we are fetching from it at the moment.

If you have an alternative approach for this - I don't mind changing this piece.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BlockImporterService.scala
AnastasiiaL: I agree, this is not supposed to go to the codebase, it was here for the sake of our discussions, that were based on those pieces

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/HeadersFetcher.scala
AnastasiiaL: the purpose of this command is unchanged regardless the parameter being Left or Right, having 2 commands would only add cumbersomeness imo. But I'm fine with whatever the team prefers. What do you think @LeonorLunatech @jvdp @AurelienRichez ?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
AnastasiiaL: I'm not sure I know what's a tagged number🤔

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
AnastasiiaL: this method is fully taken from the current codebase, and it seems to be working. Do you want me to rewrite it?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
AnastasiiaL: agree

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
AnastasiiaL: right

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/HeadersFetcher.scala
AurelienRichez: I have a slight preference for having two explicit commands `FetchHeadersByNumber` and `FetchHeadersByHash` instead of using Either to overload an existing command. It will lead to a few more lines, but I find it easier to understand. I note that we have other instances of this pattern in `GetBlockHeaders` case class.

src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
jvdp: An immutable set of peers? Then the logic that operate on the set can be unit tested, separate from where it is kept as state.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/ExecutionSync.scala
jvdp: >     2\. what outgoing messages do you mean? those that were previously going to BlockImporter will now go to the ChainManager, that's mentioned in the comments, unless you mean something else

What about the other `ProgressProtocol` messages and messages to the fetcher?


>     3\. of course we can try :) what logic from this actor would you extract?

I propose a straight forward transformation into a case class with values for every bit of state and transition methods for the different messages that cause state changes. A bit like `BlockFetcherState`.



>     4\. I'm not sure I follow here: move the code out where? the API won't change much indeed, it's just gonna be a bit simpler and message propagating will follow another path, but the function of this actor will stay largely the same (unless we decide to change that)

I meant to replace the current `RegularSync` actor gradually, instead of all at once.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/HeadersFetcher.scala
jvdp: This `Either` might've been leaked from the json RPC. I think that's not the best reason to do it this way but I'm otherwise neutral on it.

Do we actually need both, though? Aren't we always going to fetch by hash?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/ExecutionSync.scala
AnastasiiaL: > What about the other ProgressProtocol messages and messages to the fetcher?

ah incoming messages you meant, I don't see a point in ProgressProtocol messages. do you? it's only used for logging, why not to log in place?

And the only message that was going to the fetcher was used to keep the fetcher state's blocks queue aware of the mined and checkpointed  blocks. This is no longer our intention with the new design.

> I propose a straight forward transformation into a case class with values for every bit of state and transition methods for the different messages that cause state changes. A bit like BlockFetcherState.

I still don't see what you're trying to achieve, BlockFetcherState is introduced for handling a state and a validation logic, while ExecutionSync/RegularSync is just an actor passing the messages around with no logic being executed inside an actor. 

But feel free to create your version of it, then we can discuss it.

> I meant to replace the current RegularSync actor gradually, instead of all at once.

We don't replace it atm, the decision was just to have this class in our codebase as a hint of how the things will be glued later, the transition of this part was not discussed yet.


I don't mind not merging this at all, if the purpose of it (or a gain) is not clear

src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
AnastasiiaL: and how to make sure that every Nth peer replaces the oldest one from the set? reimplement lruSet?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/HeadersFetcher.scala
AnastasiiaL: we need to support both, yes

src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
jvdp: Yes, and using immutable data structures.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
AnastasiiaL: actually thinking more on that - we wouldn't know the hash of the block we want to fetch from the peer if we don't have that block ourselves. And the point of fetching from multiple peers is exactly to get alternative chains, so we do need to fetch by number here

AurelienRichez (approved):

dzajkowski (approved):

PR #1070, opened 2021-07-21
Feature/etcm 967 db consistency check

dzajkowski:
Check if the db is consistent. It has been observed that dbs get broken on sagano. This should support figuring out when it happens.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/nodebuilder/StdNode.scala
+++ b/src/main/scala/io/iohk/ethereum/nodebuilder/StdNode.scala
@@ -29,6 +29,8 @@ abstract class BaseNode extends Node {
 
     loadGenesisData()
 
+    verifyStorageConsistency()
+
     startPeerManager()
 
     startPortForwarding()
@@ -57,6 +59,22 @@ abstract class BaseNode extends Node {
 
   private[this] def loadGenesisData(): Unit =
     if (!Config.testmode) genesisDataLoader.loadGenesisData()
+
+  private[this] def verifyStorageConsistency(): Unit = {
+    val step = 1000
+    val bbn = storagesInstance.storages.appStateStorage.getBestBlockNumber()
+    val bnms = storagesInstance.storages.blockNumberMappingStorage
+    val bhs = storagesInstance.storages.blockHeadersStorage
+    Range(0, bbn.intValue, step).foreach { idx =>
+      (for {
+        hash <- bnms.get(idx)
+        _ <- bhs.get(hash)
+      } yield ()).fold {
+        log.error("Database seems to be in inconsistent state, shutting down")
+        shutdown()
+      }(_ => ())
+    }
+  }
 
   private[this] def startPeerManager(): Unit = peerManager ! PeerManagerActor.StartConnecting
 
--- a/src/main/scala/io/iohk/ethereum/nodebuilder/StdNode.scala
+++ b/src/main/scala/io/iohk/ethereum/nodebuilder/StdNode.scala
@@ -13,6 +13,7 @@ import io.iohk.ethereum.metrics.MetricsConfig
 import io.iohk.ethereum.network.PeerManagerActor
 import io.iohk.ethereum.network.ServerActor
 import io.iohk.ethereum.network.discovery.PeerDiscoveryManager
+import io.iohk.ethereum.nodebuilder.tooling.StorageConsistencyChecker
 import io.iohk.ethereum.utils.Config
 
 /** A standard node is everything Ethereum prescribes except the mining algorithm,
@@ -29,7 +30,12 @@ abstract class BaseNode extends Node {
 
     loadGenesisData()
 
-    verifyStorageConsistency()
+    StorageConsistencyChecker.checkStorageConsistency(
+      storagesInstance.storages.appStateStorage.getBestBlockNumber(),
+      storagesInstance.storages.blockNumberMappingStorage,
+      storagesInstance.storages.blockHeadersStorage,
+      shutdown
+    )(log)
 
     startPeerManager()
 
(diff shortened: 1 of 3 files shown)


src/main/scala/io/iohk/ethereum/nodebuilder/StdNode.scala
```diff
@@ -58,6 +60,22 @@ abstract class BaseNode extends Node {
...
+    val step = 1000
+    val bbn = storagesInstance.storages.appStateStorage.getBestBlockNumber()
+    val bnms = storagesInstance.storages.blockNumberMappingStorage
+    val bhs = storagesInstance.storages.blockHeadersStorage
+    Range(0, bbn.intValue, step).foreach { idx =>
+      (for {
+        hash <- bnms.get(idx)
+        _ <- bhs.get(hash)
+      } yield ()).fold {
+        log.error("Database seems to be in inconsistent state, shutting down")
```
LeonorLunatech: shouldn't we include information on how the user should act when this happens? I guess how to delete the DB?

src/main/scala/io/iohk/ethereum/nodebuilder/tooling/StorageConsistencyChecker.scala
```diff
@@ -0,0 +1,29 @@
...
+      blockHeadersStorage: BlockHeadersStorage,
+      shutdown: ShutdownOp,
+      step: Int = DefaultStep
+  )(implicit log: Logger): Unit =
+    Range(0, bestBlockNumber.intValue, step).foreach { idx =>
+      (for {
+        hash <- blockNumberMappingStorage.get(idx)
+        _ <- blockHeadersStorage.get(hash)
+      } yield ()).fold {
+        log.error("Database seems to be in inconsistent state, shutting down")
```
leo-bogastry, line 25: Shouldn't we include information on how the user should act when this happens? I guess how to delete the DB?

leo-bogastry:
Is the long-term plan to be able to have Mantis self fix this consistency?

src/main/scala/io/iohk/ethereum/nodebuilder/tooling/StorageConsistencyChecker.scala
dzajkowski: Is this a good place to add such documentation for the client? Not convinced, I think we have a mantis docs project.
We don't have a policy for such situations.

dzajkowski:
> Is the long-term plan to be able to have Mantis self fix this consistency?

@LeonorLunatech depends, we don't know what the setting is. We did not have a user report this as an issue, so it might be test env specific.

src/main/scala/io/iohk/ethereum/nodebuilder/tooling/StorageConsistencyChecker.scala
leo-bogastry: In that case advising the user to check the website. But I think that saying nothing is the worse option :)

leo-bogastry:
> > Is the long-term plan to be able to have Mantis self fix this consistency?
> 
> @LeonorLunatech depends, we don't know what the setting is. We did not have a user report this as an issue, so it might be test env specific.

Ok, i get that. In that case, because I'm also wondering if there's a performance impact, would it make sense to have this enabled only when running in test networks?

src/main/scala/io/iohk/ethereum/nodebuilder/StdNode.scala
```diff
@@ -24,28 +28,37 @@ import io.iohk.ethereum.utils.Config
...
 
-  private[this] def startPeerManager(): Unit = peerManager ! PeerManagerActor.StartConnecting
+  def start(): Unit = {
+    startMetricsClient()
 
-  private[this] def startServer(): Unit = server ! ServerActor.StartServer(networkConfig.Server.listenAddress)
+    loadGenesisData()
 
-  private[this] def startSyncController(): Unit = syncController ! SyncProtocol.Start
+    StorageConsistencyChecker.checkStorageConsistency(
```
lukasz-golebiewski: Do you mind wrapping this in a `startConsistencyCheck()` or something similar?

src/main/scala/io/iohk/ethereum/nodebuilder/tooling/PeriodicConsistencyCheck.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  def check(): Behavior[ConsistencyCheck] = Behaviors.receiveMessage { case Tick =>
+    log.debug("Running a storageConsistency check")
+    StorageConsistencyChecker.checkStorageConsistency(
+      appStateStorage.getBestBlockNumber(),
+      blockNumberMappingStorage,
+      blockHeadersStorage,
+      shutdown
+    )(log)
+    timers.startSingleTimer(Tick, 5.seconds)
```
lukasz-golebiewski: It's a bit confusing. Are we doing the check every 10 minutes (https://github.com/input-output-hk/mantis/pull/1070/files#diff-a8c02d6306b1c231f0efde1b3f46fa9cc6659899bfa882c317479432df489a7aR23) or 5 seconds? 
It's probably a good idea to make this configurable, unless it's just a temporary measure

src/main/scala/io/iohk/ethereum/nodebuilder/tooling/StorageConsistencyChecker.scala
```diff
@@ -0,0 +1,29 @@
...
+  val DefaultStep = 1000
+
+  def checkStorageConsistency(
+      bestBlockNumber: BigInt,
+      blockNumberMappingStorage: BlockNumberMappingStorage,
+      blockHeadersStorage: BlockHeadersStorage,
+      shutdown: ShutdownOp,
+      step: Int = DefaultStep
+  )(implicit log: Logger): Unit =
+    Range(0, bestBlockNumber.intValue, step).foreach { idx =>
```
lukasz-golebiewski, line 20: I'm wondering if it makes sense to check every 1000th block. I imagine a case where the check passes but the db is inconsistent.
Perhaps it would be worthwhile to occasionally check the whole range?

src/main/scala/io/iohk/ethereum/nodebuilder/tooling/PeriodicConsistencyCheck.scala
dzajkowski: good catch, this should be a method

src/main/scala/io/iohk/ethereum/nodebuilder/tooling/StorageConsistencyChecker.scala
dzajkowski: this is a first pass. I want to start identifying the circumstances when clients loose consistency.

lukasz-golebiewski (approved):

src/main/scala/io/iohk/ethereum/nodebuilder/tooling/StorageConsistencyChecker.scala
AurelienRichez: I think we should log the missing block's hash and number. I'm not sure it would really helps us when this happens, but it does not hurt.

AurelienRichez (approved):

strauss-m (approved):

PR #1073, opened 2021-07-22
[ETCM-1042] Switch to a case class implementation for blockchain branch

AurelienRichez:
# Description

Instead of having a interface which tries to contains every method for a branch, we change it to be a simple case class. It will be more flexible for future changes ; e.g. implement some function for the canonical branch but not for other kinds of branch (or the other way around), without changing a central interface.

This PR

 - makes the blockchain implementation only contain the hash and the number of the tip (before it was the whole header, but it was not necessary)
 - uses a case class with a sealed trait, instead of a trait with methods to implement


src/main/scala/io/iohk/ethereum/domain/branch/Branch.scala
```diff
@@ -2,17 +2,8 @@ package io.iohk.ethereum.domain.branch
 
 import akka.util.ByteString
 
-import io.iohk.ethereum.domain.Block
+sealed trait Branch
 
-/** An interface to manipulate blockchain branches */
-trait Branch {
+case class BestBranchSubset(tipBlockHash: ByteString, tipBlockNumber: BigInt) extends Branch
```
leo-bogastry: Why did you add the suffix `Subset` here?

AurelienRichez: It works with any subset on the best branch so I wanted to make it clear and kept the name in from my initial draft of the case class representation. The idea was that we could use it later to define a fork as a tip and a best branch subset which is common to the fork and our current best branch. 

But I agree that it might be confusing and not really relevant to our use case for now. I'll just name it BestBranch as before

leo-bogastry (approved):

src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
```diff
@@ -71,16 +71,13 @@ class BlockchainReader(
...
-          this
-        )
-      }
-      .getOrElse(EmptyBranch$)
+  def getBestBranch(): Branch = {
+    val number = getBestBlockNumber()
+    blockNumberMappingStorage
+      .get(number)
+      .map(hash => BestBranch(hash, number))
+      .getOrElse(EmptyBranch)
```
AnastasiiaL, line 80: a fallback to the second best branch may be more suitable than an empty branch

src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
```diff
@@ -113,6 +110,38 @@ class BlockchainReader(
...
+        for {
+          hash <- getHashByBlockNumber(number)
+          block <- getBlockByHash(hash)
+        } yield block
+      } else None
+    case EmptyBranch => None
+  }
+
+  /** Returns a block hash for the block at the given height if any */
+  def getHashByBlockNumber(branch: Branch, number: BigInt): Option[ByteString] = branch match {
```
AnastasiiaL, line 125: what do you think of using a type alias for ByteString when we are talking about a hash?

src/main/scala/io/iohk/ethereum/domain/branch/Branch.scala
```diff
@@ -2,17 +2,8 @@ package io.iohk.ethereum.domain.branch
 
 import akka.util.ByteString
 
-import io.iohk.ethereum.domain.Block
+sealed trait Branch
 
-/** An interface to manipulate blockchain branches */
-trait Branch {
+case class BestBranch(tipBlockHash: ByteString, tipBlockNumber: BigInt) extends Branch
```
AnastasiiaL, line 7: I can imagine this case class being used for all the branches, not only the best one

src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
AurelienRichez: I added a fallback like in `getBestBlock` to be more consistent

src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
AurelienRichez: I would prefer a value class or a tagged type, because the type alias would just be "cosmetic". But it would probably be out of scope to introduce the new type in this PR.

src/main/scala/io/iohk/ethereum/domain/branch/Branch.scala
AurelienRichez: It does not contain data specific to the best branch, yes. Right now it is more a marker type to tell the code that we can use the block number -> hash index.

src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
```diff
@@ -71,16 +71,14 @@ class BlockchainReader(
...
-          blockNumberMappingStorage,
-          this
-        )
-      }
-      .getOrElse(EmptyBranch$)
+  def getBestBranch(): Branch = {
+    val number = getBestBlockNumber()
+    blockNumberMappingStorage
+      .get(number)
+      .orElse(blockNumberMappingStorage.get(appStateStorage.getBestBlockNumber()))
```
jvdp, line 78: What's this for? This is a work-around not related to this change, right?

src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
```diff
@@ -113,6 +111,38 @@ class BlockchainReader(
   def genesisBlock: Block =
     getBlockByNumber(0).get
 
+  /** Returns a block inside this branch based on its number */
+  def getBlockByNumber(branch: Branch, number: BigInt): Option[Block] = branch match {
+    case BestBranch(_, tipBlockNumber) =>
+      if (tipBlockNumber >= number && number >= 0) {
```
jvdp: Could be nice as a guard on the `case`.

src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
```diff
@@ -113,6 +111,38 @@ class BlockchainReader(
...
+
+    case EmptyBranch => None
+  }
+
+  /** Checks if given block hash is in this chain. (i.e. is an ancestor of the tip block) */
+  def isInChain(branch: Branch, hash: ByteString): Boolean = branch match {
+    case BestBranch(_, tipBlockNumber) =>
+      (for {
+        header <- getBlockHeaderByHash(hash) if header.number <= tipBlockNumber
+        hash <- getHashByBlockNumber(branch, header.number)
```
jvdp: You are shadowing `hash` here; perhaps you can specify the names a bit more?

Also, this relies on using the "hash by block number" index as an authority of "this chain". How do you envision that to work in the new architecture? Will this index be kept up to date eg. when reorganising?

src/main/scala/io/iohk/ethereum/domain/branch/Branch.scala
```diff
@@ -2,17 +2,8 @@ package io.iohk.ethereum.domain.branch
...
-  /** Returns a block inside this branch based on its number */
-  def getBlockByNumber(number: BigInt): Option[Block]
-
-  /** Returns a block hash for the block at the given height if any */
-  def getHashByBlockNumber(number: BigInt): Option[ByteString]
-
-  /** Checks if given block hash is in this chain. (i.e. is an ancestor of the tip block) */
-  def isInChain(hash: ByteString): Boolean
-}
+case object EmptyBranch extends Branch
```
jvdp, line 9: So we'll have multiple types of branch later on, right? For the purposes of the code you're changing now (branch as produced by `getBestBranch`) you could also name the trait `BestBranch` and rename empty branch to `Genesis`, which resolves to a block number of 1. (Or is it 0?)

jvdp (approved):

src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
AurelienRichez: Yes it is not related to this change.
It's related to [anastasiia's comment](https://github.com/input-output-hk/mantis/pull/1073/files/3a61349d1244a5c16b27278508989033a583ea95#r678129983). It allows to be consistent with `getBestBlock` which fallbacks to what is in the storage if the block from memory does not exists.

src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
AurelienRichez: Good catch. Fixed

For now we have to keep the index updated indeed. Right now `isInChain` is used by `BranchResolution.resolveBranch` to check if a sequence of blocks has a parent inside the canonical chain (i.e. if the sequence is rooted). In the end, we would like for mantis to be able to synchronize without maintaining that index. It would still be there to help json rpc calls, but mantis would not have to wait for it to be rebuilt to continue its work.

That said, in the new design, we won't really care that a parent is part of the chain, but we will care that a block is rooted so that we can compare its weight to the current best block.

If we want to do that, I think it depend on how we store incoming blocks: 
 - if only rooted block are stored (i.e. we store block once they form a branch), then if it is in storage, all is good and we can compare it with the best branch.
 - if we store every block it gets trickier. I think in that case we will have to add some block metadata which tells us if a block is rooted or not so that we can easily tell "the parent is rooted so this block is also rooted", probably with different types of blocks.

src/main/scala/io/iohk/ethereum/domain/BlockchainReader.scala
AurelienRichez: I rewrote it to have the guard, however I refrained from having a `case _ =>` below because I would like to get the exhaustive check to fail clearly in case we modify the trait later on.

src/main/scala/io/iohk/ethereum/domain/branch/Branch.scala
AurelienRichez: I think we will have several types yes (well, best branch and "other" branch). I don't really know what to do with the `EmptyBranch` as I return it when even the genesis is not found (normally that can happen only during tests). I wonder if we should not add the genesis as a parameter to the block storage, to ensure that it is always there.

PR #1074, opened 2021-07-22
[ETCM-912] Magneto gas changes in EXT* codes family

lukasz-golebiewski:
# Description
This PR adds the prerequisites necessary for enabling gas cost changes in the EVM related to EIP-2929.
It also contains changes in the EXT* code family

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/BlockchainConfigForEvm.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/BlockchainConfigForEvm.scala
@@ -6,6 +6,7 @@ import io.iohk.ethereum.vm.BlockchainConfigForEvm.EtcForks.Agharta
 import io.iohk.ethereum.vm.BlockchainConfigForEvm.EtcForks.Atlantis
 import io.iohk.ethereum.vm.BlockchainConfigForEvm.EtcForks.BeforeAtlantis
 import io.iohk.ethereum.vm.BlockchainConfigForEvm.EtcForks.EtcFork
+import io.iohk.ethereum.vm.BlockchainConfigForEvm.EtcForks.Magneto
 import io.iohk.ethereum.vm.BlockchainConfigForEvm.EtcForks.Phoenix
 import io.iohk.ethereum.vm.BlockchainConfigForEvm.EthForks.BeforeByzantium
 import io.iohk.ethereum.vm.BlockchainConfigForEvm.EthForks.Byzantium
@@ -35,13 +36,15 @@ case class BlockchainConfigForEvm(
     aghartaBlockNumber: BigInt,
     petersburgBlockNumber: BigInt,
     phoenixBlockNumber: BigInt,
+    magnetoBlockNumber: BigInt,
     chainId: Byte
 ) {
   def etcForkForBlockNumber(blockNumber: BigInt): EtcFork = blockNumber match {
     case _ if blockNumber < atlantisBlockNumber => BeforeAtlantis
     case _ if blockNumber < aghartaBlockNumber  => Atlantis
     case _ if blockNumber < phoenixBlockNumber  => Agharta
-    case _ if blockNumber >= phoenixBlockNumber => Phoenix
+    case _ if blockNumber < magnetoBlockNumber  => Phoenix
+    case _ if blockNumber >= magnetoBlockNumber => Magneto
   }
 
   def ethForkForBlockNumber(blockNumber: BigInt): BlockchainConfigForEvm.EthForks.Value = blockNumber match {
@@ -82,6 +85,7 @@ object BlockchainConfigForEvm {
       aghartaBlockNumber = forkBlockNumbers.aghartaBlockNumber,
       petersburgBlockNumber = forkBlockNumbers.petersburgBlockNumber,
       phoenixBlockNumber = forkBlockNumbers.phoenixBlockNumber,
+      magnetoBlockNumber = forkBlockNumbers.magnetoBlockNumber,
       chainId = chainId
     )
   }
--- a/src/main/scala/io/iohk/ethereum/vm/BlockchainConfigForEvm.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/BlockchainConfigForEvm.scala
@@ -68,6 +68,8 @@ object BlockchainConfigForEvm {
     val BeforeByzantium, Byzantium, Constantinople, Petersburg, Istanbul, Berlin = Value
   }
 
+  def isEip2929Enabled(etcFork: EtcFork): Boolean = etcFork >= EtcForks.Magneto
+
   def apply(blockchainConfig: BlockchainConfig): BlockchainConfigForEvm = {
     import blockchainConfig._
     BlockchainConfigForEvm(
--- a/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/OpCode.scala
@@ -235,27 +235,6 @@ sealed trait ConstGas { self: OpCode =>
   protected def varGas[W <: WorldStateProxy[W, S], S <: Storage[S]](state: ProgramState[W, S]): BigInt = 0
 }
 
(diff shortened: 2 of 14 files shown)


src/test/scala/io/iohk/ethereum/vm/CreateOpcodeSpec.scala
```diff
@@ -248,6 +248,13 @@ class CreateOpcodeSpec extends AnyWordSpec with Matchers with ScalaCheckProperty
       "leave return buffer empty" in {
         result.stateOut.returnData shouldEqual ByteString.empty
       }
+
+      "add the new contract to accessed_addresses" in {
+        result.world.getGuaranteedAccount(newAccountAddress())
+        val addr = newAccountAddress()
```
strauss-m, line 253: I would have called result.world.getGuaranteedAccount with the val addr instead re-calling newAccountAddress().
For create2, this seems to not be significant, but for create, there is a slight side effect with the nonce being modified.

Unless there is a good reason to call multiple time newAccountAddress()

```
val addr = newAccountAddress()
result.world.getGuaranteedAccount(addr)
result.stateOut.accessedAddresses should contain(addr)

```

lukasz-golebiewski: Good catch!

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -452,28 +454,62 @@ case object CODECOPY extends OpCode(0x39, 3, 0, _.G_verylow) {
...
     val (Seq(_, memOffset, _, size), _) = state.stack.pop(4)
     val memCost = state.config.calcMemCost(state.memory.size, memOffset, size)
     val copyCost = state.config.feeSchedule.G_copy * wordsForBytes(size)
-    memCost + copyCost
+
+    val currentBlockNumber = state.env.blockHeader.number
+    val etcFork = state.config.blockchainConfig.etcForkForBlockNumber(currentBlockNumber)
+    val eip2929Enabled = isEip2929Enabled(etcFork)
+
+    val accessCost = if (eip2929Enabled) {
```
AurelienRichez: Could we factorize this bit of code with `EXTCODESIZE` ?  (well it's a draft so it might already be planned)

src/main/scala/io/iohk/ethereum/vm/ProgramState.scala
```diff
@@ -126,6 +133,12 @@ case class ProgramState[W <: WorldStateProxy[W, S], S <: Storage[S]](
   def revert(data: ByteString): ProgramState[W, S] =
     copy(error = Some(RevertOccurs), returnData = data, halted = true)
 
+  def addAccessedAddress(addr: Address): ProgramState[W, S] =
```
AurelienRichez, line 138: Note: There seems to be two conventions in the same class here for naming (`doSomething`: e.g. `refundGas` vs `withSomething`: e.g. `withAddressToDelete` ). 

I personally slightly prefer `withAccessedAddress` to be clear that the structure is immutable. But I would agree that it also adds another ambiguity about whether the new instance accessed addresses are replaced by the passed value or if the value is appended, so I'm not fixed on that one.

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
lukasz-golebiewski: Yes, this is planned

src/main/scala/io/iohk/ethereum/vm/ProgramState.scala
lukasz-golebiewski: I understand `with*` functions as replacing the old value with the new one in the newly created copy. `add*` functions don't replace, but add a new item to the already existing collection. 

To tell whether the function modifies the structure on which it is called it should be enough to check it's signature. A function which does the modification in place probably doesn't need to return anything (maybe a Boolean stating whether the update was successful?). A pure function without side effects is going to return the same type as the object on which it is called (as it is in this case).

src/main/scala/io/iohk/ethereum/vm/BlockchainConfigForEvm.scala
```diff
@@ -35,13 +36,15 @@ case class BlockchainConfigForEvm(
...
+    magnetoBlockNumber: BigInt,
     chainId: Byte
 ) {
   def etcForkForBlockNumber(blockNumber: BigInt): EtcFork = blockNumber match {
     case _ if blockNumber < atlantisBlockNumber => BeforeAtlantis
     case _ if blockNumber < aghartaBlockNumber  => Atlantis
     case _ if blockNumber < phoenixBlockNumber  => Agharta
-    case _ if blockNumber >= phoenixBlockNumber => Phoenix
+    case _ if blockNumber < magnetoBlockNumber  => Phoenix
+    case _ if blockNumber >= magnetoBlockNumber => Magneto
```
dzajkowski, line 47: do you mind adding a task for creating the same mapping for eth berlin?

lukasz-golebiewski: It is done

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -184,27 +184,78 @@ abstract class OpCode(val code: Byte, val delta: Int, val alpha: Int, val constG
...
+  )
+
   protected def varGas[W <: WorldStateProxy[W, S], S <: Storage[S]](state: ProgramState[W, S]): BigInt
 
   protected def exec[W <: WorldStateProxy[W, S], S <: Storage[S]](state: ProgramState[W, S]): ProgramState[W, S]
 
   protected def availableInContext[W <: WorldStateProxy[W, S], S <: Storage[S]]: ProgramState[W, S] => Boolean = _ =>
     true
+
+  protected def isEip2929Enabled(etcFork: EtcFork): Boolean = etcFork >= EtcForks.Magneto
```
dzajkowski: wouldn't this fit in `object`?

src/main/scala/io/iohk/ethereum/vm/OpCode.scala
```diff
@@ -184,21 +184,51 @@ abstract class OpCode(val code: Byte, val delta: Int, val alpha: Int, val constG
...
+}
+
+trait AddrAccessGas { self: OpCode =>
+
+  private def coldGasFn: FeeSchedule => BigInt = _.G_cold_account_access
+  private def warmGasFn: FeeSchedule => BigInt = _.G_warm_storage_read
+
+  override protected def constGas[W <: WorldStateProxy[W, S], S <: Storage[S]](state: ProgramState[W, S]): BigInt = {
```
AurelienRichez: I find it confusing that `constGas` depends on the program state because now it is not "const" anymore (granted before it was not completely constant but it depended only on the current fee schedule). It blurs the line between what is considered constant gas cost and variable gas cost. And now it is not clear for someone wanting to change the gas costs where the change should be made. 

I see that one problem is that `EXTCODECOPY` already has a varGas defined so we would need some kind of special code there. We could make `AddrAccessGas` use `varGas` instead:  in `EXTCODECOPY`, with something like ` + super.varGas(state)` inside `varGas` implementation (I admit using `super` is not really conventional, I don't have any simpler idea at the moment).

What do you think ? There might be another reason I don't see to use constGas instead of varGas.

lukasz-golebiewski: I assumed that the difference between the two was that `constGas` only picks an appropriate value from the fee schedule and `varGas` additionally needs to perform some calculations which depend on the accessed data itself (e.g. it's size). 

Would it be enough to add this ^ to the scaladoc?

lukasz-golebiewski: We could also rename `constGas` to `baseGas` if this helps

AurelienRichez (approved):

strauss-m (approved):

PR #1077, opened 2021-07-26
[ETCM-1044] add best block hash into storage

AurelienRichez:
# Description

This was originally from a spike (ETCM-1051), but I ended up implementing ETCM-1044 while poking at the code. 

Make `getBestBlock` use `getBlockByHash`.

For this I added the best block hash along with the bestBlockNumber, so that when we want to get the best block number we can just get it from its hash and not from the block number mapping.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
+++ b/src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
@@ -1,12 +1,19 @@
 package io.iohk.ethereum.db.storage
 
+import akka.util.ByteString
+import boopickle.Default.Unpickle
+import boopickle.Pickler
+import boopickle.DefaultBasic._
+
 import java.math.BigInteger
-
 import scala.collection.immutable.ArraySeq
-
 import io.iohk.ethereum.db.dataSource.DataSource
 import io.iohk.ethereum.db.dataSource.DataSourceBatchUpdate
 import io.iohk.ethereum.db.storage.AppStateStorage._
+import io.iohk.ethereum.domain.appstate.BestBlockInfo
+import io.iohk.ethereum.utils.ByteUtils.{byteSequenceToBuffer, compactPickledBytes}
+import io.iohk.ethereum.utils.Hex
+import io.iohk.ethereum.utils.Picklers._
 
 /** This class is used to store app state variables
   *   Key: see AppStateStorage.Keys
@@ -26,6 +33,16 @@ class AppStateStorage(val dataSource: DataSource) extends TransactionalKeyValueS
 
   def getBestBlockNumber(): BigInt =
     getBigInt(Keys.BestBlockNumber)
+
+  def getBestBlockData(): BestBlockInfo =
+    BestBlockInfo( // FIXME default value for hash ?
+      get(Keys.BestBlockHash).map(v => ByteString(Hex.decode(v))).getOrElse(ByteString.empty),
+      getBigInt(Keys.BestBlockNumber)
+    )
+
+  def putBestBlockData(b: BestBlockInfo): DataSourceBatchUpdate =
+    put(Keys.BestBlockNumber, b.number.toString)
+      .and(put(Keys.BestBlockHash, Hex.toHexString(b.hash.toArray)))
 
   def putBestBlockNumber(bestBlockNumber: BigInt): DataSourceBatchUpdate =
     put(Keys.BestBlockNumber, bestBlockNumber.toString)
@@ -72,9 +89,17 @@ object AppStateStorage {
 
   object Keys {
     val BestBlockNumber = "BestBlockNumber"
+    val BestBlockHash = "BestBlockHash"
     val FastSyncDone = "FastSyncDone"
     val EstimatedHighestBlock = "EstimatedHighestBlock"
     val SyncStartingBlock = "SyncStartingBlock"
     val LatestCheckpointBlockNumber = "LatestCheckpointBlockNumber"
   }
+
+  implicit private val bestBlockDataPickler: Pickler[BestBlockInfo] = generatePickler[BestBlockInfo]
+  private val bestBlockDataSerializer = (bestBlockData: BestBlockInfo) =>
+    compactPickledBytes(Pickle.intoBytes(bestBlockData))
+  private val bestBlockDataDeserializer =
+    (byteSequenceToBuffer _).andThen(Unpickle[BestBlockInfo].fromBytes)
+
 }
--- a/src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
+++ b/src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
@@ -6,18 +6,11 @@ import akka.util.ByteString
 
 import scala.collection.immutable.ArraySeq
 
-import boopickle.Default.Unpickle
-import boopickle.DefaultBasic._
-import boopickle.Pickler
-
 import io.iohk.ethereum.db.dataSource.DataSource
 import io.iohk.ethereum.db.dataSource.DataSourceBatchUpdate
 import io.iohk.ethereum.db.storage.AppStateStorage._
 import io.iohk.ethereum.domain.appstate.BestBlockInfo
-import io.iohk.ethereum.utils.ByteUtils.byteSequenceToBuffer
-import io.iohk.ethereum.utils.ByteUtils.compactPickledBytes
 import io.iohk.ethereum.utils.Hex
-import io.iohk.ethereum.utils.Picklers._
 
 /** This class is used to store app state variables
   *   Key: see AppStateStorage.Keys
--- a/src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
+++ b/src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
@@ -1,17 +1,21 @@
 package io.iohk.ethereum.db.storage
 
-import akka.util.ByteString
-import boopickle.Default.Unpickle
-import boopickle.Pickler
-import boopickle.DefaultBasic._
-
 import java.math.BigInteger
+
+import akka.util.ByteString
+
 import scala.collection.immutable.ArraySeq
+
+import boopickle.Default.Unpickle
+import boopickle.DefaultBasic._
+import boopickle.Pickler
+
 import io.iohk.ethereum.db.dataSource.DataSource
 import io.iohk.ethereum.db.dataSource.DataSourceBatchUpdate
 import io.iohk.ethereum.db.storage.AppStateStorage._
 import io.iohk.ethereum.domain.appstate.BestBlockInfo
-import io.iohk.ethereum.utils.ByteUtils.{byteSequenceToBuffer, compactPickledBytes}
+import io.iohk.ethereum.utils.ByteUtils.byteSequenceToBuffer
+import io.iohk.ethereum.utils.ByteUtils.compactPickledBytes
 import io.iohk.ethereum.utils.Hex
 import io.iohk.ethereum.utils.Picklers._
 
@@ -34,7 +38,7 @@ class AppStateStorage(val dataSource: DataSource) extends TransactionalKeyValueS
   def getBestBlockNumber(): BigInt =
     getBigInt(Keys.BestBlockNumber)
 
-  def getBestBlockData(): BestBlockInfo =
+  def getBestBlockInfo(): BestBlockInfo =
     BestBlockInfo( // FIXME default value for hash ?
       get(Keys.BestBlockHash).map(v => ByteString(Hex.decode(v))).getOrElse(ByteString.empty),
       getBigInt(Keys.BestBlockNumber)
@@ -95,11 +99,5 @@ object AppStateStorage {
     val SyncStartingBlock = "SyncStartingBlock"
     val LatestCheckpointBlockNumber = "LatestCheckpointBlockNumber"
   }
-
-  implicit private val bestBlockDataPickler: Pickler[BestBlockInfo] = generatePickler[BestBlockInfo]
-  private val bestBlockDataSerializer = (bestBlockData: BestBlockInfo) =>
-    compactPickledBytes(Pickle.intoBytes(bestBlockData))
-  private val bestBlockDataDeserializer =
-    (byteSequenceToBuffer _).andThen(Unpickle[BestBlockInfo].fromBytes)
 
 }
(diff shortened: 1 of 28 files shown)


src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
```diff
@@ -27,6 +38,16 @@ class AppStateStorage(val dataSource: DataSource) extends TransactionalKeyValueS
   def getBestBlockNumber(): BigInt =
     getBigInt(Keys.BestBlockNumber)
 
+  def getBestBlockInfo(): BestBlockInfo =
+    BestBlockInfo( // FIXME default value for hash ?
+      get(Keys.BestBlockHash).map(v => ByteString(Hex.decode(v))).getOrElse(ByteString.empty),
```
AurelienRichez, line 36: I don't really know if there is a decent value to return when the hash is not there. This is a case of "should not happen" because we always have a genesis when running the application. But it does happen during tests.

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -188,7 +200,8 @@ class BlockchainImpl(
         removeBlockNumberMapping(block.number)
       else blockNumberMappingStorage.emptyBatchUpdate
 
-    val newBestBlockNumber: BigInt = (bestBlockNumber - 1).max(0)
+    val potientialNewBestBlockNumber: BigInt = (block.number - 1).max(0)
+    val potientialNewBestBlockHash: ByteString = block.header.parentHash
```
leo-bogastry: There's a typo in `potential` in both variables! Should be `potentialNewBestBlockNumber` and `potentialNewBestBlockHash `

leo-bogastry: Shouldn't you use `potentialNewBestBlockNumber` in line 255 as well? And get rid of `val bestBlockNumber`?

AurelienRichez: I don't think so. It was already `bestBlockNumber` before and not `NewBestBlockNumber`.

I think my changes here deserve some explanations: 
By reading the code and how it is called, I saw that it actually assumes that the block we are deleting is the best block because it said that `newBestBlockNumber` is now `bestBlockNumber - 1` (so `bestBlockNumber == block.number`). I tried to make it clearer and less error prone by saying that the new best block number is actually `block.number - 1`.

Fixed the typo

leo-bogastry: It's the fact that `bestBlockNumber` and `potentialNewBestBlockNumber` don't derive from each other anymore, makes me a bit uneasy. I think you ended up by adding two versions of getting the current best block

src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
strauss-m: Is it because the tests setting are incomplete (missing genesis, empty storage, other ?), and is it considered ok ?

As for the default value, I would eventually go the Option way.

But it raises the issue of how to handle the default value (be it the actual empty ByteString or a None). If every piece of code handling the potential missing hash is unable to cope with it, it may be worth making it mandatory to have a proper genesis each time, and raise a hard-error at a single point during the getBestBlockData()

src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
```diff
@@ -27,6 +34,16 @@ class AppStateStorage(val dataSource: DataSource) extends TransactionalKeyValueS
   def getBestBlockNumber(): BigInt =
     getBigInt(Keys.BestBlockNumber)
 
+  def getBestBlockData(): BestBlockInfo =
+    BestBlockInfo( // FIXME default value for hash ?
+      get(Keys.BestBlockHash).map(v => ByteString(Hex.decode(v))).getOrElse(ByteString.empty),
+      getBigInt(Keys.BestBlockNumber)
+    )
+
+  def putBestBlockData(b: BestBlockInfo): DataSourceBatchUpdate =
```
strauss-m: I wonder if the method would not have been better named: putBestBlock**Info** instead of **Data** to keep it coherent with the parameter type

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerEthSpec.scala
```diff
@@ -413,10 +413,10 @@ class JsonRpcControllerEthSpec
...
 
   it should "eth_gasPrice" in new JsonRpcControllerFixture {
-    blockchainWriter
-      .storeBlock(Block(Fixtures.Blocks.Block3125369.header.copy(number = 42), Fixtures.Blocks.Block3125369.body))
-      .commit()
-    blockchain.saveBestKnownBlocks(ByteString.empty, 42)
+    private val block: Block =
+      Block(Fixtures.Blocks.Block3125369.header.copy(number = 42), Fixtures.Blocks.Block3125369.body)
+    blockchainWriter.storeBlock(block).commit()
+    blockchain.saveBestKnownBlocks(block.hash, 42)
```
strauss-m: Since the block is now available, replace 42 by block.number.

Actually, I wonder if we shouldn't provide a blockchain.saveBestKnownBlock**Info**(**block: Block**) method instead of / side by side with the saveBestKnowBlocks(hash, number) (and eventually rename it to saveBestKnowBlockInfo too, because as it is, I don't see it related to multiple block*s*, but only to on block*Info*)
I raise this as most of the time, I see the call is done by extracting values from a given block, and at the blockchain level, I don't see any evil with directly handling a block (no added dependency)

src/test/scala/io/iohk/ethereum/domain/BlockchainReaderSpec.scala
```diff
@@ -0,0 +1,19 @@
...
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import io.iohk.ethereum.Fixtures
+import io.iohk.ethereum.blockchain.sync.EphemBlockchainTestSetup
+
+class BlockchainReaderSpec extends AnyFlatSpec with Matchers {
+
+  "BlockchainReader" should "be able to get the best block after it was stored by BlockchainWriter" in new EphemBlockchainTestSetup {
+    val validBlock = Fixtures.Blocks.ValidBlock.block
```
strauss-m: proposal: it looks like this could be transformed into a more generic property based test (replace the fixture.block with a generated one), unless it's hard to generate a valid block usable by the blockwriter.

src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
AurelienRichez: > Is it because the tests setting are incomplete (missing genesis, empty storage, other ?), and is it considered ok ?

I think the reason is that it is not strictly necessary to actually have a genesis in storage for some test because of the mocking, and setting up a real genesis / making it mandatory in the design was not done as a consequence. In this particular case as 0 was a pretty decent default value (it is the genesis number after all) it was not really visible. But now storing the genesis hash makes the problem more apparent. 

This inconsistency does not really happen outside of tests because the application loads genesis on startup, so it's "mostly" ok. 

> As for the default value, I would eventually go the Option way.

I would personally prefer to have a design that makes this inconsistency impossible as you said to be sure that we have a genesis (probably having the Storage take some information about the genesis in its constructor). I'm not sure about the impact this would have though as for instance ETS  requires us to change the genesis at runtime.

Returning an option would mean that we would have to handle the `None` case everywhere, which really only means the the application is in an inconsistent state (In which case the application should probably crash).

src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
AurelienRichez: Completely agree. This is an oversight

src/test/scala/io/iohk/ethereum/jsonrpc/JsonRpcControllerEthSpec.scala
AurelienRichez: There is already a `save` function that does something similar. I'm not sure why some test use it or not so I kept the same structure. But I think once we have a proper consensus and a central point where the best chain is decided it will be clearer what to do about that.

src/test/scala/io/iohk/ethereum/domain/BlockchainReaderSpec.scala
AurelienRichez: It could yes. I'm not really sure what would be the added value of a property test there though as is it quite simple (just storing and retrieving and no real logic involved). Maybe we should have a rule of thumb on what would need a property test.

strauss-m (approved):

src/test/scala/io/iohk/ethereum/domain/BlockchainReaderSpec.scala
jvdp: Could any of the randomized data (or size of the data) impact the property you want to test? Maybe edge cases?

src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
jvdp: Would it help to have to genesis block available by other means (eg. passed in as parameter) instead of from storage? That might actually be the more straightforward approach: genesis data is loaded from config when the node starts and is then written to the database. So loading it from the database is a bit of a detour.

jvdp:
Do we still need to store best block number when we could fetch the header by hash and read the number from there?

src/main/scala/io/iohk/ethereum/db/storage/AppStateStorage.scala
AurelienRichez: Yes, what I think about is having some information about genesis passed as a parameter. Not only for the app state storage, but also in other storage such as bodies and headers. This way we can ensure that the genesis is there and remove some options in the code (like in `getBestBlock`).

I think that doing it in this PR would be too big of a change even if we only do the AppStateStorage part. I created this followup ticket to detail what would need to be done : https://jira.iohk.io/browse/ETCM-1090.

src/test/scala/io/iohk/ethereum/domain/BlockchainReaderSpec.scala
AurelienRichez: I added a property test instead

AurelienRichez:
> Do we still need to store best block number when we could fetch the header by hash and read the number from there?

Yes, we should be able to get rid of the block number in the app state storage. It is used in a few places right now so it cannot be done in one step. I'm still not sure that we should do it though, but I don't see any reason to not favor consistency right now so I'll create a ticket for this.

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
AurelienRichez: In fact this part of the function will be removed after we merge #1092 so I think we should wait for #1092, fix the conflicts in this PR and we will be OK.

leo-bogastry: 👍

leo-bogastry: #1092 is merged!

leo-bogastry (approved):

src/main/scala/io/iohk/ethereum/nodebuilder/StdNode.scala
```diff
@@ -132,6 +134,24 @@ abstract class BaseNode extends Node {
     tryAndLogFailure(() => Metrics.get().close())
     tryAndLogFailure(() => storagesInstance.dataSource.close())
   }
+
+  def fixDatabase(): Unit = {
```
AurelienRichez, line 140: I added this to be able to maintain compatibility.
The problem is that when we start with an existing database (say staging), we don't have the best block hash stored. This ensures on startup that the database is consistent. 
@LeonorLunatech, @strauss-m, do you think it is an acceptable compromise ?

leo-bogastry: 👍 Let as know how the tests went

AurelienRichez: It worked as expected and staging is mining normally 👍

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -187,8 +200,8 @@ class BlockchainImpl(
       into the case of having an incomplete best block and so an inconsistent db
      */
     val bestBlockNumberUpdates =
-      if (appStateStorage.getBestBlockNumber() > newBestBlockNumber)
-        appStateStorage.putBestBlockNumber(newBestBlockNumber)
+      if (appStateStorage.getBestBlockNumber() > potentialNewBestBlockNumber)
```
AnastasiiaL, line 203: do we want to keep this condition?
what if the `getBestBlockNumber` returns a number smaller than the potentialNewBestBlockNumber? we won't update it?

src/main/scala/io/iohk/ethereum/domain/appstate/BestBlockInfo.scala
```diff
@@ -0,0 +1,5 @@
+package io.iohk.ethereum.domain.appstate
+
+import akka.util.ByteString
+
+case class BestBlockInfo(hash: ByteString, number: BigInt)
```
leo-bogastry: Maybe this could be called just `BlockInfo`? Just to make it more generic. I think I need to use this in ETCM-1081

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
AurelienRichez: As currently mantis is storing only the best chain, the guarantee is that we have a path from genesis up to the best block.

So the logic as I understand it is as follow: 
  - if we are deleting a block with a number lower or equal to the current best block, then we change the current best block because otherwise there would be a "hole" in the chain up to the current best block, so the best block can only be the parent of the block we are deleting.
  - if we are deleting a block with a number higher than what is considered the current best block, then it does not affect the current best block because otherwise it would mean extending the chain while we are just deleting a block (also we are not sure that we have a path to a random block above the best block number so this ensure consistency). I don't really know when this scenario happens to be honest, but I guess it might happen if we are deleting a series of block out of order.
  
So as of now I think the implementation still makes sense. But indeed it can not be used as is when we start storing more than the current best chain.

AnastasiiaL (approved):

PR #1078, opened 2021-07-27
[ETCM-1058] Replace block import by consensus

leo-bogastry:
# Description

Replace class BlockImport by a Consensus class, reusing current code. 

Added a new trait Consensus and class ConsensusImpl
ConsensusImpl reuses existing code (with some refactoring)
The refactoring is done is on separate commits to make it easier to evaluate

# Testing
Regular Sync was tested with mainnet
Regular Sync / mining still to be tested in staging


src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
```diff
@@ -0,0 +1,347 @@
...
+    Task
+      .evalOnce(importBlockToTop(block, currentBestBlock.header.number, currentWeight))
+      .executeOn(blockExecutionScheduler)
+
+  /** *
+    * Open for discussion: this is code that was in BlockImport. Even thought is tested (before) that only one block
+    * is being added to the blockchain, this code assumes that the import may be of more than one block (a branch)
+    * and because it is assuming that it may be dealing with a branch the code to handle errors assumes several scenarios.
+    * Is there is reason for this over-complication?
+    */
```
leo-bogastry, line 162: I added this comment because I think this deserves a discussion! What are your thoughts?

src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
```diff
@@ -1,364 +1,344 @@
...
-    newBranch: List[Block],
-    weights: List[ChainWeight]
-) extends BlockImportResult
-
-case class BlockImportFailed(error: String) extends BlockImportResult
-
-case class BlockImportFailedDueToMissingNode(reason: MissingNodeException) extends BlockImportResult
-
-case object UnknownParent extends BlockImportResult
+//package io.iohk.ethereum.ledger
```
dzajkowski: note: if the differ could pick this up as a rename it would be easy to see what changed between the files. as is it treats the file as changed and 'ConsensusImpl' as new.

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
AurelienRichez: I don't know the history to be able to know why it was like that. But in our case we will probably want to import several blocks in one run, so I think we might run into a similar scenario where only a part of the branch executes. 
But I agree that in that case it does not really make sense as we check that the block is the direct child of the current best block 🤔.
I think we can keep it as it is right now, as we know we will rewrite this part to fit our design anyway.

src/main/scala/io/iohk/ethereum/ledger/BlockImport.scala
leo-bogastry: The refactoring of that code is in separate commits

src/main/scala/io/iohk/ethereum/testmode/TestModeComponentsProvider.scala
```diff
@@ -4,35 +4,34 @@ import akka.util.ByteString
...
     evmCodeStorage: EvmCodeStorage,
-    syncConfig: SyncConfig,
     validationExecutionContext: Scheduler,
     miningConfig: MiningConfig,
     vm: VMImpl,
     node: TestNode
 ) {
 
-  def blockImport(
+  def evaluateBranchBlock(
```
AurelienRichez: `getConsensus` or `consensus` ? 
As it is the usage is quite confusing: 
```
          .evaluateBranchBlock(preimageCache)
          .evaluateBranchBlock(value)
```

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
dzajkowski:
```scala
    val executionResult = for {
      topBlock <- blockQueue.enqueueBlock(block, bestBlockNumber)
      topBlocks = blockQueue.getBranch(topBlock.hash, dequeue = true)
      (executed, errors) = blockExecution.executeAndValidateBlocks(topBlocks, currentWeight)
    } yield (executed, errors, topBlocks)
```

pretty sure the assumption is that this line:
`blockQueue.getBranch(topBlock.hash, dequeue = true)`
can return a branch that has not been stored or executed.

this is the internal caching that BlockImport is doing.

src/main/scala/io/iohk/ethereum/testmode/TestModeComponentsProvider.scala
leo-bogastry: You are totally right. Fixed

AurelienRichez (approved):

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
```diff
@@ -1,81 +1,131 @@
...
-    *         - [[io.iohk.ethereum.ledger.DuplicateBlock]] - block already exists either in the main chain or in the queue
-    *         - [[io.iohk.ethereum.ledger.BlockImportFailed]] - block failed to execute (when importing to top or reorganising the chain)
+    *   - [[BlockImportedToTop]] - if the block was added as the new best block
+    *   - [[BlockEnqueued]]      - block is stored in the [[io.iohk.ethereum.ledger.BlockQueue]]
+    *   - [[ChainReorganised]]   - a better new branch was found causing chain reorganisation
+    *   - [[DuplicateBlock]]     - block already exists either in the main chain or in the queue
+    *   - [[BlockImportFailed]]  - block failed to execute (when importing to top or reorganising the chain)
     */
-  def importBlock(
+  override def evaluateBranchBlock(
```
jvdp, line 66: Why is it called `evaluate`, the term for this elsewhere is `execute`, right?

Also, the comments and result type still talk of "importing", while we should probably drop that term everywhere?

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
```diff
@@ -1,81 +1,131 @@
...
+      bestBlock.header.hashAsHexString
+    )
+    Task.now(
+      BlockImportFailed(
+        s"Couldn't get total difficulty for current best block with hash: ${bestBlock.header.hashAsHexString}"
+      )
+    )
+  }
+
+  private def returnNoBestBlock(): Task[BlockImportFailed] = {
```
jvdp, line 124: Why is this split out into a method? Doesn't seem to do much.

(If anything should be split out: `BlockImportFailed("Couldn't find the current best block")` could be a `val`.)

Perhaps we can reuse these error descriptions as log messages as well?

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
```diff
@@ -1,81 +1,131 @@
...
+    } else {
+      reorganiseOrEnqueue(block, bestBlock, weight)
+    }
+    importResult.foreach(measureBlockMetrics)
+    importResult
+  }
+
+  private def doBlockPreValidation(block: Block)(implicit
+      blockchainConfig: BlockchainConfig
+  ): Task[Either[ValidationBeforeExecError, BlockExecutionSuccess]] =
```
jvdp, line 102: So this "pre-validation" code was duplicated, right? That's nice to refactor 👍

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
```diff
@@ -1,81 +1,131 @@
...
+    Task
+      .evalOnce {
+        val validationResult = blockValidation.validateBlockBeforeExecution(block)
+        validationResult.left.foreach { error =>
```
jvdp: While at it this can be improved using `tap` from cats-effect.

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
```diff
@@ -1,81 +1,131 @@
...
+    *   - [[BlockEnqueued]]      - block is stored in the [[io.iohk.ethereum.ledger.BlockQueue]]
+    *   - [[ChainReorganised]]   - a better new branch was found causing chain reorganisation
+    *   - [[DuplicateBlock]]     - block already exists either in the main chain or in the queue
+    *   - [[BlockImportFailed]]  - block failed to execute (when importing to top or reorganising the chain)
     */
-  def importBlock(
+  override def evaluateBranchBlock(
       block: Block
   )(implicit blockExecutionScheduler: Scheduler, blockchainConfig: BlockchainConfig): Task[BlockImportResult] =
     blockchainReader.getBestBlock() match {
```
jvdp, line 69: I wonder if we can flatten this `match` structure somehow to make this clearer? Although moving all this code out already helps a bunch.

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
```diff
@@ -1,81 +1,131 @@
...
+import io.iohk.ethereum.ledger.BlockQueue
 import io.iohk.ethereum.ledger.BlockQueue.Leaf
+import io.iohk.ethereum.ledger.BlockValidation
 import io.iohk.ethereum.mpt.MerklePatriciaTrie.MissingNodeException
 import io.iohk.ethereum.utils.BlockchainConfig
 import io.iohk.ethereum.utils.ByteStringUtils
 import io.iohk.ethereum.utils.Logger
 
-class BlockImport(
+class ConsensusImpl(
```
jvdp, line 43: Does it need to be split into `Consensus` and `ConsensusImpl`?

src/test/scala/io/iohk/ethereum/consensus/ConsensusSpec.scala
```diff
@@ -99,7 +111,39 @@ class BlockImportSpec extends AnyFlatSpec with Matchers with ScalaFutures {
 
     (blockQueue.removeSubtree _).expects(*)
 
-    whenReady(blockImport.importBlock(block).runToFuture)(_ shouldBe a[BlockImportFailed])
+    whenReady(consensus.evaluateBranchBlock(block).runToFuture)(
+      _ shouldBe BlockImportFailed("MPTError(io.iohk.ethereum.mpt.MerklePatriciaTrie$MPTException: Invalid Node)")
+    )
+  }
+
+  it should "handle no best block available error when importing to top" in new ImportBlockTestSetup {
```
jvdp, line 119: Did you write these two from scratch?

jvdp (approved):
Do have some questions but wouldn't want to block further work on this.

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
leo-bogastry: One is a trait the other a class. It will help with testing

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
leo-bogastry: `The term elsewhere is execute, right`? No. Where did you get that idea from? Actually me and Aurelién are struggling to give this method a proper name, because it can have several outcomes. I removed the name `import` exactly because as said in the description it `Tries to import`

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
leo-bogastry: I agree with you, still does not look great. With the follow-up ticket ETCM-1069 this will improve :)

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
leo-bogastry: Yes, whatever happened a pre-validation was being done

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
leo-bogastry: It was mainly to make the main method `evaluateBranchBlock` more readable, by moving the detail of logging the error and creating the actual error description

src/test/scala/io/iohk/ethereum/consensus/ConsensusSpec.scala
leo-bogastry: I used the existing tests as a blue print

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
leo-bogastry: Thanks! Done

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
jvdp: There is the `BlockExecution` class for example. So this could be considered "branch execution"?

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
jvdp: So I would argue it is less readable now since those methods actually do so little.

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
jvdp: Does it though? It didn't get in the way with `BlockImport` before.

This way does make code navigation a lot more annoying.

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
leo-bogastry: It is not guaranteed that there will be a block or branch execution, the block may end up by just being added to the BlockQueue

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
leo-bogastry: Just look at the `private[ledger] val blockExecution: BlockExecution` (previous line 28). That was being accessed directly in the tests. 
Hopefully we can improve the tests much more in the future

PR #1079, opened 2021-07-27
[KAIZEN] Turn some warnings into errors

lukasz-golebiewski:


build.sbt
```diff
@@ -63,12 +63,15 @@ def commonSettings(projectName: String): Seq[sbt.Def.Setting[_]] = Seq(
     "-unchecked",
     "-deprecation",
     "-feature",
-    //    "-Xfatal-warnings", // disabled until unused are removed
+    // https://www.scala-lang.org/2021/01/12/configuring-and-suppressing-warnings.html
+    // cat={warning-name}:ws prints a summary with the number of warnings of the given type
+    // any:e turns all remaining warnings into errors
+    "-Wconf:cat=deprecation:ws,cat=lint-package-object-classes:ws,cat=unused:ws,cat=lint-infer-any:ws,cat=lint-byname-implicit:ws,cat=other-match-analysis:ws,any:e",
```
lukasz-golebiewski: We should aim to gradually make the list of `cat={}:ws` thinner

AurelienRichez: I think it's great to reduce the noise during dev. I think a way to make it display the full warning easily would be useful though. Something like an env var (but maybe there is a more elegant way):

```suggestion
    if (sys.env.get("MANTIS_FULL_WARNS").contains("true"))
      ""
    else
      "-Wconf:cat=deprecation:ws,cat=lint-package-object-classes:ws,cat=unused:ws,cat=lint-infer-any:ws,cat=lint-byname-implicit:ws,cat=other-match-analysis:ws,any:e",
```

usage: `MANTIS_FULL_WARNS=true sbt compile`

lukasz-golebiewski: Sounds good, but I would be explicit and put `"-Wconf:any:w"` in there

AurelienRichez (approved):

dzajkowski (approved):

lukasz-golebiewski:
Closing in favor of https://github.com/input-output-hk/mantis/pull/1081

PR #1082, opened 2021-07-29
[ETCM-1060] rewrite ommers validation to avoid using getBlockByNumber

AurelienRichez:
# Description

This PR rewrites `getNBlocksBack` in OmmersValidation so that it does not use `getBlockByNumber`.

 - It reduces our dependency on the block number mapping storage: The other parts that are using `getBlockByNumber` are out of the core flow of mantis (json rpc calls). This means that it opens a path to be able to switch the best branch just by changing a hash in the storage, and having the block number mapping built somewhere else. We can insert new blocks without rebuilding the index each time. 
 - We can validate ommers for any block without considering if it is on top of the best chain or not. In the end it will allow to have only one implementation instead of treating alternative branches are a special case (currently we have a different way of retrieving blocks if they are in the blockQueue).

### DIFF
--- a/src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
@@ -33,8 +33,14 @@ trait OmmersValidator {
     val bestBranch = blockchainReader.getBestBranch()
     val getNBlocksBack: (ByteString, Int) => List[Block] =
       (_, n) =>
-        ((blockNumber - n) until blockNumber).toList
-          .flatMap(nb => blockchainReader.getBlockByNumber(bestBranch, nb))
+        Iterator
+          .iterate(blockchainReader.getBlockByHash(parentHash))(
+            _.flatMap(block => blockchainReader.getBlockByHash(block.header.parentHash))
+          )
+          .take(n)
+          .collect { case Some(block) => block }
+          .toList
+          .reverse
 
     validate(parentHash, blockNumber, ommers, getBlockHeaderByHash, getNBlocksBack)
   }
--- a/src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
@@ -32,9 +32,9 @@ trait OmmersValidator {
     val getBlockHeaderByHash: ByteString => Option[BlockHeader] = blockchainReader.getBlockHeaderByHash
     val bestBranch = blockchainReader.getBestBranch()
     val getNBlocksBack: (ByteString, Int) => List[Block] =
-      (_, n) =>
+      (tailBlockHash, n) =>
         Iterator
-          .iterate(blockchainReader.getBlockByHash(parentHash))(
+          .iterate(blockchainReader.getBlockByHash(tailBlockHash))(
             _.flatMap(block => blockchainReader.getBlockByHash(block.header.parentHash))
           )
           .take(n)
--- a/src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
@@ -30,15 +30,15 @@ trait OmmersValidator {
   )(implicit blockchainConfig: BlockchainConfig): Either[OmmersError, OmmersValid] = {
 
     val getBlockHeaderByHash: ByteString => Option[BlockHeader] = blockchainReader.getBlockHeaderByHash
-    val bestBranch = blockchainReader.getBestBranch()
     val getNBlocksBack: (ByteString, Int) => List[Block] =
       (tailBlockHash, n) =>
         Iterator
           .iterate(blockchainReader.getBlockByHash(tailBlockHash))(
-            _.flatMap(block => blockchainReader.getBlockByHash(block.header.parentHash))
+            _.filter(_.number > 0) // avoid trying to fetch parent of genesis
+              .flatMap(block => blockchainReader.getBlockByHash(block.header.parentHash))
           )
-          .take(n)
           .collect { case Some(block) => block }
+          .take(n)
           .toList
           .reverse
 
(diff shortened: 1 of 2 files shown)


leo-bogastry (approved):

src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
```diff
@@ -30,11 +30,16 @@ trait OmmersValidator {
...
     val getNBlocksBack: (ByteString, Int) => List[Block] =
       (_, n) =>
-        ((blockNumber - n) until blockNumber).toList
-          .flatMap(nb => bestBranch.getBlockByNumber(nb))
+        Iterator
+          .iterate(blockchainReader.getBlockByHash(parentHash))(
+            _.flatMap(block => blockchainReader.getBlockByHash(block.header.parentHash))
+          )
+          .take(n)
+          .collect { case Some(block) => block }
```
strauss-m, line 40: Is the consistency of the chain checked elsewhere ?
In case of getBlockByHash() returning a None, we may end up with shorter than expected block list, but it doesn't seem to be checked (within the StdOmmerValidator at least).

src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
```diff
@@ -30,11 +30,16 @@ trait OmmersValidator {
...
+          .toList
```
strauss-m, line 42: The refactor doesn't change this point, so consider it an open question: do we already have an idea of the number of lists and their size, is there a memory risk somehow ?

src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
```diff
@@ -30,11 +30,16 @@ trait OmmersValidator {
   )(implicit blockchainConfig: BlockchainConfig): Either[OmmersError, OmmersValid] = {
 
     val getBlockHeaderByHash: ByteString => Option[BlockHeader] = blockchainReader.getBlockHeaderByHash
-    val bestBranch = blockchainReader.getBestBranch()
     val getNBlocksBack: (ByteString, Int) => List[Block] =
       (_, n) =>
```
strauss-m: Actually, the first parameter should not be ignored, especially now that we are using hashes in the blocks getter function.
The end result should be the same, as the methods calling getNBBlockBack() are passing the parentHash as parameter.

Proposal pseudo code:

```
    val getNBlocksBack: (ByteString, Int) => List[Block] =
      (tailBlockHash, n) =>
        Iterator
          .iterate(blockchainReader.getBlockByHash(tailBlockHash))(
            _.flatMap(block => blockchainReader.getBlockByHash(block.header.parentHash))
          )
          .take(n)
          .collect { case Some(block) => block }
          .toList
          .reverse

```

src/main/scala/io/iohk/ethereum/ledger/BlockValidation.scala
```diff
@@ -39,15 +39,17 @@ class BlockValidation(
...
+        case Some(highestBlockInStorage) =>
           // We already have |block +: queuedBlocks|
           val remaining = n - queuedBlocks.length - 1
-
-          val numbers = (block.header.number - remaining) until block.header.number
-          val bestBranch = blockchainReader.getBestBranch()
-          val blocks =
-            (numbers.toList.flatMap(nb => bestBranch.getBlockByNumber(nb)) :+ block) ::: queuedBlocks
-          blocks
+          val remainingBlocks = Iterator
```
strauss-m, line 45: same remarks as in OmmersValidation

src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
AurelienRichez: Currently, the assumption is that only the canonical chain is stored. So we should have a parent (otherwise it means that the app is in an inconsistent state)

src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
AurelienRichez: Yes, the list is actually bounded to 6 elements. So there should not be a memory problem.
We want to check that a given block was not already used as an ommer in the parents blocks, and an ommer is only valid for 6 generations, so don't have to go back further than that.

src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
AurelienRichez: Indeed. If the result is the same, what do you think about removing the parameter as it essentially complexify the interface needlessly ?

src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
strauss-m: _Maybe to ease the refactor effort, and catch problem as soon as possible, it could be beneficial to add locally a check that raise something strong enough to make it visible, something that would manifest itself with failed tests for exemple (a log.error could just be ignored)_

EDIT: on second thought, it would be a bad idea if this piece of code is called for every block. It would prevent synching the first blocks received after genesis when a client is launched for the first time with an empty state.

Wild idea: introduce a strict mode (activable on dev station and on CI) for this kind of (should not happen, but refactors can break this assumption)

src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
strauss-m: It depends if one let the lambda provider or the caller have control of the chain range.

However, it seems that `BlockValidation.getNBlocksBackFromChainOrQueue()` implements this interface, and is actually using the hash parameter, so we may have to keep it that way, and make every implementing function uses its parameters.

src/main/scala/io/iohk/ethereum/consensus/pow/validators/OmmersValidator.scala
AurelienRichez: We already have a rudimentary consistency check in #1070. 

Indeed, for the first few blocks it would be expected to have less than `n` blocks. I will try to make that more explicit in the code.

strauss-m (approved):

PR #1086, opened 2021-08-02
fix buildkite checking out mantis-automation

nrdxp:
# Description

Buildkite is erroneously checking out mantis-automation somehow do to some strange hook bug.

# Proposed Solution

A clean checkout looks like it is required to keep the repo's separate for whatever reason. Also @AurelienRichez's fix to add a `- wait` statement also was compounding the issue, since this caused the first step to not run at all, and skip directly to mantis automation step.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,3 +1,6 @@
+env:
+  BUILDKITE_CLEAN_CHECKOUT: "true"
+
 steps:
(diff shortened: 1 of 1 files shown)


AurelienRichez (approved):
👍

ghost (approved):

dzajkowski:
I think this needs to be parked until we figure out how to run these tests in a less disruptive way.

PR #1087, opened 2021-08-02
[ETCM-1053] extend PeerEventBus to support Akka Streams

jvdp:
... and use it for the Peer case class as specified in the ticket.

However, I'm a bit skeptical we want to go down the route of managing and materializing a separate source per actor. Instead I think it will be more practical to use one source with the messages from different peers. That way we won't need to worry about setup and teardown and precise synchronisation of these things. Look at the added spec for an example of this (there's a `Thread.sleep` in there.)

A possible improvement would be to materialize a future that completes when the subscription is performed (via ask-pattern with the `PeerEventBusActor`) to aid in this synchronisation. I was playing around with this but it only made sense for the specs, and I think we should avoid this kind of stuff for production code.

There's also a small bugfix in there to unsubscribe when the subscribing actor dies, which might cause memory leaks otherwise.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/network/PeerEventBusActor.scala
+++ b/src/main/scala/io/iohk/ethereum/network/PeerEventBusActor.scala
@@ -1,9 +1,17 @@
 package io.iohk.ethereum.network
 
+import akka.NotUsed
 import akka.actor.Actor
 import akka.actor.ActorRef
 import akka.actor.Props
+import akka.actor.Terminated
 import akka.event.ActorEventBus
+import akka.stream.OverflowStrategy
+import akka.stream.WatchedActorTerminatedException
+import akka.stream.scaladsl.Source
+import akka.util.Timeout
+
+import scala.concurrent.Future
 
 import io.iohk.ethereum.network.PeerEventBusActor.PeerEvent.MessageFromPeer
 import io.iohk.ethereum.network.PeerEventBusActor.PeerEvent.PeerDisconnected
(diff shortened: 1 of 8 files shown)


src/main/scala/io/iohk/ethereum/network/PeerEventBusActor.scala
```diff
@@ -1,20 +1,55 @@
...
+    *   actor terminates.
+    *
+    *   Note:
+    *   - subscription is asynchronous so it may miss messages when starting.
+    *   - it does not complete when a specified peerId disconnects.
+    */
+  def messageSource(peerEventBus: ActorRef, messageClassifier: MessageClassifier): Source[MessageFromPeer, NotUsed] =
+    Source
+      .fromMaterializer { (mat, _) =>
+        import mat.executionContext
```
leo-bogastry: this `import mat.executionContext` does not seem to be needed, i was able to run tests without this import

src/main/scala/io/iohk/ethereum/network/PeerEventBusActor.scala
```diff
@@ -1,20 +1,55 @@
...
+        val (actorRef, src) = Source
+          .actorRef[MessageFromPeer](1, OverflowStrategy.fail)
```
leo-bogastry: this `actorRef` is deprecated. Have you tried to use the other variant they mention in documentation?

src/test/scala/io/iohk/ethereum/network/PeerEventBusActorSpec.scala
```diff
@@ -46,6 +56,33 @@ class PeerEventBusActorSpec extends AnyFlatSpec with Matchers {
...
+
+    val completeOnTermination =
+      Flow[MessageFromPeer].recoverWithRetries(1, { case _: WatchedActorTerminatedException => Source.empty })
+
+    val stream1 =
+      PeerEventBusActor.messageSource(peerEventBusActor, classifier1).via(completeOnTermination).runWith(Sink.seq)
+    val stream2 =
+      PeerEventBusActor.messageSource(peerEventBusActor, classifier2).via(completeOnTermination).runWith(Sink.seq)
+
+    Thread.sleep(100) // stream is not subscribed right away
```
AurelienRichez: I managed to remove the sleep with a probe that forwards messages, and adding an `expectMsgType[Subscribe]` before publishing.

I don't know if it is a really good idea, but it seems to fit the bill to allow waiting for subscribe without modifying production code only for that purpose.
  
```scala
    val classifier1 = MessageClassifier(Set(Ping.code), PeerSelector.WithId(PeerId("1")))
    val classifier2 = MessageClassifier(Set(Ping.code), PeerSelector.AllPeers)
    val peerEventBusProbe = TestProbe()(system)
    peerEventBusProbe.setAutoPilot { (sender: ActorRef, msg: Any) =>
      peerEventBusActor.tell(msg, sender)
      TestActor.KeepRunning
    }

    val completeOnTermination =
      Flow[MessageFromPeer].recoverWithRetries(1, { case _: WatchedActorTerminatedException => Source.empty })

    val stream1 =
      PeerEventBusActor.messageSource(peerEventBusProbe.ref, classifier1).via(completeOnTermination).runWith(Sink.seq)

    peerEventBusProbe.expectMsgType[Subscribe]
    val stream2 =
      PeerEventBusActor.messageSource(peerEventBusProbe.ref, classifier2).via(completeOnTermination).runWith(Sink.seq)
    peerEventBusProbe.expectMsgType[Subscribe]

    val msgFromPeer = MessageFromPeer(Ping(), PeerId("1"))
    peerEventBusActor ! PeerEventBusActor.Publish(msgFromPeer)

    val msgFromPeer2 = MessageFromPeer(Ping(), PeerId("99"))
    peerEventBusActor ! PeerEventBusActor.Publish(msgFromPeer2)

    peerEventBusProbe.ref ! PoisonPill

    whenReady(stream1)(_ shouldEqual Seq(msgFromPeer))
    whenReady(stream2)(_ shouldEqual Seq(msgFromPeer, msgFromPeer2))
 ```

src/main/scala/io/iohk/ethereum/network/PeerEventBusActor.scala
jvdp: Good catch!

src/main/scala/io/iohk/ethereum/network/PeerEventBusActor.scala
jvdp: Missed that, will have a look. 👍

src/test/scala/io/iohk/ethereum/network/PeerEventBusActorSpec.scala
jvdp: Ohh that is clever, I like it!

AurelienRichez (approved):

leo-bogastry (approved):

PR #1088, opened 2021-08-03
disable ERC20 tests for now

AurelienRichez:
# Description

Following the problems with ERC20 test hanging. I disabled them temporarily to make easier to merge PRs while we are discussing a solution.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -7,8 +7,7 @@ steps:
     agents:
       queue: project42
     timeout_in_minutes: 60
-  - wait
-  - label: "Mantis Automation"
-    command:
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -7,8 +7,7 @@ steps:
     agents:
       queue: project42
     timeout_in_minutes: 60
-  - wait
```
strauss-m, line 10: is it expected to have lost the wait while commenting out the erc20 tests command ?

leo-bogastry (approved):

dzajkowski (approved):

.buildkite/pipeline.yml
AurelienRichez: The wait was added by me in an attempt to limit the impact of the ERC20 tests which was causing a weird bug with buildkite fetching from the wrong repository. But it was more like a workaround and was removed in #1086 (a real fix for that specific issue).

PR #1090, opened 2021-08-06
[ETCM-987] Fix saving the best block number without adding a mapping between block number and block hash

leo-bogastry:
# Description
Calling `appStateStorage.putBestBlockNumber` that makes an update to the best block number without calling `blockNumberMappingStorage.put` to add the mapping between the block number and block hash can result on a `None` when calling `blockchain.getBestBlock`.
I wanted to add these two calls together in the blockchain or blockchainwriter class, but it is not that straight forward.

In order to be able to test this modification already I applied the fix in FastSync and created ETCM-1089 for a proper refactor

# Testing
This has been tested against mainnet.

### DIFF
--- a/src/it/scala/io/iohk/ethereum/sync/util/FastSyncItSpecUtils.scala
+++ b/src/it/scala/io/iohk/ethereum/sync/util/FastSyncItSpecUtils.scala
@@ -34,6 +34,7 @@ object FastSyncItSpecUtils {
       FastSync.props(
         storagesInstance.storages.fastSyncStateStorage,
         storagesInstance.storages.appStateStorage,
+        storagesInstance.storages.blockNumberMappingStorage,
         bl,
         blockchainReader,
         blockchainWriter,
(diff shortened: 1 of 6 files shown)


strauss-m (approved):

AnastasiiaL (approved):

jvdp (approved):
I don't disagree with this change but I feel the main improvement it does is mention the TODO in the code that this logic needs to move to a blockchain-storage related component. It won't affect the actual bug because that doesn't have much to do with fast sync, as discussed.

leo-bogastry:
> I don't disagree with this change but I feel the main improvement it does is mention the TODO in the code that this logic needs to move to a blockchain-storage related component. It won't affect the actual bug because that doesn't have much to do with fast sync, as discussed.

I think you are probably right. The modification shows that is was not just about having cache and storage in synchronization, different storages should also be always syncronized

PR #1091, opened 2021-08-06
[ETCM-1069-partial] Create ConsensusAdapter with prevalidation

AurelienRichez:
# Description

This PR introduces the `ConsensusAdapter`. It will do much more later but the modifier code is getting quite big so I prefer to create a first small PR with a partial implementation and have another PR(s) with the bigger changes. 

The `ConsensusAdapter` wraps the Consensus. For now, it only does prevalidation and duplicate checks.

### DIFF
--- a/src/main/scala/io/iohk/ethereum/consensus/ConsensusAdapter.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/ConsensusAdapter.scala
@@ -1,0 +1,17 @@
+package io.iohk.ethereum.consensus
+
+import io.iohk.ethereum.blockchain.sync.regular.BlockImportResult
+import io.iohk.ethereum.domain.Block
+import io.iohk.ethereum.utils.BlockchainConfig
+import monix.eval.Task
+import monix.execution.Scheduler
+
+/** This is a temporary class to isolate the real Consensus and extract responsibilities which should not
+  * be part of the consensus in the final design, but are currently needed.
+  */
+class ConsensusAdapter(consensus: Consensus) {
+  def evaluateBranchBlock(
+      block: Block
+  )(implicit blockExecutionScheduler: Scheduler, blockchainConfig: BlockchainConfig): Task[BlockImportResult] =
+    consensus.evaluateBranch(Seq(block))
+}
--- a/src/main/scala/io/iohk/ethereum/consensus/ConsensusAdapter.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/ConsensusAdapter.scala
@@ -44,6 +44,7 @@ class ConsensusAdapter(
           }
         }
       case None =>
+        log.error("Couldn't find the current best block")
         Task.now(BlockImportFailed("Couldn't find the current best block"))
     }
 
--- a/src/main/scala/io/iohk/ethereum/consensus/ConsensusAdapter.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/ConsensusAdapter.scala
@@ -1,17 +1,58 @@
 package io.iohk.ethereum.consensus
 
-import io.iohk.ethereum.blockchain.sync.regular.BlockImportResult
-import io.iohk.ethereum.domain.Block
-import io.iohk.ethereum.utils.BlockchainConfig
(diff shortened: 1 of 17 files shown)


src/test/scala/io/iohk/ethereum/consensus/ConsensusSpec.scala
```diff
@@ -400,6 +400,109 @@ class ConsensusSpec extends AnyFlatSpec with Matchers with ScalaFutures {
     blockchainReader.getBestBlock().get shouldEqual newBlock3WithOmmer
   }
 
+  it should "dequeue blocks where there is an execution error" in new EphemBlockchain {
```
AurelienRichez, line 348: Those tests are not strictly necessary in that PR as it does not touch the blockQueue but they make sure that I did not change the way the blockQueue is handled here and in the next PR.

src/main/scala/io/iohk/ethereum/consensus/ConsensusAdapter.scala
```diff
@@ -0,0 +1,68 @@
...
+        } else {
+          doBlockPreValidation(block).flatMap {
+            case Left(error) =>
+              Task.now(BlockImportFailed(error.reason.toString))
+            case Right(BlockExecutionSuccess) =>
+              consensus.evaluateBranch(Seq(block))
+          }
+        }
+      case None =>
+        Task.now(BlockImportFailed("Couldn't find the current best block"))
```
leo-bogastry, line 49: Don't forget to also add a `log.error` here

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
```diff
@@ -63,26 +61,19 @@ class ConsensusImpl(
...
     *   - [[BlockImportFailed]]  - block failed to execute (when importing to top or reorganising the chain)
     */
-  override def evaluateBranchBlock(
-      block: Block
-  )(implicit blockExecutionScheduler: Scheduler, blockchainConfig: BlockchainConfig): Task[BlockImportResult] =
+  override def evaluateBranch(
+      branch: Seq[Block]
+  )(implicit blockExecutionScheduler: Scheduler, blockchainConfig: BlockchainConfig): Task[BlockImportResult] = {
+    val block = branch.head
     blockchainReader.getBestBlock() match {
```
leo-bogastry, line 68: The `blockchainReader.getBestBlock()` was also moved to `ConsensusAdapter`, so I think you can remove it from here and simplify the code

src/test/scala/io/iohk/ethereum/consensus/ConsensusSpec.scala
leo-bogastry: 👍

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
AurelienRichez: I did not want to make the assumption that the caller will provide the best block, so I think we should keep it this way even if it is indeed a bit wasteful.

src/main/scala/io/iohk/ethereum/consensus/ConsensusAdapter.scala
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+  private def doBlockPreValidation(block: Block)(implicit
+      blockchainConfig: BlockchainConfig
+  ): Task[Either[ValidationBeforeExecError, BlockExecutionSuccess]] =
+    Task
+      .evalOnce(blockValidation.validateBlockBeforeExecution(block))
+      .tap {
+        case Left(error) =>
+          log.error("Error while validating block with hash {} before execution: {}", block.hash, error.reason)
```
AnastasiiaL: error here is of type ValidationBeforeExecError right? then it should be `error.reason.toString`

AurelienRichez: But `toString` will be called by the logger, right ?

AurelienRichez: I added the toString and also wrote block hashes in hex instead of plain ByteString

AnastasiiaL (approved):

leo-bogastry (approved):

PR #1092, opened 2021-08-09
[ETCM-1048] Remove bestKnownBlockAndLatestCheckpoint cache

leo-bogastry:
# Description
Continuing the effort of removing the desynchronization between caches and RocksDB the `AtomicReference[BestBlockLatestCheckpointNumbers]` used as cache was removed and replaced by persistence in RocksDB.

# Testing
This branch has been tested against mainnet, deployed in Staging and tested against the ERC20 tests. Report here: https://buildkite.com/input-output-hk/mantis-automation/builds/7468#_


src/test/scala/io/iohk/ethereum/consensus/ConsensusSpec.scala
```diff
@@ -197,61 +197,6 @@ class ConsensusSpec extends AnyFlatSpec with Matchers with ScalaFutures {
     blockQueue.isQueued(oldBlock3.header.hash) shouldBe true
   }
 
-  it should "get best stored block after reorganisation of the longer chain to a shorter one if desync state happened between cache and db" in new EphemBlockchain {
```
leo-bogastry, line 200: This test was removed because testing synchonization between cache and DB does not make sense anymore

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
```diff
@@ -68,7 +68,7 @@ class SyncController(
   def start(): Unit = {
     import syncConfig.doFastSync
 
-    appStateStorage.putSyncStartingBlock(appStateStorage.getBestBlockNumber())
+    appStateStorage.putSyncStartingBlock(appStateStorage.getBestBlockNumber()).commit
```
jvdp: What's this change for? (Also elsewhere.)

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -565,16 +565,13 @@ class FastSync(
...
     // TODO [ETCM-676]: Move to blockchain and make sure it's atomic
-    private def discardLastBlocks(startBlock: BigInt, blocksToDiscard: Int): Unit = {
+    private def discardLastBlocks(startBlock: BigInt, blocksToDiscard: Int): Unit =
       // TODO (maybe ETCM-77): Manage last checkpoint number too
-      appStateStorage.putBestBlockNumber((startBlock - blocksToDiscard - 1).max(0)).commit()
-
       (startBlock to ((startBlock - blocksToDiscard).max(1)) by -1).foreach { n =>
         blockchainReader.getBlockHeaderByNumber(n).foreach { headerToRemove =>
-          blockchain.removeBlock(headerToRemove.hash, withState = false)
+          blockchain.removeBlock(headerToRemove.hash)
```
jvdp, line 574: Just to remark: perhaps this parameter was also added to prevent issues with write amplification, but we won't need to actually remove blocks anymore (or at least from sync logic) when we can save tentative chains / blocks. (Just need to make sure the cleanup is done in a decently efficient way.)

jvdp (approved):
Nice simplification 👍

src/main/scala/io/iohk/ethereum/blockchain/sync/SyncController.scala
leo-bogastry: without the `commit` the data is not saved in the DB

jvdp: Ah so this was actually a bugfix?

Also, doesn't it require the `()`?

leo-bogastry: I added the `()`, but it works both ways.
I don't know if an actual bug resulted from that missing `commit()`, but I know that these put method return a `DataSourceBatchUpdate`. And this allows to chain several put calls and then commit then atomically. In this case the `commit()` was clearly forgotten

src/test/scala/io/iohk/ethereum/ObjectGenerators.scala
```diff
@@ -33,6 +33,8 @@ trait ObjectGenerators {
 
   def intGen(min: Int, max: Int): Gen[Int] = Gen.choose(min, max)
 
+  def posIntGen(min: Int, max: Int): Gen[Int] = Gen.choose(min, max).suchThat(_ > 0)
```
AurelienRichez, line 36: why do we need this ? would `intGen(1, max: Int)` not work ? Or did it use 0 during shrinking ? 😞

AurelienRichez (approved):

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
```diff
@@ -567,16 +567,13 @@ class FastSync(
     }
 
     // TODO [ETCM-676]: Move to blockchain and make sure it's atomic
-    private def discardLastBlocks(startBlock: BigInt, blocksToDiscard: Int): Unit = {
+    private def discardLastBlocks(startBlock: BigInt, blocksToDiscard: Int): Unit =
       // TODO (maybe ETCM-77): Manage last checkpoint number too
-      appStateStorage.putBestBlockNumber((startBlock - blocksToDiscard - 1).max(0)).commit()
```
AnastasiiaL, line 572: why is saving to the storage removed here?🤔

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
```diff
@@ -77,15 +77,13 @@ class BlockchainImpl(
...
-    blockchainMetadata: BlockchainMetadata
+    blockchainReader: BlockchainReader
 ) extends Blockchain
     with Logger {
 
   override def getChainWeightByHash(blockhash: ByteString): Option[ChainWeight] = chainWeightStorage.get(blockhash)
 
-  override def getLatestCheckpointBlockNumber(): BigInt =
-    blockchainMetadata.bestKnownBlockAndLatestCheckpoint.get().latestCheckpointNumber
+  override def getLatestCheckpointBlockNumber(): BigInt = appStateStorage.getLatestCheckpointBlockNumber()
```
AnastasiiaL, line 86: if `getBestBlockNumber()` moved to the `blockchainReader`, then I think `getLatestCheckpointBlockNumber()` should also go there

src/test/scala/io/iohk/ethereum/ObjectGenerators.scala
leo-bogastry: Oh I forgot about this issue. I don't know why but this test started failing consistently because it was generating negative numbers

src/main/scala/io/iohk/ethereum/domain/Blockchain.scala
leo-bogastry: You are right. I created this ticket https://jira.iohk.io/browse/ETCM-1092

src/main/scala/io/iohk/ethereum/blockchain/sync/fast/FastSync.scala
leo-bogastry: After I did all the changes I had a bunch of tests failing in FastSyncItSpec, because the best blocknumber was always a bit inferior than expected, and the issue was that line.
Now that we don't have the caches anymore and always save to storage that instructions was bringing issues.

PR #1093, opened 2021-08-10
[ETCM-1045] Move getChainWeight to BlockchainReader

leo-bogastry:
# Description

In order to reduce the scope of the Blockchain class, getChainWeight was moved to BlockchainReader

# Testing
This branch was deployed in Staging

### DIFF
--- a/src/test/scala/io/iohk/ethereum/ledger/LedgerTestSetup.scala
+++ b/src/test/scala/io/iohk/ethereum/ledger/LedgerTestSetup.scala
@@ -151,7 +151,6 @@ trait TestSetup extends SecureRandomBuilder with EphemBlockchainTestSetup {
 
   def applyChanges(
       stateRootHash: ByteString,
-      blockchainStorages: BlockchainStorages,
       changes: Seq[(Address, Changes)]
   ): ByteString = {
     val initialWorld = InMemoryWorldStateProxy(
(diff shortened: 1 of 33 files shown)


src/test/scala/io/iohk/ethereum/domain/BlockchainSpec.scala
```diff
@@ -214,7 +214,7 @@ class BlockchainSpec extends AnyFlatSpec with Matchers with ScalaCheckPropertyCh
   }
 
   it should "return correct best block number after saving and rolling back blocks" in new TestSetup {
-    forAll(posIntGen(min = 1, max = maxNumberBlocksToImport)) { numberBlocksToImport =>
+    forAll(intGen(min = 1, max = maxNumberBlocksToImport)) { numberBlocksToImport =>
```
leo-bogastry: I did not understand why this test was failing in a previous PR, the generated number sometimes where negative so I added the `posIntGen` that had a clause filtering negative numbers.
In this branch this test was failing again sometimes so, looking into it, I realized that now that the cache was removed in previous PR's the way the value `expectedPersistedBestBlock` (line 257) was being calculated made no sense and simplified the test

AurelienRichez (approved):

AnastasiiaL (approved):

src/test/scala/io/iohk/ethereum/ledger/LedgerTestSetup.scala
```diff
@@ -479,7 +478,7 @@ trait EphemBlockchain extends TestSetupWithVmAndValidators with MockFactory {
 trait CheckpointHelpers {
   private val sampleCheckpoint = ObjectGenerators.fakeCheckpointGen(3, 3).sample.get
 
-  def getCheckpointBlock(parent: Block, difficulty: BigInt, checkpoint: Checkpoint = sampleCheckpoint): Block =
```
ghost, line 482: @LeonorLunatech what is the reason for this change?

leo-bogastry: The parameter `difficulty` was not being used

leo-bogastry: We have several warnings being output when the code compile, and we are trying to take care of them. This was one of them

PR #1094, opened 2021-08-11
[ETCM-921] transaction with access list

strauss-m:
Make `SignedTransaction` wrap `Transaction` to allow `TypedTransactions`.

This PR is split from https://github.com/input-output-hk/mantis/pull/1061

Included in this branch are:
- ETCM-911: new Typed Transaction envelop
- ETCM-921: new Type 1 transaction (payload + rlp encoder/decoder)

Follow up tickets are expected:
- ETCM-1095: receipt implementation
- ETCM-1096: signature/validation process enhancement for EIP-2930

### DIFF
--- a/rlp/src/main/scala/io/iohk/ethereum/rlp/package.scala
+++ b/rlp/src/main/scala/io/iohk/ethereum/rlp/package.scala
@@ -44,6 +44,8 @@ package object rlp {
     override def toString: String = s"RLPValue(${Hex.toHexString(bytes)})"
   }
 
+  case class PrefixedRLPEncodable(prefix: Byte, prefixedRLPEncodeable: RLPEncodeable) extends RLPEncodeable
+
   trait RLPEncoder[T] {
     def encode(obj: T): RLPEncodeable
   }
--- a/rlp/src/main/scala/io/iohk/ethereum/rlp/package.scala
+++ b/rlp/src/main/scala/io/iohk/ethereum/rlp/package.scala
@@ -44,7 +44,22 @@ package object rlp {
     override def toString: String = s"RLPValue(${Hex.toHexString(bytes)})"
   }
 
-  case class PrefixedRLPEncodable(prefix: Byte, prefixedRLPEncodeable: RLPEncodeable) extends RLPEncodeable
+  /** Modelise a RLPEncodable that should be binary prefixed by a raw byte.
+    *
+    * When converting this RLPEncodable to byte, the resulting value will be:
+    * prefix || prefixedRLPEncodable.toByte
+    * where || is the binary concatenation symbol.
+    *
+    * To be able to read back the data, use TypedTransaction.TypedTransactionsRLPAggregator
+    *
+    * This is for example used for typed transaction and typed receipt.
+    *
+    * @param prefix the raw byte
+    * @param prefixedRLPEncodeable the RLPEncodable to prefix with
+    */
+  case class PrefixedRLPEncodable(prefix: Byte, prefixedRLPEncodeable: RLPEncodeable) extends RLPEncodeable {
+    require(prefix <= 0x07f, "prefix should be lower than 0x7f")
+  }
 
   trait RLPEncoder[T] {
     def encode(obj: T): RLPEncodeable
--- a/rlp/src/main/scala/io/iohk/ethereum/rlp/package.scala
+++ b/rlp/src/main/scala/io/iohk/ethereum/rlp/package.scala
@@ -58,7 +58,7 @@ package object rlp {
     * @param prefixedRLPEncodeable the RLPEncodable to prefix with
     */
   case class PrefixedRLPEncodable(prefix: Byte, prefixedRLPEncodeable: RLPEncodeable) extends RLPEncodeable {
-    require(prefix <= 0x07f, "prefix should be lower than 0x7f")
+    require(prefix >= 0, "prefix should be in the range [0; 0x7f]")
   }
 
   trait RLPEncoder[T] {
--- a/src/main/scala/io/iohk/ethereum/network/p2p/messages/BaseETH6XMessages.scala
+++ b/src/main/scala/io/iohk/ethereum/network/p2p/messages/BaseETH6XMessages.scala
@@ -13,6 +13,7 @@ import io.iohk.ethereum.rlp.RLPImplicitConversions._
 import io.iohk.ethereum.rlp.RLPImplicits._
 import io.iohk.ethereum.rlp._
 import io.iohk.ethereum.utils.ByteStringUtils.ByteStringOps
+import io.iohk.ethereum.utils.Config
 
 object BaseETH6XMessages {
   object Status {
@@ -62,7 +63,7 @@ object BaseETH6XMessages {
         case r: RLPList if r.items.isEmpty => AccessListItem(null, List.empty)
 
         case RLPList(rlpAddress, rlpStorageKeys: RLPList) =>
-          val address = rlpAddress.decodeAs[Address]("address ")
+          val address = rlpAddress.decodeAs[Address]("address")
           val storageKeys = fromRlpList[BigInt](rlpStorageKeys).toList
           AccessListItem(address, storageKeys)
       }
@@ -154,7 +155,7 @@ object BaseETH6XMessages {
 
   object SignedTransactions {
 
-    lazy val chainId: Byte = 1.toByte //Config.blockchains.blockchainConfig.chainId
+    lazy val chainId: Byte = Config.blockchains.blockchainConfig.chainId
 
     implicit class SignedTransactionEnc(val signedTx: SignedTransaction) extends RLPSerializable {
 
--- a/src/main/scala/io/iohk/ethereum/network/p2p/messages/BaseETH6XMessages.scala
+++ b/src/main/scala/io/iohk/ethereum/network/p2p/messages/BaseETH6XMessages.scala
@@ -8,11 +8,11 @@ import io.iohk.ethereum.domain.BlockHeaderImplicits._
 import io.iohk.ethereum.domain._
 import io.iohk.ethereum.network.p2p.Message
 import io.iohk.ethereum.network.p2p.MessageSerializableImplicit
+import io.iohk.ethereum.rlp.RLPCodec.Ops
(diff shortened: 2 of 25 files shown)


src/test/scala/io/iohk/ethereum/domain/TransactionSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+      ByteString(Hex.decode("5544")),
+      Nil
+    )
+    val stx = SignedTransaction.apply(
+      tx = tx,
+      ECDSASignature
+        .fromBytes(
+          ByteString(
+            Hex.decode(
+              "c9519f4f2b30335884581971573fadf60c6204f59a911df35ee8a540456b266032f1e8e2c5dd761f9e4f88f41c8310aeaba26a8bfcdacfedfa12ec3862d3752101"
```
lukasz-golebiewski, line 78: How do we know this value is correct? Is this specified somewhere in the EIP?

strauss-m: Dom tooks those values directly from core-geth own tests.

See https://github.com/ethereum/go-ethereum/blob/master/core/types/transaction_test.go#L71

lukasz-golebiewski: Gotcha! It would probably make sense to add a github link to those tests for posterity in a comment (pointing to a concrete commit, rather than master, in case they get moved)

lukasz-golebiewski (approved):

src/main/scala/io/iohk/ethereum/network/p2p/messages/BaseETH6XMessages.scala
```diff
@@ -50,6 +50,24 @@ object BaseETH6XMessages {
...
+  implicit val accessListItemCodec: RLPCodec[AccessListItem] =
+    RLPCodec.instance[AccessListItem](
+      { case AccessListItem(address, storageKeys) =>
+        RLPList(address, toRlpList(storageKeys))
+      },
+      {
+        case r: RLPList if r.items.isEmpty => AccessListItem(null, List.empty)
+
+        case RLPList(rlpAddress, rlpStorageKeys: RLPList) =>
+          val address = rlpAddress.decodeAs[Address]("address ")
```
leo-bogastry: there's a space in `"address "`

strauss-m: good catch, fixed

rlp/src/main/scala/io/iohk/ethereum/rlp/package.scala
```diff
@@ -44,6 +44,23 @@ package object rlp {
...
+    *
+    * To be able to read back the data, use TypedTransaction.TypedTransactionsRLPAggregator
+    *
+    * This is for example used for typed transaction and typed receipt.
+    *
+    * @param prefix the raw byte
+    * @param prefixedRLPEncodeable the RLPEncodable to prefix with
+    */
+  case class PrefixedRLPEncodable(prefix: Byte, prefixedRLPEncodeable: RLPEncodeable) extends RLPEncodeable {
+    require(prefix <= 0x07f, "prefix should be lower than 0x7f")
```
lukasz-golebiewski:
```
scala> 0x07f.toByte
val res0: Byte = 127

scala> 128.toByte <= 0x7f.toByte
val res3: Boolean = true
```

Looks like all `Byte`s are smaller than `0x7f` so it makes more sense to check whether `prefix` is larger than 0

src/main/scala/io/iohk/ethereum/network/p2p/messages/BaseETH6XMessages.scala
```diff
@@ -134,26 +158,76 @@ object BaseETH6XMessages {
...
+        * - all other RLPEncodable are kept unchanged
+        *
+        * This is the responsibility of the RLPDecoder to insert this meaning into its RLPList, when appropriate.
+        *
+        * @return a Seq of TypedTransaction enriched RLPEncodable
+        */
+      def toTypedRLPEncodables: Seq[RLPEncodeable] =
+        encodables match {
+          case Seq() => Seq()
+          case Seq(RLPValue(v), rlpList: RLPList, tail @ _*) if v.length == 1 && v.head < 0x7f =>
```
lukasz-golebiewski: Do you think it would make sense not to repeat the magic `0x7f` in multiple places but putting it for example in a wrapper value class for `Byte`?

lukasz-golebiewski (commented):
Looks great overall, aside from a few remarks

rlp/src/main/scala/io/iohk/ethereum/rlp/package.scala
strauss-m: good catch, I'll change the check and add proper testing

src/main/scala/io/iohk/ethereum/network/p2p/messages/BaseETH6XMessages.scala
strauss-m: good idea, I've extracted the magic number and added the proposed wrapper

lukasz-golebiewski (approved):

rlp/src/main/scala/io/iohk/ethereum/rlp/package.scala
```diff
@@ -44,6 +44,23 @@ package object rlp {
...
+    *
+    * To be able to read back the data, use TypedTransaction.TypedTransactionsRLPAggregator
+    *
+    * This is for example used for typed transaction and typed receipt.
+    *
+    * @param prefix the raw byte
+    * @param prefixedRLPEncodeable the RLPEncodable to prefix with
+    */
+  case class PrefixedRLPEncodable(prefix: Byte, prefixedRLPEncodeable: RLPEncodeable) extends RLPEncodeable {
+    require(prefix >= 0, "prefix should be in the range [0; 0x7f]")
```
dzajkowski, line 61: isn't this up to 0xFF?

strauss-m: 0x7f is the limit for two reasons:
- EIP-2718: TransactionType is a positive unsigned 8-bit number between 0 and 0x7f that represents the type of the transcation
- rlp encoding spec: for a single byte whose value is in the [0x00, 0x7f] range, that byte is its own RLP encoding

That's what allows us to consider a single RLPValue [0; 0x7f] as a single raw byte transaction type (and vice-versa) 

Since Byte is a signed type, the range [0; 0x7f] check become a simple positive condition.

src/main/scala/io/iohk/ethereum/network/p2p/messages/BaseETH6XMessages.scala
```diff
@@ -50,6 +52,24 @@ object BaseETH6XMessages {
...
+    RLPCodec.instance[AccessListItem](
+      { case AccessListItem(address, storageKeys) =>
+        RLPList(address, toRlpList(storageKeys))
+      },
+      {
+        case r: RLPList if r.items.isEmpty => AccessListItem(null, List.empty)
+
+        case RLPList(rlpAddress, rlpStorageKeys: RLPList) =>
+          val address = rlpAddress.decodeAs[Address]("address")
+          val storageKeys = fromRlpList[BigInt](rlpStorageKeys).toList
```
dzajkowski, line 68: regarding this BigInt type - were you able to verify if this is good enough? maybe a block from ETC mainnet with a typed transaction and the access list populated?

src/main/scala/io/iohk/ethereum/network/p2p/messages/BaseETH6XMessages.scala
```diff
@@ -166,14 +242,60 @@ object BaseETH6XMessages {
...
       def toSignedTransactions: SignedTransactions = rawDecode(bytes) match {
-        case rlpList: RLPList => SignedTransactions(rlpList.items.map(_.toSignedTransaction))
+        case rlpList: RLPList => SignedTransactions(rlpList.items.toTypedRLPEncodables.map(_.toSignedTransaction))
         case _                => throw new RuntimeException("Cannot decode SignedTransactions")
       }
     }
 
     implicit class SignedTransactionRlpEncodableDec(val rlpEncodeable: RLPEncodeable) extends AnyVal {
+
+      // scalastyle:off method.length
```
dzajkowski, line 256: is there a matching :on in the file? can't find it

strauss-m: No, I was thinking that the style annotation was on a per-method scope.
Maybe the best way is to actually split the method ? 
I have mixed feeling about this: on one hand, the method is really long, but on the other one, it's quite straite-forward to read, as the lenght is due to the case class formatting, not actual statements.
However, once we start adding new transactions, we'll need to split it, so maybe doing it now is a good time. Do you have any preference (split vs scalastyle:off) ?

src/main/scala/io/iohk/ethereum/network/p2p/messages/BaseETH6XMessages.scala
strauss-m: I didn't try to find one, but it feel like a good real-life test case until we connect to mainnet. 
I'll give it a look.

Concerning the StorageKey type, I've seen both BigInt and UInt256 in mantis code, as well as Hash (byte[32]) in go-ethereum. Maybe we should introduce (in another ticket ?) a specific StorageKey type similar to the Address one, with a 32 bytes length

src/test/scala/io/iohk/ethereum/domain/TransactionSpec.scala
```diff
@@ -0,0 +1,142 @@
...
+    x shouldBe expected
+
+  }
+  it should "correctly serialize to EIP1598 rlp" in {}
+
+  // plan:
+  // TODO ReceiptSpec
+  // TODO decode string to case class
+  // TODO adjust runVM in ledger
+  // TODO align how geth and besu treat typed transactions when sent via rpc before magneto HF
```
dzajkowski: maybe drop these? no need to have the plan in develop

dzajkowski (approved):
LGTM

PR #1096, opened 2021-08-11
[ETCM-1069] consensus adapter

AurelienRichez:
# Description

Depends on #1091

Change the ConsensusImpl so that  it does not depend on the block queue anymore. The ConsensusAdapter ensures compatibility with the existing code. 

It also define a `ConsensusResult` sum type which is the return type we get when we call the consensus. But it could be simpler when we get rid of the Adapter because I had to add more fined grained information about what happened during the consensus to allow the adapter to be able to faithfully reproduce the same behaviour as before.

# Testing
Branch was tested against ETC and deployed in Staging successfully.
ERC20 tests run successfully in Staging: https://buildkite.com/input-output-hk/mantis-automation/builds/9936#_

### DIFF
--- a/src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
@@ -77,28 +77,40 @@ trait Consensus {
 
 object Consensus {
   /* This return type for consensus is probably overcomplicated for now because some information is needed
-   * to keep the compatibility with the current code (particularly for the block queue handling).
+   * to keep the compatibility with the current code (particularly for the block queue handling), and be able
+   * to translate the values to BlockImportResult.
    * In particular:
    *  - `blockToEnqueue` fields won't be needed if the block are already stored in memory
    *  - The distinction between ExtendedCurrentBestBranch and SelectedNewBestBranch won't really be useful
    *  because there will be no need to put back the old branch into the block queue in case of reorganisation
-   *  - `ConsensusErrorDueToMissingNode` would mean that the application is in an inconsistent state
+   *  - `ConsensusErrorDueToMissingNode` and  `ConsensusError` would mean that the application is in an
+   *  inconsistent state. Unless there is a reason to think that mantis would self heal when that happens, I
+   *  don't think there is a reason to add them here.
    */
 
   sealed trait ConsensusResult
 
+  /** The new branch was selected and it extended the best branch. */
   case class ExtendedCurrentBestBranch(blockImportData: List[BlockData]) extends ConsensusResult
+
+  /** The new branch was selected and it extended the best branch, but it did not execute completely. */
   case class ExtendedCurrentBestBranchPartially(blockImportData: List[BlockData], failureBranch: BranchExecutionFailure)
       extends ConsensusResult
 
+  /** The new branch was selected but was not an extension of the best branch. */
   case class SelectedNewBestBranch(oldBranch: List[Block], newBranch: List[Block], weights: List[ChainWeight])
       extends ConsensusResult
 
+  /** The proposed new branch was not better than the current best one. */
   case object KeptCurrentBestBranch extends ConsensusResult
 
+  /** A block in the branch cannot be executed. */
   case class BranchExecutionFailure(blockToEnqueue: List[Block], failingBlockHash: ByteString, error: String)
       extends ConsensusResult
 
+  /** An error external the the blocks in the branch occured, which prevents the branch from being executed.
+    * Usually this is due to an inconsistency in the database.
+    */
   case class ConsensusError(blockToEnqueue: List[Block], err: String) extends ConsensusResult
   case class ConsensusErrorDueToMissingNode(blockToEnqueue: List[Block], reason: MissingNodeException)
       extends ConsensusResult
--- a/src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
@@ -96,7 +96,8 @@ object Consensus {
 
   case object KeptCurrentBestBranch extends ConsensusResult
 
-  case class BranchExecutionFailure(failingBlockHash: ByteString, error: String) extends ConsensusResult
+  case class BranchExecutionFailure(blockToEnqueue: List[Block], failingBlockHash: ByteString, error: String)
+      extends ConsensusResult
 
   case class ConsensusError(blockToEnqueue: List[Block], err: String) extends ConsensusResult
   case class ConsensusErrorDueToMissingNode(blockToEnqueue: List[Block], reason: MissingNodeException)
--- a/src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
+++ b/src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
@@ -1,10 +1,15 @@
 package io.iohk.ethereum.consensus
 
+import cats.data.NonEmptyList
+
 import monix.eval.Task
 import monix.execution.Scheduler
 
-import io.iohk.ethereum.blockchain.sync.regular.BlockImportResult
+import io.iohk.ethereum.consensus.Consensus.ConsensusResult
 import io.iohk.ethereum.domain.Block
+import io.iohk.ethereum.domain.ChainWeight
+import io.iohk.ethereum.ledger.BlockData
+import io.iohk.ethereum.mpt.MerklePatriciaTrie.MissingNodeException
 import io.iohk.ethereum.utils.BlockchainConfig
 
 /** This file documents the original interface that was designed at ETCM-1018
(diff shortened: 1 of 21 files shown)


src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
```diff
@@ -28,226 +20,168 @@ import io.iohk.ethereum.ledger.BlockData
...
-      reorganiseOrEnqueue(block, bestBlock, weight)
+  ): Task[ConsensusResult] = {
+    val parentHash = branch.head.header.parentHash
+    val parentWeight = blockchain
+      .getChainWeightByHash(parentHash)
+      .getOrElse(throw new Error("Inconsistent database"))
+
+    val consensusResult: Task[ConsensusResult] =
+      if (newBranchWeight(branch, parentWeight) > currentBestBlockWeight) {
+        if (currentBestBlock.isParentOf(branch.head)) {
```
leo-bogastry: Wouldn't it make more sense to do this tests first (`currentBestBlock.isParentOf(branch.head)`) and only after calculate the `newBranchWeight` if necessary?

leo-bogastry: Also we could save a read on getting `parentWeight`, and only do it if needed

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
```diff
@@ -28,226 +20,168 @@ import io.iohk.ethereum.ledger.BlockData
...
+        )
     }
-    importResult.foreach(measureBlockMetrics)
-    importResult
+
+  private def reorganise(newBranch: NonEmptyList[Block], parentWeight: ChainWeight, parentHash: ByteString)(implicit
+      blockchainConfig: BlockchainConfig
+  ): ConsensusResult = {
+
+    val bestNumber = blockchainReader.getBestBlockNumber()
```
leo-bogastry: The caller method has the info on the best block, so you can just pass it as an argument here

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
```diff
@@ -258,7 +192,6 @@ class ConsensusImpl(
     * @param executedBlocks sub-sequence of new branch that was executed correctly
```
leo-bogastry, line 238: you can remove the `@param newBranch` from this documentation

src/main/scala/io/iohk/ethereum/consensus/ConsensusAdapter.scala
```diff
@@ -35,15 +50,54 @@ class ConsensusAdapter(
...
+          returnNoTotalDifficulty(bestBlock)
         } else {
           doBlockPreValidation(block).flatMap {
             case Left(error) =>
               Task.now(BlockImportFailed(error.reason.toString))
             case Right(BlockExecutionSuccess) =>
-              consensus.evaluateBranch(Seq(block))
+              enqueueAndGetBranch(block, bestBlock.number) match {
+                case None =>
+                  Task.now(BlockEnqueued)
```
leo-bogastry: Can you please add a comment explaining this scenario is when the block is enqueue but in a branch not rooted?   I had to go check why a None would end up in a `BlockEnqueued` :)

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
```diff
@@ -28,226 +20,168 @@ import io.iohk.ethereum.ledger.BlockData
...
-  ): Task[BlockImportResult] = {
-    val importResult = if (isPossibleNewBestBlock(block.header, bestBlock.header)) {
-      importToTop(block, bestBlock, weight)
-    } else {
-      reorganiseOrEnqueue(block, bestBlock, weight)
+  ): Task[ConsensusResult] = {
+    val parentHash = branch.head.header.parentHash
+    val parentWeight = blockchain
+      .getChainWeightByHash(parentHash)
+      .getOrElse(throw new Error("Inconsistent database"))
```
leo-bogastry: Can you please add more info to the error? Like block hash/number

src/main/scala/io/iohk/ethereum/consensus/ConsensusAdapter.scala
```diff
@@ -35,15 +50,54 @@ class ConsensusAdapter(
...
+                case Some(newBranch) =>
+                  consensus
+                    .evaluateBranch(newBranch)
```
leo-bogastry: This method got really big, can you maybe separate this whole chunk of branch evaluation?

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
```diff
@@ -28,226 +20,168 @@ import io.iohk.ethereum.ledger.BlockData
...
     blockExecution: BlockExecution
 ) extends Consensus
     with Logger {
 
-  /** Tries to import the block as the new best block in the chain or enqueue it for later processing.
-    *
-    * @param block                   block to be imported
+  /** Try to set the given branch as the new best branch if it is better than the current best
+    * branch.
+    * @param branch                  the new branch as a sorted list of blocks. It's parent must
```
leo-bogastry: Just a typo in `It's` instead of `its`

src/test/scala/io/iohk/ethereum/consensus/ConsensusImplSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+import io.iohk.ethereum.consensus.Consensus.SelectedNewBestBranch
+import io.iohk.ethereum.domain.Block
+import io.iohk.ethereum.domain.ChainWeight
+import io.iohk.ethereum.ledger.BlockData
+import io.iohk.ethereum.ledger.BlockExecution
+import io.iohk.ethereum.utils.BlockchainConfig
+
+class ConsensusImplSpec extends AnyFlatSpec with Matchers with ScalaFutures {
+  import ConsensusImplSpec._
+  "Consensus" should "extend the current best chain" in new ConsensusSetup {
```
leo-bogastry, line 31: Maybe some more tests for the scenarios where errors happen?

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
AurelienRichez: Actually, I should maybe change this. I did it at the beginning and forgot to rewrite it. Ideally, I would prefer to just throw when that kind of things happens, because it's not the Consensus job to handle inconsistencies that should not happen in the first place. But currently the way of doing is returning a type to represent the error, so I should do that to stay compatible.

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
AurelienRichez: Yes. I note however that in the final design we would not really care about the fact that we are extending the branch or changing the branch, so this condition is there for compatibility but I think in the end it might just be the weight comparison.

src/main/scala/io/iohk/ethereum/consensus/ConsensusAdapter.scala
AurelienRichez: I extracted the big match in another method.

src/main/scala/io/iohk/ethereum/consensus/ConsensusImpl.scala
AurelienRichez: I rewrote the method and added the hash of the block and its number in the message.

leo-bogastry: Easier to read now 👍

src/main/scala/io/iohk/ethereum/consensus/Consensus.scala
```diff
@@ -67,3 +74,44 @@ trait Consensus {
...
+   *  don't think there is a reason to add them here.
+   */
+
+  sealed trait ConsensusResult
+
+  /** The new branch was selected and it extended the best branch. */
+  case class ExtendedCurrentBestBranch(blockImportData: List[BlockData]) extends ConsensusResult
+
+  /** The new branch was selected and it extended the best branch, but it did not execute completely. */
+  case class ExtendedCurrentBestBranchPartially(blockImportData: List[BlockData], failureBranch: BranchExecutionFailure)
```
AnastasiiaL, line 97: from the name I'd understand that the branch was partially extended, but instead it holds the branch that was extended but not executed yet. So I'd suggest `PendingExtendedCurrentBestBranch` or `NotVerifiedExtendedCurrentBestBranch`

leo-bogastry: It holds the part of the branch that was executed in the first parameter `blockImportData` and in the second parameter `failureBranch` the part of the branch that failed the execution. The last block from `blockImportData` is saved as best block, so the current best chain get extended until that point

AnastasiiaL: then `ExtendedCurrentBestBranchWithDiscardedPart` or `ExtendedCurrentBestBranchWithFailedPart`?

leo-bogastry: I really think `partially` serves it purpose well here. Would `ExtendedPartiallyCurrentBestBranch` be more clear?

AnastasiiaL: if the name is clear for everyone but me - just leave it as is. I know now what it means

jvdp (approved):

AnastasiiaL (approved):

PR #1101, opened 2021-08-17
Etcm 1055 Adjust PeersClient.nextBestBlock()

AnastasiiaL:
# Description

Adjust PeersClient.nextBestBlock(). It should be able to make sure the next best peer has a different best block, so we don't fetch in parallel identical branches

### DIFF
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/PeerComparator.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/PeerComparator.scala
@@ -1,0 +1,9 @@
+package io.iohk.ethereum.blockchain
+
+import io.iohk.ethereum.network.EtcPeerManagerActor.PeerInfo
+
+object PeerComparator {
+
+  def doPeersHaveSameBestBlock(peerInfo1: PeerInfo, peerInfo2: PeerInfo): Boolean =
+    peerInfo1.bestBlockHash == peerInfo2.bestBlockHash
+}
--- a/src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
+++ b/src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
@@ -14,6 +14,7 @@ import scala.concurrent.ExecutionContext
 import scala.jdk.CollectionConverters._
 import scala.reflect.ClassTag
 
+import io.iohk.ethereum.blockchain.PeerComparator
 import io.iohk.ethereum.blockchain.sync.Blacklist.BlacklistReason
 import io.iohk.ethereum.blockchain.sync.PeerListSupportNg.PeerWithInfo
 import io.iohk.ethereum.network.EtcPeerManagerActor.PeerInfo
@@ -42,6 +43,10 @@ class PeersClient(
 
   val statusSchedule: Cancellable =
     scheduler.scheduleWithFixedDelay(syncConfig.printStatusInterval, syncConfig.printStatusInterval, self, PrintStatus)
+
+  val numberOfPeersToFetchFrom = syncConfig.peersToFetchFrom
+
+  val activeFetchingNodes: mutable.Set[PeerWithInfo] = lruSet(numberOfPeersToFetchFrom)
 
   def receive: Receive = running(Map())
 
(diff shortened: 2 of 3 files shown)


src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
```diff
@@ -209,18 +209,23 @@ object PeersClient {
...
+      .collect { case PeerWithInfo(peer, peerInfo @ PeerInfo(_, _, true, _, _)) =>
+        (peer, peerInfo)
       }
 
     val peer =
       peersToUse
-        .filter { case (peer, _) => !activeFetchingNodes.contains(peer) }
-        .maxByOption { case (_, weight) => weight }
+        .filter { case (peer, _) => !activeFetchingNodes.map(_.peer).contains(peer) }
+        .filter { case (_, peerInfo) =>
```
leo-bogastry: I think a `filterNot` would be more readable:
```
.filterNot { case (peer, _) => activeFetchingNodes.map(_.peer).contains(peer) }
.filterNot { case (_, peerInfo) => activeFetchingNodes.map(_.peerInfo).exists(PeerComparator.doPeersHaveSameBestBlock(peerInfo, _))
```

src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
```diff
@@ -209,18 +209,23 @@ object PeersClient {
...
+        (peer, peerInfo)
       }
 
     val peer =
       peersToUse
-        .filter { case (peer, _) => !activeFetchingNodes.contains(peer) }
-        .maxByOption { case (_, weight) => weight }
+        .filterNot { case (peer, _) => activeFetchingNodes.map(_.peer).contains(peer) }
+        .filterNot { case (_, peerInfo) =>
+          activeFetchingNodes.map(_.peerInfo).exists(PeerComparator.doPeersHaveSameBestBlock(peerInfo, _))
```
jvdp, line 226: In theory, this line implies the same as the line above, right? Cause a peer will have the same best block as itself?

src/main/scala/io/iohk/ethereum/blockchain/sync/PeerComparator.scala
```diff
@@ -0,0 +1,9 @@
+package io.iohk.ethereum.blockchain
+
+import io.iohk.ethereum.network.EtcPeerManagerActor.PeerInfo
+
+object PeerComparator {
+
+  def doPeersHaveSameBestBlock(peerInfo1: PeerInfo, peerInfo2: PeerInfo): Boolean =
+    peerInfo1.bestBlockHash == peerInfo2.bestBlockHash
```
jvdp, line 8: This looks really simple; wasn't it extracted a bit prematurely?

jvdp (approved):

jvdp:
How will this interact with selecting peers from which to request blocks? They won't be matched up with corresponding one yet, right?

src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
AnastasiiaL: in theory the peers are progressing forward. So if we added the peer to the activeFetchingNodes 10min ago, the peersToDownloadFrom can already contain same peer with updated best block

src/main/scala/io/iohk/ethereum/blockchain/sync/PeerComparator.scala
AnastasiiaL: I agree it's too simple to be extracted, but this was explicitly stated on the story during the refinement that the comparison logic should not be a part of the PeersClient. Among other reasons for switching to another comparison logic

AnastasiiaL:
> How will this interact with selecting peers from which to request blocks? They won't be matched up with corresponding one yet, right?

this was only for enabling a pool of selected peers. How this will be called was not refined yet

src/main/scala/io/iohk/ethereum/blockchain/sync/PeersClient.scala
jvdp: Ahh ok, that's fun... Immutable objects in a mutable datastructure.

AnastasiiaL: at this point, the data structure is already immutable :)

leo-bogastry (approved):

PR #1102, opened 2021-08-17
Prevent retesteth from hanging the CI build

jvdp:
Basically by wrapping it with `timeout 5m`. In addition I bumped the verbosity so it will be easier to see in the logs what's going on.

### DIFF
--- a/test-ets.sh
+++ b/test-ets.sh
@@ -14,7 +14,7 @@ final_exit_code=0
 
 function run_and_annotate {
   echo "running retesteth $1"
-  ets/retesteth -t "$1" 2>&1 | tee "retesteth-$1-log.txt"
+  ets/retesteth -t "$1" -- --verbosity 3 2>&1 | tee "retesteth-$1-log.txt"
   exit_code=$?
   echo "retesteth $1 exit code: $exit_code"
 
--- a/test-ets.sh
+++ b/test-ets.sh
@@ -14,7 +14,7 @@ final_exit_code=0
 
 function run_and_annotate {
   echo "running retesteth $1"
-  ets/retesteth -t "$1" &> "retesteth-$1-log.txt"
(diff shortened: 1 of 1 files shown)


test-ets.sh
```diff
@@ -14,7 +14,7 @@ final_exit_code=0
 
 function run_and_annotate {
   echo "running retesteth $1"
-  ets/retesteth -t "$1" &> "retesteth-$1-log.txt"
```
lukasz-golebiewski, line 17: Could you also `tee` the output to the console so that it's possible to follow what's going on in buildkite?

lukasz-golebiewski (approved):

dzajkowski (approved):

test-ets.sh
```diff
@@ -14,7 +14,7 @@ final_exit_code=0
...
+  timeout 5m ets/retesteth -t "$1" -- --verbosity 3 &> "retesteth-$1-log.txt"
```
leo-bogastry: According to the logs of the build, the ETS tests did run and completed, but the top summary says 
```
retesteth: GeneralStateTests -- --

retesteth crashed; check the artifacts
```
Maybe we need a longer timeout?

leo-bogastry:
It took three runs of the ETS tests, but with the timeout builkite now show the complete final report on the ETS tests

test-ets.sh
jvdp: Yes I introduced that in the first commit in this branch actually, to do debugging. But with the timeout the output should eventually end up as artifact. And for developing on ETS it should probably be run locally.

What do you think?

test-ets.sh
jvdp: ETS used to run very quickly (whole suite in a minute or so), I'm kinda curious why it's drastically slower now.

PR #1103, opened 2021-08-18
[ETCM-1095] eip2930 receipt

strauss-m:
# Description

Introduce Receipt for Type 01 transactions

# Proposed Solution

Type 01 receipt content is the same than legacy ones, but prefixed with the transaction type byte

### DIFF
--- a/src/main/scala/io/iohk/ethereum/domain/Receipt.scala
+++ b/src/main/scala/io/iohk/ethereum/domain/Receipt.scala
@@ -6,24 +6,51 @@ import org.bouncycastle.util.encoders.Hex
 
 import io.iohk.ethereum.mpt.ByteArraySerializable
 
+sealed trait Receipt {
+  def postTransactionStateHash: TransactionOutcome
+  def cumulativeGasUsed: BigInt
+  def logsBloomFilter: ByteString
+  def logs: Seq[TxLogEntry]
+}
+
+// shared structure for EIP-2930, EIP-1559
+abstract class TypedLegacyReceipt(transactionTypeId: Byte, val delegateReceipt: LegacyReceipt) extends Receipt {
+  def postTransactionStateHash: TransactionOutcome = delegateReceipt.postTransactionStateHash
+  def cumulativeGasUsed: BigInt = delegateReceipt.cumulativeGasUsed
+  def logsBloomFilter: ByteString = delegateReceipt.logsBloomFilter
+  def logs: Seq[TxLogEntry] = delegateReceipt.logs
+}
+
 object Receipt {
 
   val byteArraySerializable: ByteArraySerializable[Receipt] = new ByteArraySerializable[Receipt] {
+
     import io.iohk.ethereum.network.p2p.messages.ETH63.ReceiptImplicits._
 
     override def fromBytes(bytes: Array[Byte]): Receipt = bytes.toReceipt
 
     override def toBytes(input: Receipt): Array[Byte] = input.toBytes
   }
+}
 
+object LegacyReceipt {
   def withHashOutcome(
       postTransactionStateHash: ByteString,
       cumulativeGasUsed: BigInt,
       logsBloomFilter: ByteString,
       logs: Seq[TxLogEntry]
-  ): Receipt =
-    Receipt(HashOutcome(postTransactionStateHash), cumulativeGasUsed, logsBloomFilter, logs)
+  ): LegacyReceipt =
+    LegacyReceipt(HashOutcome(postTransactionStateHash), cumulativeGasUsed, logsBloomFilter, logs)
+}
 
+object Type01Receipt {
+  def withHashOutcome(
+      postTransactionStateHash: ByteString,
+      cumulativeGasUsed: BigInt,
+      logsBloomFilter: ByteString,
+      logs: Seq[TxLogEntry]
+  ): Type01Receipt =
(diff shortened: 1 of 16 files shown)


lukasz-golebiewski (approved):

leo-bogastry (approved):

src/main/scala/io/iohk/ethereum/domain/Receipt.scala
```diff
@@ -35,24 +62,33 @@ object Receipt {
   *
   * More description: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-658.md
   */
-case class Receipt(
+case class LegacyReceipt(
```
AurelienRichez, line 65: I'm not really sure about the name because it is still used in the new receipts while "legacy" sounds like it is only there as a necessary evil and should not really be used anymore. On the other hand I guess it makes it clear that it corresponds to a `LegacyTransaction`. 

It is up for discussion, but we could do it like this: 

```scala
  case class ReceiptBody(
    postTransactionStateHash: TransactionOutcome, 
    cumulativeGasUsed: BigInt, 
    logsBloomFilter: ByteString, 
    logs: Seq[TxLogEntry]
  )
  
  sealed trait TransactionType
  case object Legacy extends TransactionType
  case object Type01 extends TransactionType

  case class Receipt(transactionType: TransactionType, body: ReceiptBody)
```

What do you think ?

src/main/scala/io/iohk/ethereum/domain/Transaction.scala
```diff
@@ -22,15 +22,31 @@ sealed trait Transaction {
...
   val LegacyThresholdLowerBound: Int = 0xc0
   val LegacyThresholdUpperBound: Int = 0xfe
+
+  def withGasLimit(gl: BigInt): Transaction => Transaction = {
+    case tx: LegacyTransaction         => tx.copy(gasLimit = gl)
+    case tx: TransactionWithAccessList => tx.copy(gasLimit = gl)
+  }
+
+  implicit class TransactionTypeValidator(val transactionType: Byte) extends AnyVal {
+    def isValidTransactionType: Boolean = transactionType >= MinAllowedType && transactionType <= MaxAllowedType
```
AurelienRichez: I'm not really a big fan of implicit conversion if we can avoid it because it can be surprising where it is used. Also, currently only `Type01` is allowed. Wouldn't it be more safe to just allow that for now ? 

Or maybe we can move that to the encoding/networking part because the fact that a transaction type is encoded as a number from 0 to 0x7f is not really relevant in the domain. We can have transaction types as an enumeration to make it impossible to have an invalid transaction type in the domain part, and let the networking part make the conversion.

src/main/scala/io/iohk/ethereum/domain/Transaction.scala
```diff
@@ -22,15 +22,31 @@ sealed trait Transaction {
...
-  val Type01: Int = 1
+  val Type01: Byte = 1.toByte
+
+  val MinAllowedType: Byte = 0
+  val MaxAllowedType: Byte = 0x7f
+
   val LegacyThresholdLowerBound: Int = 0xc0
   val LegacyThresholdUpperBound: Int = 0xfe
+
+  def withGasLimit(gl: BigInt): Transaction => Transaction = {
```
AurelienRichez: why not make this function a method in `Transaction` ?

src/test/scala/io/iohk/ethereum/ObjectGenerators.scala
```diff
@@ -69,20 +69,33 @@ trait ObjectGenerators {
...
     logsBloomFilter = ByteString(logsBloomFilter),
     logs = Seq()
   )
 
+  def type01ReceiptGen(): Gen[Type01Receipt] = legacyReceiptGen().map(Type01Receipt(_))
+
   def addressGen: Gen[Address] = byteArrayOfNItemsGen(20).map(Address(_))
 
-  def transactionGen(): Gen[LegacyTransaction] = for {
+  def accessListItemGen(): Gen[AccessListItem] = for {
```
AurelienRichez: nitpick: `def addressGen: Gen[Address] ` does not have an argument list so I think we should keep the same style 
```suggestion
  def accessListItemGen: Gen[AccessListItem] = for {
```

src/main/scala/io/iohk/ethereum/domain/Transaction.scala
strauss-m: I agree with the TransactionType enum:
- only check when decoding RLP data
- easier validation (transactionTypeByte is one of enum values)

As for the validation not being currently restricted to type01, I got two choices:
- current implementation: can accept un-handled transaction type, but don't need to change for new transaction
- proposed implementation: only valid for supported transaction types, but need to be updated for every new transaction

Your version seems more safe.

src/main/scala/io/iohk/ethereum/domain/Receipt.scala
strauss-m: I was reluctant to normalize Receipt as being always the same for every transactions, current and coming, but Legacy, TX01 and TX02 share the same receipt payload.

Even without enforcing / implying that every receipts will end up with the current form, we can do something like:
- rename LegacyReceipt -> ReceiptBody
- rename TypedLegacyReceipt -> TypedReceipt (with byte changed to enum, the same one used for Transaction, see next comment)
- keep Type01Receipt extends TypedReceipt(ReceiptBody(...))

AurelienRichez (approved):

PR #1104, opened 2021-08-19
[ETCM-1066] Add EIP-2929 VM integration tests

lukasz-golebiewski:
# Description
Adds tests for gas changes introduced in EIP-2929 that span across multiple opcodes

Also fixes a bug in which `G_cold_account_access` was charged incorrectly for storage access instead of `G_cold_sload`

### DIFF
--- a/src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpecPostEip2929.scala
+++ b/src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpecPostEip2929.scala
@@ -1,16 +1,23 @@
 package io.iohk.ethereum.vm
 
+import akka.util.ByteString
+import akka.util.ByteString.{empty => bEmpty}
+
+import org.bouncycastle.util.encoders.Hex
 import org.scalacheck.Arbitrary
 import org.scalacheck.Gen
 import org.scalatest.funsuite.AnyFunSuite
 import org.scalatest.matchers.should.Matchers
 import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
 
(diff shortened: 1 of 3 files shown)


src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpecPostEip2929.scala
```diff
@@ -203,13 +210,13 @@ class OpCodeGasSpecPostEip2929 extends AnyFunSuite with OpCodeTesting with Match
     val table = Table[UInt256, UInt256, Boolean, BigInt, BigInt](
       ("offset", "value", "alreadyAccessed", "startGas", "expectedGasConsumption"),
       (0, 1, true, G_callstipend + 1, G_sload),
-      (0, 1, false, G_sload + G_cold_account_access, G_sload + G_cold_account_access),
+      (0, 1, false, G_callstipend + 1, G_sload + G_cold_sload),
```
strauss-m, line 213: I was wondering why the '+1' here, but the yellow paper answered me: the call will fail if there is less than or EQUALS to G_callstipend gas left.

strauss-m (approved):

src/test/scala/io/iohk/ethereum/vm/OpCodeGasSpecPostEip2929.scala
dzajkowski: I wonder if we should start adding such tidbits of knowledge into the codebase.
//YP, page X, "quote"

dzajkowski (approved):

PR #1106, opened 2021-08-19
test bitte-ci

blaggacao:
- Remove .buildkite
- test


test-ets.sh
```diff
@@ -12,37 +12,15 @@ done
...
-  fi
-
-  summary=$(sed -n '/Total Tests Run/,$p' "retesteth-$1-log.txt")
-  if [[ -z "$summary" ]]; then
-    summary="retesteth crashed; check the artifacts"
-  fi
-  passed=$(grep -oP 'Total Tests Run: \d+' "retesteth-$1-log.txt")
-  failed=$(grep -oP 'TOTAL ERRORS DETECTED: \d+' "retesteth-$1-log.txt")
-
-  cat <<EOF | buildkite-agent annotate --context "retesteth-$1" --style "$style"
```
jvdp, line 34: Is there any Bitte CI equivalent of this? Since the ETS step currently fails (was marked "softFail" in Buildkite) it would be convenient to be able to see the actual success / failure status.

blaggacao: Unfortunately not (yet). We are trying to get a PoC running for now. You'll be able to use buildkite in parallel for a while until we can make sure Bitte CI is in a usable shape.

jvdp: Ok, that sounds good 👍 

Can you make a bitte / ci-specific version of this file (again)? I'm also making one soon for doing the same thing but locally and it would be good to distinguish these use cases.

blaggacao: You mean this https://github.com/input-output-hk/mantis/blob/mig-bitte-ci/test-ets-bitte.sh ?

jvdp: Yes exactly.

PR #1108, opened 2021-08-20
ETCM-1023: Block fetching and -building flow

jvdp:
Implements a flow that construct blocks from messages from the `PeerEventBus`, and hooks it into `RegularSync` via a temporary block buffer data structure and flow (as proof of concept.)

[update 23/8]: Just tested this on Sagano and it regular-syncs up to the best block and then remains stuck there. So it's close.


src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
```diff
@@ -89,4 +96,59 @@ class FetcherService(validator: BlockValidator, blockchainReader: BlockchainRead
         else
           bodiesAreOrderedSubsetOfRequested(remainingHeaders, respondedBodies, matchedBlocks)
     }
+
+}
+
+object FetcherService {
+  type Hashes = Seq[ByteString]
+
+  type BlockIdentifier = (ByteString, ByteString)
```
leo-bogastry: In this case you are not really making profit of types, I would suggest `type BlockIdentifier = (TransactionsRoot, OmmersHash)` (defining the missing types of course)

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BranchBuffer.scala
```diff
@@ -0,0 +1,49 @@
...
+    copy(byParent = byParent + (block.parentHash -> block), branchFound = Queue.empty)
+      .branchFrom(trunk match {
+        case BestBranch(tipBlockHash, _) => tipBlockHash
+        case _                           => ByteString.empty
+      })
+
+  @tailrec
+  private def branchFrom(hash: Hash): BranchBuffer =
+    byParent.get(hash) match {
+      case None        => this
```
leo-bogastry, line 36: I guess that if it receives a block that is from another branch it dies here because the `byParent` map won't have the block parent of this new block

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSync.scala
```diff
@@ -78,22 +96,49 @@ class RegularSync(
...
+    PeerEventBusActor
+      .messageSource(
+        peerEventBus,
+        PeerEventBusActor.SubscriptionClassifier
+          .MessageClassifier(
+            Set(Codes.BlockBodiesCode, Codes.BlockHeadersCode),
+            PeerEventBusActor.PeerSelector.AllPeers
+          )
+      )
+      .buffer(256, OverflowStrategy.fail)
```
leo-bogastry, line 131: I guess in the future this should be configurable, in case we need to tweak it?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BranchBuffer.scala
jvdp: Yep, so next step is to make this a bit smarter.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSync.scala
jvdp: Well that depends a bit on what the full logic will look like later, since incoming data depends on requests. So if processing lags a bit (for whatever reason), fewer requests for more data will be done. This is something we'll be able to reason about.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BranchBuffer.scala
```diff
@@ -0,0 +1,49 @@
...
+import io.iohk.ethereum.domain.branch.Branch
+
+/** Naive & temporary branch buffer implementation with some serious limitations:
+  * - it is not able to store multiple competing branches at the same time.
+  * - it will only find branches from the best branch tip.
+  *
+  * @param byParent in-memory buffer of blocks retrievable by parentHash. Only one block per parent is kept, last received wins.
+  * @param branchFound branch found from given best branch, as a list of blocks. They are removed from the buffer.
+  */
+case class BranchBuffer(byParent: Map[Hash, Block] = Map.empty, branchFound: Queue[Block] = Queue.empty) {
```
AnastasiiaL, line 25: why do we need a branch buffer if we immediately store everything?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
```diff
@@ -89,4 +96,60 @@ class FetcherService(validator: BlockValidator, blockchainReader: BlockchainRead
...
+    } yield Block(header, body)
+  }
+
+  /** State of block fetching stream after processing a given incoming message with block headers or bodies
+    *
+    * @param outstanding headers that are yet to be matched to bodies
+    * @param bodiesRequest information for requesting bodies corresponding to newly outstanding headers
+    * @param result blocks produced by matching received headers with received bodies
+    */
+  case class FetchState(
```
AnastasiiaL, line 121: so in the end we still can't get rid of the state? this was one of the goals we defined. We still can do it by using sync processing

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSync.scala
```diff
@@ -78,22 +96,49 @@ class RegularSync(
...
 
   override def receive: Receive = running(
     ProgressState(startedFetching = false, initialBlock = 0, currentBlock = 0, bestKnownNetworkBlock = 0)
   )
 
+  private def startNewFlow() =
+    PeerEventBusActor
+      .messageSource(
+        peerEventBus,
+        PeerEventBusActor.SubscriptionClassifier
```
AnastasiiaL, line 124: if you're replacing the PeerRequestHandler here then PeerDisconnectedClassifier is missing

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/FetcherService.scala
jvdp: So this isn't a state in the imperative 'shared mutable state' sense; it's kept by the `scan` operator. This is a more functional approach.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSync.scala
jvdp: It's not replacing PeerRequestHandler yet. What's missing is a piece of `Flow` that manages the sending of headers requests and handling those disconnection and best block messages (and more?)

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BranchBuffer.scala
jvdp: We don't immediately store everything; only rooted blocks / branches, with the reasoning that these blocks have passed the proof of work threshold and so should be relatively safe to store without DoS concerns.

(With the caveat that we should also not store super old blocks; we have been discussing a "settled block" concept that we can also use for this.)

AnastasiiaL: got it. how much work do you think it takes to turn this into a structure that supports parallel branches? and can the current one be used in any way as is?

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/RegularSync.scala
AnastasiiaL: but it's gonna replace it in future? just pointing out what should be there if we proceed with this

jvdp: I sorta misspoke. It's largely replacing `PeerRequestHandler` but by replacing `BlockFetcher`. FastSync (until rewritten) still uses `PeerRequestHandler` so it's not replacing it in that sense.

src/main/scala/io/iohk/ethereum/blockchain/sync/regular/BranchBuffer.scala
jvdp: The current one is able to sync Sagano fully, but then might get stuck on a dead branch.

I think I'm close to one that will work for parallel branches for my current ticket (1059.)

jvdp:
So after some discussion we've decided to abandon the `PeerEventBus`-based block stream and stick with the plans for the `FetcherService` method, but it's left in for now to test the better `BranchBuffer` during development.

PR #1109, opened 2021-08-20
nix: remove old niv code paths & various fixes

nrdxp:
* Use flake-compat so sources are same for stable/unstable users
* completely remove niv
* add envrc for direnv users
* add retesteth to devshell, and unify flake devShell and shell.nix

# Description

I was requested to add retesteth to the Nix shell, and I discovered a lot of inconsistency and stale code, so I cleaned it up as best I could.

# Proposed Solution

We don't use Niv, so remove all mention of it, pull sources exclusively using the flake, and use flake-compat to pull them out for stable Nix users (like the buildkite).

The most important thing is to make sure CI still passes since we no longer use some of the old niv pins, and also to make sure I didn't overlook anything. Once it is green, it should be ready to merge.


nix/compat.nix
```diff
@@ -0,0 +1,197 @@
+# Compatibility function to allow flakes to be used by
```
jvdp: Is this taken from somewhere? If so, might be useful to include some info to keep track of it.

jvdp (approved):

nix/compat.nix
nrdxp: yes:
https://github.com/edolstra/flake-compat

I can just use a builtin fetch function to grab it instead of dumping it in, I was just in a rush I guess :smile:

leo-bogastry (approved):

nix/compat.nix
nrdxp: I updated the expression to pull it with builtins.fetchurl

PR #1111, opened 2021-08-24
[ETCM-1049] Berlin config

lukasz-golebiewski:
# Description
This PR adds changes necessary for syncing beyond the ETH Berlin fork

### DIFF
--- a/src/main/scala/io/iohk/ethereum/vm/BlockchainConfigForEvm.scala
+++ b/src/main/scala/io/iohk/ethereum/vm/BlockchainConfigForEvm.scala
@@ -13,6 +13,7 @@ import io.iohk.ethereum.vm.BlockchainConfigForEvm.EthForks.Byzantium
 import io.iohk.ethereum.vm.BlockchainConfigForEvm.EthForks.Constantinople
 import io.iohk.ethereum.vm.BlockchainConfigForEvm.EthForks.Istanbul
 import io.iohk.ethereum.vm.BlockchainConfigForEvm.EthForks.Petersburg
+import io.iohk.ethereum.vm.BlockchainConfigForEvm.EthForks.Berlin
 
 /** A subset of [[io.iohk.ethereum.utils.BlockchainConfig]] that is required for instantiating an [[EvmConfig]]
   * Note that `accountStartNonce` is required for a [[WorldStateProxy]] implementation that is used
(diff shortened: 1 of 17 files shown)


src/main/scala/io/iohk/ethereum/vm/BlockchainConfigForEvm.scala
```diff
@@ -68,7 +71,9 @@ object BlockchainConfigForEvm {
     val BeforeByzantium, Byzantium, Constantinople, Petersburg, Istanbul, Berlin = Value
   }
 
-  def isEip2929Enabled(etcFork: EtcFork): Boolean = etcFork >= EtcForks.Magneto
+  def isEip2929Enabled(etcFork: EtcFork, ethFork: BlockchainConfigForEvm.EthForks.Value): Boolean =
```
strauss-m, line 74: It could be nicer to use EthForks.EthFork instead of EthForks.Value (or better with the proper import just use EthFork) to mimick the EtcFork type declaration

strauss-m (approved):

src/main/scala/io/iohk/ethereum/vm/BlockchainConfigForEvm.scala
lukasz-golebiewski: I agree, but for some reason the compiler is unhappy with what you proposed

strauss-m: Strange, you're right. It can handle explicit EthForks.EthFork but not import + standalone EthFork

strauss-m: And for whatever reason, even thought I can't see a difference, removing the manual EthFork import, and letting idea fixes itself the missing import seems to do the tricks.

dzajkowski (approved):

jvdp:
Can we now also enable Berlin in ETS? https://github.com/input-output-hk/mantis/blob/develop/ets/config/mantis/config

lukasz-golebiewski:
> Can we now also enable Berlin in ETS? https://github.com/input-output-hk/mantis/blob/develop/ets/config/mantis/config

I think we need to wait for all other changes related to Magneto/Berlin

PR #1114, opened 2021-08-27
[ETCM-1082] Add block metadata storage

leo-bogastry:
# Description
In order to have explicit and easily queryable information on whether blocks have already been successfully executed, a new storage `BlockMetadaStorage` (and `BlockMetadataProxy` ) was added.
This storage can easily be extended in the future if we need more metadata block information.

# Proposed Solution
Added `BlockMetadaStorage`, `BlockMetadataProxy` and modified `ConsensusImpl` to first check for already executed blocks and only executed those blocks not yet executed.
A small step of this implementation is missing, which is to add the initial block metadata with `isExecuted=false`, because it can only be done once saving blocks is extracted from BlockExecutions and done in a chains manager layers. Follow-up ticket created (ETCM-1138)
The PR is divided into several commits, that help with the review.

# Testing
Tested against ETC mainnet
Deployed in Staging and tested against RPC20 tests
A few ETS tests are failing, see comment below

### DIFF
--- a/src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
+++ b/src/it/scala/io/iohk/ethereum/ledger/BlockImporterItSpec.scala
@@ -53,7 +53,50 @@ class BlockImporterItSpec
     testScheduler.awaitTermination(60.second)
   }
 
-  "BlockImporter" should "not discard blocks of the main chain if the reorganisation failed" in new TestFixture() {
+  "BlockImporter" should "import a block to the top of the main chain" in new StartedImportFixture {
+    // Fixture imports up to oldBlock4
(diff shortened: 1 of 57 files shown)


leo-bogastry:
An extra 36 tests of the ETS suite are failing. If they are run individually, in fresh mantis instance, they do pass
Also, if in `BlockchainWriter.save` I modify the instruction I added that set a block as executed from `true` to `false`, then the ETS suite geos back to the previous number of failing tests.
The reason why I added the `blockMetadataProxy.putBlockIsExecuted(block.header.hash, isExecuted = false)` in BlockchainWriter.save is because we currently only save blocks that were executed and this way all can be committed atomically.
All other tests are running well, including ERC20 in Staging.
Let me know your thoughts

PR #1117, opened 2021-09-07
[ETCM-1161] revert FastSync to version before header skeleton

jvdp:
It turns out it's just about as fast and more reliable this way.


src/it/scala/io/iohk/ethereum/sync/FastSyncItSpec.scala
```diff
@@ -234,7 +234,7 @@ class FastSyncItSpec extends FlatSpecBase with Matchers with BeforeAndAfterAll {
     )
   }
 
-  it should "blacklist peer on Invalid batch last header number" in customTestCaseResourceM(
+  ignore should "blacklist peer on Invalid batch last header number" in customTestCaseResourceM(
```
AnastasiiaL: why ignoring these two? if the functionality is removed the tests can be removed also, otherwise it's gonna be confusing

AnastasiiaL (approved):

dzajkowski (approved):

AurelienRichez (approved):

PR #1119, opened 2021-09-08
[ETCM-1165] use tx1 accesslist in evm

strauss-m:
# Description

Use Transaction Type 01 access list to populate the EVM accessed* fields in ProgramState

### DIFF
--- a/src/main/scala/io/iohk/ethereum/extvm/VMServer.scala
+++ b/src/main/scala/io/iohk/ethereum/extvm/VMServer.scala
@@ -152,7 +152,7 @@ class VMServer(messageHandler: MessageHandler) extends Logger {
       initialAddressesToDelete = Set(),
       evmConfig = vmConfig,
       originalWorld = world,
-      // FIXME add access list to CallContext
+      // TODO ETCM-1202 use access list from CallContext
       warmAddresses = Set.empty,
       warmStorage = Set.empty
     )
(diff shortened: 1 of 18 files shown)


src/main/scala/io/iohk/ethereum/extvm/VMServer.scala
```diff
@@ -151,7 +151,10 @@ class VMServer(messageHandler: MessageHandler) extends Logger {
       world = world,
       initialAddressesToDelete = Set(),
       evmConfig = vmConfig,
-      originalWorld = world
+      originalWorld = world,
+      // FIXME add access list to CallContext
```
AurelienRichez: I created a ticket ETCM-1202 for this so we can add it there
```suggestion
      // TODO ETCM-1202 use access list from CallContext
```

src/main/scala/io/iohk/ethereum/vm/EvmConfig.scala
```diff
@@ -265,6 +270,10 @@ object FeeSchedule {
     override val G_cold_sload = 2100
     override val G_cold_account_access = 2600
     override val G_warm_storage_read = 100
+
+    // note: the access list does not exist until magneto hard fork
```
AurelienRichez: in fact this is also true for the values just above (up to `G_cold_sload`) so we can move that comment up.

AurelienRichez (approved):

AnastasiiaL (approved):

